PR #1, opened 2021-03-16
Adding some preliminary rustdocs

smklein:


smklein:
This is something of a grab-bag of updates, but I figured I'd send it along.

Basically, I'm trying to fill out rust docs for portions of Propolis as I read through. The structure is making more sense to me, though I've left some comments in-line where I had some questions.

I'm really interested in committing more - I haven't worked with hypervisors much before, and this codebase looks super cool!

propolis/src/block.rs
```diff
@@ -25,13 +28,19 @@ pub enum BlockResult {
...
 
+/// Trait indicating that a type may be used as a request to a block device.
 pub trait BlockReq: Send + Sync + 'static {
+    /// Type of operation being issued.
     fn oper(&self) -> BlockOp;
+    /// Offset within the block device, in bytes.
     fn offset(&self) -> usize;
+    /// Returns the next region of memory within a request to a block device.
     fn next_buf(&mut self) -> Option<GuestRegion>;
+    /// Signals to the guest that a block operation has been completed.
```
pfmooney, line 39: It's really signalling to the driver emulation that the operation is complete.  It's up to the emulation how/if the guest should be notified.

propolis/src/block.rs
```diff
@@ -57,6 +72,9 @@ pub struct PlainBdev<R: BlockReq> {
     cond: Condvar,
 }
 impl<R: BlockReq> PlainBdev<R> {
+    /// Creates a new block device from a device at `path`.
+    ///
+    /// TODO: Not yet implemented for character devices.
```
pfmooney, line 77: Not sure that's worth noting here ATM

propolis/src/block.rs
```diff
@@ -149,6 +167,9 @@ impl<R: BlockReq> PlainBdev<R> {
         }
         BlockResult::Success
     }
+
+    /// Spawns a new thread named `name` on the dispatcher `disp` which
+    /// begins processing incoming requests.
```
pfmooney, line 172: "incoming requests" is a little too specific.  That thread can do whatever it wants

propolis/src/chardev/sock.rs
```diff
@@ -137,6 +137,10 @@ impl UDSock {
                 if e.kind() != ErrorKind::AddrInUse {
                     return Err(e);
                 }
+                eprintln!(
+                    "Couldn't connect to {}, removing and trying again...",
```
pfmooney, line 141: Considering it's a `bind(3)` call, not `connect(3)`, this verbiage is a little off

propolis/src/common.rs
```diff
@@ -44,12 +44,24 @@ enum ROInner<'a> {
...
 }
 impl<'a> ReadOp<'a> {
+    /// Initializes a new read operation from a buffer.
+    ///
+    /// # Arguments
+    ///
+    /// - `op_offset`: An auxiliary offset stored within the operation.
+    /// - `buf`: A buffer which represents the "sink" of the read operation.
+    // TODO: Is there a better explanation for `op_offset`? It doesn't seem to
+    // be used in any calculations within this operation.
```
pfmooney, line 64: op_offset is used to specify the offset of the operation within a given region.  It's not used in the calculations so much as being "along for the ride" as metadata that the emulation can use, rather than needing another function argument to pass along the offset.

propolis/src/common.rs
```diff
@@ -60,6 +72,20 @@ impl<'a> ReadOp<'a> {
...
+    /// # Arguments
+    ///
+    /// - `op_offset`: Offset of the child operation. Does not need to correlate
+    /// to the `parent` operation's offset.
+    /// - `parent`: The operation from which this operation is being split.
+    /// - `range`: The location within the parent operation to be moved
+    /// to the child.
+    // TODO: Why doesn't this impact the `write_offset` of the parent, or take
+    // that into consideration? What if the parent already wrote past this
+    // region?
```
pfmooney, line 88: I started tracking read_offset/write_offset in the hopes that I could, for example, verify that the data of the operation was entirely consumed when the emulation was finished.  That got a little tricky considering how the `new_child` behavior can split arbitrarily.  I didn't feel like tracking extents or a bitfield or something quite yet.  In this case, when something splits off a new child, it's expected that the consumer of that ReadOp/WriteOp will intend to consume the whole thing (and thus start from a zeroed read/write offset)

propolis/src/dispatch/mod.rs
```diff
@@ -361,6 +387,10 @@ impl DispCtx {
     ) -> DispCtx {
         DispCtx { mctx, event: EventCtx::new(edisp), ctrl, di: inner }
     }
+
+    /// Returns true if the function holding this [`DispCtx`] object
+    /// should yield control back to the dispatcher.
+    // TODO: Why "&mut self", not just "&self"?
```
pfmooney, line 393: The lifetime requirement means there would be no "dangling borrows" of resources from this DispCtx at the time of yield.

propolis/src/dispatch/mod.rs
```diff
@@ -375,12 +405,20 @@ impl DispCtx {
         cb(&inst)
     }
 
+    /// Halts the instance asssociated with the dispatcher.
+    // TODO: What prevents re-invoking this method concurrently / repeatedly?
+    // Would that be safe? Kinda seems like invalid state transitions would
+    // cause a panic.
```
pfmooney, line 411: Halt (and reset, for your subsequent comment) are still new additions and the interfaces are being polished

propolis/src/exits.rs
```diff
@@ -1,13 +1,21 @@
...
     vm_entry, vm_entry_cmds, vm_entry_payload, vm_exit, vm_exitcode,
 };
 
+/// Describes the reason for exiting execution of a vCPU.
 pub struct VmExit {
+    /// The instruction pointer of the guest at the time of exit.
     pub rip: u64,
+    /// The length of the instruction which triggered the exit.
+    /// Zero means unknown.
+    // TODO: Unsure about this; is this description accurate?
```
pfmooney, line 16: For exits which are due to an instruction (say an attempted MMIO), it's the length of that instruction.  There are other exits reasons which are not related to the currently executing guest instruction which would result in that field being 0.

propolis/src/exits.rs
```diff
@@ -67,6 +76,11 @@ impl From<&vm_exit> for VmExitKind {
             Ok(c) => c,
         };
         match code {
+            vm_exitcode::VM_EXITCODE_VMX => {
+                // TODO: Populate and use this exit type.
+                let _vmx = unsafe { &exit.u.vmx };
+                VmExitKind::Vmx
+            }
```
pfmooney, line 83: Make this separate from the rustdoc change?

propolis/src/instance.rs
```diff
@@ -9,14 +11,27 @@ use crate::inventory::Inventory;
...
 
+/// States of operation for an instance.
 #[derive(Copy, Clone, Eq, PartialEq, Debug)]
 pub enum State {
+    /// Initial state. Instances cannot return to this state
+    /// after transitioning away from it.
     Initialize,
+    /// The instance is booting. Instances automatically
+    /// transition to boot after initialization, unless they're
+    /// being explicitly destroyed.
```
pfmooney, line 22: For the time being, I would avoid the commentary about the automatic state changing, considering this is still pretty early and subject to change

propolis/src/instance.rs
```diff
@@ -88,6 +108,7 @@ impl Instance {
         Ok(this)
     }
 
+    // TODO: Why is this separate from creating an instance?
```
pfmooney, line 111: Initialization might take place over a longer period.  One could imagine in the dropshot version that individual disks and NICs are attached separately, in their own API calls.  I did not want to do it all in one go.

propolis/src/intr_pins.rs
```diff
@@ -17,9 +19,13 @@ pub trait IntrPin: Send + Sync + 'static {
     }
 }
 
+/// Describes the operation to take with an interrupt pin.
 pub enum PinOp {
+    /// Turn the interrupt on.
     Assert,
+    /// Turn the interrupt off.
     Deassert,
+    /// Turn the interrupt on, then off.
```
pfmooney, line 28: nit: not a huge fan of "turn on, turn off" vs assert/deassert

propolis/src/lib.rs
```diff
@@ -98,6 +98,9 @@ pub fn vcpu_run_loop(mut vcpu: VcpuHdl, ctx: &mut DispCtx) {
                 println!("wrmsr({:x}, {:x})", msr, val);
                 next_entry = VmEntry::Run
             }
+            VmExitKind::Vmx => {
+                panic!("VMX exits not yet handled");
+            }
```
pfmooney, line 103: separate from rustdocs change

propolis/src/vcpu.rs
```diff
@@ -1,24 +1,30 @@
...
     pub(crate) fn new(hdl: Arc<VmmHdl>, id: i32) -> Self {
         Self { hdl, id }
     }
 
+    /// ID of the virtual CPU.
     pub fn cpuid(&self) -> i32 {
         self.id
     }
 
+    /// Resets the capabilities of the virtual CPU.
```
pfmooney, line 27: Not so much resets, as it does "sets"

propolis/src/vmm/hdl.rs
```diff
@@ -129,12 +167,24 @@ impl VmmHdl {
         if let Some(name) = segname {
             let name_raw = name.as_bytes();
 
+            // XXX: Does this name need to be null-terminated?
+            // It's crossing an FFI boundary, and C won't have any
+            // way to distinguish the length.
```
pfmooney, line 172: We start with the name array filled with zeroes, and the `< SEG_NAME_LEN` ensures that at least one NUL will remain at the end of the copied bytes.

propolis/src/exits.rs
smklein: Done

propolis/src/lib.rs
smklein: Done

propolis/src/vmm/hdl.rs
```diff
@@ -156,18 +206,24 @@ impl VmmHdl {
...
 
         assert!(devoff.offset >= 0);
         Ok(devoff.offset as usize)
     }
 
+    /// Maps a memory segment into propolis' address space.
+    ///
+    /// Returns a pointer to the mapped segment, if successful.
+    // TODO: Could wrap this in an object which unmaps on drop?
+    // TODO: Unsafe docs
```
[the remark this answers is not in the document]
smklein: Ack, will remove the comments for now.

propolis/src/dispatch/mod.rs
pfmooney: I mean to resources exposed by the `DispCtx`.  I think it's plausible that things like the `memctx` other resource maps are updated with less lock-heavy implementation in the future.  The dispatcher yield point is a natural fit for how/when to update them.  The `&mut` prevents the worker thread from holding onto references which might become stale.

This example shows a case where those references held across the `check_yield()` aren't allowed:
```rust
    fn process_loop(&self, ctx: &mut DispCtx) {
        let mut reqs = self.reqs.lock().unwrap();
        loop {
            let mem = ctx.mctx.memctx();

            let some_data: u32 = mem.read(GuestAddr(0x1000)).unwrap();
            if ctx.check_yield() {
                break;
            }
            let more_data: u32 = mem.read(GuestAddr(0x1000)).unwrap();
...
```
The compiler balks about the mutable borrow at `check_yield`.

Switched up a little bit, the story is different, since `memctx` wasn't held across the yield point:
```rust
    fn process_loop(&self, ctx: &mut DispCtx) {
        let mut reqs = self.reqs.lock().unwrap();
        loop {
            let some_data: u32 = ctx.mctx.memctx().read(GuestAddr(0x1000)).unwrap();
            if ctx.check_yield() {
                break;
            }
            let more_data: u32 = ctx.mctx.memctx().read(GuestAddr(0x1000)).unwrap();
...
```

smklein: Interesting - I think the intent here is a little more clear to me: if the processing loop decides to yield, it should not be holding context from before the yield point, and forcing access to a `&mut self` method in `DispCtx` forces all other lifetimes referencing `DispCtx` to end (is that right?).

I think this goal makes sense, although I'm struggling a little with this mechanism for enforcement.

1) The `check_yield` method isn't *really* what you're trying to use as the "end of `&DispCtx` lifetimes" - the *yield* point would be where you want the lifetime to end.
2) A caller could choose to yield unconditionally, in which case the `&mut self` enforcement would never occur.
3) Even in the cases mentioned above, the yield happens to be implemented as "return from the processing function" (basically the `break` statement). In those cases, it isn't possible to "access `some_data`, yield, and then access `more_data`" - if the yield happens, we've already left the loop. I think this is similar to point (1) though, of focusing on "what does a yield look like" rather than "what does a check-if-we-should-yield look like".

propolis/src/instance.rs
```diff
@@ -152,6 +164,8 @@ impl Instance {
         });
     }
 
+    /// Registers  callback, `func`, whichi si invoked whenever a state
```
pfmooney, line 167: nit: typo

propolis/src/common.rs
```diff
@@ -208,6 +204,13 @@ pub struct WriteOp<'a> {
     read_offset: usize,
 }
 impl<'a> WriteOp<'a> {
+    /// Initializes a new write operation from a buffer.
+    ///
+    /// # Arguments
+    ///
+    /// - `op_offset`: An auxiliary offset stored within the operation,
+    /// identifying the region where `buf` should be stored.
```
pfmooney, line 212: The `op_offset` does not at all relate to an offset within `buf`.  It's the offset of the operation relative to the emulated resource, not the underlying storage (hence why it can be manipulated in `new_child` and such)

propolis/src/block.rs
```diff
@@ -36,7 +36,7 @@ pub trait BlockReq: Send + Sync + 'static {
     fn offset(&self) -> usize;
     /// Returns the next region of memory within a request to a block device.
     fn next_buf(&mut self) -> Option<GuestRegion>;
-    /// Signals to the guest that a block operation has been completed.
+    /// Signals to the driver emulation that a block operation has been completed.
```
pfmooney, line 39: nit: "device emulation" rather than "driver emulation"?

propolis/src/hw/rtc.rs
```diff
@@ -53,8 +53,7 @@ impl Rtc {
         let high_bytes = high_chunks.to_le_bytes();
         hdl.rtc_write(MEM_OFF_HIGH, high_bytes[0])?;
         hdl.rtc_write(MEM_OFF_HIGH + 1, high_bytes[1])?;
-        // XXX Shouldn't this be "high_bytes[2]"?
-        hdl.rtc_write(MEM_OFF_HIGH + 2, high_bytes[1])?;
+        hdl.rtc_write(MEM_OFF_HIGH + 2, high_bytes[2])?;
```
pfmooney, line 56: Can probably just commit this by itself (which would mean little more context included about the fix too, rather than being lumped in with all your docs work)

propolis/src/instance.rs
smklein: Fixed

propolis/src/block.rs
smklein: Fixed

propolis/src/hw/rtc.rs
smklein: Done

propolis/src/dispatch/mod.rs
smklein: Spoke offline (well, online, you know how it is) and resolved this - I was totally wrong, `break` isn't the yield, `check_yield` is the yield, so it makes sense that it needs to act on `&mut self`.

PR #5, opened 2021-04-05
Conversion to "Mapping" objects, reducing unsafe

smklein:
Reduces usage of unsafe, and relocates it to a dedicated module for handling guest memory mappings.

The heart of this patch exists in propolis::vmm::mapping, which introduces the following:

- Mapping: An owned memory mapping. Unmaps on drop.
- SubMapping: A borrowed memory mapping (think "str" vs "String").
- GuardSpace: A region of guard space, from which underlying memory regions may be allocated.

Additionally, there is the introduction of a new wrapper around the VMM:
- VmmFile: Basically just a File, but provides guarantees that the file will not be truncated (requires unsafe).

### DIFF
--- a/propolis/src/hw/virtio/queue.rs
+++ b/propolis/src/hw/virtio/queue.rs
@@ -389,9 +389,8 @@ impl Chain {
         };
         let mut done = 0;
         let total = self.for_remaining_type(true, |addr, len| {
-            let len = usize::min(raw.len(), len);
-            let remain = &mut raw[done..len];
-            if let Some(copied) = mem.read_into(addr, remain) {
+            let remain = &mut raw[done..];
+            if let Some(copied) = mem.read_into(addr, remain, len) {
                 let need_more = copied != remain.len();
 
                 done += copied;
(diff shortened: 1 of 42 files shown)


propolis/src/vmm/machine.rs
```diff
@@ -446,74 +407,34 @@ impl Builder {
...
-            .last();
-        if let Some((start, len, (_, _))) = last_mem_seg {
-            Ok(start + len)
-        } else {
-            Err(Error::new(ErrorKind::InvalidData, "missing guest memory"))
-        }
-    }
-
     fn prep_mem_map(&self, hdl: &VmmHdl) -> Result<ASpace<MapEnt>> {
-        let last_sysmem = self.last_sysmem_addr()?;
```
smklein, line 465: So, I'm not sure I fully understand this previous mechanism of "calculate guard space based on last sysmem address".

- Doesn't this calculation ignore ROM?
- Isn't it implicitly dependent on the order of SysMem registration?

The new version just sums the memory, and adds space for guard pages.

propolis/src/hw/virtio/queue.rs
```diff
@@ -381,7 +381,7 @@ impl Chain {
         if (self.read_stat.bytes_remain as usize) < item_sz {
             return false;
         }
-        // SAFETY: We assume the mutable item reference we have received is
+        // Safety: We assume the mutable item reference we have received is
```
smklein, line 384: This is minor, but I s/SAFETY/Safety for consistency regarding unsafe comments. If you prefer screamy case, that's fine too, just trying to be uniform.

propolis/src/vmm/machine.rs
smklein: Followed-up offline, using a new allocator within GuardSpace for this now, and calculating the "highest seen guest address" as the total size of that region.

PR #7, opened 2021-04-23
Minimal Dropshot Server

smklein:
- Adds APIs to create instance, change state, and interact with the serial console remotely
- Adds structure to the "Inventory" class for expanded future use

### DIFF
--- a/propolis-cli/src/main.rs
+++ b/propolis-cli/src/main.rs
@@ -120,6 +120,8 @@ fn main() {
         let hdl = machine.get_hdl();
         let chipset = hw::chipset::i440fx::I440Fx::create(Arc::clone(&hdl));
         chipset.attach(mctx);
+        let chipset_id = inv.register_root(chipset.clone(), "chipset".to_string())
+            .map_err(|e| -> std::io::Error { e.into() })?;
 
         // UARTs
         let com1 = LpcUart::new(chipset.irq_pin(ibmpc::IRQ_COM1).unwrap());
@@ -143,15 +145,20 @@ fn main() {
         LpcUart::attach(&com2, pio, ibmpc::PORT_COM2);
         LpcUart::attach(&com3, pio, ibmpc::PORT_COM3);
         LpcUart::attach(&com4, pio, ibmpc::PORT_COM4);
-        inv.register(com1, "com1".to_string());
-        inv.register(com2, "com2".to_string());
-        inv.register(com3, "com3".to_string());
-        inv.register(com4, "com4".to_string());
+        inv.register(chipset_id, com1, "com1".to_string())
+            .map_err(|e| -> std::io::Error { e.into() })?;
+        inv.register(chipset_id, com2, "com2".to_string())
+            .map_err(|e| -> std::io::Error { e.into() })?;
+        inv.register(chipset_id, com3, "com3".to_string())
+            .map_err(|e| -> std::io::Error { e.into() })?;
+        inv.register(chipset_id, com4, "com4".to_string())
+            .map_err(|e| -> std::io::Error { e.into() })?;
 
         // PS/2
         let ps2_ctrl = PS2Ctrl::create();
         ps2_ctrl.attach(pio, chipset.as_ref());
-        inv.register(ps2_ctrl, "ps2_ctrl".to_string());
+        inv.register(chipset_id, ps2_ctrl, "ps2_ctrl".to_string())
+            .map_err(|e| -> std::io::Error { e.into() })?;
 
         let debug = std::fs::File::create("debug.out").unwrap();
         let buffered = std::io::LineWriter::new(debug);
@@ -159,7 +166,8 @@ fn main() {
             Some(Box::new(buffered) as Box<dyn std::io::Write + Send>),
             pio,
         );
-        inv.register(dbg, "debug".to_string());
+        inv.register(chipset_id, dbg, "debug".to_string())
+            .map_err(|e| -> std::io::Error { e.into() })?;
 
         for (name, dev) in config.devs() {
             let driver = &dev.driver as &str;
@@ -211,7 +219,6 @@ fn main() {
         // with all pci devices attached, place their BARs and wire up access to PCI
         // configuration space
         chipset.pci_finalize(&ctx);
-        inv.register(chipset, "chipset".to_string());
 
         let mut fwcfg = hw::qemu::fwcfg::FwCfgBuilder::new();
         fwcfg
@@ -227,8 +234,10 @@ fn main() {
         let fwcfg_dev = fwcfg.finalize();
         fwcfg_dev.attach(pio);
 
-        inv.register(fwcfg_dev, "fwcfg".to_string());
-        inv.register(ramfb, "ramfb".to_string());
+        inv.register(chipset_id, fwcfg_dev, "fwcfg".to_string())
+            .map_err(|e| -> std::io::Error { e.into() })?;
+        inv.register(chipset_id, ramfb, "ramfb".to_string())
+            .map_err(|e| -> std::io::Error { e.into() })?;
 
         let ncpu = mctx.max_cpus();
         for id in 0..ncpu {
--- a/propolis-server/rust-toolchain
+++ b/propolis-server/rust-toolchain
@@ -1,0 +1,1 @@
+nightly
--- a/propolis-server/src/api.rs
+++ b/propolis-server/src/api.rs
@@ -1,0 +1,190 @@
+// TODO: REMOVE ME
+#![allow(dead_code)]
+
+// use bitflags::bitflags;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct InstancePathParams {
+    pub instance_id: Uuid,
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct InstanceEnsureRequest {
+    pub properties: InstanceProperties,
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct InstanceEnsureResponse {}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct InstanceGetResponse {
+    pub instance: Instance,
+}
+
+/// Requested state of an Instance.
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct InstanceStateChange {
+    pub state: InstanceStateRequested,
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub enum InstanceStateRequested {
+    Run,
+    Stop,
+    Reboot,
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct InstanceSerialRequest {
+    pub bytes: Vec<u8>,
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct InstanceSerialResponse {
+    pub bytes: Vec<u8>,
+}
+
+/// Current state of an Instance.
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub enum InstanceState {
+    Creating,
+    Starting,
+    Running,
+    Stopping,
+    Stopped,
+    Repairing,
+    Failed,
+    Destroyed,
+}
+
+#[derive(Clone, Deserialize, Serialize, JsonSchema)]
+pub struct InstanceProperties {
+    pub generation_id: u64,
+
+    /// Unique identifier for this Instance.
+    pub id: Uuid,
+    /// Human-readable name of the Instance.
+    pub name: String,
+    /// Free-form text description of an Instance.
+    pub description: String,
+
+    /// ID of the image used to initialize this Instance.
+    pub image_id: Uuid,
+    /// ID of the bootrom used to initialize this Instance.
+    pub bootrom_id: Uuid,
+    /// Size of memory allocated to the Instance, in MB.
+    pub memory: u64,
+    /// Number of vCPUs to be allocated to the Instance.
+    pub vcpus: u8,
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct Instance {
+    pub properties: InstanceProperties,
+    pub state: InstanceState,
+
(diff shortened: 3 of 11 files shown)


propolis-cli/src/main.rs
```diff
@@ -120,6 +120,8 @@ fn main() {
         let hdl = machine.get_hdl();
         let chipset = hw::chipset::i440fx::I440Fx::create(Arc::clone(&hdl));
         chipset.attach(mctx);
+        let chipset_id = inv.register_root(chipset.clone(), "chipset".to_string())
```
smklein, line 123: Changes in this file should be limited to the conversion to the new interface API

propolis-server/rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly
```
smklein, line 1: Not actually using USDT in the server yet, but this seems like a prereq

propolis-server/src/api.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct Instance {
+    pub properties: InstanceProperties,
+    pub state: InstanceState,
+
+    pub disks: Vec<DiskAttachment>,
+    pub nics: Vec<NicAttachment>,
+}
```
smklein, line 87: This portion is being "used", so to speak.

The rest is copied from RFD 152, but still very much in flux.

propolis-server/src/config.rs
```diff
@@ -0,0 +1,71 @@
+use std::collections::{btree_map, BTreeMap};
```
smklein, line 1: Here's an example config file I'm using:

```
bootrom = "/export/home/smklein/OVMF_CODE.fd"

[dev.block0]
driver = "pci-virtio-block"
disk = "/export/home/smklein/alpine2.iso"
pci-path = "0.4.0"

[dev.net0]
driver = "pci-virtio-viona"
vnic = "vnic_prop0"
pci-path = "0.5.0"
```

It's somewhat of a kludge; I expect the devices to eventually removed as we migrate to a distributed device.

propolis-server/src/main.rs
```diff
@@ -0,0 +1,351 @@
...
+
+#[endpoint {
+    method = PUT,
+    path = "/instances/{instance_id}/serial",
+}]
+async fn instance_serial(
+    rqctx: Arc<RequestContext<ServerContext>>,
+    path_params: Path<api::InstancePathParams>,
+    request: TypedBody<api::InstanceSerialRequest>,
+) -> Result<HttpResponseOk<api::InstanceSerialResponse>, HttpError> {
```
smklein, line 280: This is a short-polling mega hack. I *can* split it into two separate interfaces (where the client becomes a server) and both sides long-poll, but it feels like we should just invest that time in websockets instead.

propolis/src/inventory.rs
```diff
@@ -1,63 +1,714 @@
...
+    fn insert_into_parent(
+        &mut self,
+        parent_id: EntityID,
+        child_id: EntityID,
+    ) -> Result<(), RegistrationError> {
+        let parent = self
+            .entities
+            .get_mut(&parent_id)
+            .ok_or_else(|| RegistrationError::MissingParent(parent_id))?;
+        parent.children.insert(child_id);
```
bnaecker, line 273: Does this need to return an `Err` if the child is already registered? The behavior on the set won't matter, and this function is internal, but it might help uphold the invariants of the public API.

propolis-server/src/main.rs
```diff
@@ -0,0 +1,346 @@
...
+
+    let properties = request.into_inner().properties;
+    if path_params.into_inner().instance_id != properties.id {
+        return Err(HttpError::for_internal_error(
+            "UUID mismatch (path did not match struct)".to_string(),
+        ));
+    }
+
+    // Create the instance.
+    let lowmem = (properties.memory * 1024 * 1024) as usize;
```
bnaecker, line 110: This is MiB, but the comment in `api::InstanceProeprties` above indicates that field is in MB.

bnaecker (commented):
This is great, thanks for all the work here. I have a few questions and clarifications, but nothing blocking.

propolis-server/rust-toolchain
pfmooney: It's used in `propolis` the library, so it is a prereq

propolis-server/src/initializer.rs
```diff
@@ -0,0 +1,287 @@
...
+        path: P,
+    ) -> Result<(), Error> {
+        let (romfp, rom_len) = open_bootrom(path.as_ref())
+            .unwrap_or_else(|e| panic!("Cannot open bootrom: {}", e));
+        self.machine.populate_rom("bootrom", |mapping| {
+            let mapping = mapping.as_ref();
+            if mapping.len() < rom_len {
+                return Err(Error::new(ErrorKind::InvalidData, "rom too long"));
+            }
+            let offset = mapping.len() - rom_len;
```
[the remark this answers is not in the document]
pfmooney: x86 CPUs boot from a specific address (`0xfffffff0`) which is supposed to be at the end of the bootrom.  For that reason, we use the size of the bootrom to determine how it is placed in the physical address space for correct placement.

propolis-server/src/api.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub block_count: u64,
+
+    /// Block size (bytes).
+    pub block_size: u32,
+
+    /// Storage interface.
+    pub interface: DiskType,
+}
+
+// TODO: Struggling to make this struct derive "JsonSchema"
```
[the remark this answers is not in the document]
smklein: It would be nice for the disk flags to be... well, flags, since there may be any combination of them. Agreed that we *could* use a u32 for this and then do the conversion ourselves (which is what we're more-or-less currently doing) to a stronger type.

However, that leaves the (minor) problem of "the u32 could be interpreted as something else by the server", or "something that was not `DiskFlags` could be supplied by the caller".

Totally agreed that "just use u32 on the wire" is a valid workaround, but avoiding the type conversion (if bitflags could derive `JsonSchema`) would remove the possibility of bad type casting.

propolis-server/src/api.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// Human-readable name of the Instance.
+    pub name: String,
+    /// Free-form text description of an Instance.
+    pub description: String,
+
+    /// ID of the image used to initialize this Instance.
+    pub image_id: Uuid,
+    /// ID of the bootrom used to initialize this Instance.
+    pub bootrom_id: Uuid,
+    /// Size of memory allocated to the Instance, in MB.
```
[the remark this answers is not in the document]
smklein: You're totally right - I was too fast and loose here, I'll tighten up the referenced allocations (MiB > MB!).

I do think there is an advantage to using MiB over bytes - we don't really want to enable byte-granularity allocations, so by calling this MiB, it's harder to misuse the API (sure, we can round up if someone asks for a fraction of a MiB, but it would be nice to just prevent that weird request in the first place).

On the other hand, consistency is nice for using "individual bytes" throughout. I could be persuaded either way,  but I'll proooobably stick with MiB in the meantime?

propolis-server/src/initializer.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+pub fn build_instance(
+    name: &str,
+    max_cpu: u8,
+    lowmem: usize,
+) -> Result<Arc<Instance>> {
+    let builder = Builder::new(name, true)?
+        .max_cpus(max_cpu)?
+        .add_mem_region(0, lowmem, Prot::ALL, "lowmem")?
```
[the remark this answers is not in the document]
smklein: The non-answer-but-sorta-truth is https://github.com/oxidecomputer/propolis/blob/master/propolis-cli/src/main.rs#L42-L65

@pfmooney please correct me, but I *think* some of it is arbitrary, and other parts are embedded within the expectations of the machine's layout.

- `vmm::MAX_SYSMEM` and `vmm:MAX_PHYSMEM` are documented as "arbitrary"
- The ROM region starting behind `0x1_0000_0000` is almost certainly related to the `0xFFFF_FFF0` x86 boot start constraint (though I'm not 100% sure why there's that `0x10` byte gap; I guess x86 CPUs actually like to boot into the last 16 bytes of bootrom).
- For the rest (mmio regions) I *think* they're arbitrary. If there are constraints, it would definitely be good to encode them.

propolis/src/inventory.rs
```diff
@@ -1,63 +1,714 @@
...
+    ///
+    /// Only one root device may exist in the inventory at once.
+    pub fn register_root<T: Entity>(
+        &mut self,
+        ent: Arc<T>,
+        name: String,
+    ) -> Result<EntityID, RegistrationError> {
+        let rec = Record::new(None, ent, name);
+
+        if self.root.is_some() {
```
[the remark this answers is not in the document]
smklein: good call; done

propolis/src/inventory.rs
```diff
@@ -1,63 +1,714 @@
...
+        self.entities.get(&id)?.concrete()
+    }
+
+    /// Access an entity's record by ID.
+    pub fn get(&self, id: EntityID) -> Option<&Record> {
+        let record = self.entities.get(&id)?;
+        Some(record)
+    }
+
+    /// Removes an entity from the inventory.
```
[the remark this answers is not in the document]
smklein: Updated

propolis/src/inventory.rs
smklein: The ordering (in the public api) guarantees that any conflict (e.g., child already registered) will return an error before getting here - I don't really want to return an error, because that error handling code would be impossible to get to.

I could change this to an assertion? `assert!(parent.children.insert(child_id))` would demand that invariant, but IMO also make it more clear that it should never be reachable from the public API

propolis-server/src/main.rs
smklein: yup - updated docs to `MiB`

smklein:
FYI @pfmooney - you asked about the usage of `UDSock` from the server, and if we could remove it (since the socket is an unnecessary intermediate step).

That is done with this latest patchset: [serial.rs](https://github.com/oxidecomputer/propolis/pull/7/files#diff-5e167ae595b5b7c680fa65b52f97c0249d8df511bfe384ea52afd1402a5ce35f).

`Serial` is a structure which sits atop a serial device, adding buffering and an asynchronous API. I've included a handful of tests which poke and prod at the "buffer full" behavior, but this should be a good primitive for when we move to the connection-based API.

To be explicit, the interfaces (inherited due to implementation of `AsyncWrite` and `AsyncRead`) I expect we'll use for the serial device are [`AsyncWriteExt`](https://docs.rs/futures/0.3.14/futures/io/trait.AsyncWriteExt.html) and [`AsyncReadExt`](https://docs.rs/tokio/0.2.2/tokio/io/trait.AsyncReadExt.html).

These provide interfaces equivalent to:
```rust
async fn read(&mut self, buf: &mut [u8]) -> io::Result<usize>;
async fn write(&mut self, buf: &[u8]) -> io::Result<usize>;
```

My implementations of `poll_read` and `poll_write` return immediately when data/space (data for read, space for write) exists in the buffer, and yields the task until the underlying source/sink makes room in the pipeline (via a [`Waker`](https://docs.rs/futures/0.3.14/futures/task/struct.Waker.html) object, which is the standard for futures).

I know we discussed the possibility of a structure which could "avoid reading in a tight loop when there is output from serial, and balance latency with bandwidth". I built a version with a worker task within the `Serial` structure (with fancy timer mechanics, to only make the output "readable" after a timer fired), but kinda realized this is a primitive would could be built *on top of* a buffering class, but probably shouldn't be inside it. I think this would let a caller make their own decision re: backoff - calling `serial.read(...).await` in a tight loop is fairly cheap, and if a caller wants to add their own criterion for backoff before transmitting the results across a network (based on time, buffering, or some combination) they can.

propolis-server/src/api.rs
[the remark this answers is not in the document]
bnaecker: I agree with all that, for sure. Maybe I put this comment in an unhelpful place, but I was trying to ask why the workaround below specified the flag as `const u32` rather than just defining `enum DiskFlags` with the appropriate variants and `From` impl. But I think I understand now -- you're writing it in a way such that, if the `bitflags!` bug can be resolved, the consuming code need not change. Good idea, apologies for the confusing comment here!

propolis/src/inventory.rs
bnaecker: Assert sounds good to me -- I agree handling an error gracefully here is both difficult and not really correct. But having the assert would be nice if we change the API or make some other mistake.

propolis-server/src/api.rs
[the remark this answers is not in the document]
bnaecker: Yeah, MiB sounds like a good approach here. The `ApiByteCount` type supports constructions from mebibytes, among other things, which we could leverage to make sure the conversion from bytes in the public API is an appropriate multiple of an MiB.

bnaecker (commented):
Thanks for clarifying some of those details! Otherwise LGTM.

pfmooney (commented):
Some of these comments were queued prior to the update (GH code review is Not The Best), so I'll just post what I had now

propolis-server/src/initializer.rs
```diff
@@ -0,0 +1,287 @@
...
+            vcpu.reboot_state().unwrap();
+            vcpu.activate().unwrap();
+            // Set BSP to start up
+            if id == 0 {
+                vcpu.set_run_state(bhyve_api::VRS_RUN).unwrap();
+                vcpu.set_reg(bhyve_api::vm_reg_name::VM_REG_GUEST_RIP, 0xfff0)
+                    .unwrap();
+            }
+        }
+        Ok(())
```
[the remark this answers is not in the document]
smklein: Split to "initialize_cpus"

propolis/src/inventory.rs
```diff
@@ -1,63 +1,714 @@
...
+    }
+    fn as_pci(&self) -> Option<&dyn Pci> {
+        None
+    }
+    fn as_chipset(&self) -> Option<&dyn Chipset> {
+        None
+    }
+    fn as_uart(&self) -> Option<&dyn Uart> {
+        None
+    }
```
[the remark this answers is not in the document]
smklein: Updated

PR #9, opened 2021-04-28
Add workspace and github action

smklein:


smklein:
This should cover building, testing, formatting and docs (though "doc warnings" are ignored). Happy to add any extra checks you'd like, but this seems to be working.

Also, making a single workspace to build everything is sorta the easiest option - even locally (rather than running tons of separate checks for different sub-projects).

smklein:
Also: I dunno how familiar you are with GitHub tooling, but FWIW, these checks will not stop us from merging. Branch protection is an option in "Settings" (to which I do not have access; happy to set it up if I get added).

pfmooney:
I don't think we need branch protection or anything like that for now.  The checks on a PR should be adequate.

PR #10, opened 2021-05-04
Create Propolis Client Crate

smklein:
This PR updates the propolis client crate to add some simple HTTP wrappers.

It leans heavily on "reqwest" to actually handle the transfer.

Known deficiencies:
- Not really handling redirects right now
- HTTP, not HTTPS - but https://docs.rs/reqwest/0.11.3/reqwest/blocking/struct.ClientBuilder.html#method.https_only looks like a great way to improve on this
- Serial console uses short-polling.


propolis-client/src/lib.rs
```diff
@@ -1,6 +1,152 @@
...
+// Sends a "request", awaits "response", and parses the body
+// into a deserializable type.
+async fn send_and_parse_response<T: DeserializeOwned>(
+    request: reqwest::RequestBuilder,
+) -> Result<T, Error> {
+    send_and_check_ok(request).await?.json().await.map_err(|e| e.into())
+}
+
+impl Client {
+    pub fn new(address: SocketAddr, log: Logger) -> Client {
```
luqmana, line 55: Should this take a `T: ToSocketAddrs`? Makes it a little more ergonomic in what a user can pass in.

propolis-client/src/lib.rs
```diff
@@ -1,6 +1,152 @@
...
+        Client {
+            client: reqwest::Client::new(),
+            log: log.new(o!("propolis_client address" => address.to_string())),
+            address,
+        }
+    }
+
+    async fn get<T: DeserializeOwned>(
+        &self,
+        path: String,
```
luqmana, line 65: I'd say either take a `&str` or like reqwest a `T: IntoUrl`

propolis-client/src/lib.rs
smklein: Hrmmm wouldn't this imply accepting multiple addresses, rather than just one? Seems like this would split the invocations to reqwest into multiple invocations (try one address, if it fails, try the next, etc). Is there a good wrapper for this?

propolis-client/src/lib.rs
smklein: Done!

propolis-client/src/lib.rs
luqmana: That does touch on another point I guess: the reason it returns an iterator is because `ToSocketAddrs` is implemented for strings for which it may end up needing to do name resolution and that might end up returning multiple results. It's a good question in whether we want `Client::new` to possibly be performing name resolution as well. If we don't, then I'm fine with leaving it as is and then the caller can decide to just call `.to_socket_addrs()?.next()` if they want.

propolis-client/src/lib.rs
```diff
@@ -1,6 +1,153 @@
...
+        if let Some(body) = body {
+            request = request.body(body);
+        }
+
+        send_and_parse_response(request).await
+    }
+
+    async fn put_no_response(
+        &self,
+        path: String,
```
luqmana, line 94: One more String here.

smklein: Done

PR #13, opened 2021-05-27
Fix warning by dev/release profiles from member Cargo.toml to workspace Cargo.toml.

luqmana:
Clear up the warning about "profiles for the non root package will be ignored" by moving these to the workspace root.

If we need to, we can add specific workspace member [overrides](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides) though that does come with the limitation that you can't specify `panic`, `lto` or `rpath` settings. Another limitation is if we publish one of the binary workspace members then `cargo install` does would not use the workspace root profile settings ([details](https://github.com/rust-lang/cargo/issues/8264)).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,3 +6,8 @@ members = [
   "propolis-client",
   "propolis-server",
 ]
+
+[profile.dev]
+panic = "abort"
+[profile.release]
+panic = "abort"
--- a/propolis-cli/Cargo.toml
+++ b/propolis-cli/Cargo.toml
@@ -7,11 +7,6 @@ edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
-[profile.dev]
-panic = "abort"
-[profile.release]
-panic = "abort"
-
 [dependencies]
 pico-args = "0.3"
 libc = "0.2"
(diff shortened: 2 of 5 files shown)


smklein:
> Clear up the warning about "profiles for the non root package will be ignored" by moving these to the workspace root.
> 
> If we need to, we can add specific workspace member [overrides](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides) though that does come with the limitation that you can't specify `panic`, `lto` or `rpath` settings. Another limitation is if we publish one of the binary workspace members then `cargo install` does would not use the workspace root profile settings ([details](https://github.com/rust-lang/cargo/issues/8264)).

This might be a bit overkill, what happens if we define these profiles in the workspace *and* the non-root packages? Would that resolve the `cargo install` case?

smklein (approved):

luqmana:
> > Clear up the warning about "profiles for the non root package will be ignored" by moving these to the workspace root.
> > If we need to, we can add specific workspace member [overrides](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides) though that does come with the limitation that you can't specify `panic`, `lto` or `rpath` settings. Another limitation is if we publish one of the binary workspace members then `cargo install` does would not use the workspace root profile settings ([details](https://github.com/rust-lang/cargo/issues/8264)).
> 
> This might be a bit overkill, what happens if we define these profiles in the workspace _and_ the non-root packages? Would that resolve the `cargo install` case?

In that case you'll get the same warning again about the profile settings being ignored for those non-root packages.

PR #15, opened 2021-06-02
Split propolis-server, add integration tests, fix reset behavior

smklein:


propolis-server-lib/src/server.rs
```diff
@@ -0,0 +1,433 @@
+use anyhow::Result;
```
smklein, line 1: Most of this file has been pulled out of "propolis-server/src/main.rs" below

propolis-server-lib/src/server.rs
```diff
@@ -0,0 +1,431 @@
...
+        // If we transition to the destroyed state, we should be able to
+        // free up resources for the underlying VM.
+        if matches!(next_state, propolis::instance::State::Destroy) {
+            let r = propolis::vmm::destroy_vm(id.to_string());
+            if let Err(e) = r {
+                eprintln!("Failed to destroy VM ({}): {}", id, e);
+            } else {
+                eprintln!("Destroyed VM ({})", id);
+            }
+        }
```
smklein, line 243: This bit is new, and worth taking a look at.

Rather than leaving defunct VMs sitting around, the server now explicitly destroys them. WDYT?

propolis-server-lib/tests/artifacts.rs
```diff
@@ -0,0 +1,130 @@
+use anyhow::{bail, Result};
```
smklein, line 1: This file is test-only, and allows thread-safe downloading of artifacts to be used in testing.

propolis-server-lib/tests/integration_tests.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+#[tokio::test]
+async fn test_ensure_instance_put_running() {
+    let log = initialize_log();
+    let server = initialize_server(log.clone()).await;
+
+    let client = Client::new(server.local_addr(), log);
+
+    let id = Uuid::from_str("0000002a-000c-0005-0c03-0938362b0809").unwrap();
```
smklein, line 131: These UUIDs are arbitrary but consistent - this means that if a test is failing repeatedly, we won't be generating tons of distinct VMs on a machine.

propolis-server-lib/src/server.rs
pfmooney: Destroying it rather than leaving it dangling sounds good.  Do you see a natural way for this to fit into the propolis-lib API, rather than reaching all the way in for the vmm method and ID string?

propolis-server-lib/tests/integration_tests.rs
pfmooney: Sounds good for until we have a reason for it to be otherwise

propolis-server-lib/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+dropshot = { git = "https://github.com/oxidecomputer/dropshot", branch = "main" }
+futures = "0.3"
+libc = "0.2"
+thiserror = "1.0"
+tokio = { version = "1", features = ["full"] }
+toml = "0.5"
+serde = "1.0"
+serde_derive = "1.0"
+schemars = { version = "0.8", features = [ "uuid" ] }
+structopt = { version = "0.3", default-features = false }
```
luqmana, line 23: You don't need this in propolis-server-lib as far as I can tell

propolis-server-lib/tests/artifacts.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+    Ok(())
+}
+
+// Used to ensure that the body of "setup" is only invoked once, no matter
+// how many tests exist in this test suite.
+static START: OnceCell<()> = OnceCell::const_new();
+
+/// Initialize a directory full of artifacts which may be used for testing.
+// TODO: Could actually return a reference to the Artifact objects after
```
luqmana, line 112: Towards this, you could just move the const's into setup and have it return them. That way one can't access them without calling setup.

propolis-server-lib/tests/integration_tests.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+}
+
+#[tokio::test]
+async fn test_uninitialized_server() {
+    let log = initialize_log();
+    let server = initialize_server(log.clone()).await;
+
+    let client = Client::new(server.local_addr(), log);
+    assert!(matches!(
```
luqmana, line 116: Since we're already using nightly, you can collapse these to just `assert_matches!`

pfmooney: The use of nightly is mainly to fulfill the requirements of `usdt`.  I think we should constrain our use of nightly features as much as possible (so we can get back on stable ASAP).

propolis-server-lib/Cargo.toml
smklein: Good catch; removing this and a handful of other unnecessary deps.

propolis-server-lib/tests/artifacts.rs
smklein: Sure, I'll do that (and delete the TODO, since this is done).

propolis-server-lib/tests/integration_tests.rs
smklein: I'm gonna defer making this change for the reasons Patrick mentioned, but TIL about `assert_matches`.

(Matches is already so useful, definitely interested in making it part of my idiomatic rust toolkit)

luqmana (approved):
Looks good to me!

propolis-server-lib/src/server.rs
smklein: Done - I moved this into the state machine driver in `instance.rs`.

This now requires that the invocation of `VmmHdl::destroy` act on "&self", instead of "&mut self". I'll toy around with managing the object more exclusively (perhaps behind an `Option`?). I'd really like to ensure that after destroy is called, we won't try to communicate with the VM.

PR #18, opened 2021-06-08
Update propolis / sled-agent state translation to include rebooting

smklein:
- Adds a "rebooting" state to align with https://github.com/oxidecomputer/omicron/pull/118
- Updates "quiesce" to be correctly identified as "stopping", distinct from "halt", which is "stopped"

(Sometime after this PR, I plan to align the propolis / omicron names for states, but I want them to be aligned *first* to make the rename patch nomenclature only, without any logic shifting)

### DIFF
--- a/client/src/api.rs
+++ b/client/src/api.rs
@@ -68,6 +68,7 @@ pub enum InstanceState {
     Running,
     Stopping,
     Stopped,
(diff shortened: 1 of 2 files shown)


luqmana (approved):
👍

PR #22, opened 2021-07-09
[block] Allow PlainBdevs to be explicitly read-only

smklein:
This command is particularly useful when mounting a
PlainBdev on a read-only filesystem, or in a situation
where the caller wants to ensure that modifications
cannot be made.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -185,9 +185,14 @@ fn main() {
                     let disk_path =
                         dev.options.get("disk").unwrap().as_str().unwrap();
 
+                    let readonly: bool = || -> Option<bool> {
+                        dev.options.get("readonly")?.as_str()?.parse().ok()
+                    }()
+                    .unwrap_or(false);
+
                     let plain: Arc<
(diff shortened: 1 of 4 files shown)


pfmooney (approved):

cli/src/main.rs
```diff
@@ -185,9 +185,14 @@ fn main() {
                     let disk_path =
                         dev.options.get("disk").unwrap().as_str().unwrap();
 
+                    let readonly: bool = || -> Option<bool> {
+                        dev.options.get("readonly")?.as_str()?.parse().ok()
+                    }()
+                    .unwrap_or(false);
```
pfmooney, line 191: As we get further into the config stuff, I imagine we'll want some sort of convenience access function for expected-boolean data.

smklein: Agreed - there's some tension right now between "settings that we would only ever do on local objects" vs "settings that will actually be applicable once we get e.g. cloud storage, remote images up".

100% on board with abstracting away the parsing though.

PR #25, opened 2021-07-26
Create new BlockDevBackingStore abstraction

jmpesp:
Separate ideas of BlockDev and BlockDevBackingStore:

- BlockDev is the API of a virtualized block device.
- BlockDevBackingStore fulfils the requests sent to the BlockDev.

PlainBdev does the same thing that it did before, except now requests are fulfilled by a supplied BlockDevBackingStore. New backing stores now only need to implement a new BlockDevBackingStore trait and hand it to a PlainBDev. Parts of PlainBdev that interacted with a file have been pulled out into a new FileBlockDevBackingStore.

Functionally, this only represents a refactor. The next work will implement a BlockDevBackingStore for Crucible.

There's a few TODOs here but I wanted to get some comments on this path forward with a PR.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -194,8 +194,14 @@ fn main() {
                     .unwrap_or(false);
 
                     let plain: Arc<
-                        block::PlainBdev<hw::virtio::block::Request, block::FileBlockDevBackingStore>,
-                    > = block::create_file_backed_block_device(disk_path, readonly).unwrap();
+                        block::PlainBdev<
+                            hw::virtio::block::Request,
+                            block::FileBlockDevBackingStore,
+                        >,
+                    > = block::create_file_backed_block_device(
+                        disk_path, readonly,
+                    )
+                    .unwrap();
 
                     let vioblk = hw::virtio::VirtioBlock::create(
                         0x100,
@@ -243,8 +249,10 @@ fn main() {
                         dev.options.get("readonly")?.as_str()?.parse().ok()
                     }()
                     .unwrap_or(false);
-                    let plain =
-                        block::create_file_backed_block_device(disk_path, readonly).unwrap();
+                    let plain = block::create_file_backed_block_device(
+                        disk_path, readonly,
+                    )
+                    .unwrap();
                     let ns = hw::nvme::NvmeNs::create(plain.clone());
 
                     if let Err(e) =
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -193,7 +193,9 @@ fn main() {
                     }()
                     .unwrap_or(false);
 
-                    let plain = block::PlainBdev::from_file(disk_path, readonly) .unwrap();
+                    let plain =
+                        block::PlainBdev::from_file(disk_path, readonly)
+                            .unwrap();
 
                     let vioblk = hw::virtio::VirtioBlock::create(
                         0x100,
@@ -242,7 +244,9 @@ fn main() {
                     }()
                     .unwrap_or(false);
 
-                    let plain = block::PlainBdev::from_file(disk_path, readonly).unwrap();
+                    let plain =
+                        block::PlainBdev::from_file(disk_path, readonly)
+                            .unwrap();
                     let ns = hw::nvme::NvmeNs::create(plain.clone());
 
                     if let Err(e) =
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -193,10 +193,7 @@ fn main() {
                     }()
                     .unwrap_or(false);
 
-                    let plain = block::create_file_backed_block_device(
-                        disk_path, readonly,
-                    )
-                    .unwrap();
+                    let plain = block::PlainBdev::from_file(disk_path, readonly) .unwrap();
 
                     let vioblk = hw::virtio::VirtioBlock::create(
                         0x100,
@@ -244,10 +241,8 @@ fn main() {
                         dev.options.get("readonly")?.as_str()?.parse().ok()
                     }()
                     .unwrap_or(false);
-                    let plain = block::create_file_backed_block_device(
-                        disk_path, readonly,
-                    )
-                    .unwrap();
+
+                    let plain = block::PlainBdev::from_file(disk_path, readonly).unwrap();
                     let ns = hw::nvme::NvmeNs::create(plain.clone());
 
                     if let Err(e) =
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -193,12 +193,7 @@ fn main() {
                     }()
                     .unwrap_or(false);
 
-                    let plain: Arc<
-                        block::PlainBdev<
-                            hw::virtio::block::Request,
-                            block::FileBlockDevBackingStore,
-                        >,
-                    > = block::create_file_backed_block_device(
+                    let plain = block::create_file_backed_block_device(
                         disk_path, readonly,
                     )
                     .unwrap();
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -194,8 +194,8 @@ fn main() {
                     .unwrap_or(false);
 
                     let plain: Arc<
-                        block::PlainBdev<hw::virtio::block::Request>,
-                    > = block::PlainBdev::create(disk_path, readonly).unwrap();
+                        block::PlainBdev<hw::virtio::block::Request, block::FileBlockDevBackingStore>,
+                    > = block::create_file_backed_block_device(disk_path, readonly).unwrap();
 
                     let vioblk = hw::virtio::VirtioBlock::create(
                         0x100,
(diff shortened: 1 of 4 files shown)


cli/src/main.rs
```diff
@@ -194,8 +194,14 @@ fn main() {
                     .unwrap_or(false);
 
                     let plain: Arc<
-                        block::PlainBdev<hw::virtio::block::Request>,
-                    > = block::PlainBdev::create(disk_path, readonly).unwrap();
+                        block::PlainBdev<
```
luqmana, line 197: Nit: Don't need to explicitly specify the type here. I know this is existing but given you also added a new type parameter here, I figure might as well clean it up. Should be the same as line 252 below.

propolis/src/block.rs
```diff
@@ -53,53 +54,138 @@ pub struct BlockInquiry {
...
 
         let meta = metadata(p)?;
         let is_ro = readonly || meta.permissions().readonly();
 
         let fp = OpenOptions::new().read(true).write(!is_ro).open(p)?;
-        let is_raw = fp.metadata()?.file_type().is_char_device();
+        let is_ro = readonly || meta.permissions().readonly();
+        let len = fp.metadata().unwrap().len() as usize;
+
+        Ok(FileBlockDevBackingStore { fp: fp, is_ro: is_ro, len: len })
```
luqmana, line 95: Nit: for cases where the name of the field is the same as the name of the variable you're initializing it with, you can use a shorthand where you don't specify the field name:

```rust
Ok(FileBlockDevBackingStore { fp, is_ro, len })
```

More info: https://doc.rust-lang.org/edition-guide/rust-2018/data-types/field-init-shorthand.html

propolis/src/block.rs
```diff
@@ -53,53 +54,138 @@ pub struct BlockInquiry {
...
+        readonly: bool,
+    ) -> Result<FileBlockDevBackingStore> {
         let p: &Path = path.as_ref();
 
         let meta = metadata(p)?;
         let is_ro = readonly || meta.permissions().readonly();
 
         let fp = OpenOptions::new().read(true).write(!is_ro).open(p)?;
-        let is_raw = fp.metadata()?.file_type().is_char_device();
+        let is_ro = readonly || meta.permissions().readonly();
```
luqmana, line 92: Duplicate of line 89 above?

propolis/src/block.rs
```diff
@@ -53,53 +54,138 @@ pub struct BlockInquiry {
...
+    }
+
+    pub fn get_file(&self) -> &File {
+        &self.fp
+    }
+}
+
+impl BlockDevBackingStore for FileBlockDevBackingStore {
+    fn issue_read(&self, offset: usize, sz: usize) -> Result<Vec<u8>> {
+        let mut vec = Vec::<u8>::with_capacity(sz);
```
luqmana, line 105: A bit unfortunate that we need to allocate an intermediate buffer here vs the previous behaviour of reading directly into the mapped guest memory. Could we just modify `issue_read` (and similarly `issue_write`) to be passed in a `GuestRegion`?

propolis/src/block.rs
```diff
@@ -53,53 +54,138 @@ pub struct BlockInquiry {
...
         Ok(Arc::new(this))
     }
-    fn raw_init(&mut self) {
-        // TODO: query block size, write cache, discard, etc
-        assert!(!self.is_raw);
-        let len = self.fp.metadata().unwrap().len() as usize;
-        self.sectors = len / self.block_size;
-    }
+
+    /*
```
luqmana, line 186: Nit: while rust does support block comments, generally line comments are favoured. I don't know if Oxide has any particular style guide but that's what the Rust standard library follows at least: https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#use-line-comments

propolis/src/block.rs
```diff
@@ -108,69 +194,126 @@ impl<R: BlockReq> PlainBdev<R> {
...
+
+            let result = match req.oper() {
+                BlockOp::Read => {
+                    self.process_read_request(offset, sz, &mem, buf)
+                }
+                BlockOp::Write => self.process_write_request(offset, &mem, buf),
+                BlockOp::Flush => self.process_flush(),
+            };
+
+            // TODO: should all BlockReq be attempted instead of returning early?
```
luqmana, line 226: I'm inclined to say if we fail to read/write/flush one of the BlockReqs, we should abort the IO request and signal as such to the guest.

cli/src/main.rs
jmpesp: :+1:, fixed in f5eccea

propolis/src/block.rs
jmpesp: Fixed in a413e2c

propolis/src/block.rs
jmpesp: Fixed in 8625835

propolis/src/block.rs
jmpesp: Good idea, this makes it much cleaner. Fixed in 6b79bfe.

propolis/src/block.rs
jmpesp: Yeah, I agree. Fixed in 234e243

propolis/src/block.rs
```diff
@@ -53,53 +54,121 @@ pub struct BlockInquiry {
...
+
+    pub fn get_file(&self) -> &File {
+        &self.fp
+    }
+}
+
+impl BlockDevBackingStore for FileBlockDevBackingStore {
+    fn issue_read(&self, offset: usize, sz: usize, mapping: SubMapping) -> Result<usize> {
+        let nread = mapping.pread(&self.fp, sz, offset as i64)?;
+        Ok(nread as usize)
```
luqmana, line 105: The return types should match up so you can simplify this a bit:
```suggestion
        mapping.pread(&self.fp, sz, offset as i64)
```

jmpesp: Fixed in 7d9e407

propolis/src/block.rs
jmpesp: Fixed in 285d000.

propolis/src/block.rs
```diff
@@ -202,7 +321,18 @@ impl<R: BlockReq> BlockDev<R> for PlainBdev<R> {
         BlockInquiry {
             total_size: self.sectors as u64,
             block_size: self.block_size as u32,
-            writable: !self.is_ro,
+            writable: !self.backing_store.is_ro(),
         }
     }
 }
+
+pub fn create_file_backed_block_device<R: BlockReq>(
```
luqmana, line 329: Minor: Since this is sort of a constructor for a `PlainBdev` I wonder if it should just be a static method on that type instead of a free-fn. We already have `PlainBdev::create` which just takes the backing store but we could make this another method like `from_file`. Thoughts?

propolis/src/block.rs
```diff
@@ -108,67 +181,113 @@ impl<R: BlockReq> PlainBdev<R> {
...
     }
-    fn process_read(&self, req: &mut R, ctx: &DispCtx) -> BlockResult {
+
+    // Match against individual BlockReqs and call appropriate processing functions.
+    fn process_request(&self, req: &mut R, ctx: &DispCtx) -> BlockResult {
         let mem = ctx.mctx.memctx();
 
         let mut offset = req.offset();
+
+        // TODO: are buffers guaranteed to be sector size? do BlockDevBackingStores require
```
luqmana, line 198: I'm pretty sure buffers are not required to be sector size.

propolis/src/vmm/mapping.rs
```diff
@@ -545,7 +545,7 @@ mod tests {
 
         let input: u64 = 0xDEADBEEF;
         mapping.as_ref().write(&input).unwrap();
-        let output = mapping.as_ref().read().unwrap();
+        let output: u64 = mapping.as_ref().read().unwrap();
```
luqmana, line 548: What changed to need this now?

luqmana:
Thanks @jmpesp! Besides some minor nits lgtm!

propolis/src/block.rs
jmpesp: Nice, done in 6d1389c

propolis/src/block.rs
```diff
@@ -108,67 +181,113 @@ impl<R: BlockReq> PlainBdev<R> {
...
+
+        if let Some(mapping) = mem.readable_region(&buf) {
+            let nwritten = self.backing_store.issue_write(offset, sz, mapping);
+            match nwritten {
+                Ok(nwritten) => {
+                    assert_eq!(nwritten as usize, sz);
+                    BlockResult::Success
+                }
+                Err(_) => {
+                    // TODO: Error writing to guest memory
```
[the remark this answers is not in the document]
jmpesp: Fixed in 4379308

propolis/src/vmm/mapping.rs
jmpesp: Strange, I saw a compile error previously, but I just removed it and did not see any problem. Reverted in 45f0ee9.

propolis/src/block.rs
jmpesp: Removed comment in cf484fc

jmpesp:
Is this ok to merge?

PR #26, opened 2021-07-26
Cross the illumos 13833 flag day

pfmooney:
This updates the bhyve-api (and propolis consumer) for the ioctl changes in the [13833](https://code.illumos.org/c/illumos-gate/+/1522) patchset.

### DIFF
--- a/bhyve-api/src/structs.rs
+++ b/bhyve-api/src/structs.rs
@@ -277,3 +277,33 @@ pub struct vm_run_state {
     pub sipi_vector: u8,
     pub _pad: [u8; 3],
 }
+
+pub const VM_MAX_NAMELEN: usize = 128;
+pub const VM_MAX_SEG_NAMELEN: usize = 128;
+
+// Allocate guest memory segments from existing reservoir capacity, rather than
+// attempting to create transient allocations.
+pub const VCF_RESERVOIR_MEM: u64 = 1;
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct vm_create_req {
+    pub name: [u8; VM_MAX_NAMELEN],
+    pub flags: u64,
+}
+impl Default for vm_create_req {
+    fn default() -> Self {
+        Self { name: [0u8; VM_MAX_NAMELEN], flags: 0 }
+    }
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct vm_destroy_req {
+    pub name: [u8; VM_MAX_NAMELEN],
+}
+impl Default for vm_destroy_req {
+    fn default() -> Self {
+        Self { name: [0u8; VM_MAX_NAMELEN] }
+    }
+}
--- a/propolis/src/vmm/hdl.rs
+++ b/propolis/src/vmm/hdl.rs
@@ -8,7 +8,6 @@
 //! object which represents a single VM.
 
 use super::mapping::*;
-use std::ffi::CString;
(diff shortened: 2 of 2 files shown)


pfmooney:
(This is pending [the RTI](https://illumos.topicbox.com/groups/advocates/T8c6c6ac6586dec72-M1d9fb3275304e7609da498f0/rti-13833-want-bhyve-memory-reservoir-and-friends) of course)

bhyve-api/src/structs.rs
```diff
@@ -277,3 +277,33 @@ pub struct vm_run_state {
     pub sipi_vector: u8,
     pub _pad: [u8; 3],
 }
+
+pub const VM_MAX_NAMELEN: usize = 128;
+pub const VM_MAX_SEG_NAMELEN: usize = 128;
+
+// Allocate guest memory segments from existing reservoir capacity, rather than
+// attempting to create transient allocations.
+pub const VCF_RESERVOIR_MEM: u64 = 1;
```
smklein, line 286: IMO it would be useful to label this as a flag, somehow. That's not immediately apparently without reading the corresponding illumos change.

propolis/src/vmm/hdl.rs
```diff
@@ -33,35 +32,40 @@ pub fn create_vm(name: impl AsRef<str>, force: bool) -> Result<VmmHdl> {
...
         .open(bhyve_api::VMM_CTL_PATH)?;
-    let namestr = CString::new(name)
-        .or_else(|_x| Err(Error::from_raw_os_error(libc::EINVAL)))?;
-    let nameptr = namestr.as_ptr();
     let ctlfd = ctl.as_raw_fd();
 
-    let res = unsafe { libc::ioctl(ctlfd, bhyve_api::VMM_CREATE_VM, nameptr) };
+    let mut req = bhyve_api::vm_create_req::default();
+    req.name[..name.len()].copy_from_slice(name.as_bytes());
+    let res = unsafe { libc::ioctl(ctlfd, bhyve_api::VMM_CREATE_VM, &req) };
```
smklein, line 46: (nit - kinda up to you)

Here and below, we have name length checks before copying the name into the C struct.
I think it makes sense to keep those two things as tightly coupled as possible, which could be done by an `impl`-ed `fn` that both checks the length and copies from a slice. I think this would make it harder for callers to:

- Check against the wrong length (off by ones are super easy here)
- Forget checking the length at all

smklein (approved):

propolis/src/vmm/hdl.rs
pfmooney: Added some additional comment context on that.

bhyve-api/src/structs.rs
pfmooney: Added additional comment context

propolis/src/vmm/hdl.rs
pfmooney: I've tightened it up a bit.  Returning the error directly to the client was belt-and-suspenders anyways.  I also considered a `new` fn on those two structs which took the name and did the appropriate copying, since it doesn't make sense to create a struct w/o a name, if that makes more sense

pfmooney:
FYI: illumos 13833 (& friends) was [integrated](https://github.com/illumos/illumos-gate/commit/b57f5d3e6a2df8d435e606797cf3934811848343), so this patch will be required on machines running a kernel featuring that work.

PR #27, opened 2021-07-28
[client][server] NIC instantiation via HTTP API

smklein:
Minimal API to request NICs over the HTTP API.

Unfinished things / open questions:
- Listing all attached NICs and their statuses. We *could* just store this info at the "server" layer, so we'd basically just return whatever information was requested previously. However, IMO it would be better to have a way of querying the device inventory to be able to return more dynamic device info.
- Assigning PCI slots. Currently, we're using a *super* hardcoded slot schema; feedback definitely welcome here.

### DIFF
--- a/client/src/api.rs
+++ b/client/src/api.rs
@@ -16,6 +16,7 @@ pub struct InstancePathParams {
 #[derive(Clone, Deserialize, Serialize, JsonSchema)]
 pub struct InstanceEnsureRequest {
     pub properties: InstanceProperties,
+    pub nics: Vec<NetworkInterfaceRequest>,
 }
 
 #[derive(Clone, Deserialize, Serialize, JsonSchema)]
@@ -89,7 +90,7 @@ pub struct Instance {
     pub state: InstanceState,
 
     pub disks: Vec<DiskAttachment>,
-    pub nics: Vec<NicAttachment>,
+    pub nics: Vec<NetworkInterface>,
 }
 
 /// Describes how to connect to one or more storage agent services.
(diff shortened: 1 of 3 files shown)


server/src/lib/server.rs
```diff
@@ -128,6 +128,24 @@ fn propolis_to_api_state(
...
+//
+// For all the devices requested by Nexus (network interfaces, disks, etc),
+// we'd like to assign a stable PCI slot, such that re-allocating these
+// devices on a new instance of propolis produces the same guest-visible
+// BDFs.
+fn slot_to_bdf(slot: api::PciSlot) -> Result<pci::Bdf> {
+    match slot.0 {
+        // NOTE: This "5" is 100% arbitrary; it just shows we'd map to different
+        // slots as new devices are requested.
+        n if n < 5 => Ok(pci::Bdf::new(0, n, 0)),
```
pfmooney, line 141: Slots 0 and 1 are taken by the hostbridge and chipset, respectively.  I might make sense for each class of devices (NICs, block volumes, etc) to each have its own index and corresponding mapping.  Perhaps this?:
0x8 - 0xf: NICs 0-7
0x10 - 0x17 - blockdevs 0.7

smklein: Updated to use this slot mapping scheme.

client/src/api.rs
```diff
@@ -176,21 +177,26 @@ pub struct DiskAttachment {
     pub state: DiskAttachmentState,
 }
 
+/// A stable index which is translated by Propolis
+/// into a PCI BDF, visible to the guest.
+#[derive(Copy, Clone, Deserialize, Serialize, JsonSchema)]
+pub struct PciSlot(pub u8);
```
pfmooney, line 183: Perhaps even leave PCI out of the name?  It is kind of implied by the device types we're using. :shrug:

pfmooney (approved):

PR #28, opened 2021-08-03
[virtio] Avoid advertising MTU feature, cope with dladm failure

smklein:


propolis/src/hw/virtio/viona.rs
```diff
@@ -126,7 +129,10 @@ impl VirtioDevice for VirtioViona {
     }
     fn device_get_features(&self) -> u32 {
         let mut feat = VIRTIO_NET_F_MAC;
-        feat |= VIRTIO_NET_F_MTU;
+        // NOTE: We have dropped the "VIRTIO_NET_F_MTU" flag from feat.
+        //
+        // https://www.illumos.org/issues/13992
+        // feat |= VIRTIO_NET_F_MTU;
```
pfmooney, line 135: Expose this conditionally based on `self.mtu.is_some()`?

smklein: Done

pfmooney (approved):

propolis/src/hw/virtio/viona.rs
pfmooney: Thanks

PR #29, opened 2021-08-11
Add highmem handling

pfmooney:
I forgot to wire up highmem handling when I was doing memory stuff.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -45,8 +45,9 @@ fn build_instance(
     name: &str,
     max_cpu: u8,
     lowmem: usize,
+    highmem: usize,
 ) -> Result<Arc<Instance>> {
-    let builder = Builder::new(name, true)?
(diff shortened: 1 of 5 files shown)


cli/src/main.rs
```diff
@@ -91,10 +100,15 @@ fn main() {
     let config = parse_args();
 
     let vm_name = config.get_name();
-    let lowmem: usize = config.get_mem() * 1024 * 1024;
     let cpus = config.get_cpus();
 
-    let inst = build_instance(vm_name, cpus, lowmem).unwrap();
+    const GB: usize = 1024 * 1024 * 1024;
```
dancrossnyc, line 105: Just as an aside, we should constants for things like this squirreled away at a higher level. No action here, just a note for the future.

dancrossnyc (approved):

luqmana (approved):

PR #30, opened 2021-08-23
Overhaul VM lifecycle handling

pfmooney:
Took a swing at getting instance reset working.  There are still a bunch of rough edges to address, but this could be a reasonable starting point for some of that effort.  The README update is still pending integration of the requisite patch in [illumos 14024](https://code.illumos.org/c/illumos-gate/+/1660)

### DIFF
--- a/bhyve-api/src/structs.rs
+++ b/bhyve-api/src/structs.rs
@@ -1,5 +1,7 @@
-use libc::size_t;
 use std::os::raw::{c_int, c_uint, c_void};
+
+use bitflags::bitflags;
+use libc::size_t;
 
 // 3:0 - segment type
 /// Descriptor type flag (0 = system, 1 = code/data)
@@ -269,6 +271,28 @@ pub struct vm_capability {
     pub captype: c_int,
     pub capval: c_int,
     pub allcpus: c_int,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone, Default)]
+pub struct vm_suspend {
+    /// Acceptable values defined by `vm_suspend_how`
+    pub how: u32,
+}
+
+// bit definitions for `vm_reinit.flags`
+bitflags! {
+    #[repr(C)]
+    #[derive(Default)]
+    pub struct VmReinitFlags: u64 {
+        const FORCE_SUSPEND = (1 << 0);
+    }
+}
+
+#[repr(C)]
+#[derive(Copy, Clone, Default)]
+pub struct vm_reinit {
+    pub flags: VmReinitFlags,
 }
 
 #[repr(C)]
--- a/propolis/src/hw/chipset/i440fx.rs
+++ b/propolis/src/hw/chipset/i440fx.rs
@@ -6,6 +6,7 @@ use crate::common::*;
 use crate::dispatch::DispCtx;
 use crate::hw::ibmpc;
 use crate::hw::pci::{self, Bdf, INTxPinID, PioCfgDecoder};
+use crate::instance;
 use crate::intr_pins::{IntrPin, LegacyPIC, LegacyPin};
 use crate::pio::{PioBus, PioDev};
 use crate::util::regmap::RegMap;
@@ -13,6 +14,13 @@ use crate::util::self_arc::*;
 use crate::vmm::{MachineCtx, VmmHdl};
 
 use lazy_static::lazy_static;
+
+const HB_DEV: u8 = 0;
+const HB_FUNC: u8 = 0;
+const LPC_DEV: u8 = 1;
+const LPC_FUNC: u8 = 0;
+const PM_DEV: u8 = 1;
+const PM_FUNC: u8 = 3;
 
 pub struct I440Fx {
     pic: Arc<LegacyPIC>,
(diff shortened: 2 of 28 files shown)


bhyve-api/src/structs.rs
```diff
@@ -271,6 +271,23 @@ pub struct vm_capability {
...
 
+#[repr(C)]
+#[derive(Copy, Clone, Default)]
+pub struct vm_suspend {
+    /// Acceptable values defined by `vm_suspend_how`
+    pub how: u32,
+}
+
+// bit definitions for `vm_reinit.flags`
+pub const VM_REINIT_F_FORCE_SUSPEND: u64 = 1 << 0;
```
luqmana, line 282: Not to pick on this addition, but we should try to make more use of bitstruct/bitflags in propolis in general.

propolis/src/hw/chipset/i440fx.rs
```diff
@@ -163,7 +164,19 @@ impl SelfArc for I440Fx {
...
+impl Entity for I440Fx {
+    fn state_transition(
+        &self,
+        next: instance::State,
+        target: Option<instance::State>,
+        ctx: &DispCtx,
+    ) {
+        // TODO: make it easier to do this automatically
+        let bus = self.pci_bus.lock().unwrap();
+        let pm = bus.device_at(1, 3).unwrap().as_devinst().unwrap();
```
luqmana, line 176: Could we use a constant or something for these instead of magic numbers

propolis/src/hw/chipset/i440fx.rs
```diff
@@ -652,6 +675,13 @@ impl Piix3PM {
             PmReg::Reserved => {}
         }
     }
+    fn reset(&self, ctx: &DispCtx) {
+        let mut regs = self.regs.lock().unwrap();
+        regs.reset();
+        // Make sure PM timer is attached to the right IO port
+        // TODO: error handling?
+        ctx.mctx.hdl().pmtmr_locate(PMBASE_DEFAULT + 0x8).unwrap();
```
luqmana, line 683: This one too

propolis/src/instance.rs
```diff
@@ -38,6 +38,7 @@ impl State {
             return true;
         }
         match (from, to) {
+            // Anything can set us on the road to destruction
```
luqmana, line 41: 😆

propolis/src/instance.rs
```diff
@@ -121,181 +168,283 @@ impl Instance {
...
+                inner.suspend_info = Some((kind, source));
+                self.set_target_state_locked(inner, State::Reset)
+            }
+            SuspendKind::Halt => {
+                if matches!(inner.suspend_info, Some((SuspendKind::Halt, _))) {
+                    return Ok(());
+                }
+                if inner.suspend_info.is_none() {
+                    let hdl = inner.machine.get_hdl();
+                    let _ = hdl.suspend(
```
luqmana, line 263: Why do we only issue the suspend here if there isn't any `suspend_info`?

pfmooney: Since these represent hard resets or poweroffs, we don't allow competing suspend requests to be issued.  Once an instance is progressing through a suspend, the bhyve kernel component will reject further requests until the instance undergoes a `VM_REINIT` operation.

propolis/src/hw/chipset/i440fx.rs
pfmooney: I do plan on a different method of access for these chipset owned devices, but until then, sure.

propolis/src/hw/chipset/i440fx.rs
pfmooney: Done

bhyve-api/src/structs.rs
pfmooney: Gotta start somewhere.  Done

propolis/src/inventory.rs
```diff
@@ -435,28 +391,19 @@ impl Record {
...
-    fn as_pci(&self) -> Option<&dyn Pci> {
-        None
-    }
-    fn as_chipset(&self) -> Option<&dyn Chipset> {
-        None
-    }
-    fn as_chardev(&self) -> Option<&dyn CharDev> {
-        None
+    #[allow(unused_variables)]
+    fn state_transition(
```
smklein, line 401: (definitely not blocking, but just for discussion)

The current expectation with this API is that we'l give entities a single blocking moment in time to react to state transitions, right?

Do we view it as limiting that the reaction to transitions is effectively serialized and synchronous? I'm wondering if at some point we'd want to be able to communicate to devices that an event is happening - like "migration start" - and let multiple devices begin acting on that event concurrently.

(Admittedly, this would also be possible with the current API, just with more states)

pfmooney:
Are there any other significant concerns or comments on this?  As I've been working on the tokio/async integration for propolis-the-library, it has overlapped with many of these areas, so I may just rebase on top of this if folks think it's acceptable.

luqmana (approved):
Looks good to me. 👍🏿 Should be fine to merge whenever the illumos changes get in.

pfmooney:
> Looks good to me. 👍🏿 Should be fine to merge whenever the illumos changes get in.

Thanks for the review.  I'll get the async work posted as soon as it's behaving.

propolis/usdt.d
```diff
@@ -13,4 +13,11 @@ provider propolis {
 
 	/* vm_exit(vcpuid, rip, code) */
 	probe vm_exit(uint32_t, uint64_t, uint32_t);
+
+	/* virtio_vq_notify(virtio_dev_addr, virtqueue_id) */
+	probe virtio_vq_notify(uint64_t, uint16_t);
+	/* virtio_vq_pop(vq_addr, avail_idx) */
+	probe virtio_vq_pop(uint64_t, uint16_t);
+	/* virtio_vq_pop(vq_addr, used_idx, used_len) */
```
luqmana, line 21:
```suggestion
	/* virtio_vq_push(vq_addr, used_idx, used_len) */
```

pfmooney:
With [14024](https://www.illumos.org/issues/14024) on the glide scope for integration (RTI sent) I'm planning to land this as well as #47 once that is integrated into illumos-gate.

pfmooney:
Integrated via cdf64cd90be5ff695ea0a33f23ad8ba8bb9440cf

PR #31, opened 2021-08-25
feature(asm) requires nightly, so add rust-toolchain.toml

jmpesp:


rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "nightly"
+components = ["rustc", "rustfmt"]
```
pfmooney, line 3: What are the ramifications of spelling out the covered components.  Should we bring the other `rust-toolchain` files in the repo in line with this?

jmpesp: I just pushed c536c095d3d28b5c83a044b4f398d76b10e019fc which specifies `profile = "default"` instead, in line with https://rust-lang.github.io/rustup/concepts/profiles.html.

I'll change the crucible rust-toolchain file to match this too.

dancrossnyc: Squash that commit down into this one? No need to push both commits.

jmpesp: Done

PR #32, opened 2021-08-25
only run integration tests on illumos based systems

jmpesp:


pfmooney (approved):
Is there anything we should do, restructuring-wise, to make this easier going forward? (So as the test count expands, it's not on the author to remember the `cfg()` exclusion?)

dancrossnyc (approved):

smklein:
> > Is there anything we should do, restructuring-wise, to make this easier going forward? (So as the test count expands, it's not on the author to remember the `cfg()` exclusion?)
> 
> Perhaps put the tests into a module that one can wrap in a `cfg`?

Seems reasonable to me - apologies for not including this when I wrote the integration tests, but they're definitely intended to only run on illumos!

jmpesp:
41b4d267885674bef32be6d761cd42055cd73911 creates a module and wraps it in the cfg instead.

PR #33, opened 2021-08-25
Get to a clean clippy output

jmpesp:
Apply clippy suggestions until they're all gone :)

### DIFF
--- a/propolis/src/hw/pci/device.rs
+++ b/propolis/src/hw/pci/device.rs
@@ -1096,7 +1096,7 @@ impl MsixCfg {
                         }
                         MsixCapReg::TableOff => {
                             // table always at offset 0 for now
-                            ro.write_u32(0 | self.bar as u8 as u32);
+                            ro.write_u32(self.bar as u8 as u32);
                         }
                         MsixCapReg::PbaOff => {
                             ro.write_u32(self.pba_off | self.bar as u8 as u32);
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -220,8 +220,7 @@ fn main() {
                     );
                     chipset.pci_attach(bdf.unwrap(), vioblk);
 
-                    plain
-                        .start_dispatch(format!("bdev-{} thread", name), &disp);
+                    plain.start_dispatch(format!("bdev-{} thread", name), disp);
                 }
                 "pci-virtio-viona" => {
                     let vnic_name =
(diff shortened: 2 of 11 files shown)


andrewjstone (approved):
LGTM. Thanks for the fixes @jmpesp

propolis/src/hw/pci/device.rs
```diff
@@ -1096,47 +1096,42 @@ impl MsixCfg {
...
+                            ro.write_u32(self.bar as u8 as u32);
                         }
                         MsixCapReg::PbaOff => {
                             ro.write_u32(self.pba_off | self.bar as u8 as u32);
                         }
                     }
                 }
                 RWOp::Write(wo) => {
-                    match id {
-                        MsixCapReg::MsgCtrl => {
```
pfmooney, line 1108: This was written as a match to be consistent with all of the other register handling

propolis/src/hw/uart/base.rs
```diff
@@ -98,10 +98,11 @@ impl Uart {
                 self.reg_intr_enable = data;
                 // Although not specified in the datasheet, some consumers expect a THRE
                 // interrupt to be raised when toggling that on in IER.
-                if old & IER_ETBEI == 0 && data & IER_ETBEI != 0 {
-                    if self.tx_fifo.is_empty() {
-                        self.thre_intr = true
-                    }
```
pfmooney, line 104: I think this is more readable as-is, since it separates the register value evaluation from how the device state factors into the behavior.  The compiler will optimize this into the same result anyways.

dancrossnyc:
> I think the clippy lints can be used as a guideline, but I am not targetting 100% clippy-clean for propolis. I do not want to litter the codebase with exclusions (like `len_without_empty`).

For whatever small amount it's worth, I've found being "clippy clean" to be very useful. Giving authority for those things over to a tool just eliminates entire categories of contention (similarly with formatting). If there are exceptions that are universally agreed upon, one can often wrap them up into crate-level defaults, to avoid pragmas at the call site.

> In particular, I'm not fond of some of its control-flow suggestions, when the code has been written a certain way for readability or consistency.

I actually kinda like some of the suggestions, which is, I suppose, an illustration of how these things can be subjective.

pfmooney:
> If there are exceptions that are universally agreed upon, one can often wrap them up into crate-level defaults, to avoid pragmas at the call site.

I'm certainly onboard with determining which lints we're going to ignore and excluding them at the top level.  `len_without_is_empty` and whatever insists on implementing `Default` come to mind.

> I actually kinda like some of the suggestions, which is, I suppose, an illustration of how these things can be subjective.

It makes suggestions I agree with (like around `ok_or` vs `ok_or_else`), but its changes to nested if statements really rubs me the wrong way.  I don't care how the compiler chooses to optimize that code, but if I separate them out, I'm doing so for a purpose.  It's true, that's subjective, but I'm not willing to cede that sort of decision to a tool yet.

smklein:
https://github.com/rust-lang/cargo/issues/5034 is still in progress, but it should make crate-wide lint toggling a little easier.

In the meantime, there are some options for using a "subset of clippy", if we want to remain clean:
- Module-wide allow statements, like `#![allow(clippy::style)]`
- Clippy command line arguments. We use `cargo clippy -- -D warnings -A clippy::style` in omicron, and have that verified by a bot - it helps for a lot of minor optimization / readability checks, but isn't nearly as picky as default clippy.

I know "where to draw the line on clippy" is subjective, but IMO reducing it to an agreeable subset is a good place to start - the positives outweigh the noise easily when pared down.

PR #35, opened 2021-08-27
Initialize backing stores separately from devices

jmpesp:
Initialize backing stores separately from devices, and refer to them by name in device init instead.

Before:

    [dev.block0]
    driver = "pci-virtio-block"
    disk = "/home/james/disk.img"
    pci-path = "0.4.0"

After:

    [backing_store.bootimg]
    type = "file"
    path = "/home/james/disk.img"

    [dev.block0]
    driver = "pci-virtio-block"
    backing_store = "bootimg"
    pci-path = "0.4.0"

The code initializing backing stores now lives in one place. An equivalent change should happen for the server code at some point too (in server/src/lib/server.rs and server/src/lib/initializer.rs).

### DIFF
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -11,6 +11,9 @@ struct Top {
 
     #[serde(default, rename = "dev")]
     devices: BTreeMap<String, Device>,
+
+    #[serde(default, rename = "backing_store")]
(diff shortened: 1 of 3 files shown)


jmpesp:
As discussed in our meeting, I'm going to try to merge the BlockDev and BackingStore traits instead.

PR #36, opened 2021-08-30
Put blockdev and backingstore back together again

jmpesp:
Instead of 3 layers:

    - device (virtio, nvme)
    - blockdev (plainbdev)
    - backing store (file, crucible, etc)

Fold into 2 layers:

    - device (virtio, nvme)
    - blockdev (plainbdev)
    
I've also committed a commented out test that would work if there was a mock / test Dispatcher.

### DIFF
--- a/propolis/src/block.rs
+++ b/propolis/src/block.rs
@@ -14,7 +14,7 @@ use crate::dispatch::{DispCtx, Dispatcher};
 use crate::vmm::{MemCtx, SubMapping};
 
 /// Type of operations which may be issued to a virtual block device.
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum BlockOp {
     Flush,
     Read,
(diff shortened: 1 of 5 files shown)


propolis/src/block.rs
```diff
@@ -59,140 +62,34 @@ pub trait BlockDev<R: BlockReq>: Send + Sync + 'static {
...
-
-    /// some backing stores may have block size requirements. optionally return a
-    /// block size in bytes.
-    fn block_size(&self) -> Option<usize>;
-}
-
-/// A block device implementation backed by an underlying file.
-pub struct FileBackingStore {
+/// Standard [`BlockDev`] implementation.
+pub struct PlainBdev<R: BlockReq> {
```
luqmana, line 66: Should we rename this now to indicate it is file backed? `FileBdev` or something. Then as we add others like `CruicibleBdev` it'll be consistent.

propolis/src/block.rs
```diff
@@ -217,102 +114,100 @@ impl<R: BlockReq, S: BackingStore> PlainBdev<R, S> {
...
-                    BlockResult::Success
+        bufs: Vec<GuestRegion>,
+    ) -> Result<BlockResult> {
+        let mappings: Vec<SubMapping> = bufs
+            .iter()
+            .map(|buf| {
+                if is_read {
+                    mem.writable_region(buf).unwrap()
+                } else {
+                    mem.readable_region(buf).unwrap()
```
luqmana, line 157: The unwraps aren't ideal. You could instead collect into a `Option<Vec<_>>` which would give back a None if any of the mappings were None.

jmpesp: nice, done in 11152cd

propolis/src/block.rs
jmpesp: done in e53f8a3

luqmana (approved):
Thanks James!

PR #37, opened 2021-08-31
Decouple blockdev init from device init

jmpesp:
Currently, virtio and nvme assume that a file will be the backing block device. This PR separates the block device config from the device config, allowing separate initialization.

Note that `start_dispatch` had to move to the BlockDev trait for the dynamic type stuff to work.

Also pulls out some config bits common to both the cli and the server.

This is a breaking API change from the perspective of the server's ensure instance request. Previously, the TOML config would look like:

```
[dev.block1]
driver = "pci-virtio-block"
disk = "alpine.img"
pci-path = "0.6.0"
```

Now:

```
[block_dev.alpine]
type = "file"
path = "/home/james/alpine.img"

[dev.block1]
driver = "pci-virtio-block"
block_dev = "alpine"
pci-path = "0.6.0"
```


jmpesp:
As discussed out of band, the server and cli config may deviate further in the future, so this PR shouldn't make them common.

jmpesp:
Note this may not be a breaking API change after all - in the server, block devices are initialized out of `server_context.config`, which is provided via a cfg file when the server boots.

server/src/lib/config.rs
```diff
@@ -47,6 +52,14 @@ impl Config {
     pub fn devs(&self) -> IterDevs {
         IterDevs { inner: self.devices.iter() }
     }
+
+    pub fn block_dev<R: propolis::block::BlockReq>(
+        &self,
+        name: &str,
+    ) -> Arc<dyn propolis::block::BlockDev<R>> {
+        let entry = self.block_devs.get(name).unwrap();
```
smklein, line 60: I'm willing to allow slightly more tolerance in the CLI tools for unwrap - because they're for testing anyway - but for anything in the server, I'd really like to be cautious about unwraps. 

Both here and in `BlockDevice::block_dev()` below, there are many conditions in which we could unwrap and panic - can we propagate the result upwards instead?

server/src/lib/config.rs
```diff
@@ -69,11 +82,43 @@ impl Device {
...
+pub struct BlockDevice {
+    #[serde(default, rename = "type")]
+    pub bdtype: String,
+
+    #[serde(flatten, default)]
+    pub options: BTreeMap<String, toml::Value>,
+}
+
+impl BlockDevice {
+    pub fn block_dev<R: propolis::block::BlockReq>(
```
smklein, line 95: IMO it's weird to have a function named "block device" on a structure named "block device" - it's not intuitive what this does based on name.

Can we call this `create_block_device`, perhaps?

smklein (approved):

cli/src/config.rs
```diff
@@ -48,10 +45,60 @@ impl Config {
...
+            }
+        }
+    }
+}
+
+/// Iterator returned from [`Config::devs`] which allows iteration over
+/// all [`Device`] objects.
 pub struct IterDevs<'a> {
-    inner: btree_map::Iter<'a, String, Device>,
+    pub inner: btree_map::Iter<'a, String, Device>,
```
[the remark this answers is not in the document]
jmpesp: it doesn't, no - fixed in 08eb899

server/src/lib/config.rs
```diff
@@ -69,11 +82,43 @@ impl Device {
...
+            }
+        }
+    }
+}
+
 /// Iterator returned from [`Config::devs`] which allows iteration over
 /// all [`Device`] objects.
 pub struct IterDevs<'a> {
-    inner: btree_map::Iter<'a, String, Device>,
+    pub inner: btree_map::Iter<'a, String, Device>,
```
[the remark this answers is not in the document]
jmpesp: it doesn't, no - fixed in 08eb899

server/src/lib/config.rs
jmpesp: agreed, fixed in 9036f7a

server/src/lib/server.rs
```diff
@@ -265,22 +265,33 @@ async fn instance_ensure(
...
-                                ErrorKind::InvalidData,
-                                "Cannot parse disk path",
-                            )
-                        })?;
+                        let block_dev_name = dev
+                            .options
+                            .get("block_dev")
+                            .unwrap()
+                            .as_str()
+                            .unwrap();
```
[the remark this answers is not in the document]
jmpesp: agreed, fixed in 84c04b9

server/src/lib/config.rs
jmpesp: agreed, fixed in e3b4ea9

smklein (approved):
LGTM - thanks!

PR #38, opened 2021-09-01
nvme: zero out offset of prp2 if in prp list

jmpesp:
prp2 points to a prp list if the size of the operation is large enough.

In the 1.0e spec, prp2 "may also have a non-zero offset within the memory page", but it also says "If this entry is in the PRP List then the Offset portion of this field shall be cleared to 0h".

This commit clears the offset, fixing errors seen when using NVMe emulation.

### DIFF
--- a/propolis/src/hw/nvme/cmds.rs
+++ b/propolis/src/hw/nvme/cmds.rs
@@ -613,7 +613,7 @@ impl PrpIter<'_> {
                      * within the command, shall be Qword aligned and may also have a non-zero offset within
                      * the memory page.
                      */
-                    if ((self.prp2 % 8) != 0) {
+                    if (self.prp2 % 8) != 0 {
                         return Err("PRP2 not Qword aligned!");
                     }
 
@@ -626,10 +626,7 @@ impl PrpIter<'_> {
                      * then the Offset portion of this field shall be cleared to 0h.
                      */
                     let idx = (self.prp2 & PAGE_OFFSET as u64) / 8;
-                    PrpNext::List(
-                        self.prp2 & (PAGE_MASK as u64),
-                        idx as u16,
-                    )
+                    PrpNext::List(self.prp2 & (PAGE_MASK as u64), idx as u16)
                 };
                 (self.prp1, size, next)
             }
--- a/propolis/src/hw/nvme/cmds.rs
+++ b/propolis/src/hw/nvme/cmds.rs
@@ -537,7 +537,7 @@ enum PrpNext {
     Done,
 }
 
-/// The max number of entries in a single Physical Region Page (PRP) List
+/// The last valid PRP list entry index in a single Physical Region Page (PRP) List
 ///
 /// 512 64-bit entries in a PRP list
 /// Note this relies on a Page Size of 4k (2^12).
@@ -545,7 +545,7 @@ enum PrpNext {
 ///     - 2 because PRP entries are expected to be 32-bit aligned.
 ///
 /// See NVMe 1.0e Section 4.3 Physical Region Page Entry and List
-const PRP_LIST_MAX: u16 = 511; // XXX: 512?
+const PRP_LIST_MAX: u16 = 511;
 
 /// A helper object for iterator over a single, 2 or a list of PRPs.
 pub struct PrpIter<'a> {
@@ -591,8 +591,8 @@ impl PrpIter<'_> {
         assert!(self.error.is_none());
 
         // PRP Entry Layout
-        // | 63                                      n + 1 | n . . . . . . . 2 1 0 |
-        // |         page base address                     |      offset     | 0 0 |
+        // | 63 . . . . . . . . . . . . . . . . . . . n + 1 | n . . . . . . 2 | 1 0 |
+        // |         page base address                      |      offset     | 0 0 |
         let (addr, size, next) = match self.next {
             PrpNext::Prp1 => {
                 // The first PRP entry contained within the command may have a non-zero offset
@@ -613,7 +613,9 @@ impl PrpIter<'_> {
                      * within the command, shall be Qword aligned and may also have a non-zero offset within
                      * the memory page.
                      */
-                    assert!((self.prp2 % 8) == 0);
+                    if ((self.prp2 % 8) != 0) {
+                        return Err("PRP2 not Qword aligned!");
+                    }
 
                     /*
                      * If this entry is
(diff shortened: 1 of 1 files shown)


propolis/src/hw/nvme/cmds.rs
```diff
@@ -545,14 +545,15 @@ enum PrpNext {
 ///     - 2 because PRP entries are expected to be 32-bit aligned.
 ///
 /// See NVMe 1.0e Section 4.3 Physical Region Page Entry and List
-const PRP_LIST_MAX: u16 = 511;
+const PRP_LIST_MAX: u16 = 511; // XXX: 512?
```
luqmana, line 548: Hmm, as written you're right it is a little confusing. This is more accurately the last valid PRP list entry index (which start at 0). Maybe it'd be better to set this to 512 then just `- 1` as necessary below.

propolis/src/hw/nvme/cmds.rs
```diff
@@ -601,8 +607,27 @@ impl PrpIter<'_> {
...
+                     * If this entry is
+                     *
+                     * - not the first PRP entry in the command, or
+                     * - in the PRP List
+                     *
+                     * then the Offset portion of this field shall be cleared to 0h.
+                     */
+                    let idx = (self.prp2 & PAGE_OFFSET as u64) / 8;
+                    PrpNext::List(
+                        self.prp2 - (self.prp2 & PAGE_OFFSET as u64),
```
luqmana, line 628:
```suggestion
                        self.prp2 & (PAGE_MASK as u64),
```

propolis/src/hw/nvme/cmds.rs
```diff
@@ -589,8 +590,13 @@ impl PrpIter<'_> {
         assert!(self.remain > 0);
         assert!(self.error.is_none());
 
+        // PRP Entry Layout
+        // | 63                                      n + 1 | n . . . . . . . 2 1 0 |
+        // |         page base address                     |      offset     | 0 0 |
```
luqmana, line 595:
```suggestion
        // | 63                                      n + 1 | n . . . . . . 2 | 1 0 |
        // |         page base address                     |      offset     | 0 0 |
```

Small nit: just lining it up

luqmana (approved):
Thanks @jmpesp, nice catch!

propolis/src/hw/nvme/cmds.rs
```diff
@@ -601,8 +607,27 @@ impl PrpIter<'_> {
...
                 } else {
-                    let list_off = (self.prp2 & PAGE_OFFSET as u64) / 8;
-                    PrpNext::List(self.prp2, list_off as u16)
+                    /*
+                     * The first PRP List entry (i.e. the first pointer to a memory page containing
+                     * additional PRP entries) that if present is contained in the PRP Entry 2 location
+                     * within the command, shall be Qword aligned and may also have a non-zero offset within
+                     * the memory page.
+                     */
+                    assert!((self.prp2 % 8) == 0);
```
pfmooney, line 616: If this is a value controlled by the guest, then `assert`ing if it is not as-expected is rather dangerous, as broken behavior on their part leads to an immediate reset (propolis crash)

jmpesp: Right, addressed in eb141578c3763d1e1ed035fe307159b20ae8e5fe

propolis/src/hw/nvme/cmds.rs
jmpesp: I clarified the comment in eb141578c3763d1e1ed035fe307159b20ae8e5fe

propolis/src/hw/nvme/cmds.rs
jmpesp: suggestion done in eb141578c3763d1e1ed035fe307159b20ae8e5fe

propolis/src/hw/nvme/cmds.rs
jmpesp: suggestion done in eb141578c3763d1e1ed035fe307159b20ae8e5fe

propolis/src/hw/nvme/cmds.rs
```diff
@@ -601,8 +607,26 @@ impl PrpIter<'_> {
...
+                    /*
+                     * If this entry is
+                     *
+                     * - not the first PRP entry in the command, or
+                     * - in the PRP List
+                     *
+                     * then the Offset portion of this field shall be cleared to 0h.
+                     */
+                    let idx = (self.prp2 & PAGE_OFFSET as u64) / 8;
+                    PrpNext::List(self.prp2 & (PAGE_MASK as u64), idx as u16)
```
pfmooney, line 629: Is this constraint of the list base address proper?  I believe the base pointer to the PRP list can be offset within the page:

> Dependent on the command definition, the first PRP entry contained within the command may have a non-
zero  offset  within  the  memory  page.    The  first  PRP  List  entry  (i.e.  the  first  pointer  to  a  memory  page 
containing additional PRP entries) that if present is typically contained in the PRP Entry 2 location within 
the command, shall be Qword aligned and may also have a non-zero offset within the memory page.

jmpesp: I'm not 100% sure.

The spec says what you quoted, and at the *same time* says (4.3, figure 10) that

> If this entry is not the first PRP entry in the command or in the PRP List then the Offset portion of this field shall be cleared to 0h.

I interpreted this to mean the following

- that prp2 can always come in with a non-zero offset
  - if not pointing to a list, then this is an offset into a page
  - if pointing to a list, that offset is interpreted as an index
- if prp2 is pointing to a list, then the offset should be zeroed.

I know that the tests I was running to validate this interpretation (https://fight-flash-fraud.readthedocs.io/en/latest/ on alpine, and separately an openbsd install) would report corruption (or panic) without this patch, and would work with it. That's not great justification but I believe it validates the interpretation.

pfmooney: When the PRP2 value is referring to a PRP List, I believe the value itself does not count as a "PRP entry" (since it's not pointing to memory to be consumed in the operation).  Looking at the nvme emulation in both qemu and the C bhyve userspace, neither is masking PRP2 like that when using it to locate the PRP List.

jmpesp: The reason this works is because we're constructing an enum:

```
                     let idx = (self.prp2 & PAGE_OFFSET as u64) / 8;
                     PrpNext::List(self.prp2 & (PAGE_MASK as u64), idx as u16)
```
that we later pull apart and use like this:
```
             PrpNext::List(base, idx) => {
                 assert!(idx <= PRP_LIST_MAX);
                 let entry_addr = base + (idx as u64) * 8;
```

If we don't mask off prp2, then base contains the offset, and idx adds the offset to that, pointing to an invalid memory location.

tl;dr my interpretation of the spec was wrong, but the fix was right :) I'll make a PR correcting the comments.

PR #41, opened 2021-09-03
nvme: correct comments from PR #38

jmpesp:
The fix from #38 was good but I interpreted the spec incorrectly. Correct the comments to explain how the fix actually works.

### DIFF
--- a/propolis/src/hw/nvme/cmds.rs
+++ b/propolis/src/hw/nvme/cmds.rs
@@ -607,19 +607,34 @@ impl PrpIter<'_> {
                     // entry which should be present in PRP2
                     PrpNext::Prp2
                 } else {
-                    // The first PRP List entry (i.e. the first pointer to a memory page containing
-                    // additional PRP entries) that if present is contained in the PRP Entry 2
-                    // location within the command, shall be Qword aligned and may also have a
-                    // non-zero offset within the memory page.
+                    // If the remaining length is larger than the page size, PRP2 points to a list.
+                    //
+                    // The first PRP List entry:
+                    // - shall be Qword aligned, and
+                    // - may also have a non-zero offset within the memory page.
                     if (self.prp2 % 8) != 0 {
                         return Err("PRP2 not Qword aligned!");
                     }
 
-                    // PRP2 is allowed to have a non-zero offset into the memory page, but mask it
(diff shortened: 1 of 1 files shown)


luqmana (approved):

propolis/src/hw/nvme/cmds.rs
```diff
@@ -607,24 +607,17 @@ impl PrpIter<'_> {
...
-                    /*
-                     * The first PRP List entry (i.e. the first pointer to a memory page containing
-                     * additional PRP entries) that if present is contained in the PRP Entry 2 location
-                     * within the command, shall be Qword aligned and may also have a non-zero offset within
-                     * the memory page.
-                     */
+                    // The first PRP List entry (i.e. the first pointer to a memory page containing
+                    // additional PRP entries) that if present is contained in the PRP Entry 2
+                    // location within the command, shall be Qword aligned and may also have a
+                    // non-zero offset within the memory page.
```
pfmooney, line 613: The phrasing on this is a bit awkward.  It seems like it needs at least a comma, and perhaps be split up.

propolis/src/hw/nvme/cmds.rs
```diff
@@ -607,24 +607,17 @@ impl PrpIter<'_> {
...
-                     * If this entry is
-                     *
-                     * - not the first PRP entry in the command, or
-                     * - in the PRP List
-                     *
-                     * then the Offset portion of this field shall be cleared to 0h.
-                     */
+                    // PRP2 is allowed to have a non-zero offset into the memory page, but mask it
+                    // out when creating the base part of the PrpNext::List enum - the offset will
+                    // be passed with idx.
```
pfmooney, line 620: The phrasing is confusing here too.  "Allowed to have a non-zero offset, but mask it out" seems almost like it's being discarded.  Perhaps make it clear that the list can start at an offset in the page, but must not cross the page boundary, and that any offset is preserved to index into that region.

pfmooney (approved):
Thanks for making those clearer.

One minor style request: Make the commit message more explanatory about the change, rather than referencing the previous PR - something like "nvme: clarify PrpIter behavior for PRP Lists " or whatever

jmpesp:
:+1:

PR #45, opened 2021-09-07
NVMe: Use bitstruct for defining/manipulating controller "registers" and other structures.

luqmana:
Pulled in the [`bitstruct`](https://github.com/dancrossnyc/rust-bitstruct) crate to define the controller register structures. Now instead of one-off fields we've added in an ad-hoc way, we can manipulate the corresponding registers in a way that follows directly from the spec. Makes manipulating the parts we need a lot more clear than twiddling the individual bits. Also fixed a few spots where I got it wrong previously.

As a new example, the last commit also adds a first stab at handling shutdown notifications (aka linux won't hang a bit anymore during shutdown as the driver fruitlessly spun waiting for us to say we finished with shutdown processing).

### DIFF
--- a/propolis/src/hw/nvme/mod.rs
+++ b/propolis/src/hw/nvme/mod.rs
@@ -1,5 +1,6 @@
 use std::collections::BTreeMap;
 use std::convert::TryInto;
+use std::mem;
 use std::sync::{Arc, Mutex, MutexGuard};
 
 use crate::common::*;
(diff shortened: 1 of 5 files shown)


propolis/src/hw/nvme/mod.rs
```diff
@@ -316,13 +305,73 @@ impl PciNvme {
             ..Default::default()
         });
 
+        // We have unit tests that these are 16 and 64 bytes, respectively
+        // But just make sure as we specify these as powers of 2 in places
+        debug_assert_eq!(mem::size_of::<RawCompletion>().count_ones(), 1);
+        debug_assert_eq!(mem::size_of::<RawSubmission>().count_ones(), 1);
```
pfmooney, line 311: style nit: this could assert `is_power_of_two()` instead of `count_ones()`, right?

propolis/src/hw/nvme/mod.rs
```diff
@@ -1,5 +1,6 @@
 use std::collections::BTreeMap;
 use std::convert::TryInto;
+use std::mem;
```
pfmooney, line 3: style nit: maybe just import `size_of` directly, since that's the only thing being used (and is unambiguous in the file)

pfmooney (commented):
Looks like some nice clean-up.

pfmooney (approved):

dancrossnyc:
I love to see bitstruct used here!

dancrossnyc (approved):

PR #47, opened 2021-09-09
 Integrate tokio runtime into propolis dispatcher

pfmooney:
Building off the back of the lifecycle work (#30), this integrates a tokio runtime into the propolis dispatcher, and makes async tasks with access to machine resources (through a `DispCtx`) a first class notion.

The viona driver is the first test case for its use, but the Block interface(s) are presumably next in line to use the functionality.


pfmooney:
If this were gerrit, it would have only the latest commit, but I'm not versed enough in GH these days to know how to do that (or if it's even possible)

README.md
```diff
@@ -10,8 +10,8 @@ interface presented by the bhyve kernel component, running on recent illumos
 bits is required.
 
 At a minimum, the build must include the fix for
-[13275](https://www.illumos.org/issues/13275). (Present since commit
-[2606939](https://github.com/illumos/illumos-gate/commit/2606939d92dd3044a9851b2930ebf533c3c03892))
+[14024](https://www.illumos.org/issues/14024). (Present since commit
+[XXXXXXX](https://github.com/illumos/illumos-gate/commit/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX))
```
smklein, line 14: Should this be XXX'd out?

propolis/src/dispatch/sync_tasks.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// Spawns a new dedicated worker thread named `name` which invokes
+    /// `func` on `data`.
+    ///
+    /// An optional `wake` function may be supplied, when invoked, this
+    /// function should trigger the worker to move to a barrier point.
+    pub fn spawn(
+        &self,
+        name: String,
+        func: Box<SyncFn>,
+        wake: Option<Box<WakeFn>>,
```
smklein, line 99: Dunno how much you care - but for these functions operating on `Box<FunctionType>`, you could pretty easily use function-local generics and avoid boxing.

If we used the following:

```rust
pub type SyncFn = FnOnce(&mut SyncCtx) + Send + 'static;
```
We'd already be specifying the `Send` and `'static` constraints necessary to move these functions around.

https://doc.rust-lang.org/rust-by-example/fn/closures/input_parameters.html

propolis/src/dispatch/async_tasks.rs
```diff
@@ -0,0 +1,168 @@
...
+impl AsyncInner {
+    fn next_id(&mut self) -> AsyncTaskId {
+        self.next_id += 1;
+        AsyncTaskId { val: self.next_id }
+    }
+}
+
+pub(super) struct AsyncDispatch {
+    inner: Mutex<AsyncInner>,
+    rt: Mutex<Option<Runtime>>,
```
smklein, line 39: 1. Why let this be optional? It seems like it should always exist, if I understand the constructor usage correctly
2. Would that avoid the need for the `Mutex`? Most `Runtime` methods act on `&self`, not `&mut self`.

propolis/src/hw/virtio/viona.rs
```diff
@@ -368,6 +362,104 @@ impl VionaHdl {
...
+// This is an ugly hack to work around tokio's inability to poll for event
+// readiness on states other than POLLIN/POLLOUT, since viona communicates
+// changes to in-kernel ring interrupt state with POLLRDBAND.  In the short
+// term, we can translate that to POLLIN using nested epoll.  The viona fd is
+// added to an epoll handle, subscribing to EPOLLRDBAND.  When that condition is
+// met for the device, epoll will generate an event, making the epoll fd itself
+// readable.  We can subscribe to that using the normal tokio event system.
+//
+// In the long term, viona should probably move to something like eventfd to
+// make polling on those ring interrupt events more accessible.
```
smklein, line 374: What about patching mio?

https://github.com/tokio-rs/mio/blob/master/src/sys/unix/selector/epoll.rs seems to consider "readable" to be "EPOLLIN" or "EPOLLRDHUP" - why wouldn't `EPOLLRDBAND` be part of this too?

propolis/src/instance.rs
```diff
@@ -408,6 +408,11 @@ impl Instance {
 
             next_state = state.next_transition(inner.state_target);
         }
+
+        // Explicitly drop the instance-held reference to the Machine.  If all
+        // the worker tasks (sync or async) are complete, this should be the
+        // last holder and result in its destruction.
+        let _ = inner.machine.take();
```
smklein, line 415: Why do we need to do this explicitly? Forcing all access to the machine to go through an extra `unwrap` is kinda gross - wouldn't this happen implicitly as the last reference to the instance goes out of scope?

README.md
pfmooney: It's is here, as it is in #30, because that change has not gone upstream yet, so I don't have the relevant commit id.

propolis/src/dispatch/async_tasks.rs
pfmooney: I ran into issues where the Runtime was dropped from an async context in propolis-server (When the `Instance` -> `Dispatcher` -> `Runtime` drop chain occurred).  That's why it's in an option so I can drop it separately from a known sync context.

propolis/src/instance.rs
pfmooney: This is a bit complicated (and perhaps would benefit from a longer comment) : I need the `Machine` to get dropped before the `Instance`, so the `mmap()` areas it owns are `munmap()`ed prior to the `Instance` doing a destroy of the vmm handle.

luqmana:
> If this were gerrit, it would have only the latest commit, but I'm not versed enough in GH these days to know how to do that (or if it's even possible)

You could change the base branch from master to lifecycle and that should give just the changes on top.

propolis/src/hw/virtio/viona.rs
pfmooney: I'm a bit grumpy that there's no way to actually differentiate them as a consumer of `mio`.  My eventual plan is to change how viona does these notifications.  `POLLRDBAND` is rather awkward.

propolis/src/dispatch/sync_tasks.rs
pfmooney: I wasn't wild about monomorphizing the innards of the dispatcher over every possible consumer.  The boxing wasn't a concern (at least from a perf standpoint) since a thread is being spawned for this anyways.  The async equivalent to this had to be generic (rather than `Box<dyn>`) because I simply couldn't figure out type machinations to accomplish it that pleased the compiler (or that didn't rely on the unstable async callback feature)

propolis/src/dispatch/mod.rs
```diff
@@ -50,361 +35,156 @@ impl Dispatcher {
...
-    ///
-    /// Panics if the dispatcher has already been associated with
-    /// an instance.
-    pub(crate) fn assoc_instance(&self, inst: Weak<instance::Instance>) {
-        let res = self.inner.inst.lock().unwrap().replace(inst);
-        assert!(res.is_none());
     }
 
     /// Spawns a new dedicated worker thread named `name` which invokes
     /// `func` on `data`.
```
luqmana, line 66: Nit: update comment as we don't take a `data` anymore

propolis/src/chardev/sock.rs
```diff
@@ -1,137 +1,117 @@
...
-    fn drive(&mut self) {
-        if let Some(sink) = self.sink.as_ref() {
-            while let Some(b) = self.buf.pop_front() {
-                if !sink.sink_write(b) {
-                    self.buf.push_front(b);
-                    break;
+        let sink = sink.unwrap();
+        loop {
+            let mut buf = [0u8];
+            loop {
```
luqmana, line 31: Since we're already in an async context, can't we replace the nested loop with a `AsyncReadExt::read` call:

```rust
use tokio::io::AsyncReadExt;
...
if client.read(&mut buf).await? == 0 {
    return Ok(());
}
```

luqmana: Ah, that needs &mut to client, right.

pfmooney: I'm iterating on changes to the chardev drivers to dogfood this async stuff.  Hopefully it'll address your thoughts there.

pfmooney:
The latest commit will probably go in separately, but it's a continuation of my dog-fooding the async interfaces to see how well/poorly they work for device emulation.  Feedback welcome!

pfmooney:
Merged up through 4022decc52832e7c31462f493101adebd9b181a7

PR #49, opened 2021-09-24
Add propolis-server-cli tool to more easily manipulate propolis-server instances.

luqmana:
This adds a new bin crate (`propolis-server-cli`) meant to ease interacting with `propolis-server` so you don't need to keep batch/curl scripts handy.

Usage:
```
propolis-server-cli 0.1.0
A simple CLI tool to manipulate propolis-server

USAGE:
    propolis-server-cli [FLAGS] [OPTIONS] --server <server> <SUBCOMMAND>

FLAGS:
    -d, --debug      Enable debugging
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -p, --port <port>        propolis-server port [default: 12400]
    -s, --server <server>    propolis-server address

SUBCOMMANDS:
    get       Get the properties of a propolis instance
    help      Prints this message or the help of the given subcommand(s)
    new       Create a new propolis instance
    serial    Drop to a Serial console connected to the instance
    state     Transition the instance to a new state
```

### DIFF
--- a/server/src/lib/server.rs
+++ b/server/src/lib/server.rs
@@ -25,6 +25,7 @@ use tokio_tungstenite::tungstenite::{
     self, handshake, protocol::Role, Message,
 };
 use tokio_tungstenite::WebSocketStream;
+use uuid::Uuid;
 
 use propolis::dispatch::DispCtx;
 use propolis::hw::chipset::Chipset;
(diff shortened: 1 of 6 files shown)


server/src/lib/server.rs
```diff
@@ -361,6 +362,32 @@ async fn instance_ensure(
     Ok(HttpResponseCreated(api::InstanceEnsureResponse {}))
 }
 
+#[endpoint {
+    method = GET,
+    path = "/instances/{instance_id}/uuid",
+    unpublished = true,
```
luqmana, line 368: cc @smklein had to add this endpoint to make the tool a bit more ergonomic to allow for interacting with instances by name instead of uuid.

smklein: Seems reasonable; this is trivial to maintain in the server so that tradeoff seems worth it.

PR #51, opened 2021-10-03
Update README to reflect current state of affairs.

luqmana:
Update the example TOML configs as well as mentioning the different crates (propolis, cli, standalone).


smklein (approved):
Thanks for updating this!

PR #56, opened 2021-10-12
Add Crucible as a block backend

jmpesp:
Implement block::Backend for new CrucibleBackend type. Copy Driver from
the FileBackend code, with the intention in the future of submitting
work asynchronously.

Add Crucible as optional - it's not necessary to compile with it.

Note dispatcher spawned task names have to be globally unique! Added
a print of {:?} to FileBackend code.


propolis/src/block/crucible.rs
```diff
@@ -0,0 +1,341 @@
...
+        CrucibleBackend::_create(disp, targets, read_only, key)
+            .map_err(map_crucible_error_to_io)
+    }
+
+    fn _create(
+        disp: &Dispatcher,
+        targets: Vec<SocketAddrV4>,
+        read_only: bool,
+        key: Option<String>,
+    ) -> anyhow::Result<Arc<Self>, crucible::CrucibleError> {
```
pfmooney, line 50: Per the [rust crate zoo](https://github.com/oxidecomputer/meta/blob/master/engineering/rust-crate-zoo.adoc#anyhow) I'd been lead to believe that `thiserror` was generally to be used in libraries while `anyhow` was for applications.  Is crucible planning to follow that pattern? (Here, propolis is fulfilling the library role, so I'd expect us to use `thiserror` once we start working on the whole error situation)

propolis/src/block/crucible.rs
```diff
@@ -0,0 +1,341 @@
...
+            if let Some(req) = guard.pop_front() {
+                drop(guard);
+                let ctx = sctx.dispctx();
+                match process_request(self.guest.clone(), &req, &ctx) {
+                    Ok(_) => req.complete(block::Result::Success, &ctx),
+                    Err(_) => req.complete(block::Result::Failure, &ctx),
+                }
+            } else {
+                // wait until more requests are available
+                self.idle_threads.add_permits(1);
```
pfmooney, line 163: As I found with some other code I was working on, this needs a guard, as a quiesce request can cause multiple trips over the wait condition, causing permits in excess of the thread count to accrue.

server/src/lib/config.rs
```diff
@@ -144,6 +148,77 @@ impl BlockDevice {
...
+                            .to_string()
+                            .parse()
+                            .map_err(|e| {
+                                ParseError::AsError(
+                                    "x".to_string(),
+                                    format!("parse(): {:?}", e),
+                                )
+                            })
+                    })
+                    .collect::<Result<Vec<SocketAddrV4>, ParseError>>()?;
```
pfmooney, line 187: Not a problem for right now, but boy-howdy is this verbose.  Let's put our heads together on how we might improve the ergonomics for config stuff down the line, since these cases will only become more complex.

propolis/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-libc = "0.2"
+libc = "0.2.103"
```
luqmana, line 11: Curious, was there something specific in .103 that was added?

propolis/Cargo.toml
```diff
@@ -21,6 +21,11 @@ viona_api = { path = "../viona-api" }
 usdt = "0.1.11"
 tokio = { version = "1", features = ["full"] }
 futures = "0.3"
+crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "adb2662b", optional = true }
```
luqmana, line 24: Would it make sense to have a propolis branch or similar on the crucible side?

propolis/Cargo.toml
pfmooney: We should have specified the patch version when the preadv/pwritev stuff was added.  Otherwise local cargo lock files would potentially cause build issues (at least while we're not checking in Cargo.lock)

server/src/lib/config.rs
```diff
@@ -144,6 +148,77 @@ impl BlockDevice {
...
+                    .iter()
+                    .map(|x| {
+                        x.as_str()
+                            .ok_or_else(|| {
+                                ParseError::AsError(
+                                    "x".to_string(),
+                                    "as_str".to_string(),
+                                )
+                            })?
+                            .to_string()
```
luqmana, line 178: I think you can just omit this (here and a few other places below)? `parse` is defined on `&str` anyways.

propolis/Cargo.toml
jmpesp: It wouldn't compile without 0.2.102 explicitly set (0.2 didn't work), and I bumped it when I saw the latest

propolis/Cargo.toml
jmpesp: I'd rather it go off a rev of main

propolis/src/block/crucible.rs
pfmooney: Sounds good.  In the mean time, I don't think you need to hold this PR up for oxidecomputer/crucible#85.

propolis/Cargo.toml
```diff
@@ -21,6 +21,11 @@ viona_api = { path = "../viona-api" }
...
+anyhow = "1"
```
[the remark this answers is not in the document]
jmpesp: I'm only including it for crucible Guest return types in _create - it will be removed after https://github.com/oxidecomputer/crucible/issues/85 is addressed.

server/src/lib/config.rs
```diff
@@ -144,6 +148,76 @@ impl BlockDevice {
...
+                                ParseError::AsError(
+                                    "x".to_string(),
+                                    "as_str".to_string(),
+                                )
+                            })?
+                            .to_string())
+                    });
+
+                let key: Option<String> =
+                    if let Some(key) = key { Some(key?) } else { None };
```
luqmana, line 208: You can replace this by adding `.map_or(Ok(None), |r| r.map(Some))?;` to the above statement.

propolis/src/block/crucible.rs
jmpesp: I think it has to do with the wake function not resetting (or otherwise interacting) with the idle_threads semaphore, though I'm not sure what the right action to take is. I didn't have time to go further into this today.

propolis/src/block/crucible.rs
```diff
@@ -0,0 +1,341 @@
...
+                mapping.len(),
+            );
+        }
+
+        nread += mapping.len();
+    }
+
+    let offset = guest.byte_offset_to_block(offset)?;
+
+    let mut waiter = guest.write(offset, crucible::Bytes::from(vec))?;
```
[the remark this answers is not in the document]
pfmooney: Seems reasonable to circle back on some of this perf-related stuff after initial integration

propolis/src/block/file.rs
```diff
@@ -173,7 +173,7 @@ impl Driver {
 
             let _ = disp
                 .spawn_sync(
-                    format!("file bdev worker {}", i),
+                    format!("file {:?} bdev worker {}", self.fp, i),
```
pfmooney, line 176: This is no longer necessary after bf46fe953310b93e06df8f1948cb5d5496b6a2cf, but it's probably still good to have distinct names.  The `Debug` output of the `File` might get a little longer for posix_setname, though.  Have you checked to make sure this is fine?

rmustacc: FYI, the limit varies per operating system (as does truncation behavior). The limit on helios is 32. Given that's our primary deployment vector you may want to optimize for that rather than the lowest common denominator (but you'll want to verify truncation behavior on other platforms through that interface).

propolis/src/block/crucible.rs
```diff
@@ -0,0 +1,353 @@
...
+        };
+
+        // XXX Crucible uses std::sync::mpsc::Receiver, not
+        // tokio::sync::mpsc::Receiver, so use tokio::task::block_in_place here.
+        // Remove that when Crucible changes over to the tokio mpsc.
+
+        // After up_main has returned successfully, wait for active negotiation
+        let uuid = tokio::task::block_in_place(|| guest.query_upstairs_uuid())?;
+
+        println!("Calling activate for {:?}", uuid);
```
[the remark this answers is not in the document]
jmpesp: Done in dd0bd517b0738108a4bbf590af020e2d2b6344ac.

pfmooney (approved):
Let's get this landed so more integrated testing is possible.  It'll be easier to iterate from there.

PR #65, opened 2021-11-01
Flatten virtio emulation structure

pfmooney:
This flattens the data structures involved in emulating virtio devices.  The intent is to make migration of state easier by providing access to all data associated with a given device, rather than having it be hidden inside "inner" types.

### DIFF
--- a/server/src/lib/initializer.rs
+++ b/server/src/lib/initializer.rs
@@ -205,15 +205,14 @@ impl<'a> MachineInitializer<'a> {
         be_register: ChildRegister,
     ) -> Result<(), Error> {
         let be_info = backend.info();
-        let vioblk = virtio::VirtioBlock::create(0x100, be_info);
+        let vioblk = virtio::PciVirtioBlock::new(0x100, be_info);
         let id = self
             .inv
             .register(&vioblk, format!("vioblk-{}", bdf), None)
             .map_err(|e| -> std::io::Error { e.into() })?;
         let _ = self.inv.register_child(be_register, id).unwrap();
 
-        let blk = vioblk.inner_dev::<virtio::block::VirtioBlock>();
-        backend.attach(blk, self.disp);
+        backend.attach(vioblk.clone(), self.disp);
         chipset.device().pci_attach(bdf, vioblk);
 
         Ok(())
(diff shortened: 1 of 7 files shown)


server/src/lib/initializer.rs
```diff
@@ -205,15 +205,14 @@ impl<'a> MachineInitializer<'a> {
...
         let be_info = backend.info();
-        let vioblk = virtio::VirtioBlock::create(0x100, be_info);
+        let vioblk = virtio::PciVirtioBlock::new(0x100, be_info);
         let id = self
             .inv
             .register(&vioblk, format!("vioblk-{}", bdf), None)
             .map_err(|e| -> std::io::Error { e.into() })?;
         let _ = self.inv.register_child(be_register, id).unwrap();
 
-        let blk = vioblk.inner_dev::<virtio::block::VirtioBlock>();
```
luqmana, line 215: 🎉

standalone/src/main.rs
```diff
@@ -212,28 +212,26 @@ fn main() {
...
                     let vnic_name =
                         dev.options.get("vnic").unwrap().as_str().unwrap();
 
-                    let viona = hw::virtio::viona::VirtioViona::create(
-                        vnic_name, 0x100, &hdl,
-                    )
-                    .unwrap();
+                    let viona =
+                        hw::virtio::PciVirtioViona::new(vnic_name, 0x100, &hdl)
+                            .unwrap();
```
luqmana, line 234: You can just replace this `.unwrap()` with `?`

luqmana (approved):
Looks good, thanks!

With this, would it be right to say that more-or-less all the emulated devices have been flattened out and should be good to go for serializing?

pfmooney:
> With this, would it be right to say that more-or-less all the emulated devices have been flattened out and should be good to go for serializing?

The flattening was intended to make that serialization easier, yeah.

standalone/src/main.rs
pfmooney: Yes indeed

pfmooney:
Merged as 18c9c89a10b064d1f4f253545953f212f94da82a

PR #66, opened 2021-11-03
Updates `usdt` dependency

bnaecker:
- `usdt` v0.1.11 -> v0.1.18
- Moves probe definition from separate D file to inline probes
- Adds feature flag `dtrace-probes` to the main `propolis` crate, which
  enables the probes. The same feature has been added to
  `propolis-server` and `propolis-standalone`, forwarding to the
  `propolis` feature. The features are all on by default, so the
  behavior remains the same as before. Opting out of the features with
  `--no-default-features` allows building with a stable toolchain.


propolis/Cargo.toml
```diff
@@ -18,14 +18,20 @@ thiserror = "1.0"
 bhyve_api = { path = "../bhyve-api" }
 dladm = { path = "../dladm" }
 viona_api = { path = "../viona-api" }
-usdt = "0.1.11"
+usdt = { version = "0.1.18", default-features = false }
 tokio = { version = "1", features = ["full"] }
 futures = "0.3"
 slog = "2.7"
+serde = { version = "1" }
+serde_json = { version = "1", optional = true }
```
pfmooney, line 26: So `serde_json` is required for the inline probes even though they aren't (currently) using `Serialize` types in their arguments?

bnaecker: It's not actually, no. This was hanging around from local testing with a previous version of the `usdt` crate. Thanks for catching, it's removed in 91f3f61.

pfmooney (approved):
Thanks for all the work on usdt

PR #69, opened 2021-11-16
Add migration endpoints to propolis-server

luqmana:
This adds the endpoints described in [RFD 71](https://rfd.shared.oxide.computer/rfd/0071) along with a first pass for some of the setup.

**New Endpoints:**
1. `/instances/{src-uuid}/migrate/start` — This is the endpoint called by the new (destination) propolis-server instance when its ready to begin the migration process. This request requires no body as we'll perform an HTTP upgrade to reuse the underlying TCP socket as a bidirectional channel to perform the migration over.
2. `/instances/{src-uuid}/migrate/status` — This endpoint allows querying either the source or destination on the current progress of a migration.

**Modified Endpoints:**
1. `/instances/{dst-uuid}` — This is the existing "ensure" endpoint, which has been extended with an optional blob describing a source VM from which a new VM will be populated.

### DIFF
--- a/server/src/lib/migrate.rs
+++ b/server/src/lib/migrate.rs
@@ -391,7 +391,7 @@ pub async fn dest_initiate(
     let conn = hyper::upgrade::on(res).await?;
 
     let migrate_context = Arc::new(MigrateContext {
-        migration_id: migration_id.clone(),
+        migration_id,
         state: RwLock::new(MigrationState::Sync),
     });
 
--- a/server/src/lib/migrate.rs
+++ b/server/src/lib/migrate.rs
@@ -1,10 +1,12 @@
 use std::sync::Arc;
 
 use dropshot::{HttpError, RequestContext};
-use hyper::{header, Body, Method, Response, StatusCode};
+use hyper::{header, upgrade::Upgraded, Body, Method, Response, StatusCode};
+use propolis::instance::Instance;
 use propolis_client::api;
 use slog::{error, info, o};
 use thiserror::Error;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use uuid::Uuid;
 
 use crate::server::Context;
@@ -58,6 +60,10 @@ pub enum MigrateError {
 
     #[error("unexpected Uuid")]
     UuidMismatch,
+
+    #[error("protocol error")]
+    // TODO: just for testing rn
+    Protocol,
 }
 
 impl MigrateError {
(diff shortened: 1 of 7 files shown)


server/src/lib/migrate.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+async fn source_migrate_task(
+    _instance: Arc<Instance>,
+    mut conn: Upgraded,
+    log: slog::Logger,
+) -> Result<(), MigrateError> {
+    info!(log, "Enter Migrate Task");
+
+    // TODO: actual migration protocol, for now just send some stuff back and forth
```
luqmana, line 112: cc @dancrossnyc this would be where we plug in the tokio codec Framed stuff to wrap this `Upgraded` object.

server/src/lib/migrate.rs
```diff
@@ -0,0 +1,308 @@
...
+            src_protocol
+        );
+        return Err(MigrateError::incompatible(src_protocol, dst_protocol));
+    }
+
+    // Now co-opt the socket for the migration protocol
+    let mut conn = hyper::upgrade::on(res).await?;
+
+    // Good to go, ready to migrate from the source via `conn`
+    // TODO: wrap in a tokio codec::Framed or such
```
luqmana, line 294: @dancrossnyc and here for the destination side.

pfmooney:
I assume a5bdf813276f8c6b7160e3f3f8ef7bc6ba72d940 is due to the stabilization of the `asm` stuff in the latest toolchain?  Perhaps land that (now) on its own with a little note in README about the MSRV requirement?

luqmana:
> I assume [a5bdf81](https://github.com/oxidecomputer/propolis/commit/a5bdf813276f8c6b7160e3f3f8ef7bc6ba72d940) is due to the stabilization of the `asm` stuff in the latest toolchain? Perhaps land that (now) on its own with a little note in README about the MSRV requirement?

Oh, no we already have a top-level rust-toolchain.toml which is set to nightly so that hasn't changed. I was just annoyed needed to change multiple other files while testing something earlier. But yup, I'll pull that outta this PR.

server/src/lib/migrate.rs
dancrossnyc: Cool; thanks.

server/src/lib/migrate.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    let upgrade = hyper::upgrade::on(&mut *request);
+    let instance = context.instance.clone();
+
+    // We've successfully negotiated a migration protocol w/ the destination.
+    // Now, we spawn a new task to handle the actual migration over the upgraded socket
+    let task = tokio::spawn(async move {
+        let conn = match upgrade.await {
+            Ok(upgraded) => upgraded,
```
dancrossnyc, line 197: Could we perhaps call this `conn` or `socket` or something? I probably just need more caffeine, but I was thinking that this referred to the upgrade case for LM. :-D

luqmana: I already do have a `conn` there but I just added a comment to make it more clear. Let me know if that's enough.

pfmooney:
I see no reason why we should get this landed, provided it'll make it easier to integrate with the other pending bits for LM.

> `/instances/{dst-uuid}` — This is the existing "ensure" endpoint, which has been extended with an optional blob describing a source VM from which a new VM will be populated.

An item for later follow-up, but we might end up wanting a separate endpoint for creating an instance for the purpose of "rehydration" during a migration.  The expectations there might be different enough to split those cases up. *shrug*

Thanks for putting this all together.

dancrossnyc:
@pfmooney I think there was a typo in your comment, "I see no reason why we shouldn't get this landed" ?

pfmooney:
> @pfmooney I think there was a typo in your comment, "I see no reason why we shouldn't get this landed" ?

And this is why I shouldn't be up so late.

dancrossnyc:
> And this is why I shouldn't be up so late.

:-) The meaning was clear from context, though.

PR #70, opened 2021-11-16
Expose more migration state

pfmooney:


propolis/src/inventory.rs
```diff
@@ -483,258 +494,3 @@ impl From<EntityID> for u64 {
         id.num
     }
 }
-
-#[cfg(test)]
-mod tests {
```
pfmooney, line 488: It seems pretty likely that the hierarchy and/or interfacing for the inventory will continue to change as we get further along with the migration stuff, so I didn't want to burn a bunch of time on keeping tests in lockstep until its future was more certain

propolis/src/hw/bhyve/atpic.rs
```diff
@@ -0,0 +1,35 @@
+use std::sync::Arc;
+
+use crate::dispatch::DispCtx;
+use crate::inventory::Entity;
+use crate::migrate::Migrate;
+
+use erased_serde::Serialize;
+
+pub struct BhyveAtPic {}
```
luqmana, line 9: Are the `atpic` devices even necessary for the types of VMs we want to run? My understanding was that atpic is for managing the pair of 8259 chips in legacy systems.

propolis/src/hw/bhyve/atpic.rs
```diff
@@ -0,0 +1,35 @@
...
+impl BhyveAtPic {
+    pub fn create() -> Arc<Self> {
+        Arc::new(Self {})
+    }
+}
+
+impl Entity for BhyveAtPic {
+    fn type_name(&self) -> &'static str {
```
luqmana, line 17: Not specific to this PR, but we should just make this an associated const that's a `&'static str`.

propolis/src/hw/bhyve/atpit.rs
```diff
@@ -0,0 +1,35 @@
+use std::sync::Arc;
+
+use crate::dispatch::DispCtx;
+use crate::inventory::Entity;
+use crate::migrate::Migrate;
+
+use erased_serde::Serialize;
+
+pub struct BhyveAtPit {}
```
luqmana, line 9: Similarly here, I'd expect providing tsc or at the very least hpet rather than pit.

propolis/src/hw/bhyve/pmtimer.rs
```diff
@@ -0,0 +1,39 @@
+use std::sync::Arc;
+
+use crate::dispatch::DispCtx;
+use crate::inventory::Entity;
+use crate::migrate::Migrate;
+
+use erased_serde::Serialize;
+
+pub struct BhyvePmTimer {}
```
luqmana, line 9: Does this represent the ACPI Power Management Timer?

propolis/src/inventory.rs
```diff
@@ -177,19 +172,35 @@ struct InventoryInner {
 }
 
 impl InventoryInner {
-    /// Registers a new entity with the inventory.
+    /// Registers a new entity (and its children) with the inventory.
     ///
     /// Returns an error if the object has already been registered.
     /// Returns an error if the parent is not registered.
```
luqmana, line 178: Is this comment still accurate since we don't take a `parent_id` parameter anymore?

propolis/src/inventory.rs
```diff
@@ -75,27 +75,22 @@ impl Inventory {
...
-                )
-                .unwrap();
-            }
-        }
-        Ok(res)
+    /// Registers a new entity, bearing an instance name, with the inventory.
+    ///
+    /// Returns an error if the object has already been registered.
+    /// Returns an error if the parent is not registered.
+    pub fn register_instance<T: Entity>(
```
luqmana, line 87: It looks like you basically unnested a bunch of devices as part of using this method but not all of them are implicitly registered in their (former) parent's `child_register` method.

luqmana (commented):
Looks good overall, left a couple of comments.

propolis/src/hw/bhyve/atpic.rs
pfmooney: It's in the kernel emulation and not really disabled, so yeah, we need it still (sadly).

propolis/src/hw/bhyve/atpic.rs
pfmooney: Is that possible with trait objects?  A `const fn` definitely wasn't.

propolis/src/hw/bhyve/atpit.rs
pfmooney: Again, it's emulated in kernel, so we'll need to migrate the state.  The PIT is often used for initial timer calibration at startup.

propolis/src/hw/bhyve/atpic.rs
luqmana: Ah right, that's true we do need it to be object safe.

propolis/src/hw/bhyve/pmtimer.rs
pfmooney: Aye, it does.

propolis/src/inventory.rs
pfmooney: Will fix

propolis/src/inventory.rs
pfmooney: `register_child` was left for cases such as block devices where the entity does not hold a reference to the child when it itself is being added to the registry.

I'll remove the note about the missing parent, since that error shouldn't apply here for now.

propolis/src/hw/bhyve/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+
+pub type KernelDefaults = (
+    Arc<BhyveAtPic>,
+    Arc<BhyveAtPit>,
+    Arc<BhyveHpet>,
+    Arc<BhyveIoApic>,
+    Arc<BhyveRtc>,
+);
+
+/// Create cohort of devices emulated by bhyve kernel component
```
pfmooney, line 25: I guess I could have advertised this fact a little more, but these are all devices which are emulated (by default) by the kernel vmm portion of bhyve.  They're here not because we're going to do emulation in userspace, but so we can gather up all their state for a migration.

luqmana (approved):

pfmooney:
Merged in 2135ae418d1c0be8342fd1a6669eeb8e2e231ae2. Thanks for looking it over.

PR #72, opened 2021-11-18
Read Crucible disks from a JSON file

jmpesp:
Change the cli so that it reads Crucible disks from a JSON file instead
of trying to construct them out of command line arguments.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,14 +1,17 @@
+use std::fs::File;
+use std::io::BufReader;
+use std::path::PathBuf;
 use std::{
     net::{IpAddr, SocketAddr, ToSocketAddrs},
     os::unix::prelude::AsRawFd,
 };
 
-use anyhow::{anyhow, bail, Context};
+use anyhow::{anyhow, Context};
 use futures::{SinkExt, StreamExt};
 use propolis_client::{
     api::{
         DiskRequest, InstanceEnsureRequest, InstanceProperties,
-        InstanceStateRequested, Slot,
+        InstanceStateRequested,
     },
     Client,
 };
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -55,9 +58,9 @@ enum Command {
         #[structopt(short, default_value = "1024")]
         memory: u64,
 
-        // Specify multiple disks in groups of 3 targets
-        #[structopt(long, use_delimiter = true)]
-        crucible_target: Vec<SocketAddr>,
+        // file with a JSON array of DiskRequest structs
+        #[structopt(long, parse(from_os_str))]
+        crucible_disks: PathBuf,
```
pfmooney, line 63: Does this mean that having a crucible disk present is now required?

jmpesp: This is my mistake, made it optional in ec97ee3.

cli/src/main.rs
```diff
@@ -144,28 +147,15 @@ async fn new_instance(
...
-            name: format!("d{}", disk_i),
-            address: target_group.to_vec(),
-            slot: Slot(disk_i),
-            read_only: false,
-            key: None,
-            gen: 0,
-        });
-        disk_i += 1;
+    for disk in &disks {
+        assert_eq!(disk.address.len(), 3);
```
pfmooney, line 151: Should this perhaps be part of the `parse_crucible_disks`, and then the rest of the validation be left to the API we're calling into?

jmpesp: Moved target validation server side in 00009e6

jmpesp:
@pfmooney let me know if there's any further work here :)

pfmooney:
> @pfmooney let me know if there's any further work here :)

ship it

PR #74, opened 2021-11-23
In DiskRequest, specify virtio or nvme

jmpesp:
Also address some clippy warnings:
- remove unused total size field
- address redundant to_string warnings

### DIFF
--- a/client/src/api.rs
+++ b/client/src/api.rs
@@ -175,6 +175,7 @@ pub struct DiskRequest {
     pub read_only: bool,
     pub key: Option<String>,
     pub gen: u64,
+    pub device: String,
 }
 
 #[derive(Clone, Deserialize, Serialize, JsonSchema)]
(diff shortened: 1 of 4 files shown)


client/src/api.rs
```diff
@@ -175,6 +175,7 @@ pub struct DiskRequest {
     pub read_only: bool,
     pub key: Option<String>,
     pub gen: u64,
+    pub device: String,
```
pfmooney, line 178: meta question: Is there anything we want to do to coordinate changes to the client api?

pfmooney (approved):

client/src/api.rs
jmpesp: I'm not aware if that's an answered question or not. I know that versioning has been discussed a few times, and that there's some relevant RFDs too.

At this stage of development, at least a change like this will cause a compile failure instead of a silent runtime failure. In that vein, I verified that Omicron will build with this change (they use a rev versioned dependency anyway :) ).

PR #75, opened 2021-11-23
Initial Live Migration Protocol

dancrossnyc:
This is the initial implementation of live migration protocol.
For the first edit, it is mostly stubs.

Signed-off-by: Dan Cross <cross@oxidecomputer.com>


dancrossnyc:
This is still very much a WIP, but it's getting large; what's there now is mostly scaffolding with the basics of most of the protocol elements. Please kick the tires and let me know if there are large structural issues that need to be addressed; some of those may be deferred beyond the Friday demo, but better to get eyes on now.

pfmooney:
> This is still very much a WIP, but it's getting large; what's there now is mostly scaffolding with the basics of most of the protocol elements. Please kick the tires and let me know if there are large structural issues that need to be addressed; some of those may be deferred beyond the Friday demo, but better to get eyes on now.

I see no reason to delay landing it so we can both get all the needed pieces in place for the demo and then iterate on it "in the gate" afterwards.

PR #76, opened 2021-11-25
Arbitrary nvme block sizes + various fixes

jmpesp:
Support arbitrary block sizes in nvme code, otherwise defining a
crucible volume with 4k blocks causes kernel panics.

In the cli, use Ctrl-A to send a raw Ctrl-C, otherwise Ctrl-C kills the
cli.

In nvme, use a random alphanumeric serial number, otherwise Linux kernel
complains that multiple devices share the same (null) serial number.

In server, parse nvme in hardcoded server toml too.

In crucible, log errors if they happen.

### DIFF
--- a/propolis/src/hw/nvme/mod.rs
+++ b/propolis/src/hw/nvme/mod.rs
@@ -10,6 +10,7 @@ use crate::{block, common::*};
 
 use erased_serde::Serialize;
 use lazy_static::lazy_static;
+use rand::{distributions::Alphanumeric, Rng};
 use thiserror::Error;
 
 mod admin;
@@ -23,10 +24,6 @@ use queue::{CompQueue, QueueId, SubQueue};
 
 /// The max number of MSI-X interrupts we support
 const NVME_MSIX_COUNT: u16 = 1024;
-
-/// Supported block size.
-/// TODO: Support more
-const BLOCK_SZ: u64 = 512;
 
 /// NVMe errors
 #[derive(Debug, Error)]
@@ -444,12 +441,20 @@ impl PciNvme {
         let cqes = size_of::<RawCompletion>().trailing_zeros() as u8;
         let sqes = size_of::<RawSubmission>().trailing_zeros() as u8;
 
+        // Random serial number
+        let sn: [u8; 20] = rand::thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(20)
+            .collect::<Vec<u8>>()
+            .try_into()
+            .expect("Expected vector of 20!");
+
         // Initialize the Identify structure returned when the host issues
         // an Identify Controller command.
         let ctrl_ident = bits::IdentifyController {
             vid: vendor,
             ssvid: vendor,
-            // TODO: fill out serial number
+            sn,
             ieee: [0xA8, 0x40, 0x25], // Oxide OUI
             // We use standard Completion/Submission Queue Entry structures with no extra
             // data, so required (minimum) == maximum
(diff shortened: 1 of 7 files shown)


propolis/src/hw/nvme/mod.rs
```diff
@@ -444,12 +441,20 @@ impl PciNvme {
         let cqes = size_of::<RawCompletion>().trailing_zeros() as u8;
         let sqes = size_of::<RawSubmission>().trailing_zeros() as u8;
 
+        // Random serial number
+        let sn: [u8; 20] = rand::thread_rng()
```
luqmana, line 445: We should probably accept the serial as a parameter or generate it by seeding it off the VM's UUID or something. That way a guest won't see the same disk but with different serial number on every boot.

propolis/src/hw/nvme/mod.rs
```diff
@@ -466,7 +471,7 @@ impl PciNvme {
         // Initialize the Identify structure returned when the  host issues
         // an Identify Namespace command.
         let total_bytes = binfo.total_size * binfo.block_size as u64;
-        let nsze = total_bytes / BLOCK_SZ;
+        let nsze = total_bytes / (binfo.block_size as u64);
```
luqmana, line 474: We can probably simplify this and just do:
```rust
let nsze = binfo.total_size;
```

propolis/src/hw/nvme/mod.rs
pfmooney: Agreed.  We should probably look at exposing the volume uuid via the NGUID and/or EUI64 fields for the namespace identification.

propolis/src/hw/nvme/mod.rs
jmpesp: Yeah... good catch. simplified in 14f4dc9.

propolis/src/hw/nvme/mod.rs
jmpesp: Unfortunately EUI64 shows up in version 1.1 and NGUID shows up in 1.2. Upgrading the supported NVMe version is a big lift to be included in this various fixes PR :)

Instead of the random serial, I plumbed through the name field out of DiskRequest:

	{
	  "name": "data",
	  "address": [
	    "127.0.0.1:3801",
	    "172.20.3.63:3801",
	    "172.20.3.60:3801"
	  ],
	  "slot": 2,
	  "read_only": false,
	  "gen": 0,
	  "key": "+7UAEsHnh8EphVql4gHE2hV6ddFXvwb1AcAlgqwe6Dk=",
	  "device": "nvme"
	}

and block device name out of the TOML:

	# extra disk
	[block_dev.extra]
	type = "file"
	path = "/home/james/disk.img"

	[dev.block1]
	driver = "pci-nvme"
	name = "Oxide Computer Company NVMe Device"
	block_dev = "extra"
	pci-path = "0.13.0"

nvme cli shows:

	root@localhost:~# ./nvme-cli/nvme list -o json
	Could not find pci.ids file
	Could not find pci.ids file
	{
	  "Devices" : [
	    {
	      "NameSpace" : 1,
	      "DevicePath" : "/dev/nvme0n1",
	      "Firmware" : "        ",
	      "Index" : 0,
	      "ModelNumber" : "                                        ",
	      "ProductName" : "NULL",
	      "SerialNumber" : "extra",
	      "UsedBytes" : 5368709120,
	      "MaximumLBA" : 10485760,
	      "PhysicalSize" : 5368709120,
	      "SectorSize" : 512
	    },
	    {
	      "NameSpace" : 1,
	      "DevicePath" : "/dev/nvme1n1",
	      "Firmware" : "        ",
	      "Index" : 1,
	      "ModelNumber" : "                                        ",
	      "ProductName" : "NULL",
	      "SerialNumber" : "data",
	      "UsedBytes" : 4288675840,
	      "MaximumLBA" : 1047040,
	      "PhysicalSize" : 4288675840,
	      "SectorSize" : 4096
	    }
	  ]
	}

That way the service issuing the DiskRequest can choose to use a name, or volume id, or something like that. See commit bd5c5d5.

pfmooney (approved):

PR #77, opened 2021-11-29
Expose Bhyve dirty page tracking ioctl in propolis

dancrossnyc:
Plumb the ioctl being added to upstream illumos/bhyve
into propolis.

Signed-off-by: Dan Cross <cross@oxidecomputer.com>

### DIFF
--- a/bhyve-api/src/ioctls.rs
+++ b/bhyve-api/src/ioctls.rs
@@ -88,4 +88,5 @@ pub const VM_SUSPEND_CPU: i32 = VMM_IOC_BASE | 0x1d;
 pub const VM_RESUME_CPU: i32 = VMM_IOC_BASE | 0x1e;
 
 pub const VM_TRACK_DIRTY_PAGES: i32 = VMM_IOC_BASE | 0x20;
+
 pub const VM_DEVMEM_GETOFFSET: i32 = VMM_IOC_BASE | 0xff;
--- a/bhyve-api/src/ioctls.rs
+++ b/bhyve-api/src/ioctls.rs
@@ -86,7 +86,6 @@ pub const VM_GET_TOPOLOGY: i32 = VMM_IOC_BASE | 0x1b;
 pub const VM_GET_CPUS: i32 = VMM_IOC_BASE | 0x1c;
 pub const VM_SUSPEND_CPU: i32 = VMM_IOC_BASE | 0x1d;
 pub const VM_RESUME_CPU: i32 = VMM_IOC_BASE | 0x1e;
-
 pub const VM_TRACK_DIRTY_PAGES: i32 = VMM_IOC_BASE | 0x20;
 
 pub const VM_DEVMEM_GETOFFSET: i32 = VMM_IOC_BASE | 0xff;
--- a/bhyve-api/src/ioctls.rs
+++ b/bhyve-api/src/ioctls.rs
@@ -87,4 +87,5 @@ pub const VM_GET_CPUS: i32 = VMM_IOC_BASE | 0x1c;
 pub const VM_SUSPEND_CPU: i32 = VMM_IOC_BASE | 0x1d;
 pub const VM_RESUME_CPU: i32 = VMM_IOC_BASE | 0x1e;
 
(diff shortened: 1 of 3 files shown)


propolis/src/vmm/hdl.rs
```diff
@@ -256,6 +257,26 @@ impl VmmHdl {
...
+    /// - `bitmap`: A mutable bitmap of dirty pages, one bit per guest PFN
+    /// relative to `start_gpa`.
+    pub fn track_dirty_pages(
+        &self,
+        start_gpa: u64,
+        bitmap: &mut [u8],
+    ) -> Result<()> {
+        let mut tracker = bhyve_api::vm_dirty_tracker {
+            start_gpa,
+            len: bitmap.len(),
```
pfmooney, line 274: `len` is in bytes of address space, not pages.  This should be multiplied by the `PAGESIZE` constant.

bhyve-api/src/structs.rs
```diff
@@ -319,6 +319,14 @@ pub struct vm_run_state {
     pub _pad: [u8; 3],
 }
 
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct vm_dirty_tracker {
+    pub start_gpa: u64,
+    pub len: size_t,
+    pub pfns: *mut c_void,
```
pfmooney, line 327: These should probably reflect the naming in the OS header, with the `vdt_` prefixes.

bhyve-api/src/ioctls.rs
```diff
@@ -87,4 +87,5 @@ pub const VM_GET_CPUS: i32 = VMM_IOC_BASE | 0x1c;
 pub const VM_SUSPEND_CPU: i32 = VMM_IOC_BASE | 0x1d;
 pub const VM_RESUME_CPU: i32 = VMM_IOC_BASE | 0x1e;
 
+pub const VM_TRACK_DIRTY_PAGES: i32 = VMM_IOC_BASE | 0x20;
```
pfmooney, line 90: nit: Why isn't this grouped with the entries above (`0x1e`)?

propolis/src/vmm/hdl.rs
dancrossnyc: Oops. Done. The comment in the kernel needs to change, too.

bhyve-api/src/structs.rs
dancrossnyc: Interesting; this isn't quite universal (e.g., `cpuid` in `vm_run_state` is `vcpuid` in the kernel). On reflection, it appears that `vmm_dirty_tracker` may be misnamed on several levels; perhaps it should be named `vm_dirty_tracker` and the `vdt_` prefixes removed; that would be more consistent with other Bhyve data structures. What do you think? I already need to fix the comment in the kernel.

bhyve-api/src/ioctls.rs
dancrossnyc: Because semantically it's not really related to those things (tracking dirty pages feels qualitatively different than e.g. NMI injection), and the constants in this file seem to be grouped semantically. It's also not grouped that way in the kernel. But I'm happy to change it, if you'd prefer?

bhyve-api/src/structs.rs
dancrossnyc: (Talked about this on Matrix; we'll stick with what we have and the definitions here have been changed to match the kernel.)

propolis/src/vmm/hdl.rs
dancrossnyc: Also updated the kernel in the current illumos CR.

bhyve-api/src/ioctls.rs
pfmooney: I would prefer that.

The `DEVMEM_GETOFFSET` stuff is separate because it's mainly to paper over how FreeBSD and illumos bhyve handle the direct mapping of memory objects.  The rest of the ioctls relate to running to the machine, so I would keep them grouped as such.

pfmooney (approved):

bhyve-api/src/ioctls.rs
dancrossnyc: Good to go; line deleted.

PR #78, opened 2021-11-30
update for nightly-2021-11-24

ahl:
We're moving omicron and all dependents past the asm_sym flag day. See https://github.com/oxidecomputer/omicron/pull/448

Note that this will not build on macos until https://github.com/oxidecomputer/crucible/pull/127 has been merged


ahl:
... I don't understand how this built properly on macos...

PR #80, opened 2021-12-13
Update crucible interface to support SocketAddr

leftwo:
This allows crucible downstairs targets to have either IPv4 or
IPv6 addresses.

This, once approved, will have to merge when the crucible side PR: 
https://github.com/oxidecomputer/crucible/pull/143 
Is pushed as well, as builds will fail otherwise.

### DIFF
--- a/propolis/Cargo.toml
+++ b/propolis/Cargo.toml
@@ -21,6 +21,9 @@ viona_api = { path = "../viona-api" }
 usdt = { version = "0.2.1", default-features = false }
 tokio = { version = "1", features = ["full"] }
 futures = "0.3"
+# The crucible branch should be switched back to main when the Crucible PR
+# https://github.com/oxidecomputer/crucible/pull/143 is merged back to
+# crucible main.  This is here now to allow propolis with crucible to build.
 crucible = { git = "https://github.com/oxidecomputer/crucible", branch = "ipv6", optional = true }
 anyhow = "1"
 slog = "2.7"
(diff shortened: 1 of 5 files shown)


propolis/Cargo.toml
```diff
@@ -21,7 +21,7 @@ viona_api = { path = "../viona-api" }
 usdt = { version = "0.2.1", default-features = false }
 tokio = { version = "1", features = ["full"] }
 futures = "0.3"
-crucible = { git = "https://github.com/oxidecomputer/crucible", branch = "main", optional = true }
```
leftwo, line 24: This change is here just to allow things to compile.

When this (and the crucible side) has been approved, this change can come out and main will become the branch to use for crucible.

pfmooney: Please include a short comment in the toml itself to that effect?

jmpesp (approved):
:rocket: (but I'd prefer to merge the crucible PR and change the dep back to main as a prereq)

pfmooney (approved):

pfmooney:
> rocket (but I'd prefer to merge the crucible PR and change the dep back to main as a prereq)

Whichever works best for y'all

leftwo:
> > rocket (but I'd prefer to merge the crucible PR and change the dep back to main as a prereq)
> 
> Whichever works best for y'all

They are both ready to go, so if I merge both "at the same time" things should continue to work.

PR #86, opened 2022-01-20
fix break that will be caused by merging #182

jmpesp:
Crucible PR https://github.com/oxidecomputer/crucible/pull/182 will
break compilation for propolis because new CrucibleOpts fields were
added there. Add the Default line here so that this doesn't happen in
the future, but Default was also derived for CrucibleOpts in that PR.
Unfortunately, these have to go in at the same time.


luqmana (approved):
SGTM. Could also just make a small initial change to crucible adding only the Default derive to avoid having to sync merging this with https://github.com/oxidecomputer/crucible/pull/182

PR #88, opened 2022-02-01
Plumb the "Machine" into LM.

dancrossnyc:
It took an absurdly long time to do this.

### DIFF
--- a/propolis/src/instance.rs
+++ b/propolis/src/instance.rs
@@ -221,6 +221,11 @@ impl Instance {
         func(machine, &mctx, &self.disp, &state.inv)
     }
 
+    pub fn machine(&self) -> Arc<Machine> {
+        let state = self.inner.lock().unwrap();
+        state.machine.clone().unwrap()
+    }
+
     /// Returns the state of the instance.
     pub fn current_state(&self) -> State {
         let state = self.inner.lock().unwrap();
--- a/propolis/src/instance.rs
+++ b/propolis/src/instance.rs
@@ -221,11 +221,6 @@ impl Instance {
         func(machine, &mctx, &self.disp, &state.inv)
     }
 
-    pub fn machine(&self) -> Arc<Machine> {
-        let state = self.inner.lock().unwrap();
-        state.machine.clone().unwrap()
-    }
-
     /// Returns the state of the instance.
     pub fn current_state(&self) -> State {
         let state = self.inner.lock().unwrap();
--- a/server/src/lib/migrate/destination.rs
+++ b/server/src/lib/migrate/destination.rs
@@ -2,6 +2,7 @@ use futures::{SinkExt, StreamExt};
 use std::sync::Arc;
 
 use hyper::upgrade::Upgraded;
+use propolis::dispatch::AsyncCtx;
 use propolis::instance::Instance;
 use slog::info;
 use tokio_util::codec::Framed;
@@ -15,12 +16,14 @@ type Result<T> = anyhow::Result<T, MigrateError>;
 pub async fn migrate(
     migrate_context: Arc<MigrateContext>,
     instance: Arc<Instance>,
+    async_context: AsyncCtx,
     conn: Upgraded,
     log: slog::Logger,
 ) -> Result<()> {
     let mut proto = DestinationProtocol {
-        instance,
         migrate_context,
+        instance,
+        async_context,
         conn: Framed::new(conn, codec::LiveMigrationFramer::new()),
         log,
     };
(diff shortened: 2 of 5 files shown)


server/src/lib/server.rs
```diff
@@ -498,6 +454,41 @@ async fn instance_ensure(
             ))
         })?;
 
+    // Is this part of a migration?
+    let migrate = if let Some(migrate_request) = request.migrate {
+        // We stop short of the usual intialization routines because this is
```
luqmana, line 459: Need to update this comment since it isn't true after this.

server/src/lib/migrate/destination.rs
```diff
@@ -37,7 +36,7 @@ pub async fn migrate(
 }
 
 struct DestinationProtocol {
-    _request_context: Arc<RequestContext<Context>>,
+    instance: Arc<Instance>,
```
luqmana, line 39: Nice, I forgot to plumb this through.

propolis/src/instance.rs
```diff
@@ -221,6 +221,11 @@ impl Instance {
         func(machine, &mctx, &self.disp, &state.inv)
     }
 
+    pub fn machine(&self) -> Arc<Machine> {
+        let state = self.inner.lock().unwrap();
+        state.machine.clone().unwrap()
+    }
```
luqmana, line 227: This is fine but as part of trying plumbing a migrate state to propolis and tie in the migrate tasks, I just gave them `AsyncCtx` parameters which allows grabbing the `DispCtx` which has a `MachineCtx` that just wraps a `Machine`.

https://github.com/oxidecomputer/propolis/blob/6a8a3ad4d09e0b7f225eba375115c41ddacb1f4e/server/src/lib/migrate/destination.rs#L18-L30

server/src/lib/migrate/mod.rs
```diff
@@ -331,10 +331,12 @@ pub async fn dest_initiate(
     // We've successfully negotiated a migration protocol w/ the source.
     // Now, we spawn a new task to handle the actual migration over the upgraded socket
     let mctx = migrate_context.clone();
-    let task_rqctx = rqctx.clone();
+    let context = rqctx.context().context.lock().await;
+    let context = context.as_ref().unwrap();
```
luqmana, line 335: You can avoid this `unwrap` by just matching what the source side code does.

propolis/src/instance.rs
pfmooney: That seems better than emitting the `Machine` straight out.

server/src/lib/server.rs
```diff
@@ -259,61 +259,17 @@ async fn instance_ensure(
...
-                        )
-                        .unwrap();
-                    }
-                }
-            }
-            _ => {}
-        }
+    instance.on_transition(Box::new(move |next_state, _inv, _ctx| {
         let last = (*tx.borrow()).clone();
         let _ = tx.send(StateChange { gen: last.gen + 1, state: next_state });
```
pfmooney, line 264: This will mean that emitting the `StateChange` objects through the channel will now occur before the later `on_transition` event performing the on-reboot reinitinialization of the vCPU state.  That's probably fine, though.

server/src/lib/server.rs
```diff
@@ -498,6 +454,41 @@ async fn instance_ensure(
...
+                }
+                _ => {}
+            }
+        }));
+
+        None
+    };
     instance.print();
 
     // Save the newly created instance in the server's context.
```
luqmana, line 494: Need to move this above the `if let` and `drop(context)` before `dest_initiate` as it tries to lock the Mutex again leaving us deadlocked.

propolis/src/instance.rs
dancrossnyc: Agreed. That's better in all ways.

server/src/lib/migrate/destination.rs
dancrossnyc: :-)

server/src/lib/server.rs
dancrossnyc: Ooo, that's embarrassing. Done.

server/src/lib/migrate/mod.rs
dancrossnyc: Done.

(Though it's not clear how we can get to this point without context being initialized. Anyway, defense in depth!)

server/src/lib/server.rs
```diff
@@ -498,18 +454,53 @@ async fn instance_ensure(
...
         state_watcher: rx,
         serial_task: None,
     });
+    drop(context);
+
+    // Is this part of a migration?
+    let migrate = if let Some(migrate_request) = request.migrate {
+        // This is a migrate request and so we should try to establish a
+        // connection with the source instance.
+        let rqctx = rqctx.clone();
```
luqmana, line 471: Unneeded clone?

dancrossnyc: Done.

luqmana (approved):

pfmooney (approved):

PR #89, opened 2022-02-02
Async access to machine state during migration

luqmana:
Based atop https://github.com/oxidecomputer/propolis/pull/88. This is a sketch of how to reconcile the async context we're in as part of propolis-server with propolis core.

This adds a new 'Migrating' state as part of the internal propolis state machine [1]. As part of transitioning to that state, we allow specifying a `CtxId` representing an `AsyncCtx` object associated with an async task. This lets us quiesce any other tasks but explicitly allow the migrate task to access the Instance's `DispCtx` to obtain the underlying `MachineCtx` and vmm handles.

There's still a question of how to get access to the inventory.

[1] This will also need corresponding changes in omicron once its propolis dep is updated.

### DIFF
--- a/propolis/src/dispatch/async_tasks.rs
+++ b/propolis/src/dispatch/async_tasks.rs
@@ -9,7 +9,8 @@ use tokio::task::JoinHandle;
 
 use super::{DispCtx, Dispatcher, SharedCtx, SyncFn, WakeFn};
 
-type CtxId = usize;
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct CtxId(usize);
 
 struct CtxCtrl {
     quiesce_guard: Semaphore,
@@ -37,7 +38,7 @@ impl Inner {
     fn new() -> Self {
         Self {
             state: Mutex::new(InnerState {
-                next_id: 0,
+                next_id: CtxId(0),
                 quiesced: false,
                 shutdown: false,
             }),
@@ -50,7 +51,7 @@ impl Inner {
         let mut ctrls = self.ctrls.lock().unwrap();
 
         let id = state.next_id;
-        state.next_id += 1;
+        state.next_id.0 += 1;
         let ctrl = Arc::new(CtxCtrl {
             quiesce_guard: Semaphore::new(0),
             quiesce_notify: Semaphore::new(0),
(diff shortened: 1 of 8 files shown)


propolis/src/instance.rs
```diff
@@ -437,6 +478,11 @@ impl Instance {
                     // suspend become stale.
                     inner.suspend_info = None;
                 }
+                State::Migrate(_) => {
+                    let migrate_ctx = inner.migrate_ctx.take().unwrap();
+                    // We explicitly allow the migrate task to run
+                    self.disp.release_one(migrate_ctx);
+                }
```
pfmooney, line 485: While in the long-term, it'd be nice to separate out the tasks (and/or Ctx objects dispensed to grant instance access) into different classes which dictate when they are run or paused, this seems like a reasonable intermediate step.

propolis/src/instance.rs
```diff
@@ -221,6 +244,11 @@ impl Instance {
         func(machine, &mctx, &self.disp, &state.inv)
     }
 
+    pub fn machine(&self) -> Arc<Machine> {
+        let state = self.inner.lock().unwrap();
+        state.machine.clone().unwrap()
+    }
+
```
pfmooney, line 251: Is this still used?

luqmana: Rebased and gone.

propolis/src/dispatch/async_tasks.rs
```diff
@@ -50,7 +51,7 @@ impl Inner {
         let mut ctrls = self.ctrls.lock().unwrap();
 
         let id = state.next_id;
-        state.next_id += 1;
+        state.next_id.0 += 1;
```
dancrossnyc, line 54: Hmm, could we impl Add and/or AddAssign on CtxId to avoid reaching into the tuple?

dancrossnyc (approved):
Sweet.

propolis/src/dispatch/async_tasks.rs
luqmana: That would let anyone with a `CtxId` to manufacture a different one. The adding is just an implementation detail anyways and doesn't really make sense as something to lift to the type itself.

dancrossnyc: SG

pfmooney (approved):

PR #90, opened 2022-02-08
Avoid interrupt level assertion on guest reboot

pfmooney:
When the in-kernel state is reinitialized during a reboot, all existing interrupt pin level state is discarded (from both the ATPIC and IOAPIC).  This means that any userspace-emulated devices with asserted bits at reinit time will encounter troubling de-asserting them, since the userspace state has become inconsistent with the now-cleared kernel state.  As a fix for now, I'm splitting up the event transitions into `Pre` and `Post` phases, so all of the device emulation can complete its reset before we reinitialize the in-kernel state.  I also noticed that the PS2 controller (which happened to be the device holding an asserted pin over reboot) wasn't wired up for state reset at all.

### DIFF
--- a/propolis/src/hw/ps2ctrl.rs
+++ b/propolis/src/hw/ps2ctrl.rs
@@ -284,10 +284,26 @@ impl PS2Ctrl {
                 && state.aux_port.has_output(),
         );
     }
+    fn reset(&self) {
+        let mut state = self.state.lock().unwrap();
+        state.pri_port.reset();
+        state.aux_port.reset();
+        state.resp = None;
+        state.cmd_prefix = None;
+        state.ctrl_cfg = CtrlCfg::default();
+        state.ctrl_out_port = CtrlOutPort::default();
+        for b in state.ram.iter_mut() {
+            *b = 0;
+        }
+        self.update_intr(&mut state);
+    }
 }
 impl Entity for PS2Ctrl {
     fn type_name(&self) -> &'static str {
         "lpc-ps2ctrl"
+    }
+    fn reset(&self, _ctx: &DispCtx) {
+        PS2Ctrl::reset(self);
     }
     fn migrate(&self) -> Option<&dyn Migrate> {
         Some(self)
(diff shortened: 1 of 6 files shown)


pfmooney:
Note: This was happening silently prior to [illumos #14260](https://www.illumos.org/issues/14260).

propolis/src/inventory.rs
```diff
@@ -457,12 +457,17 @@ pub trait Entity: Send + Sync + 'static {
...
+        phase: TransitionPhase,
         ctx: &DispCtx,
     ) {
     }
     /// Function dedicated to `State::Reset` event delivery so implementers do
     /// not need to create a more verbose `state_transition` implementation for
     /// emulating device reset.
+    ///
+    /// This is called as part of `TransitionPhase::Pre`.  Entities which
+    /// require logic in the `Post` phase should do so via the
```
luqmana, line 469: Would it be worth also splitting out the post phase as well? Letting devices implement `pre_reset` and/or `post_reset` as needed.

propolis/src/hw/ps2ctrl.rs
```diff
@@ -652,6 +668,10 @@ impl PS2Mouse {
     fn reset(&mut self) {
         // XXX  what should the defaults be?
         self.buf.clear();
+        self.cur_cmd = None;
```
luqmana, line 671: `PS2Kbd::reset` doesn't clear `cur_cmd` in its `reset` method, should that also be updated to match `PS2Mouse` here?

propolis/src/inventory.rs
pfmooney: Since `Entity::reset` is there for convenience, and few things (so far) need the explicit post actions, I'm inclined to just let them handle it via `Entity:state_transition`.

propolis/src/hw/ps2ctrl.rs
pfmooney: Mm, yeah.  Good catch.

luqmana (approved):

dancrossnyc (approved):

pfmooney:
The unit and integration tests all pass with this change, and my helios guest which was triggering the behavior is able to reboot without stumbling over the assert now.

PR #94, opened 2022-02-16
Facilitate direct access to VM memory objects

pfmooney:
Rather than going through the nested page tables for the guest-physical address space, use mappings which directly access the underlying memory objects.  This has a workaround for the lack of [illumos #14511](https://www.illumos.org/issues/14511), until that fix is merged.

### DIFF
--- a/propolis/src/vmm/machine.rs
+++ b/propolis/src/vmm/machine.rs
@@ -3,7 +3,7 @@
 use std::io::{Error, ErrorKind, Result};
 use std::marker::PhantomData;
 use std::mem::size_of;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
 use crate::common::{GuestAddr, GuestRegion};
 use crate::mmio::MmioBus;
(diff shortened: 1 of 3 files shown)


propolis/src/vmm/machine.rs
```diff
@@ -328,18 +364,47 @@ impl<'a> MemCtx<'a> {
             let req_offset = start - addr;
             match ent.kind {
                 MapKind::SysMem(_, prot) | MapKind::Rom(_, prot) => {
-                    // TODO: This protection check may be redundant with
-                    // mapping, which has its own memory protection tracking.
-                    if prot.contains(need_prot) {
-                        let base = ent.guest_map.as_ref().unwrap().as_ref();
-                        return base.subregion(req_offset, len);
-                    }
+                    let guest_map = ent
```
dancrossnyc, line 367: It's a stylistic nit, but I felt like the `base` local was slightly more readable. There's a big chain of methods now that's a tad rough to keep track of.

dancrossnyc (approved):

dancrossnyc:
With the associated illumos change (https://www.illumos.org/issues/14511) merged, shall we integrate this?

pfmooney:
Merged in 1cbac55da86fde4a86dc833a5c8acb44fa34c8b6

PR #95, opened 2022-02-17
Attach cloud-init ISO to instance

jmpesp:
Accept new CloudInit field as part of instance ensure request. The user
can specify meta-data, and either structured fields (which renders
to #cloud-config), or a custom base64 encoded string that renders to
user-data.

Build a CIDATA ISO from these two files and mount the bytes it as a
virtio device. This commit also adds a new in-memory block backend for
propolis.


propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,283 @@
...
+            ));
+        }
+
+        nread += mapping.len();
+    }
+
+    let mut bytes = bytes.lock().unwrap();
+    for i in 0..len {
+        bytes[offset as usize + i] = vec[i];
+    }
```
luqmana, line 247:
```suggestion
    bytes[offset..offset + len].copy_from_slice(&vec);
```

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,283 @@
...
+/// Read from bytes into guest memory
+fn process_read_request(
+    bytes: &Arc<Mutex<Vec<u8>>>,
+    offset: u64,
+    len: usize,
+    mappings: &Vec<SubMapping>,
+) -> Result<()> {
+    let bytes = bytes.lock().unwrap();
+    let data =
+        &bytes[Range { start: offset as usize, end: offset as usize + len }];
```
luqmana, line 192: Any reason explicitly using Range here vs the `..` syntax?

server/src/lib/server.rs
```diff
@@ -168,6 +177,106 @@ fn slot_to_bdf(slot: api::Slot, ty: SlotType) -> Result<pci::Bdf> {
     }
 }
 
+fn create_in_memory_cloud_init_iso(
+    log: &Logger,
+    cloud_init: &api::CloudInit,
+) -> anyhow::Result<Vec<u8>> {
+    info!(log, "creating temp dir");
+
+    let temp_dir = tempdir()?;
```
luqmana, line 186: Probably not as much of an issue as each propolis instance will run in its own zone anyways, but it'd be nice if we could do this entirely in RAM without having to create things on disk first. Worry here being a malicious actor swapping out the file between when we write and when we read it.

Anyways, don't take this is a blocker--I imagine it'd require a lot more work to re architecture the iso crate.

propolis/src/block/in_memory.rs
jmpesp: No reason, but changed in f71216e

server/src/lib/server.rs
jmpesp: Changing the mkisofs-rs crate to operate on `Read + Write + Seek` instead of `File` for the output file is straight forward but it's not immediately clear to me how to nicely do it for input files. The Opt struct of the crate takes input directories to walk for files instead of files themselves.

I was thinking of adding a "input_files_raw" option which would be a Vec of `Read + Seek` but I'm not sure I like that way.

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,279 @@
...
+
+impl InMemoryBackend {
+    pub fn create(bytes: Vec<u8>, is_ro: bool) -> Result<Arc<Self>> {
+        let len = bytes.len();
+
+        let this = Self {
+            bytes: Arc::new(Mutex::new(bytes)),
+
+            driver: Mutex::new(None),
+            worker_count: NonZeroUsize::new(1).unwrap(),
```
pfmooney, line 33: Considering the fulfillment of IO requests amounts to `bcopy()`/`memcpy()`, do we really need multiple threads here?  Really this could probably just be an async task.

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,279 @@
...
+
+            is_ro,
+            block_size: 512,
+            sectors: len / 512,
```
pfmooney, line 37: It should probably be documented that a 512b sector is demanded here.

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,279 @@
...
+    driver: Mutex<Option<Arc<Driver>>>,
+    worker_count: NonZeroUsize,
+
+    is_ro: bool,
+    block_size: usize,
+    sectors: usize,
+}
+
+impl InMemoryBackend {
+    pub fn create(bytes: Vec<u8>, is_ro: bool) -> Result<Arc<Self>> {
```
pfmooney, line 26: Unless we're going to assert that the `bytes` length is a multiple of the sector size, is there anything here which demands a `Result`, rather than simply returning the created backend?

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,279 @@
...
+                        this.cv.notify_all();
+                    }
+                }) as Box<WakeFn>)
+            } else {
+                None
+            };
+
+            let _ = disp
+                .spawn_sync(
+                    format!("file bdev {}", i),
```
pfmooney, line 163: This is not a file-based blockdev

propolis/src/block/in_memory.rs
jmpesp: made configurable in 6030809

propolis/src/block/in_memory.rs
jmpesp: checking that now in 6030809

propolis/src/block/in_memory.rs
jmpesp: changed in dc870b1

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,279 @@
...
+
+/// Read from bytes into guest memory
+fn process_read_request(
+    bytes: &Arc<Mutex<Vec<u8>>>,
+    offset: u64,
+    len: usize,
+    mappings: &Vec<SubMapping>,
+) -> Result<()> {
+    let bytes = bytes.lock().unwrap();
+    let data = &bytes[(offset as usize)..(offset as usize + len)];
```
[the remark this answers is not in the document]
jmpesp: done in 3e5a320

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,279 @@
...
+        if inner_nwritten != mapping.len() {
+            return Err(std::io::Error::new(
+                ErrorKind::Other,
+                format!(
+                    "mapping.write_bytes failed! {} vs {}",
+                    inner_nwritten,
+                    mapping.len(),
+                ),
+            ));
+        }
```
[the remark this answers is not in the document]
jmpesp: I assumed that `read_bytes` and `write_bytes` returned the number of bytes actually performed (similar to  std::io::Read::read) because of the return type of `Result<usize>`, but looking at the code it always returns the number of bytes of the buffer. In memory and crucible changed in 9a0ca32

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    let mut bytes = bytes.lock().unwrap();
+    bytes[(offset as usize)..(offset as usize + len)].copy_from_slice(&vec);
+
+    Ok(())
+}
+
+fn process_flush_request() -> Result<()> {
+    Ok(())
+}
```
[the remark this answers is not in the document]
jmpesp: elided in 3e5a320

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,279 @@
...
+    offset: u64,
+    len: usize,
+    mappings: &Vec<SubMapping>,
+) -> Result<()> {
+    let mut vec: Vec<u8> = vec![0; len];
+
+    let mut nread = 0;
+    for mapping in mappings {
+        let inner_nread =
+            mapping.read_bytes(&mut vec[nread..(nread + mapping.len())])?;
```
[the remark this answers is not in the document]
jmpesp: no more buffering in 706ca24, that's a bunch better :)

propolis/src/block/in_memory.rs
```diff
@@ -0,0 +1,279 @@
...
+
+            process_read_request(bytes, off as u64, req.len(), &maps)?;
+        }
+        block::Operation::Write(off) => {
+            let maps = req.mappings(&mem).ok_or_else(|| {
+                Error::new(ErrorKind::Other, "bad guest region")
+            })?;
+
+            process_write_request(bytes, off as u64, req.len(), &maps)?;
+        }
```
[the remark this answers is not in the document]
jmpesp: Added this check to in_memory, file, and crucible in 294439b

propolis/src/block/in_memory.rs
jmpesp: Agreed, ripped out worker_count in c7c5632

jmpesp:
> From an architectural perspective, how much cloud-init specificity do we really want to bake in to the hypervisor itself? What happens when the format changes or we want to add features?
>
> Might it make sense to do the hsfs image generation in Nexus and PUT the resultant blob to Propolis (to be injected as raw contents into this in-memory emulated device) as part of establishing the VM?

In order to get this PR moving, I can rip out that part. This PR would then be simply adding the in-memory block backend, and the cloud-init virtio initialisation stuff.

On Nexus' side of things, I don't immediately have an idea where that blob will be stored once it's created. In my testing the ISOs are 56.0 KiB and I'm not crazy about shoving that into CRDB :)

jclulow:
> On Nexus' side of things, I don't immediately have an idea where that blob will be stored once it's created. In my testing the ISOs are 56.0 KiB and I'm not crazy about shoving that into CRDB :)

I don't think we would need to store the rendered image in the database.  It could be reconstructed each time we to place/start a VM, just from the metadata that goes into it, I would think?

jmpesp:
> I don't think we would need to store the rendered image in the database. It could be reconstructed each time we place/start a VM, just from the metadata that goes into it, I would think?

Yep, that makes more sense. Twas not good thinking on my part :)

jmpesp:
@luqmana @jclulow @pfmooney please re-review, I'm going to be in this code this week and should build on top of this :)

server/src/lib/server.rs
```diff
@@ -315,6 +324,24 @@ async fn instance_ensure(
...
+                    .map_err(|e| {
+                        Error::new(ErrorKind::InvalidData, e.to_string())
+                    })?;
+
+                let bytes = base64::decode(&cloud_init_bytes).map_err(|e| {
+                    Error::new(ErrorKind::InvalidInput, e.to_string())
+                })?;
+
+                init.initialize_in_memory_virtio_from_bytes(
+                    &chipset, &bytes, bdf, true,
```
luqmana, line 339: Given this is the only use of `initialize_in_memory_virtio_from_bytes` should we just pass the vec to avoid the subsequent `to_vec` call

luqmana (approved):
👍

server/src/lib/server.rs
jmpesp: :+1: fixed in 4b883c2

jmpesp:
Attach a disk to an instance in a new "cloud init" slot, and back it
with a new in-memory block backend. The cli has been updated with an
extra option to POST a path as base64 encoded bytes.

There's some refactoring work in here as well for block backends.

PR #97, opened 2022-02-19
Update for latest crucible changes.

luqmana:
Minimal propolis side changes to get it compiling again with https://github.com/oxidecomputer/crucible/pull/207. Also update the crucible dep to point to a specific revision (*) instead of just main so that CI doesn't break between once upstream changes and when we pick up the changes.

(*) I just picked the current `main`, feel free to suggest something else if that works better.


jmpesp (approved):

jmpesp:
+1 for tracking a rev :)

PR #103, opened 2022-03-08
Create Crucible volumes from construction requests

jmpesp:
Change DiskRequest to accept only two fields: gen and a new "volume
construction request". To avoid bringing in that type at this part of
the repo, use a generic map from String to serde_json::Value.

The new Volume::construct function accepts this volume construction
request and allows for Nexus to request arbitrarily shaped volumes. This
PR also changes from Guest to the BlockIO trait to support this.

Remove creating Crucible block backends from the stand alone or server
TOML file.

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -16,3 +16,4 @@ slog = { version = "2.5", features = [ "max_level_trace", "release_max_level_deb
 structopt = "0.3"
 thiserror = "1.0"
 uuid = { version = "0.8", features = [ "serde", "v4" ] }
+crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "4090a023b77dcab7a5000057cf2c96cdbb0469b6" }
--- a/client/src/api.rs
+++ b/client/src/api.rs
@@ -6,7 +6,7 @@
 
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use std::collections::BTreeMap;
+use std::collections::HashMap;
 use std::net::SocketAddr;
 use uuid::Uuid;
 
@@ -237,7 +237,7 @@ pub struct DiskRequest {
 
     // Crucible related opts
     pub gen: u64,
-    pub volume_construction_request: BTreeMap<String, serde_json::Value>,
+    pub volume_construction_request: HashMap<String, serde_json::Value>,
 }
 
 #[derive(Clone, Deserialize, Serialize, JsonSchema)]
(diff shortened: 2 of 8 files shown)


dancrossnyc (commented):
The HashMap feels like a bit of an implementation detail. I wonder if it could be put behind a newtype to hide the representation from downstream code?

jmpesp:
> The HashMap feels like a bit of an implementation detail. I wonder if it could be put behind a newtype to hide the representation from downstream code?

I think I understand what you mean - something like this?

```
diff --git a/client/src/api.rs b/client/src/api.rs
index 8c47d23..cd06bdf 100644
--- a/client/src/api.rs
+++ b/client/src/api.rs
@@ -228,6 +228,11 @@ pub const DISK_FLAG_WRITE: u32 = 0b0000_0010;
 pub const DISK_FLAG_READ_WRITE: u32 = DISK_FLAG_READ | DISK_FLAG_WRITE;
 type DiskFlags = u32;
 
+// Nexus will send serialized Crucible::VolumeConstructionRequest, but we don't
+// want to import that type here. Instead, accept arbitrary JSON and pass that
+// through to Crucible's Volume::construct(...).
+type GenericJson = HashMap<String, serde_json::Value>;
+
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
 pub struct DiskRequest {
     pub name: String,
@@ -237,7 +242,7 @@ pub struct DiskRequest {
 
     // Crucible related opts
     pub gen: u64,
-    pub volume_construction_request: HashMap<String, serde_json::Value>,
+    pub volume_construction_request: GenericJson,
 }
 
 #[derive(Clone, Deserialize, Serialize, JsonSchema)]
 ```

jmpesp:
> Maybe I should just import VolumeConstructionRequest here?

I'm going to do this, but it's gated by deriving Clone, Debug, and JsonSchema in https://github.com/oxidecomputer/crucible/pull/235

dancrossnyc:
> > The HashMap feels like a bit of an implementation detail. I wonder if it could be put behind a newtype to hide the representation from downstream code?
> 
> I think I understand what you mean - something like this?

Yeah, basically, but I think I would have just called it a `CreationRequest` or something.

jmpesp:
> > Maybe I should just import VolumeConstructionRequest here?
> 
> I'm going to do this, but it's gated by deriving Clone, Debug, and JsonSchema in [oxidecomputer/crucible#235](https://github.com/oxidecomputer/crucible/pull/235)

Done in c5a1281b5aa284970ceacac7459612acc0f718e1

dancrossnyc (approved):

PR #104, opened 2022-03-11
Batch serial console input bytes in CLI

jmpesp:
In serial task, batch reads from tokio::io::stdin, and send that as one
Binary Websockets message instead of one u8 at a time. This fixes the
dreaded "[6;14R" junk that was seen in serial sessions when the propolis
server was zoned.

I suspect this is due to some timeout logic somewhere in the giant stack
of software that only considers ESC [ a [Control Sequence Introducer][1]
if it's followed by the other bytes "quickly" enough.

There were *many* red herrings during this dive, including the rewriting
of the serial task to use a channel instead of for stdin.read as a
tokio::select branch - reading from tokio::io::stdin isn't cancel safe
so it shouldn't be on one of those branches.

Multiple people have looked at this, and so did I, and going into it
with the assumption that a byte was being lost somewhere caused me to
read a lot of the plumbing code surrounding the UART logic. This commit
also adds some tests for copy_and_consume because I noticed there
weren't any.

This commit also simplifies server's instance_serial_task to only use
two tokio::select branches for UART -> WS, and WS -> UART.

[1]: https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_(Control_Sequence_Introducer)_sequences

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -249,29 +249,82 @@ async fn serial(
     let _raw_guard = RawTermiosGuard::stdio_guard()
         .with_context(|| anyhow!("failed to set raw mode"))?;
 
-    let mut stdin = tokio::io::stdin();
     let mut stdout = tokio::io::stdout();
-    let mut next_raw = false;
+
+    // https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_exact
+    // is not cancel safe! Meaning reads from tokio::io::stdin are not cancel
+    // safe. Spawn a separate task to read and put bytes onto this channel.
+    let (tx, mut rx) = tokio::sync::mpsc::channel(16);
+
+    tokio::spawn(async move {
+        let mut stdin = tokio::io::stdin();
+
+        // next_raw must live outside loop, because Ctrl-A should work across
+        // multiple inbuf reads.
+        let mut next_raw = false;
+
+        loop {
+            let mut inbuf = vec![0u8; 1024];
+            let n = match stdin.read(&mut inbuf).await {
+                Err(_) | Ok(0) => break,
+                Ok(n) => n,
+            };
+            inbuf.truncate(n);
+
+            // Put bytes from inbuf to outbuf, but don't send Ctrl-A unless
+            // next_raw is true.
+            let mut outbuf = vec![];
+
+            let mut exit = false;
+            for c in inbuf {
+                match c {
+                    // Ctrl-A means send next one raw
+                    b'\x01' => {
+                        if next_raw {
+                            // Ctrl-A Ctrl-A should be sent as Ctrl-A
+                            outbuf.push(c);
+                            next_raw = false;
+                        } else {
+                            next_raw = true;
+                        }
+                    }
+                    b'\x03' => {
+                        if !next_raw {
+                            // Exit on non-raw Ctrl-C
+                            exit = true;
+                            break;
+                        } else {
+                            // Otherwise send Ctrl-C
+                            outbuf.push(c);
+                            next_raw = false;
+                        }
+                    }
+                    _ => {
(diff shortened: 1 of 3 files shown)


jclulow:
What was the latency between the `ESC` and the `[` arriving in the guest?!

jmpesp:
~10ms

I only suspected timeout logic because batching fixed it, but maybe that's also incorrect.

jclulow:
I suspect we're in the busybox line editing library, around:

https://github.com/mirror/busybox/blob/master/libbb/lineedit.c#L1855-L1891

```
static void ask_terminal(void)
...
		S.sent_ESC_br6n = 1;
		fputs_stdout(ESC"[6n");
		fflush_all(); /* make terminal see it ASAP! */
...
```

This happens right after it prints the prompt:

https://github.com/mirror/busybox/blob/master/libbb/lineedit.c#L2525-L2527

Oddly enough the process of reading an entire escape sequence appears to happen within the function helpfully named `read_key()`:

https://github.com/mirror/busybox/blob/master/libbb/read_key.c#L149-L182

They do acknowledge that portions of an escape sequence may not arrive in the same `read(2)` call, and appear to be willing to wait for 50 of your earth time units between reads:

```
...
				/* Need more chars, read another one if it wouldn't block.
				 * Note that escape sequences come in as a unit,
				 * so if we block for long it's not really an escape sequence.
				 * Timeout is needed to reconnect escape sequences
				 * split up by transmission over a serial console. */
				if (safe_poll(&pfd, 1, 50) == 0) {
					/* No more data!
					 * Array is sorted from shortest to longest,
					 * we can't match anything later in array -
					 * anything later is longer than this seq.
					 * Break out of both loops. */
					goto got_all;
				}
```

Though I cannot see it explicitly written anywhere, I imagine the units are actually milliseconds, so if the two bytes arrived inside 10ms I would hope it would put them together into a sequence.  But who knows.

cli/src/main.rs
```diff
@@ -249,29 +249,82 @@ async fn serial(
...
+
+    tokio::spawn(async move {
+        let mut stdin = tokio::io::stdin();
+
+        // next_raw must live outside loop, because Ctrl-A should work across
+        // multiple inbuf reads.
+        let mut next_raw = false;
+
+        loop {
+            let mut inbuf = vec![0u8; 1024];
```
luqmana, line 267: This could just be a stack allocated array right. We don't really need to create a new vec evry iteration

propolis/src/chardev/pollers.rs
```diff
@@ -463,6 +463,93 @@ fn copy_and_consume(src: &mut Vec<u8>, dest: &mut [u8]) -> usize {
     }
 }
 
+#[test]
+fn test_copy_and_consume_1() {
```
luqmana, line 467: 💯️ on tests. Thanks!

cli/src/main.rs
jmpesp: addressed in 34e3329a53193e9bb90867176a732428e1e3fdc6

server/src/lib/server.rs
```diff
@@ -709,39 +687,39 @@ async fn instance_serial_task(
...
+
+                        while i < total {
+                            let nwritten: Option<usize> =
+                                serial.write_sink(&input[i..total], actx).await;
+
+                            match nwritten {
+                                None => {
+                                    // XXX write_sink should always return Some,
+                                    // according to the code.
+                                    panic!("write_sink should always return Some!");
```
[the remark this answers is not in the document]
jmpesp: I reverted the server changes in e1d4a542b81a91fe19388baf3f5037443f8c4349

jmpesp:
I just re-tested with reverted server changes, and it looks like only the cli changes are required to address this bug.

luqmana (approved):

PR #105, opened 2022-03-14
Let propolis-server package itself

smklein:
This change enables propolis-server to publish a zone image, rather than relying on external repos (like [omicron](https://github.com/oxidecomputer/omicron)) to do that themselves.

- This change adds a new package in the Cargo workspace named `propolis-package`, which has the sole responsibility of bundling up a `propolis-server` into the [omicron-brand zone](https://github.com/oxidecomputer/helios-omicron-brand) format.
- This also moves the SMF files and "package manifest" directly from Omicron - they are now "owned" by the Propolis repo.

This change has parity with our packaging for other zone images created outside Omicron, [similar to Crucible](https://github.com/oxidecomputer/crucible/tree/main/package).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,14 @@ members = [
   "standalone",
 ]
 
+default-members = [
+  "cli",
+  "client",
+  "propolis",
+  "server",
+  "standalone",
+]
+
 [profile.dev]
 panic = "abort"
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,9 +1,10 @@
 [workspace]
 
 members = [
-  "propolis",
   "cli",
   "client",
+  "package",
+  "propolis",
   "server",
   "standalone",
 ]
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -1,0 +1,9 @@
+[package.propolis-server]
+rust.binary_names = ["propolis-server"]
+rust.release = true
+service_name = "propolis-server"
+zone = true
+blobs = [ "alpine.iso", "OVMF_CODE.fd" ]
+[[package.propolis-server.paths]]
+from = "smf/propolis-server"
+to = "/var/svc/manifest/site/propolis-server"
--- a/package/src/main.rs
+++ b/package/src/main.rs
@@ -1,0 +1,19 @@
+// Copyright 2022 Oxide Computer Company
+
+use anyhow::Result;
+use std::fs::create_dir_all;
+use std::path::Path;
+
+#[tokio::main]
+async fn main() -> Result<()> {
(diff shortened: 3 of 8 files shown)


smklein:
Previously, Omicron had been building the propolis-server binary itself, which was *kiiiinda* hacky. It depended on it as a `dependency` in `Cargo.toml`, which let us build the binary... but I think that only worked as a side-effect the fact that `propolis-server` also exposes a library.

By distributing the building/packaging process, that hack should no longer matter.

package-manifest.toml
```diff
@@ -0,0 +1,9 @@
+[package.propolis-server]
```
smklein, line 1: https://github.com/oxidecomputer/omicron/blob/d41d070528536cf8e2e75eb08e77de7ebcc29245/package-manifest.toml#L35-L43

smf/propolis-server/config.toml
```diff
@@ -0,0 +1,24 @@
+# Configuration for propolis server.
```
smklein, line 1: https://github.com/oxidecomputer/omicron/blob/d41d070528536cf8e2e75eb08e77de7ebcc29245/package-manifest.toml#L35-L43

smklein, line 1: https://github.com/oxidecomputer/omicron/blob/main/smf/propolis-server/config.toml

Cargo.toml
```diff
@@ -1,9 +1,10 @@
 [workspace]
 
 members = [
-  "propolis",
   "cli",
   "client",
+  "package",
```
pfmooney, line 6: I request that we exclude this from the `default-members` of the workspace, so it's not built by default when iterating on propolis on a dev machine.

smklein: Done - added a `default-members` to the workspace `Cargo.toml`.

Tested by inserting some junk characters in `package/src/main.rs`:
- `cargo build` still compiles fine
- `cargo run -p propolis-package` hits the (expected) compiler error

pfmooney: Thanks!

package/src/main.rs
```diff
@@ -0,0 +1,19 @@
...
+#[tokio::main]
+async fn main() -> Result<()> {
+    let cfg = omicron_zone_package::config::parse("package-manifest.toml")?;
+
+    let output_dir = Path::new("out");
+    create_dir_all(&output_dir)?;
+
+    for package in cfg.packages.values() {
+        package.create(output_dir).await?;
+    }
```
luqmana, line 16: Do we expect to include multiple packages in propolis' manifest?

smklein: I think at the moment we're just expecting the single package out of propolis - iterating over all packages in the manifest is a bit future-proofy.

luqmana (approved):
Thanks for setting this up!

PR #106, opened 2022-03-15
Generalize Outstanding I/Os waiting logic for devices

### LINKED ISSUE #98: Generalize Outstanding I/Os waiting logic for devices
As part of https://github.com/oxidecomputer/propolis/pull/92 I added some logic to wait for outstanding I/O requests to the NVMe device while pausing it during migration. Similar logic is needed for other devices and would ideally be shared. Opening this issue to track that follow up work.

luqmana:
As part of https://github.com/oxidecomputer/propolis/pull/92 I added some logic to wait for outstanding I/O requests to the NVMe device while pausing it during migration. This generalizes the logic to apply to any emulated block device.

Also cleaned up some duplication amongst the block backends.

Fixes #98

### DIFF
--- a/propolis/src/block/mod.rs
+++ b/propolis/src/block/mod.rs
@@ -2,13 +2,14 @@
 
 use std::collections::VecDeque;
 use std::num::NonZeroUsize;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::{Arc, Condvar, Mutex};
 
 use crate::common::*;
 use crate::dispatch::{AsyncCtx, DispCtx, Dispatcher, SyncCtx, WakeFn};
 use crate::vmm::{MemCtx, SubMapping};
 
+use futures::future::BoxFuture;
 use tokio::sync::{Notify, Semaphore};
 
 mod file;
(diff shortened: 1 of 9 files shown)


propolis/src/block/mod.rs
```diff
@@ -46,9 +49,31 @@ pub type CompleteFn =
...
+
+    /// A shared notifier to indicate to the block backend handling this request once
+    /// all outstanding requests have been completed (i.e. `outstanding_reqs` hits 0).
+    /// This is optional because we don't always want to be notified when we hit 0
+    /// outstanding requests but only if the device has been paused.
+    ///
+    /// The `Request::new_*` methods explicitly initialize these but it is the backend
+    /// Notifier that will actually replace it with the correct reference.
+    /// See [`Request::attach_outstanding_notifier`].
+    outstanding_notifier: Arc<Mutex<Option<Arc<Notify>>>>,
```
pfmooney, line 76: Perhaps wrap the outstanding notification state into a purpose-built struct?  I think it would help simplify reasoning about it.

Then it could be something like
```rust
pub struct Request {
...
    outstanding_req_state: Option<Arc<OutstandingReqs>>
...
}

struct Outstanding Reqs {
    count: AtomicU64,
    notifier: Notify,
}
```

propolis/src/block/mod.rs
```diff
@@ -140,38 +205,83 @@ pub trait Device: Send + Sync + 'static {
...
+
         self.armed.store(false, Ordering::Release);
-        let res = nextf();
-        if res.is_some() {
-            // Since a result was successfully retrieved, no need to rearm the
-            // notification trigger.
-            return res;
+        if let Some(mut req) = nextf() {
+            // Update outstanding count and update `Request` to track it
+            self.outstanding_reqs.fetch_add(1, Ordering::Relaxed);
```
pfmooney, line 261: Is `Relaxed` ordering correct here?

propolis/src/block/mod.rs
```diff
@@ -188,32 +298,211 @@ impl Notifier {
...
     }
+
+    /// Stop accepting requests from the block device.
+    ///
+    /// Given there might be in-flight requests being handled by the backend,
+    /// the `Notifier::paused` method returns a future indicating
+    /// when the pause operation is complete.
+    pub fn pause(&self) {
+        // Stop responding to any requests
+        assert!(!self.paused.swap(true, Ordering::Release));
```
pfmooney, line 309: style nit: Could the result from the swap be stored in a local prior to the `assert!()`?  I think it makes it easier to read though the logic while consider the asserts just additional safety around the expected invariants.

pfmooney (commented):
Thanks for taking a swing at this

propolis/src/block/mod.rs
```diff
@@ -188,32 +298,211 @@ impl Notifier {
...
+
+        // Create a new `Notify` object and stash it in the shard reference
```
jmpesp, line 311: typo here

jmpesp (approved):
:shipit:

propolis/src/block/mod.rs
luqmana: I believe so, we just want to atomicity of the increment here and do nothing with the old value. There's no other memory location we're trying to synchronize with if a thread gets to this point before it returns from this method.

propolis/src/block/mod.rs
luqmana: Updated PR

PR #108, opened 2022-03-23
Clean up Dispatcher and self-referential bits

pfmooney:
The role of the Dispatcher code has diminished as more of the emulation bits are addressed by async rust.  As some have pointed out, this can make things appear more complicated than necessary.  Cleaning some of that up should help.


dancrossnyc (approved):
This change gives me joy.

propolis/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(clippy::style)]
+#![feature(arc_new_cyclic)]
```
jordanhendricks, line 2: 🎉

jordanhendricks (approved):
Thanks a lot! The dispatch stuff is much easier to follow now.

One small comment but otherwise LGTM 👍

propolis/src/vmm/machine.rs
```diff
@@ -188,8 +188,8 @@ pub struct MachineCtx {
 }
 
 impl MachineCtx {
```
luqmana, line 190: Doesn't have to be changed here, but is there any reason we still have `MachineCtx` which seems to just wrap an `Arc<Machine>`? `Machine` is already `pub`.

propolis/src/block/crucible.rs
```diff
@@ -31,27 +31,19 @@ pub struct CrucibleBackend {
...
     }
 
     fn _create(
-        disp: &Dispatcher,
         gen: u64,
         request: VolumeConstructionRequest,
         read_only: bool,
     ) -> anyhow::Result<Arc<Self>, crucible::CrucibleError> {
-        slog::info!(
-            disp.logger(),
```
luqmana, line 50: Should we provide some other way for the backend here to get a `logger`?

pfmooney: It has access to one via `SyncCtx` and `AsyncCtx` objects during operation.  The calling code was doing logging itself, so it seemed unnecessary to expose one here.

propolis/src/vmm/machine.rs
pfmooney: The idea behind the distinction, at least for now, is that `MachineCtx` exposes normal "VM is running" operations, such as access to memory or the mmio/pio buses.  More uncommon operations (like directly registering new listeners on said buses) is done either during init, or through the `Chipset`, but critically, not from just any old place where you have access to a `MachineCtx` (say, though a `DispCtx`)

propolis/src/dispatch/mod.rs
```diff
@@ -22,45 +20,31 @@ pub use sync_tasks::*;
...
 }
 
 impl Dispatcher {
     /// Creates a new dispatcher.
     ///
     /// # Arguments
-    /// - `vm`: The machine for which the dispatcher will handle requests.
     /// - `vcpu_fn`: A function, which will be invoked by the dispatcher,
     /// to run the CPU. This function is responsible for yielding control
     /// back to the dispatcher when requested.
```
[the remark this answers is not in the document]
pfmooney: Done

luqmana (approved):

pfmooney:
Merged

PR #111, opened 2022-04-01
Disambiguate block backend instances in inventory.

luqmana:
At some point I updated the device inventory to return an error if you tried to add the same instance twice. It does this by keeping track of what's been added so far based on the type name (via `Entity::type_name`) which is generic to all instances of that type and the instance name, which is an optional parameter to the inventory methods used for adding items.

There are certain kinds of devices we only ever expect a single instance of to be present and thus never give instance-specific names to. The block backends were mistakenly being treated as such and thus if you ever had more than one it would fail upon trying to instantiate the second one, e.g. you couldn't have two disks both backed by a file even if they were using separate files.

The fix is simple enough: fill in the instance-specific portions for each of the block backends we support today:
1. **File**: we simply use the file path for the instance name. (*)
2. **Crucible**: use the UUID associated with the specific crucible backend.
3. **In-Memory**: require specifying an instance name up front (just `"cloud-init"` today).

(*) Note, it is still possible to run into conflicts with the File backend if one uses the same file for 2 different backends. For the most part I don't think this is a scenario that's necessary to support.

### DIFF
--- a/propolis/Cargo.toml
+++ b/propolis/Cargo.toml
@@ -28,6 +28,7 @@ serde = { version = "1" }
 serde_arrays = "0.1"
 erased-serde = "0.3"
 serde_json = "1.0"
+uuid = "0.8"
 
 [dev-dependencies]
 crossbeam-channel = "0.5"
--- a/propolis/src/block/crucible.rs
+++ b/propolis/src/block/crucible.rs
@@ -71,6 +71,11 @@ impl CrucibleBackend {
 
         Ok(Arc::new(be))
     }
+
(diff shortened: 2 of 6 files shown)


jmpesp (approved):
Approved, with the non-blocking question/suggestion: did you consider making an additional trait function for this?

luqmana:
> Approved, with the non-blocking question/suggestion: did you consider making an additional trait function for this?

I'm still noodling around with different ideas on how to structure the inventory but haven't had much time to dig on that lately. This is to just make sure you can actually start propolis with a valid config :P

PR #112, opened 2022-04-08
Don't send empty websockets messages

jmpesp:
After reworking this code in a46882a4b3d57715b78fa99311ecba65a83b7069,
sending a "raw" ctrl-c by pressing ctrl-a first stopped working. Fix
this by 1) making the current code unit testable, 2) realizing that the
tests all pass but the bug still exists, and finally 3) realizing that
sending the empty websockets message may be the problem.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -230,6 +230,118 @@ async fn put_instance(
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


luqmana (approved):
even more tests! 🎊

PR #114, opened 2022-04-15
no longer can we only boot alpine

jmpesp:
Remove alpine block dev so that after the global images PR lands, there
isn't a boot device conflict.

### DIFF
--- a/smf/propolis-server/config.toml
+++ b/smf/propolis-server/config.toml
@@ -5,16 +5,6 @@
 
 bootrom = "/opt/oxide/propolis-server/blob/OVMF_CODE.fd"
 
-[block_dev.alpine_iso]
-type = "file"
-path = "/opt/oxide/propolis-server/blob/alpine.iso"
-readonly = "true"
-
-[dev.block0]
-driver = "pci-virtio-block"
-block_dev = "alpine_iso"
-pci-path = "0.4.0"
-
 # NOTE: This VNIC is here for reference, but VNICs are typically managed by the
 # Sled Agent.
 


smklein:
Finallly!

smklein (approved):

jmpesp:
I'm wary of merging this, as it will disrupt other's workflows - anyone have thoughts about this?

smklein:
Do we have images that we can host in the lab / on s3? As long as we make it low-effort to "pick an easy default", I think moving forward here seems like the right move.

FWIW, by changing the SMF file, we're basically modifying the expectations for "propolis when run by the control plane", but not really changing the developer experience for "propolis when built/run standalone".

jmpesp:
> Do we have images that we can host in the lab / on s3? As long as we make it low-effort to "pick an easy default", I think moving forward here seems like the right move.
> 
> FWIW, by changing the SMF file, we're basically modifying the expectations for "propolis when run by the control plane", but not really changing the developer experience for "propolis when built/run standalone".

You're right w.r.t running via the control plane vs standalone. The problem here is that files can be hosted in the lab or s3 but Nexus needs to construct the volume out of something. Developers will have different image definitions for their house vs the lab but if they didn't set that up this PR means they can't boot as easily anymore.

Thinking about this more, what we may need in addition to this PR is documenting how to add `/opt/oxide/propolis-server/blob/alpine.iso` the "global images" table " as a (not yet implemented) file source. [Crucible supports this](https://github.com/oxidecomputer/crucible/blob/ea22352bc3f666860e610f32028f9e76a2ea027b/upstairs/src/block_io.rs#L8-L15), we (read: I) just need to teach Nexus about it. That would be the new easy default :)

smklein:
I think I mostly agree with that reasoning - I'd add, however, that if the "create image from URL" path works well, it would be easy to just host whatever images we want at a well-known location. We could even pre-populate Nexus with this info on startup.

That seems like it would avoid the need for developers to download any images "out-of-band", they'd just get 'em through the normal "download as URL" mechanism.

jmpesp:
> Thinking about this more, what we may need in addition to this PR is documenting how to add /opt/oxide/propolis-server/blob/alpine.iso the "global images" table " as a (not yet implemented) file source. [Crucible supports this](https://github.com/oxidecomputer/crucible/blob/ea22352bc3f666860e610f32028f9e76a2ea027b/upstairs/src/block_io.rs#L8-L15), we (read: I) just need to teach Nexus about it.

Opened https://github.com/oxidecomputer/omicron/pull/975 to do this.

PR #116, opened 2022-04-15
buildomat: build tests in AWS and run them in the lab

jclulow:


jclulow:
This adds two new buildomat jobs, using a new feature: job dependencies.  The first job, `tests-build`, will build test binaries with `cargo` and compress them and emit them as job artefacts:

tests-build

The second job, `tests-run`, depends on the first using a `dependencies` directive:

```
 $ awk '!/^#/ { exit(0); } { print; }' .github/buildomat/jobs/tests-run.sh
#!/bin/bash
#:
#: name = "tests-run"
#: variety = "basic"
#: target = "lab"
#: output_rules = [
#:      "/tmp/*.log",
#: ]
#: skip_clone = true
#:
#: [dependencies.build]
#: job = "tests-build"
#:
```

The directive creates a dependency, named `build`, which resolves to the first job by its name.  The name of the dependency appears in the path within the work environment when output artefacts are coped as inputs to the second job; i.e., they appear under `/input/build/` in this case.

If the first job succeeds, the second job will run:

testsrun

Things like cancelling one or the other, or job failure, should all end up cascading through the chain.  Any log files emitted as `/tmp/*.log` within the test work environment will be saved and available to view, whether or not the tests succeed.

jclulow:
OK I rebased this and there is a new failure, so: it seems time to put it back so that we can avoid repeated new failures!

PR #117, opened 2022-04-19
Use less IDs in paths

smklein:
## Old Behavior

For most endpoints, we'd use the format: 

```
http://<IP ADDRESS>/instances/<INSTANCE UUID>/<rest of the path, maybe>
```

For many endpoints, this UUID would be redundant with values stored in `properties` struct, and a somewhat silly check would validate this. This path format added little value, other than requiring clients to supply a redundant value.

Why is the UUID redundant? Propolis' server only supports a *single* instance anyway, so the server address should be a sufficient factor for distinguishing instances.

## New Behavior

Don't bother supplying the instance UUID. Simplify a lot of "name -> UUID" pathways where this is no longer necessary.

### DIFF
--- a/server/src/lib/server.rs
+++ b/server/src/lib/server.rs
@@ -3,7 +3,7 @@
 use anyhow::Result;
 use dropshot::{
     endpoint, ApiDescription, HttpError, HttpResponseCreated, HttpResponseOk,
-    HttpResponseUpdatedNoContent, Path, RequestContext, TypedBody,
+    HttpResponseUpdatedNoContent, RequestContext, TypedBody,
 };
 use futures::future::Fuse;
 use futures::{FutureExt, SinkExt, StreamExt};
@@ -25,7 +25,6 @@ use tokio_tungstenite::tungstenite::{
     self, handshake, protocol::Role, Message,
 };
 use tokio_tungstenite::WebSocketStream;
-use uuid::Uuid;
 
 use propolis::bhyve_api;
 use propolis::dispatch::AsyncCtx;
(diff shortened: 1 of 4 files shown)


server/src/lib/server.rs
```diff
@@ -561,37 +553,10 @@ async fn instance_ensure(
...
-        HttpError::for_internal_error(
-            "Server not initialized (no instance)".to_string(),
-        )
-    })?;
-
-    if path_params.into_inner().instance_id != context.properties.name {
-        return Err(HttpError::for_internal_error(
-            "Instance name mismatch (path did not match struct)".to_string(),
-        ));
-    }
```
smklein, line 583: This use of the hostname was somewhat odd - basically, we used the client-provided hostname here, for a path which should only be accessible to the control plane.

I get why we used it - it acted as a convenient way to get the redundant UUID - but we should probably avoid referring to the instance by hostname for internal purposes, as that's could theoretically change.

Regardless, that problem no longer exists after this PR.

smklein: Side-note, I was definitely confused by seeing `path = ... {instance_id}`, when the variable is actually mean to be a string-based hostname.

smklein:
Tested in Omicron by: https://github.com/oxidecomputer/omicron/pull/950

server/src/lib/server.rs
```diff
@@ -189,28 +188,21 @@ fn slot_to_bdf(slot: api::Slot, ty: SlotType) -> Result<pci::Bdf> {
 
 #[endpoint {
     method = PUT,
-    path = "/instances/{instance_id}",
+    path = "/instances",
```
pfmooney, line 191: Does the plural `instances` make sense to preserve, considering propolis is designed to only ever run a single instance at once?

smklein: Updated to `instance` everywhere

pfmooney: Thanks for cleaning this all up.

luqmana: Missing this one in the last commit?

smklein: Okay, *now* updated to `instance` everywhere

luqmana (approved):

PR #121, opened 2022-04-22
update crucible revs

jmpesp:
fixes build conflict with tracing dep


jclulow:
FYI: the `tests-run` job will remain queued until probably Friday night, as the office has been powered down for maintenance.

jmpesp:
That's right! Also unfortunate.

PR #123, opened 2022-04-26
Update hardware IDs to match current assignments in OANA.

luqmana:
Update propolis to reflect the current set of IDs assigned in [OANA (Oxide Assigned Numbers Authority)](https://github.com/oxidecomputer/oana).

### DIFF
--- a/propolis/src/hw/nvme/mod.rs
+++ b/propolis/src/hw/nvme/mod.rs
@@ -3,6 +3,7 @@ use std::mem::size_of;
 use std::sync::{Arc, Mutex, MutexGuard};
 
 use crate::dispatch::DispCtx;
+use crate::hw::ids::OXIDE_OUI;
 use crate::hw::pci;
 use crate::migrate::{Migrate, Migrator};
 use crate::util::regmap::RegMap;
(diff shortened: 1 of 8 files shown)


propolis/src/hw/virtio/pci.rs
```diff
@@ -186,10 +185,10 @@ impl PciVirtioState {
         cfg_sz: usize,
     ) -> (Self, pci::DeviceState) {
         let mut builder = pci::Builder::new(pci::Ident {
-            vendor_id: VIRTIO_VENDOR,
+            vendor_id: VENDOR_VIRTIO,
             device_id: dev_id,
-            sub_vendor_id: VIRTIO_VENDOR,
-            sub_device_id: dev_id - 0xfff,
+            sub_vendor_id: VENDOR_OXIDE,
+            sub_device_id: super::bits::virtio_sub_dev_id(dev_id),
```
pfmooney, line 191: These could probably just be passed through as parameters from the specific device?  We could then make them configurable later, if it was required for something like testing.

propolis/src/hw/nvme/mod.rs
```diff
@@ -422,16 +424,14 @@ pub struct PciNvme {
 impl PciNvme {
     /// Create a new pci-nvme device with the given values
     pub fn create(
-        vendor: u16,
-        device: u16,
```
pfmooney, line 426: Why not keep this stuff as a parameter?

luqmana: NVMe was the only device that took them as parameters so I was just matching everywhere else.

propolis/src/hw/virtio/pci.rs
luqmana: Fair enough, but we don't have any such use cases today? What cases do you see it being useful to configure the IDs in a test?

pfmooney (approved):

PR #129, opened 2022-05-17
Add initial support for PCI-PCI bridges

gjcolombo:
### Summary & usage

Stand up basic PCI-PCI bridge support by adding two components to the PCI module:

- The `Router` maps from a bus number (as supplied during a config space access) to the bus currently addressed by that number (based on how the guest has programmed the system's PCI bridges). The chipset holds a reference to a router that it uses when dispatching PCI configuration traffic.
- The `Bridge` implements a PCI-PCI bridge device, including the logic needed to represent a type 1 PCI config space header. Writing a bus number to a bridge's secondary bus register updates the bridge's router to refer subsequent uses of that bus number to the bridge's downstream bus.

Update the server's config structures to allow a server TOML to specify PCI bridges. Each bridge has a logical downstream bus number that other devices (including other bridges) can use in their own PCI paths. For example, this topology creates one bridge attached to the root bus, a second bridge downstream of the first bridge, and an NVMe device downstream of the second bridge:

```
[[pci_bridge]]
pci-path = "0.15.0"
downstream-bus = 13

[[pci_bridge]]
pci-path = "13.1.0"
downstream-bus = 16

[dev.block1]
driver = "pci-nvme"
block_dev = "disk1"
pci-path = "16.7.0"
```

### Tests

Booted illumos and Windows guests with storage devices sitting behind multiple PCI bridges; verified that these devices are enumerated correctly.

### Future work/known issues

To limit this change's scope (somewhat--it's still a lot of new code!) the following problems are left for another day:

- Bridges should be migratable in principle (copy their registers and re-establish routing as part of device startup on the target), but this change doesn't implement a migration handler for them.
- Bridges currently have no capabilities (including MSI capabilities or PCIe-related capabilities).


propolis/src/hw/chipset/i440fx.rs
```diff
@@ -178,10 +186,19 @@ impl I440Fx {
...
+    fn pci_root_bus(&self) -> &pci::Bus {
+        &self.pci_bus
+    }
+    fn pci_router(&self) -> &Arc<pci::router::Router> {
+        &self.pci_router
+    }
+    fn pci_attach(
+        &self,
+        bus: &pci::Bus,
+        bdf: Bdf,
```
pfmooney, line 198: Having both the `&Bus` and `Bdf` here is a little weird to me.  If we're attaching a device in the context of the chipset, it makes sense to me that it (the chipset) would at that point have knowledge of the bridge topology at that point, so only the `Bdf` would be required.

Alternatively, if we're in some future state where a bridge (and associated bus) is attached on the fly later, then devices might be attached directly to it (via a `Bus` ref).

What's your take on it, considering you've spend more time researching/thinking about the topic?

propolis/src/hw/pci/device.rs
```diff
@@ -35,6 +35,17 @@ pub trait Device: Send + Sync + 'static {
             }
         }
     }
+    fn std_cfg_rw(&self, mut rwo: RWOp, ctx: &DispCtx)
+    where
+        Self: std::marker::Sized,
```
pfmooney, line 40: What required this bound, out of curiosity?

propolis/src/hw/pci/device.rs
```diff
@@ -53,13 +64,8 @@ impl<D: Device + Send + Sync + 'static> Endpoint for D {
...
         let ds = self.device_state();
-        ds.cfg_space.process(&mut rwo, |id, mut rwo| match id {
-            CfgReg::Std => {
-                STD_CFG_MAP.process(&mut rwo, |id, rwo| match rwo {
-                    RWOp::Read(ro) => ds.cfg_std_read(id, ro, ctx),
-                    RWOp::Write(wo) => ds.cfg_std_write(self, id, wo, ctx),
-                });
-            }
+        ds.cfg_space.process(&mut rwo, |id, rwo| match id {
+            CfgReg::Std => self.std_cfg_rw(rwo, ctx),
```
pfmooney, line 68: I'm not entirely certain about the necessity of this rearranging.  I imagined that a more fully fleshed out bridge device might impl `Endpoint` directly, rather than `Device`, since it'd be doing some things differently around its BARs and interrupts.

propolis/src/hw/chipset/i440fx.rs
gjcolombo: I like the second approach, so much so that I implemented something like it in an earlier draft of this PR (but then excised it from this branch to limit this PR's scope). I pushed the commits to [gjcolombo/pci-bus-locations](https://github.com/oxidecomputer/propolis/tree/gjcolombo/pci-bus-locations) if you'd like to take a peek (needs cleanup before it can merge but should give you the gist).

I think a more general problem with bridges is that buses now have two names: the "logical" name/number the bus has in the bridge configuration in the TOML (or that you specify when you attach the bridge via the API, or what-have-you) and the number used to route PCI config traffic to the bus at runtime, which aren't guaranteed to be the same (and raises questions like "is the stored number in a `Bus` the former or the latter?"). So it makes a lot of sense to me to separate the questions of "how do I identify a specific bus?" from "given some bus, how do I address a specific device/function on that bus?", especially since you might not always use a `Bdf` in the former case (e.g. the server might instead refer to a bus by some logical name or entity ID because that's what's known to its client).

propolis/src/hw/pci/bridge.rs
```diff
@@ -0,0 +1,432 @@
...
+    pci_state: DeviceState,
+    inner: Mutex<Inner>,
+}
+
+impl Bridge {
+    /// Construct a new PCI bridge with the supplied downstream bus. Updating
+    /// the bridge's secondary bus number will update the supplied router such
+    /// that it maps the new bus number to the bridge's downstream bus.
+    pub fn new(bus: Arc<Bus>, router: Arc<Router>) -> Arc<Self> {
+        let builder = Builder::new(super::Ident {
```
gjcolombo, line 113: Need to add the device ID here too now that we've allocated one for bridges.

propolis/src/hw/chipset/i440fx.rs
```diff
@@ -178,10 +186,19 @@ impl I440Fx {
...
+        dev: Arc<dyn pci::Endpoint>,
+    ) {
+        bus.attach(bdf, dev, Some(self.route_lintr(&bdf)));
```
gjcolombo, line 201: The selection of a local interrupt pin should be contingent on this being bus 0 (and possibly `route_lintr` should assert that it is), since the bus/pin relationships are described via ACPI tables and the default tables we're using will only describe them for the root bus.

propolis/src/hw/pci/bridge.rs
```diff
@@ -110,41 +100,218 @@ impl Bridge {
...
+                | StdCfgReg::SubDeviceId
+                | StdCfgReg::ProgIf
+                | StdCfgReg::RevisionId
+                | StdCfgReg::HeaderType
+                | StdCfgReg::CapPtr
+                | StdCfgReg::CacheLineSize
+                | StdCfgReg::LatencyTimer
+                | StdCfgReg::Bist
+                | StdCfgReg::Reserved => {}
+                StdCfgReg::Command => {
```
pfmooney, line 222: style nit: a newline between `Command` and the ignored writes case would make it easier to tell them apart visually

gjcolombo:
The latest commits address most of the feedback above. Highlights:

- Bridges no longer `impl pci::Device`; they just `impl Endpoint` and avoid delegating anything to `pci::device::DeviceState`. This returns the Device/Endpoint trait relationships to what they were before these changes.
- Some of the logic in `device.rs` related to building configuration spaces is now in a small separate module so that bridges can share it (I anticipate they'll need to add capabilities to their maps in short order). **N.B.** This is not a pure refactoring; the new builder has a little bit of extra logic to try to detect region conflicts that wasn't present in the old code.
- Buses are now agnostic to their own numbers.

Work left to do:

- The comment above about local interrupt pins is not fully addressed yet, because...
- I'm looking to see if there's a better way to arrange elements in a PCI topology such that in the future, it will be easier to refer to bridges' downstream buses by the "logical" names they were given when they were specified (such that a client could more easily refer to one in an API call to a running instance). Currently, the logical numbers assigned to buses are read once in server init and then totally discarded; after talking with pfmooney offline earlier this week, I suspect we can do somewhat better here. The `route_lintr` comment's disposition seems likely to depend on what I end up doing here.

propolis/src/hw/pci/device.rs
```diff
@@ -996,10 +984,7 @@ pub struct Builder {
     lintr_req: bool,
     msix_cfg: Option<Arc<MsixCfg>>,
     bars: [Option<BarDefine>; 6],
-    cfgmap: RegMap<CfgReg>,
-
-    cap_next_alloc: usize,
-    caps: Vec<Cap>,
+    cfg_builder: CfgBuilder,
```
pfmooney, line 987: Good stuff, splitting this out

standalone/src/main.rs
```diff
@@ -241,7 +245,11 @@ fn main() {
                         vnic_name, 0x100, &hdl,
                     )?;
                     inv.register_instance(&viona, bdf.to_string())?;
-                    chipset.pci_attach(chipset.pci_root_bus(), bdf, viona);
+                    chipset.pci_attach(
+                        chipset.pci_root_bus(),
```
pfmooney, line 249: If the bus in the parsed `bdf` is non-0, then the result of attaching to the root bus anyways (instead of throwing some sort of error) is surprising.

propolis/src/hw/pci/bridge.rs
```diff
@@ -110,41 +100,218 @@ impl Bridge {
...
+                    "Unexpected read of bridge config space with ID {:?}",
+                    id
+                )
+            }
+        });
+    }
+
+    fn bar_rw(&self, _bar: BarN, _rwo: RWOp, _ctx: &DispCtx) {
+        // The BARs in the PCI bridge are read-only, so nothing should ever
+        // try to dispatch an I/O to a region defined in a bridge's BAR.
```
pfmooney, line 313: nit: since the BARs are read-only, they shouldn't actually be defining any regions

propolis/src/hw/pci/bridge.rs
```diff
@@ -110,41 +100,218 @@ impl Bridge {
...
-    fn std_cfg_rw(&self, mut rwo: RWOp, ctx: &DispCtx) {
-        CFG_HEADER_MAP.process(&mut rwo, |id, rwo| match rwo {
-            RWOp::Read(ro) => match id {
-                CfgReg::Common(StdCfgReg::HeaderType) => {
-                    ro.write_u8(HEADER_TYPE_BRIDGE)
+    fn cfg_std_read(&self, id: &BridgeReg, ro: &mut ReadOp, _ctx: &DispCtx) {
+        match id {
+            BridgeReg::Common(id) => match id {
+                StdCfgReg::VendorId => ro.write_u16(BRIDGE_VENDOR_ID),
+                StdCfgReg::DeviceId => ro.write_u16(BRIDGE_DEVICE_ID),
```
pfmooney, line 125: We've generally made vendor and device IDs (and their subordinate compatriots) configurable at create time. (see `PciNvme`)

pfmooney:
I think we can punt on making `route_lintr` fancy.  Legacy pin interrupts, while useful in the context of an old chipset like 440fx, are probably not going to be a focus for PCIe testing and other bits in propolis.

propolis/src/hw/pci/bridge.rs
gjcolombo: Done.

propolis/src/hw/pci/bridge.rs
gjcolombo: Updated the comment here--let me know if you had something else in mind.

standalone/src/main.rs
gjcolombo: Done.

propolis/src/hw/pci/device.rs
gjcolombo: Thanks!

propolis/src/hw/pci/bridge.rs
gjcolombo: Done.

gjcolombo:
The newest commits (from 54e87d3 onward) refactor some things to clean up propolis-server initialization. Before these commits, PCI bridges held references to their downstream buses and the PCI router, which had a couple of gross consequences:

- Server init had to create and keep track of all the buses and their logical numbers
- The Chipset trait had to expose a new interface to let server init get a `Router` so init could create all the bridges

The new commits move bridge creation into a `Topology` object in the `pci` module that represents a collection of buses with their interconnections (e.g. bridges):

- `Topology` understands how to map bus numbers--either logical IDs produced by outside callers or "routed" bus numbers produced by PCI decoders--to the buses they refer to. PCI operations like device attachment and config space access flow from the chipset to the topology, which directs them to the right buses.
- `pci::topology` supplies a `Builder` that lets a caller describe a PCI topology in the abstract (by providing a set of `BridgeDescription`s); the builder creates all the buses, updates all the tables, and registers bridges with the inventory.

This gets most of the complexity of initializing bridges out of propolis-server and allows us not to change the `Chipset` trait. It also clarifies who owns the responsibility of distinguishing between root and non-root buses when routing configuration accesses--previously `Chipset` did this by delegating to `Router`; now `Topology` just handles all of it.

propolis/src/hw/pci/device.rs
```diff
@@ -1071,32 +1063,12 @@ impl Builder {
...
-        );
+        self.cfg_builder.add_custom(offset, len);
         self
     }
 
     fn add_cap_raw(&mut self, id: u8, len: u8) {
-        // XXX: does not pay heed to any custom cfg sections which are added via
-        // the `add_custom_cfg` interface.
-        let end = self.cap_next_alloc + 2 + len as usize;
-        // XXX: on the caller to size properly for alignment requirements
```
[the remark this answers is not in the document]
gjcolombo: On further reflection I think the first XXX is only mostly handled, in the sense that allocating a custom capability doesn't move `cap_next_alloc`, such that the caller has to keep this in mind when deciding where to put the custom regions. I put this caveat into a note in the doc comment.

propolis/src/hw/pci/bits.rs
```diff
@@ -77,3 +77,20 @@ pub const PCIE_MAX_BUSES_PER_ECAM_REGION: u16 = 256;
 /// Bitwise AND'ing an ECAM MMIO access address with this mask produces an
 /// offset in bytes at which to access the target BDF's configuration region.
 pub const MASK_ECAM_CFG_OFFSET: usize = 0xfff;
+
+/// The PCI identifiers to use for PCI-PCI bridges.
+pub const BRIDGE_VENDOR_ID: u16 = 0x1de;
+pub const BRIDGE_DEVICE_ID: u16 = 0x02;
```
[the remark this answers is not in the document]
gjcolombo: Nice! I'll rebase and pick this up.

propolis/src/hw/chipset/i440fx.rs
```diff
@@ -31,13 +32,13 @@ const PM_FUNC: u8 = 3;
 const ADDR_PCIE_ECAM_REGION: usize = 0xe000_0000;
 const LEN_PCI_ECAM_REGION: usize = 0x1000_0000;
 
-#[derive(Default, Debug, Clone, Copy)]
```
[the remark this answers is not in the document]
gjcolombo: My bad, should be fixed now. (The struct was briefly not `Copy` while I was figuring out how I wanted to pass `BridgeDescription`s around, and I didn't restore the derive when I landed on this version.)

propolis/src/hw/pci/topology.rs
```diff
@@ -0,0 +1,427 @@
...
+use crate::inventory::{Inventory, RegistrationError};
+use crate::mmio::MmioBus;
+use crate::pio::PioBus;
+
+use super::bridge::Bridge;
+use super::{bits, Bdf, Bus, BusLocation, Endpoint, LintrCfg};
+
+use thiserror::Error;
+
+/// A logical identifier for a bus in the topology. A bus's logical identifer
```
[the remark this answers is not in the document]
gjcolombo: At the risk of being Extremely Nitpicky, I think the original's right here (AIUI most style manuals specify the `'s` pluralization except for proper nouns ending in S; AP style used to differ, but @APStylebook tweeted otherwise in 2021, and as we all know Twitter is _the_ source and summit of reasoned discourse and thoughtful communication style...).

All that said I'll gladly change it if you feel strongly :)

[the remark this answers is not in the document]
luqmana: Sure, feel free to keep the 's either way.

luqmana (approved):
LGTM. Thanks @gjcolombo!

propolis/src/hw/chipset/i440fx.rs
```diff
@@ -131,37 +132,29 @@ impl I440Fx {
...
+        &self,
+        location: &BusLocation,
+    ) -> (INTxPinID, Arc<dyn IntrPin>) {
+        let intx_pin = match (location.func.get() + 1) % 4 {
             0 => INTxPinID::IntA,
             1 => INTxPinID::IntB,
             2 => INTxPinID::IntC,
             3 => INTxPinID::IntD,
-            _ => panic!(),
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
gjcolombo: @luqmana gets all the credit for this one ;)

propolis/src/hw/pci/topology.rs
```diff
@@ -0,0 +1,435 @@
...
+    /// associate themselves with the supplied port I/O and MMIO buses.
+    pub fn new(
+        inventory: &'a Inventory,
+        pio_bus: &'a Arc<PioBus>,
+        mmio_bus: &'a Arc<MmioBus>,
+    ) -> Self {
+        let mut this = Self {
+            inventory,
+            pio_bus,
+            mmio_bus,
```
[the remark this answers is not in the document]
gjcolombo: Good call. Done.

propolis/src/hw/pci/topology.rs
```diff
@@ -0,0 +1,435 @@
...
+        let topology = Arc::new(Topology {
+            buses,
+            logical_buses,
+            inner: Mutex::new(inner),
+        });
+
+        for bridge in &self.bridges {
+            let new_bridge = Bridge::new(
+                ids::pci::VENDOR_OXIDE,
+                ids::pci::PROPOLIS_BRIDGE_DEV_ID,
```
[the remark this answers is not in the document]
gjcolombo: Done (though via the description rather than as an explicit argument to `add_bridge` so that the caller is only describing the bridge in one place).

propolis/src/hw/pci/cfgspace.rs
```diff
@@ -0,0 +1,119 @@
...
+        self.check_overlap(offset as usize, len as usize);
+        self.cfgmap.define_with_flags(
+            offset as usize,
+            len as usize,
+            CfgReg::Custom(offset),
+            Flags::PASSTHRU,
+        );
+    }
+
+    /// Adds a new capability region of the supplied length at the next
```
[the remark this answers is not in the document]
gjcolombo: My pleasure!

propolis/src/hw/pci/topology.rs
```diff
@@ -0,0 +1,435 @@
...
+    LogicalBusAlreadyExists(LogicalBusId),
+
+    #[error("A PCI device was already attached at {0:?}")]
+    DeviceAlreadyAttached(Bdf),
+
+    #[error("Failed to register a bridge with error {0:?}")]
+    BridgeRegistrationError(RegistrationError),
+}
+
+impl From<PciTopologyError> for IoError {
```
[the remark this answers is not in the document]
gjcolombo: There are a couple of places in server/standalone that use it. (They call it because `Instance::initialize` takes a closure that returns an `io::Result`, and errors from this function can bubble up into those init paths; in the server case it's via `MachineInitializer::initialize_chipset`).

gjcolombo:
Thanks all for the thorough reviews!

PR #131, opened 2022-05-24
Provide the WebSocket URI for an instance's serial console stream in PropolisClient

lifning:
Since technically PropolisClient is where the layout of endpoint URIs as used by clients lives, this makes more sense to put here than in, say, sled-agent. Simply returns the string rather than making a request of any kind so as not to impose any implementation details on consumers. (needed by https://github.com/oxidecomputer/omicron/pull/1108 )

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -159,4 +159,9 @@ impl Client {
         );
         self.get(path, Some(body)).await
     }
+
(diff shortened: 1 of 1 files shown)


luqmana (approved):
Small nit aside, LGTM :+1:

PR #132, opened 2022-05-26
Expose sub-device-IDs required by legacy VirtIO

pfmooney:
OVMF was unable to attach drivers for virtio-block or virtio-net devices without this.


rmustacc:
So we should not mix the legacy device IDs with our vendor ID. So if we're reverting the changes to our subsystem, it needs to be for both the vendor and the device.

pfmooney:
> So we should not mix the legacy device IDs with our vendor ID. So if we're reverting the changes to our subsystem, it needs to be for both the vendor and the device.

Thanks for pointing that out.  Fixed.

luqmana (approved):

Issue #134, opened 2022-06-01
PCI: interrupt pin register is hard-wired to 0 even when a device has a local interrupt configured


gjcolombo:
The virtual i440FX's `pci_attach` routine may assign a local interrupt pin to the device being attached, but this assignment is never made visible to the guest: reading the interrupt pin register in a device's configuration space returns the `reg_intr_pin` value stored in the device's `DeviceState`, and this field is initialized to 0 and doesn't change. Since most (all?) of the virtual PCI devices we use have MSI-X capabilities, this tends not to matter too much in practice, since guests will end up using those, but this will break if we ever have a device that (for whatever reason) doesn't use MSI-X, or a guest that elects not to use MSI-X for some device even though the capability is advertised.

One way (_the_ way?) to fix this is to have `DeviceState::attach` update the device's `reg_intr_pin` so that future reads of the register return the expected value.

---

This can cause a guest panic on Helios:

1. Set up a Propolis VM so that it boots from a Helios guest and has a PCI-PCI bridge with a virtio-block storage device attached.
2. Boot the guest. It may panic on boot. If not, log in and run `format`.

This will produce the following warnings and then panic:

```
WARNING: vioblk2: could not count fixed interrupts
WARNING: vioblk2: interrupt allocation failed
WARNING: vioblk2: failed to complete Virtio init
```

This is a "guest elected not to use MSI-X" case: AIUI, Helios won't enable MSI-X for devices behind a PCI bridge that doesn't have a HyperTransport capability descriptor, so virtio initialization will try to fall back to using an INTx interrupt, but this will fail because no interrupt resources were allocated to the device, because those get allocated during initial PCI enumeration based on the value read from the device's interrupt pin register.

(Separate from this is the question of whether Propolis will or should even allocate a local interrupt pin to devices on a non-root bus. Today it doesn't, but this panic occurs even if it does. The key is that putting a device behind a bridge with no HT capability is a good way to convince Helios not to try to use MSI for it.)

gjcolombo:
I should add that the guest panic is a [known issue upstream](https://www.illumos.org/issues/12925) (which we trigger here by creating conditions in which the guest doesn't think it has interrupt resources available).

PR #136, opened 2022-06-03
Properly emulate IntrPin field in PCI dev

pfmooney:
This should address #134

### DIFF
--- a/propolis/src/hw/chipset/i440fx.rs
+++ b/propolis/src/hw/chipset/i440fx.rs
@@ -678,6 +678,8 @@ impl Piix3PM {
             ..Default::default()
         })
         .add_custom_cfg(PMCFG_OFFSET as u8, PMCFG_LEN as u8)
+        // ACPI device requires lintr for SCI
+        .add_lintr()
         .finish();
 
         Arc::new(Self { pci_state, regs: Mutex::new(PMRegs::default()) })
--- a/propolis/src/hw/pci/device.rs
+++ b/propolis/src/hw/pci/device.rs
@@ -161,7 +161,6 @@ pub struct Ident {
 struct State {
     reg_command: RegCmd,
     reg_intr_line: u8,
-    reg_intr_pin: u8,
 
     attach: Option<bus::Attachment>,
     bars: Bars,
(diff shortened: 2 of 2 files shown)


gjcolombo (approved):

propolis/src/hw/pci/device.rs
```diff
@@ -290,7 +288,19 @@ impl DeviceState {
                 ro.write_u8(self.state.lock().unwrap().reg_intr_line)
             }
             StdCfgReg::IntrPin => {
-                ro.write_u8(self.state.lock().unwrap().reg_intr_pin)
+                if self.lintr_support {
```
pfmooney, line 291: Ovmf was grumpy about the hostbus device presenting an interrupt pin.

propolis/src/hw/chipset/i440fx.rs
```diff
@@ -678,6 +678,8 @@ impl Piix3PM {
             ..Default::default()
         })
         .add_custom_cfg(PMCFG_OFFSET as u8, PMCFG_LEN as u8)
+        // ACPI device requires lintr for SCI
+        .add_lintr()
```
pfmooney, line 682: Prior to adding this, `lspci` (on alpine) knew of the interrupt associated with the device thanks to the ACPI table, but was still confused about it due to the missing entry here:
```
Interrupt: pin ? routed to IRQ 9
```

With the entire change, `lspci` is much happier about reporting all of the values (for an attached virtio disk and nic):
```
localhost:~# lspci -vv | grep Interrupt
        Interrupt: pin A routed to IRQ 9
        Interrupt: pin B routed to IRQ 10
        Interrupt: pin B routed to IRQ 11
```

propolis/src/hw/pci/device.rs
pfmooney: By "grumpy" I mean blowing an assert and hanging during boot:
```
...
Boot Mode:0
SetPciIntLine: PCI host bridge (00:00.0) should have no interrupts!
ASSERT /home/pmooney/git/uefi-edk2/OvmfPkg/Library/PlatformBootManagerLib/BdsPlatform.c(1128): ((BOOLEAN)(0==1))
```

pfmooney:
Merged in 8259ea3ddf49b23d81e45627f93fb8981713d909

PR #138, opened 2022-06-09
Snapshot and restore VMs with propolis-standalone.

luqmana:
This set of changes adds some new flags to `propolis-standalone` to allow taking a snapshot of a running VM, saving it to a file and being able to later restore it:

```console
$ ./target/debug/propolis-standalone --help
propolis-standalone 
Propolis command-line frontend for running a VM

USAGE:
    propolis-standalone [OPTIONS] <CONFIG|SNAPSHOT>

ARGS:
    <CONFIG|SNAPSHOT>    Either the VM config file or a previously captured snapshot image

OPTIONS:
    -h, --help        Print help information
    -r, --restore     Restore previously captured snapshot
    -s, --snapshot    Take a snapshot on Ctrl-C before exiting
```

Running the snapshot/restore requires Illumos bits currently in review: https://code.illumos.org/c/illumos-gate/+/2061 & https://code.illumos.org/c/illumos-gate/+/2157.

Also, thanks to Patrick and Greg for a lot of the device import/export bits.

Depends on: https://github.com/oxidecomputer/propolis/pull/140

### DIFF
--- a/propolis/src/hw/pci/device.rs
+++ b/propolis/src/hw/pci/device.rs
@@ -7,6 +7,7 @@ use super::{bus, BarN, Endpoint};
 use crate::common::*;
 use crate::dispatch::DispCtx;
 use crate::intr_pins::IntrPin;
+use crate::migrate::MigrateStateError;
 use crate::util::regmap::{Flags, RegMap};
 
 use lazy_static::lazy_static;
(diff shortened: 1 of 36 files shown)


propolis/src/hw/pci/device.rs
```diff
@@ -974,8 +974,10 @@ impl MsixCfg {
             entries.push(migrate::MsixEntryV1 {
                 addr: lentry.addr,
                 data: lentry.data,
-                is_pending: lentry.pending,
                 is_vec_masked: lentry.mask_vec,
+                is_func_masked: lentry.mask_func,
+                is_enabled: lentry.enabled,
```
pfmooney, line 987: Shouldn't we re-hydrate the masked-by-function and enabled state from `MsixStateV1`?

luqmana: Ah, so the per-entry values should always reflect the overall value in MsixCfgState/MsixStateV1?

pfmooney: Yeah.  That state has been propagated down to the entries so that they needn't consult the device-wide MSI-X config when making a decision about if an interrupt is masked or not.

pfmooney (approved):
Nice work on this.

PR #140, opened 2022-06-10
Expose (and export) kernel dev state

pfmooney:
This depends on the changes drafted in illumos 14261 and 14635


pfmooney:
The aforementioned illumos changes haven't landed yet, but propolis will be fine without them until then.  It means the `header-check` in `bhyve_api` will fail if run against sources without those changes.

pfmooney:
14261 has been RTIed (awaiting approval), with 14635 likely to follow soon.  I plan to land this change so that Luqman's (and others) subsequent work can go in promptly.

luqmana (approved):
💯

PR #142, opened 2022-06-13
server: make vnc port a configurable parameter

rcgoodfellow:
This PR makes the VNC port a configurable parameter. Without this, multiple Propolis instances cannot run in a single zone. This came up when updating Falcon to the latest propolis revision.

~This change introduces a new flag to Propolis server, so it may have an impact on control plane machinery creating propolis-server instances.~

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -32,6 +32,9 @@ enum Args {
 
         #[clap(name = "PROPOLIS_IP:PORT", parse(try_from_str))]
         propolis_addr: SocketAddr,
+
+        #[clap(name = "VNC_IP:PORT", parse(try_from_str))]
+        vnc_addr: SocketAddr,
     },
 }
 
@@ -56,7 +59,7 @@ async fn main() -> anyhow::Result<()> {
     match args {
         Args::OpenApi => run_openapi()
             .map_err(|e| anyhow!("Cannot generate OpenAPI spec: {}", e)),
-        Args::Run { cfg, propolis_addr } => {
+        Args::Run { cfg, propolis_addr, vnc_addr } => {
             let config = config::parse(&cfg)?;
 
             // Dropshot configuration.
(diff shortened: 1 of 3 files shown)


server/src/main.rs
```diff
@@ -34,6 +34,9 @@ enum Args {
 
         #[structopt(name = "PROPOLIS_IP:PORT", parse(try_from_str))]
         propolis_addr: SocketAddr,
+
+        #[structopt(name = "VNC_IP:PORT", parse(try_from_str))]
```
gjcolombo, line 38: Would it make sense to provide a default for this to make the new flag optional?

gjcolombo (approved):
Dang it, I knew I should've made a PR when I wrote basically this same code last week :) 

LGTM, for what that's worth--perhaps someone who knows more about the system will see pitfalls that I've missed.

rcgoodfellow:
Thanks @gjcolombo!

I'm a bit torn on the default parameter. Having a default would prevent this change from breaking existing automation, and provide an easier glide path for starting up propolis-server instances. OTOH, it also seems like a bit of a foot gun having known bad situations for default parameters.

luqmana:
cc @jordanhendricks

> OTOH, it also seems like a bit of a foot gun having known bad situations for default parameters.

I wouldn't say it's bad perse. The default usecase has been running a single instance of propolis-server per-zone so I think it's worth keeping that. If you do try to run more than one, subsequent ones should fail to bind and make it obvious. The new flag to configure it would be able to unblock you in that scenario.

jordanhendricks:
I agree with @luqmana -- I hard-coded the port initially to the standard VNC port under the expectation that only one ever propolis-server would be running in per zone, but there's certainly no reason to do that if it hinders development. Thanks for the PR! It looks good to me :+1:

server/src/main.rs
jordanhendricks: +1, it would be nice not to have to specify this every time.

rcgoodfellow:
> If you do try to run more than one, subsequent ones should fail to bind and make it obvious.

It looks like that will need to be addressed in the rfb crate. See [this issue](https://github.com/oxidecomputer/rfb/issues/5).

luqmana (approved):

PR #145, opened 2022-06-17
POST to issue a snapshot request to crucible

jmpesp:
Add endpoints to list inventory entity instance names, get a particular
entity by instance name, and add a POST endpoint to issue a snapshot
request to Crucible.

### DIFF
--- a/client/src/api.rs
+++ b/client/src/api.rs
@@ -285,7 +285,8 @@ pub enum NetworkInterfaceAttachmentState {
     Faulted,
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
-pub struct InstanceGetInventoryResult {
-    pub type_name: String,
+#[derive(Deserialize, JsonSchema)]
+pub struct SnapshotRequestPathParams {
+    pub id: Uuid,
+    pub snapshot_id: Uuid,
 }
--- a/client/src/api.rs
+++ b/client/src/api.rs
@@ -284,3 +284,8 @@ pub enum NetworkInterfaceAttachmentState {
     Detached,
     Faulted,
 }
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+pub struct InstanceGetInventoryResult {
+    pub type_name: String,
+}
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -179,22 +179,13 @@ impl Client {
         format!("ws://{}/instance/serial", self.address)
     }
 
-    pub async fn instance_get_inventory_entity(
-        &self,
-        name: String,
-    ) -> Result<api::InstanceGetInventoryResult, Error> {
-        let path =
-            format!("http://{}/instance/inventory/{}", self.address, name);
-        self.get(path, None).await
-    }
-
     pub async fn instance_issue_crucible_snapshot_request(
         &self,
         inventory_name: String,
         snapshot_name: String,
     ) -> Result<(), Error> {
         let path = format!(
-            "http://{}/instance/inventory/{}/snapshot/{}",
+            "http://{}/instance/disk/{}/snapshot/{}",
             self.address, inventory_name, snapshot_name
         );
         self.post(path, None).await
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -181,12 +181,14 @@ impl Client {
 
     pub async fn instance_issue_crucible_snapshot_request(
         &self,
-        inventory_name: String,
-        snapshot_name: String,
+        disk_id: Uuid,
+        snapshot_id: Uuid,
     ) -> Result<(), Error> {
         let path = format!(
             "http://{}/instance/disk/{}/snapshot/{}",
-            self.address, inventory_name, snapshot_name
+            self.address,
+            disk_id.to_string(),
+            snapshot_id.to_string(),
         );
         self.post(path, None).await
     }
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -89,6 +89,20 @@ impl Client {
         send_and_parse_response(request).await
     }
 
+    async fn post<T: DeserializeOwned, U: IntoUrl + std::fmt::Display>(
+        &self,
+        path: U,
+        body: Option<Body>,
+    ) -> Result<T, Error> {
+        info!(self.log, "POST request to {}", path);
+        let mut request = self.client.post(path);
+        if let Some(body) = body {
+            request = request.body(body);
+        }
+
+        send_and_parse_response(request).await
+    }
+
     async fn put_no_response<U: IntoUrl + std::fmt::Display>(
         &self,
         path: U,
@@ -163,5 +177,26 @@ impl Client {
     /// Returns the WebSocket URI to an instance's serial console stream.
     pub fn instance_serial_console_ws_uri(&self) -> String {
         format!("ws://{}/instance/serial", self.address)
+    }
+
+    pub async fn instance_get_inventory_entity(
+        &self,
+        name: String,
+    ) -> Result<api::InstanceGetInventoryResult, Error> {
+        let path =
+            format!("http://{}/instance/inventory/{}", self.address, name);
+        self.get(path, None).await
+    }
+
+    pub async fn instance_issue_crucible_snapshot_request(
+        &self,
+        inventory_name: String,
+        snapshot_name: String,
+    ) -> Result<(), Error> {
+        let path = format!(
+            "http://{}/instance/inventory/{}/snapshot/{}",
+            self.address, inventory_name, snapshot_name
+        );
+        self.post(path, None).await
     }
 }
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -186,9 +186,7 @@ impl Client {
     ) -> Result<(), Error> {
         let path = format!(
             "http://{}/instance/disk/{}/snapshot/{}",
-            self.address,
-            disk_id.to_string(),
-            snapshot_id.to_string(),
+            self.address, disk_id, snapshot_id,
         );
         self.post(path, None).await
     }
--- a/propolis/src/block/crucible.rs
+++ b/propolis/src/block/crucible.rs
@@ -83,15 +83,12 @@ impl CrucibleBackend {
         // tokio::sync::mpsc::Receiver, so use tokio::task::block_in_place here.
         // Remove that when Crucible changes over to the tokio mpsc.
         let mut waiter = tokio::task::block_in_place(|| {
-            self.block_io.flush(Some(
-                SnapshotDetails {
-                    snapshot_name,
-                }))
-        }).map_err(map_crucible_error_to_io)?;
+            self.block_io.flush(Some(SnapshotDetails { snapshot_name }))
+        })
+        .map_err(map_crucible_error_to_io)?;
 
-        tokio::task::block_in_place(|| {
-            waiter.block_wait()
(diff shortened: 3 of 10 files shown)


jmpesp:
Listing inventory entity instance names:

```
$ curl -sq 'http://[fd00:1122:3344:101::c]:12400/instance/inventory' | jq .
{
  "entity_names": [
    "pci-piix4-hb",
    "lpc-ps2ctrl",
    "lpc-uart-com2",
    "pci-nvme-0.16.0",
    "pci-piix3-lpc",
    "lpc-bhyve-atpic",
    "lpc-uart-com3",
    "pci-virtio-viona-0.8.0",
    "in-memory-cloud-init",
    "block-crucible-7716c2cd-f009-476b-9fc8-898452bc5265",
    "qemu-ramfb",
    "lpc-bhyve-ioapic",
    "lpc-bhyve-pmtimer",
    "chipset-i440fx",
    "lpc-uart-com4",
    "qemu-fwcfg",
    "lpc-bhyve-hpet",
    "pci-virtio-block-0.24.0",
    "qemu-lpc-debug",
    "lpc-bhyve-rtc",
    "pci-piix3-pm",
    "lpc-uart-com1",
    "lpc-bhyve-atpit"
  ]
}
```

Listing a specific inventory entity:

```
$ curl -sq 'http://[fd00:1122:3344:101::c]:12400/instance/inventory/block-crucible-7716c2cd-f009-476b-9fc8-898452bc5265' | jq .
{
  "type_name": "block-crucible"
}
```

Issuing a snapshot request:

```
$ curl -v 'http://[fd00:1122:3344:101::c]:12400/instance/inventory/block-crucible-7716c2cd-f009-476b-9fc8-898452bc5265/snapshot/testpost' -X POST
...
> POST /instance/inventory/block-crucible-7716c2cd-f009-476b-9fc8-898452bc5265/snapshot/testpost HTTP/1.1
> Host: [fd00:1122:3344:101::c]:12400
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-type: application/json
< x-request-id: 231394d2-406c-4c62-a75d-b71be96f57fa
< content-length: 4
< date: Fri, 17 Jun 2022 20:58:36 GMT

$ zfs list -t snapshot
NAME                                                                                                      USED  AVAIL  REFER  MOUNTPOINT
oxp_d462a7f7-b628-40fe-80ff-4e4189e2d62b/crucible/regions/b65b73ff-93d7-487a-9672-17578eb7b17b@testpost      0      -  9.99M  -
oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03/crucible/regions/31bcd208-bece-4803-8de3-08bb3ea0cb42@testpost      0      -  9.99M  -
oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03/crucible/regions/5bd73c8c-e050-4b7d-a6a2-ff118b7701f1@testpost      0      -  9.99M  -
```

luqmana:
I'm a little hesitant on exposing the inventory with this new endpoint. For the most part it is an implementation detail and not something I think we want to commit to. :/

Is there a reason we can't have nexus/sled-agent just make the request directly to the crucible downstairs instead of going through propolis-server->crucible-upstairs? Presumably we already have that crucible downstairs address accessible

gjcolombo:
I could be overlooking something, but IMO part of the trouble here is that while disks have [names](https://github.com/oxidecomputer/propolis/blob/52e0d2c5b26fea0d0d17d23b5056c82744a4ab66/client/src/api.rs#L233) and Crucible backends have [UUIDs](https://github.com/oxidecomputer/propolis/blob/52e0d2c5b26fea0d0d17d23b5056c82744a4ab66/propolis/src/block/crucible.rs#L76), at most these only partially specify what names those objects have in the inventory (e.g. the backend names are [prefixed](https://github.com/oxidecomputer/propolis/blob/52e0d2c5b26fea0d0d17d23b5056c82744a4ab66/propolis/src/block/crucible.rs#L115) with "block-crucible-" when they're registered into the inventory). I'm also unsure from a cursory reading whether Nexus knows the backend UUIDs (they're not part of the `VolumeConstructionRequest`).

I haven't written this up formally yet, but subsequent to #146 I was going to look at changing up server initialization so that devices/backends' names in the inventory match the names they have in their instance specs. In the short run, where Propolis-server generates device/backend specs from REST API inputs, we can make the device-name-to-inventory-name translation part of the contract (e.g. "if you named the disk `foo-bar-baz`, then the backend is `foo-bar-baz-backend`"). In the long run, I hope/expect Nexus will generate its own specs and pass them directly to Propolis; in that case Propolis will just use the names Nexus gives things. In either case, this should make it so that Nexus can refer to entities by name without having to ask Propolis exactly what's in the inventory. (That's all provided that I don't discover any major roadblocks when I go to try this out...)

jmpesp:
> I could be overlooking something, but IMO part of the trouble here is that while disks have names and Crucible backends have UUIDs, at most these only partially specify what names those objects have in the inventory (e.g. the backend names are prefixed with "block-crucible-" when they're registered into the inventory). I'm also unsure from a cursory reading whether Nexus knows the backend UUIDs (they're not part of the VolumeConstructionRequest).

Nexus currently doesn't know the UUIDs sent [but they will be part of the VolumeConstructionRequest soon](https://github.com/oxidecomputer/crucible/pull/358). It's true that DiskRequest is by Name, but the backend is by UUID - should I dig a bit more into accessing by name?

jmpesp:
> In either case, this should make it so that Nexus can refer to entities by name without having to ask Propolis exactly what's in the inventory. (That's all provided that I don't discover any major roadblocks when I go to try this out...)

I could remove the part of this PR that lists inventory names and only POST assuming the block crucible + UUID name. What do you think?

gjcolombo:
> I could remove the part of this PR that lists inventory names and only POST assuming the block crucible + UUID name. What do you think?

I was about to suggest just that!

> Excuse my ignorance here, but won't Nexus know the name and UUID's of all storage attached to an instance? I did not think any of that knowledge lives only with Propolis?

I think the general issue is that the Propolis API doesn't (yet) formally define how Nexus's names/UUIDs for things relate to the keys Propolis uses in its object store. That is, the fact that Propolis will name a Crucible backend `block-crucible-$UUID` (and that it can find the backend later with this name) is currently more of a Propolis implementation detail than a formal contract between Propolis and Nexus.

In my view, it's OK in the short run for Nexus to rely on this kind of implementation detail, especially if we otherwise have to add an extra round trip where Nexus asks Propolis what names it gave things. I do think Propolis should have more general naming rules in the longer run (hence my rant above about instance specs), but we don't need to block the snapshot API on that.

jmpesp:
> Excuse my ignorance here, but won't Nexus know the name and UUID's of all storage attached to an instance? I did not think any of that knowledge lives only with Propolis?

That's what I meant - nexus knows the disk UUID and can assume the block-crucible-UUID format inventory name.

luqmana:
> > I'm a little hesitant on exposing the inventory with this new endpoint. For the most part it is an implementation detail and not something I think we want to commit to. :/
> 
> Is it the /inventory endpoint that's making you hesitant? That could be ripped out in favour of the specific one. Or is it using the names in general?

Yea, I think having a crucible specific endpoint is fine and seems to be the consensus if I'm reading right.

gjcolombo:
We discussed this briefly in today's hypervisor sync. A consensus emerged around refining this further by not using the inventory at all for snapshots. Instead, `propolis_server::server::InstanceContext` can hold a mapping from upstairs IDs directly to `Arc<propolis::block::CrucibleBackend>`s (populated when the backends are created during instance init), and the snapshot API can just look in that mapping without involving the inventory at all (and so without needing Nexus/Propolis to agree on the inventory's naming conventions).

propolis/src/block/crucible.rs
```diff
@@ -78,12 +79,14 @@ impl CrucibleBackend {
...
     /// Issue a snapshot request
-    pub fn snapshot(&self, snapshot_name: String) -> Result<()> {
+    pub fn snapshot(&self, snapshot_id: Uuid) -> Result<()> {
         // XXX Crucible uses std::sync::mpsc::Receiver, not
         // tokio::sync::mpsc::Receiver, so use tokio::task::block_in_place here.
         // Remove that when Crucible changes over to the tokio mpsc.
         let mut waiter = tokio::task::block_in_place(|| {
-            self.block_io.flush(Some(SnapshotDetails { snapshot_name }))
+            self.block_io.flush(Some(SnapshotDetails {
+                snapshot_name: snapshot_id.to_string(),
```
luqmana, line 88: Nit: I realize this would be on the crucible side but is the snapshot name intended to always be a UUID or any string?

client/src/lib.rs
```diff
@@ -181,12 +181,14 @@ impl Client {
...
-        snapshot_name: String,
+        disk_id: Uuid,
+        snapshot_id: Uuid,
     ) -> Result<(), Error> {
         let path = format!(
             "http://{}/instance/disk/{}/snapshot/{}",
-            self.address, inventory_name, snapshot_name
+            self.address,
+            disk_id.to_string(),
+            snapshot_id.to_string(),
```
luqmana, line 191:
```suggestion
            disk_id,
            snapshot_id,
```

Uuid impls Display

server/src/lib/server.rs
```diff
@@ -276,6 +284,11 @@ async fn instance_ensure(
     // This initialization may be refactored to be client-controlled,
     // but it is currently hard-coded for simplicity.
 
+    let mut crucible_backends: BTreeMap<
+        Uuid,
+        Arc<propolis::block::CrucibleBackend>,
+    > = BTreeMap::new();
```
luqmana, line 290: Huh, does the inference not work here?

client/src/api.rs
```diff
@@ -284,3 +284,8 @@ pub enum NetworkInterfaceAttachmentState {
     Detached,
     Faulted,
 }
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+pub struct InstanceGetInventoryResult {
```
luqmana, line 289: Leftover inventory bits

server/src/lib/server.rs
```diff
@@ -936,6 +959,45 @@ async fn instance_migrate_status(
         .map(HttpResponseOk)
 }
 
+#[derive(Deserialize, JsonSchema)]
+struct SnapshotRequestPathParams {
```
luqmana, line 963: Move to client::api since that's generally where we kept those before we removed the redundant IDs from most endpoints.

luqmana (commented):
Thanks James! A couple small comments and looks like the inventory bits are still there but not needed anymore?

propolis/src/block/crucible.rs
jmpesp: `SnapshotDetails` accepts any string, but from Nexus we should accept snapshot UUIDs.

client/src/lib.rs
jmpesp: fixed in f302dc0

server/src/lib/server.rs
jmpesp: it does haha, fixed in c1ad064

client/src/api.rs
jmpesp: removed in c341b5d

server/src/lib/server.rs
jmpesp: done in c341b5d

luqmana (approved):

PR #146, opened 2022-06-22
client/server: initial implementation of instance specs

gjcolombo:
Define an initial version of a Propolis instance specification (as described in RFD 283) and convert propolis-server's initialization logic to operate in terms of these specs. Instead of initializing components directly from a config TOML and HTTP request parameters, convert these to an instance spec and initialize from that spec. In the future, Nexus can supply specs as parameters and Propolis can initialize instances directly from those instead of constructing them on the fly.

Work remaining after this (I'll file issues for these after merging, in case the list changes as folks give feedback):

- Add framebuffer as a specifiable entity
- Enforce the naming rules in RFD 283 (currently names must be unique within a particular device/backend type, but are not required to be unique throughout the spec)
- Register items into the inventory based on the names they had in the spec rather than generated names
- Put specs into the migration preamble and check source/target specs for compatibility
- Provide an API to create an instance from a serialized spec instead of raw parameters
- Decide how to fill in guest platform & CPU/memory topology fields in the mainboard struct

Tests: Rust tests; manually booted Helios and Windows Server VMs and checked that their devices were present.


gjcolombo:
Working on downstream commits that incorporate specs into the migration protocol has made me worried about the ergonomics of spec versioning as it's structured here. It's really important to get those right, because version upgrades/downgrades are complex, and having good structure ergonomics will make it easier to reason about them and harder to make mistakes. I also don't think we need to commit to a versioning scheme immediately at this stage of development (we'll be able to take breaking changes to the spec for the foreseeable future).

I've converted this to a draft PR while I take a couple of follow-up actions:

- Add additional observations re versioning to RFD 283.
- Simplify this change by removing the versioned instance spec enum and just having instance specs act as a freestanding type for now; I expect this to simplify the logic needed to decide if two specs are migration-compatible (since no versioning scaffolding, even the trivial kind, will be needed at first).
  - We can bring the enum back later if we decide that's warranted, or can take a different approach.

server/src/lib/initializer.rs
```diff
@@ -241,109 +256,170 @@ impl<'a> MachineInitializer<'a> {
...
+                        format!(
+                            "Backend {} not found for vNIC {}",
+                            vnic_spec.backend_name, name
+                        ),
+                    )
+                })?;
+            let bdf = pci::Bdf::new(
+                vnic_spec.pci_path.0,
+                vnic_spec.pci_path.1,
+                vnic_spec.pci_path.2,
```
luqmana, line 402: Ok, looks like we do this at least a couple times so maybe add a From impl so we don't need to repeat as much.

luqmana (commented):
Nice, thanks for tackling this Greg! Looks mostly good, left a couple small comments to get some thoughts

server/Cargo.toml
```diff
@@ -47,6 +47,7 @@ uuid = "1.0.0"
 base64 = "0.13"
 
 [dev-dependencies]
+crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "8314eeddd228ec0d76cefa40c4a41d3e2611ac18" }
```
pfmooney, line 50: Why the specific version dep?

server/src/lib/config.rs
```diff
@@ -240,6 +232,30 @@ impl<'a> Iterator for IterDevs<'a> {
...
+    inner: btree_map::Iter<'a, String, BlockDevice>,
+}
+
+impl<'a> Iterator for IterBlockDevs<'a> {
+    type Item = (&'a String, &'a BlockDevice);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next()
+    }
+}
```
pfmooney, line 245: Is there an advantage to this wrapper type, vs returning the `btree_map:Iter` directly from `block_devices`?

client/src/instance_spec/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+#[derive(Clone, Copy, Deserialize, Serialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct SerialPort {
+    /// The serial port number for this port.
+    pub num: SerialPortNumber,
+
+    /// The initial discard discipline for this port. If true, the device will
+    /// not buffer bytes written from the guest until the port's discipline
+    /// changes.
+    pub autodiscard: bool,
```
pfmooney, line 192: Perhaps something for later down the line, but I'm not certain we should be exposing this in such a manner.  Serial ports with no connected backend can probably be auto-discard by default.  Those with a connected backend can let it decide. (I suspect?)

gjcolombo:
Thanks @luqmana and @pfmooney! The newest commits should address all the feedback above--I'll elaborate on a few things in the individual comment threads. 

I'll also get started on rebasing/resolving conflicts (I don't expect any major changes in approach, just need to patch a few things up), but I wanted to let you go ahead and let you have another look before doing any force pushes.

client/src/instance_spec/mod.rs
```diff
@@ -0,0 +1,224 @@
...
+#[serde(deny_unknown_fields)]
+pub struct InstanceSpec {
+    pub board: Board,
+
+    pub storage_devices: BTreeMap<String, StorageDevice>,
+    pub storage_backends: BTreeMap<String, StorageBackend>,
+    pub network_devices: BTreeMap<String, NetworkDevice>,
+    pub network_backends: BTreeMap<String, NetworkBackend>,
+    pub serial_ports: BTreeMap<String, SerialPort>,
+    pub pci_pci_bridges: BTreeMap<String, PciPciBridge>,
```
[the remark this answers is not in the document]
gjcolombo: I tried it both ways and went with the type alias for simplicity. Specs don't (yet) have naming restrictions like Omicron's, so they don't need a special `TryFrom` implementation to reject invalid names; it seemed to me that in that scenario, just using the type alias avoided a lot of conversion/Deref/etc. boilerplate that it seemed like I'd otherwise need.

server/src/lib/initializer.rs
gjcolombo: This was mildly annoying at first because `propolis-server` is the only crate that imports both the `PciPath` and `Bdf` types, but because both are external types, it can't `impl From<PciPath> for Bdf`. Rather than try to resolve that in the server itself, I added a small `types` crate to hold `PciPath` so that the Propolis lib crate can include it and define the From impl. 

If there's a better way to do this, let me know--this seemed simpler than adding a wrapper newtype in the server but I'm not at all sure I didn't overlook an even better approach.

In the longer term we could probably even move the `Bdf` struct out of the lib and into `types` and do all the device/function validation on that single type. But I'm inclined to let the dust settle on this change before doing much more refactoring with it.

server/Cargo.toml
gjcolombo: I chose this since it's the same version propolis and propolis-client are using, but I ended up resolving this by re-exporting the `VolumeConstructionRequest` type (the reason server needs this) from the instance spec library itself and not adding this dependency at all. (The semantics I really want are "the same VolumeConstructionRequest definition that the spec library used," and from what I could gather the `pub use` re-export was the most idiomatic way to achieve that.)

client/src/instance_spec/mod.rs
gjcolombo: Done--this makes it easier to reason about whether two serial ports are compatible for migration purposes, so I think it makes sense just to do this now.

types/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+use std::str::FromStr;
+
+use serde::{Deserialize, Serialize};
+
+/// A PCI bus/device/function tuple. Supports conversion from a string formatted
+/// as "B.D.F", e.g. "0.7.0".
+#[derive(
+    Clone, Copy, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Debug,
+)]
+pub struct PciPath(pub u8, pub u8, pub u8);
```
pfmooney, line 18: Unless we're going to further validate this, then exposing the latter two fields publicly means allowing invalid BDFs (since function must be 0-7 and device/slice must be 0-31)

gjcolombo: These should still be validated by the Propolis lib/server, in the sense that the lib only accepts `pci::Bdf`s and the conversion from `PciPath` to `Bdf` is fallible and will reject invalid device/function numbers. But it's not too hard to just reject invalid paths up front, and IMO doing that carries some other advantages, so I tightened up the bolts as follows:

- This struct's fields are now private, and it has a constructor that verifies that the device and function are in range.
- It also now serializes as "B.D.F" instead of "PciPath(B, D, F)"; this lets the deserializer just forward to `from_str` and settles on "B.D.F" as the One True Way to represent a string-format BDF. (I also added some serialization unit tests.)

gjcolombo:
Rebased onto master. There are some additional changes in `server/src/lib/initializer.rs` to reconcile the changes from #145 (which made init store Crucible backends into a map) with this PR (which basically rewrote initialization). Tested these by starting a local server, using the `--crucible-disks` CLI option to direct it to create Crucible disks during startup, POSTing to the snapshot API, and verifying that it could find backends in the map iff I passed it valid upstairs IDs.

types/src/lib.rs
```diff
@@ -44,6 +48,16 @@ impl PciPath {
...
+
+    pub fn bus(&self) -> u8 {
+        self.0
+    }
+    pub fn device(&self) -> u8 {
+        self.1
+    }
+    pub fn function(&self) -> u8 {
+        self.2
+    }
```
luqmana, line 60: Probably not a hot-path but since this is in a separate crate, worth marking with `#[inline]`?

luqmana (approved):

client/src/instance_spec.rs
```diff
@@ -0,0 +1,287 @@
...
+
+/// A kind of storage backend: a connection to on-sled resources or other
+/// services that provide the functions storage devices need to implement their
+/// contracts.
+#[derive(Clone, Deserialize, Serialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub enum StorageBackendKind {
+    /// A Crucible-backed device, containing a generation number and a
+    /// construction request.
+    Crucible { gen: u64, req: CrucibleRequestContents },
```
leftwo, line 159: Eventually, the generation number field needs to come from and be stored in Nexus.  When that code is written, will we be able to just plug it in here?  Is that what the "gen" field is for?

gjcolombo: That should Just Work provided that those changes don't change anything about the [`DiskRequest`](https://github.com/oxidecomputer/propolis/blob/23a38f4a1df4afcba3905f94ebbf312d0d187c7d/client/src/api.rs#L232)s that Propolis gets.

The init process used to [pass these fields to Crucible init](https://github.com/oxidecomputer/propolis/blob/23a38f4a1df4afcba3905f94ebbf312d0d187c7d/client/src/api.rs#L232) directly from the `DiskRequest`. This PR  [bounces them through the instance spec first](https://github.com/oxidecomputer/propolis/blob/182249b956f3c4fcdb8a05a571b55137476fb1d8/server/src/lib/initializer.rs#L269), but that change should be totally transparent to Nexus--the `DiskRequest` itself should be unchanged, and we should still pass its parameters to Crucible in the same way (barring any mistakes on my end, of course...).

client/src/instance_spec.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    /// A device backed by a file on the host machine. The payload is a path to
+    /// this file.
+    File { path: String },
+
+    /// A device backed by an in-memory buffer in the VMM process. The payload
+    /// is the buffer's contents.
+    InMemory { bytes: Vec<u8> },
```
pfmooney, line 167: Do we want the actual memory contents of the in-memory disk to be part of its specification here, or simply something that is copied as part of any migration?

client/src/instance_spec.rs
```diff
@@ -0,0 +1,287 @@
...
+// BackendKind and DeviceKind enums that storage devices do, but they can
+// be added if needed in future versions.
+
+/// A network backend, specifically a virtual NIC on the host system for which
+/// Virtio network adapter integration is enabled.
+#[derive(Clone, Deserialize, Serialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct NetworkBackend {
+    /// The name of the vNIC to connect to.
+    pub vnic_name: String,
```
pfmooney, line 218: Do we want to include a `kind` here, in case we expand beyond viona at some point?

server/src/lib/initializer.rs
```diff
@@ -237,111 +255,183 @@ impl<'a> MachineInitializer<'a> {
...
+            let bdf: pci::Bdf = vnic_spec.pci_path.try_into().map_err(|e| {
+                Error::new(
+                    ErrorKind::InvalidInput,
+                    format!("Couldn't get PCI BDF for vNIC {}: {}", name, e),
+                )
+            })?;
+            let hdl = self.machine.get_hdl();
+            let viona = virtio::PciVirtioViona::new(
+                &backend_spec.vnic_name,
+                0x100,
```
pfmooney, line 428: We should figure out how we want to encode the max ring size here, or frankly, any/all parameters of viona.  (this can be a follow-up)

types/src/lib.rs
pfmooney: Likely a rounding error when it comes to perf, compared to everything else going on

client/src/instance_spec.rs
gjcolombo: This is here more for instance init than for migration. In the (admittedly hypothetical) future where `instance_ensure` takes a spec instead of collections of disk, NIC, etc. requests, the spec needs some way to specify what the disk's initial contents should be. (In the current code, doing it this way lets the server initialize this backend only from the spec without having to peek at the disk requests themselves to get this data, which simplifies the code somewhat.)

During state migration, I'd expect the source in-memory backend to save the disk contents for the target to restore (especially if the source disk might've changed). 

In the preamble compatibility check (as I've drafted it), only the backend types matter--the contents are ignored because they can legitimately differ between source and target (e.g. in Crucible generation numbers). That makes it wasteful to serialize and send the disk contents, but I suspect we can deal with that in other ways (e.g. provide a "serialize for compatibility" helper that omits nonessential data).

pfmooney: It just seems odd to me to include the contents in the spec.

gjcolombo: Having examined this more, I think including the contents streamlines the instance initialization code, but the complexity will come back in the form of additional logic to strip them from specs before sending them for migration (which I think I would have to do to look myself in the mirror--I hadn't realized just how much larger the contents are than the rest of the spec). So I've gone ahead and broken this up a bit (and re-tested by creating some local VMs with a cloud-init image).

PR #147, opened 2022-06-23
Add stat framework to Propolis

leftwo:
This is an initial pass at putting some Oximeter stats into propolis, dragging Crucible along for the ride.
Feedback welcome about this strategy for metrics, and how might we improve it.

On instance create, a `ProducerRegistry` can be created that can be passed to libraries that wish to record stats.  This allows Propolis to not have to know what stats a library wishes to register.

A new struct, `InstanceMetrics`, has been added to `Context`.  This is used for tracking stats for an instance as well as containing the `ProducerRegistry` that libraries can use to register with Oximeter/Nexus for their own stats.

A new optional `metrics` field in `InstanceEnsureRequest` that, if is is present, tells propolis server if it should setup metrics at the provided registration address. The propolis CLI also has this same optional argument.

The CrucibleBackend Volume::construct() now takes a `ProducerRegistry` that if present can be 
used to register metrics with Oximeter.

This requires changes in both  crucible and omicron, and will have to be merged accordingly.

In this PR, we are creating a standalone metrics endpoint different from the propolis API.

Things that need completion or additional issues before we can merge this:

- [x] Decision on what to do if ensure_instance can't contact Nexus/Oximeter for registration.
- [x] Use DNS to determine registration address (updated this to just expect it from sled-agent).
- [x] Required changes in Crucible merged to main
- [x] More comments on how other libraries can add metrics.
- [x] Removal of the cargo patch path to crucible, update crucible git rev.


leftwo:
This won't build in CI because it still has a custom path to Crucible.

leftwo:
From hypervisor huddle: 
Would it be possible to make this an optional dependency?

How are the metrics that are recorded being stored in the database?  Specifically, would it be "easy" for someone who wanted to look at IO over time (where perhaps there were several power cycles) to be able to read out data and construct a graph.  The concern here is to be sure that we don't store our data in the database in a way that causes a common use of the data to be difficult to extract.

leftwo:
If the DNS names are working for Oximeter, I'll add that in place of the hard coded address that is here now, but other than that, I think the general shape of this is ready for any comments. 
The crucible side still needs to land and once that does I'll update this PR

server/src/lib/stats.rs
```diff
@@ -0,0 +1,163 @@
...
+/// This starts a dropshot server, and then registers the PropStatOuter
+/// producer with Oximeter.
+/// Once registered, we return the server to the caller.
+pub async fn prop_oximeter(
+    id: Uuid,
+    my_address: SocketAddr,
+    plog: Logger,
+) -> anyhow::Result<Server> {
+    // TODO: Replace with proper oximeter registration
+    let ra = IpAddr::from_str("fd00:1122:3344:101::3").unwrap();
```
leftwo, line 103: This hard coded address will need to change.
Either, have it passed in (from somewhere??) or do the DNS lookup ourselves.

leftwo: I updated the propolis side to expect the metric registration address from sled agent.  This pushes the burden of figuring out what the address is over to sled-agent, who should have that information.

propolis/Cargo.toml
```diff
@@ -21,7 +21,8 @@ viona_api = { path = "../viona-api" }
 usdt = { version = "0.3.2", default-features = false }
 tokio = { version = "1", features = ["full"] }
 futures = "0.3"
-crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "fed3e8ca7762130ee146fc516a4ef6eed2b91629", optional = true }
+crucible = { git = "https://github.com/oxidecomputer/crucible", branch = "producemetric", optional = true }
+oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "main", optional = true }
```
gjcolombo, line 25: Just to make sure I'm not misreading: it looks like `oximeter` is only optional if `crucible` is disabled; otherwise the Crucible module will be missing the `ProducerRegistry` type. Is that right? (I'm trying to reason through how we might keep this module optional when we extend metrics collection to more than just Crucible.)

server/src/lib/stats.rs
```diff
@@ -0,0 +1,160 @@
...
+        let datum = pso.run_count.datum_mut();
+        *datum += 1;
+    }
+}
+
+// This trait is what is called to update the data to send to Oximeter.
+// It is called on whatever interval was specified when setting up the
+// connection to Oximeter.  Since we get a lock in here (and on every
+// IO, don't call this too frequently, for some value of frequently that
+// I'm not sure of.
```
gjcolombo, line 74: Nit: might be good to have a `const OXIMETER_STATS_INTERVAL` somewhere with the warning about querying too frequently nearby (or just move this down by line 126 where the interval is provided).

server/src/lib/stats.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::sync::{Arc, Mutex};
+use uuid::Uuid;
+
+// These structs are used to construct the desired stats for Oximeter.
+#[derive(Debug, Copy, Clone, Target)]
+struct InstanceUuid {
+    pub uuid: Uuid,
+}
+#[derive(Debug, Default, Copy, Clone, Metric)]
+pub struct Rebooted {
```
gjcolombo, line 22: Nit: this excludes guest OS-initiated reboots, so at some point we might want to clarify the name of this metric. Not worth blocking this PR for, but maybe worth filing a tracking issue for deciding exactly what server-wide metrics we want to export & making sure they're labeled/computed properly.

client/src/api.rs
```diff
@@ -33,6 +33,9 @@ pub struct InstanceEnsureRequest {
 
     // base64 encoded cloud-init ISO
     pub cloud_init_bytes: Option<String>,
+
+    // If provided, register here to post metrics to Oximeter
+    pub metrics: Option<SocketAddr>,
```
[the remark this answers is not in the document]
leftwo: So in my first pass I tried something like this, passing metric info to propolis-server on the command line because it does seem like that makes the most sense.  Doing this involved merging the stats endpoint into the existing propolis dropshot server, and that resulted in two issues I then had to solve and required a bit of wrestling with Oximeter to get it to let me create a shared dropshot server that both propolis-server could use, and would be available to serve metrics for Oximeter to collect data from.  I worked a bit with Ben Naecker on this, and we ended up thinking it would be easier to detangle the two services from each other.

The first issue resulting from above is that I need the UUID of the instance to register with Oximeter, and I don't have that when the propolis-server first starts (It comes later in the `instance_ensure()` call).  This requirement comes from the way the initial stat collection registration inside of Oximeter works.

The second issue is, in order to pass down the "ProducerRegistry" object to any library (like crucible) I have to clone it (from the propolis-server dropshot endpoint that is created in server/src/main.rs).  Once I have this clone, I need to attach it to something that `instance_ensure()` can access when it runs.  This can probably be part of `server::Context`, but I think I still need the `Arc<Mutex<Option` around it as I believe Context is shared in a way that requires it.

A third minor issue is that Oximeter will be requesting stats from the propolis-server endpoint every X seconds, and I could not figure out how to turn off just the metrics endpoint from the server log, resulting in a bunch of collection messages.

propolis/Cargo.toml
leftwo: I'm actually not 100% sure, but my intent here was trying to allow propolis standalone to build and not have to pull in both crucible and then oximeter.

Crucible needs to know about ProducerRegistry, and now propolis-server will also need to know about it, even if Crucible is not present.

server/src/lib/stats.rs
leftwo: Yeah, that's a good point about not seeing OS level reboots.  All this metric is really counting is how many times someone called the "/instance/state" endpoint with the "Reset" state. 

I was hoping with these changes to lay the groundwork that myself or others could come back and build on to create more meaningful stats inside Propolis.  These changes just open the door.  I tried to pick something that was both simple and easy to see change.  But, yeah, I agree the name here is not the best choice.

leftwo: https://github.com/oxidecomputer/propolis/issues/153

client/src/api.rs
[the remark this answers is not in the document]
luqmana: How about moving this back to being a command-line/server option but start server the metrics server in ensure?

server/src/lib/stats.rs
leftwo: Updated in 395d4db.
I tried to improve the comments a bit more as well.

server/src/lib/stats.rs
leftwo: I also updated the name in 395d4db.

client/src/api.rs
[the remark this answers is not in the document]
leftwo: > How about moving this back to being a command-line/server option but start server the metrics server in ensure?

The metric server does start inside instance_ensure(), but that start only happens if the `InstanceEnsureRequest` `metrics` field contains the metric endpoint (nexus internal API).  I should have mentioned that requirement above.

The three things I need to start the metrics endpoint are:
1. The UUID for the instance.    <-- Comes from Nexus in `api::InstanceEnsureRequest`
2. The I.P. address of the propolis instance.       <-- Comes from server/src/main.rs command line option
3. The I.P. address and port for Nexus (so I can register).   <-- Also from Nexus, though I could potentially use DNS

I can make an option to the server when starting that could request that a metric endpoint will be later started in `instance_insure()`.  To do that I would make `propolis_addr` an`Option` and only start metrics if both `propolis_addr.is_some() && request.metrics.is_some()`.

This would mean to enable metrics, propolis server would have to be started with the metrics option, and the InstanceEnsureRequest has the metric registration.  

Does that sound like the right way to do this?  I'm not sure if wanting a command line option for the server to enable metrics was what you  wanted, or if it was just an idea.

server/src/lib/server.rs
```diff
@@ -93,6 +96,12 @@ pub(crate) struct InstanceContext {
         Mutex<BTreeMap<Uuid, Arc<propolis::block::CrucibleBackend>>>,
 }
 
+#[derive(Debug, Clone)]
+pub struct InstanceMetrics {
+    pub(crate) producer_registry: Arc<Mutex<Option<ProducerRegistry>>>,
+    pub(crate) pso: Arc<Mutex<Option<PropStatOuter>>>,
```
luqmana, line 102: It seems to me these `Arc<Mutex<...>>` are a bit redundant as both `ProducerRegistry` and `PropStatOuter` already contain `Arc`'s themselves and have `Clone` impls that do the right thing if you just passed them around directly.

I would also just remove the `Option`'s here and instead wrap `InstanceMetrics` as a whole with `Option` in `Context` below.

propolis/src/block/crucible.rs
```diff
@@ -35,21 +36,23 @@ impl CrucibleBackend {
         gen: u64,
         request: VolumeConstructionRequest,
         read_only: bool,
+        producer_registry: Arc<tokio::sync::Mutex<Option<ProducerRegistry>>>,
```
luqmana, line 39: Should be possible to just take `Option<ProducerRegistry>` here. It doesn't seem like we would want crucible to be modifying it behind the `Mutex`. See other comment re: Arc.

client/src/api.rs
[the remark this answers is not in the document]
luqmana: Right, (1) makes sense as part of the ensure request and no argument with (2). I meant having (3) passed to propolis on startup because that's not really a property of the instance per-se.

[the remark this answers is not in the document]
leftwo: > Right, (1) makes sense as part of the ensure request and no argument with (2). I meant having (3) passed to propolis on startup because that's not really a property of the instance per-se.

Ah, yeah, Okay. Thanks for breaking it down.

So,
On startup of propolis-server if I find a metric_register address arg, I will include that metric_register address and the propolis_address as part of the `server::Context` that is created.

Then, in `instance_ensure()`, the presence of metric_register and propolis_address will indicate that it is time to spin up the dropshot endpoint for metrics and, along with the instance UUID I have all the info I need.

I believe this means I can take the `metrics` field out of `propolis_client::api::InstanceEnsureRequest` as that is no longer required, as the decision for metrics has already been made.

This does sound like an improvement, and I think it also satisfies the original concern brought up by @gjcolombo that started this whole thing!  Yeah!

server/src/lib/server.rs
leftwo: Yes, I agree on the redundant part.

I tried to get things to work without them, but either I could not figure it out, or something about the way the `server::Context` was passed around was preventing just a straight `Option<T>` here.  I'll give it another try and if I can't figure it out then I'll come back for help :)

luqmana: Yea, the ownership of things gets complicated. Feel free to ping me if you run into anything gnarly.

client/src/api.rs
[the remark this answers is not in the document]
leftwo: I've implemented this idea, so this should be addressed now in the latest commit.

server/src/lib/server.rs
```diff
@@ -243,6 +262,56 @@ async fn instance_ensure(
         }));
     }
 
+    // Determine if we need to setup the metrics endpoint or not.
+    if request.metrics.is_some() {
```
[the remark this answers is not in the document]
leftwo: This section was reworked, so I don't think the request applies after the changes.

server/src/lib/server.rs
leftwo: The latest commit does improve this, (thanks for your help out of band) and I was also able to keep `producer_registry` only inside `instance_ensure()` so things got simpler as well.

propolis/src/block/crucible.rs
leftwo: Yep!  Thanks for your suggestions here.

gjcolombo (approved):
Looks great from my end--thanks for the changes!

leftwo:
This now points to the latest crucible rev in main, where all required changes are.

standalone/Cargo.toml
```diff
@@ -27,3 +27,4 @@ slog-term = "2.7"
 default = ["dtrace-probes"]
 dtrace-probes = ["propolis/dtrace-probes"]
 crucible = ["propolis/crucible"]
```
pfmooney, line 29: Now that the crucible backend is using oximeter, shouldn't the feature list `propolis/oximeter` itself here?

propolis/Cargo.toml
```diff
@@ -21,7 +21,8 @@ viona_api = { path = "../viona-api" }
 usdt = { version = "0.3.2", default-features = false }
 tokio = { version = "1", features = ["full"] }
 futures = "0.3"
-crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "fed3e8ca7762130ee146fc516a4ef6eed2b91629", optional = true }
+crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "2add0de8489f1d4de901bfe98fc28b0a6efcc3ea", optional = true }
+oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "main", optional = true }
```
pfmooney, line 25: Since these are getting long, maybe define them as `[dependencies.crucible]` style sections?

standalone/Cargo.toml
leftwo: Updated

propolis/Cargo.toml
leftwo: Updated

leftwo:
@gjcolombo  I merged with your changes like we discussed in chat.  Let me know if that looks good.

I tested these changes and stats are still produced and oximeter collects them.

PR #148, opened 2022-06-23
fix existing clippy warnings from 52e0d2c

gjcolombo:
`cargo clippy` has some concision suggestions and flags a couple of unnecessary `clone`s. This PR tidies these up.


pfmooney (approved):
Would omit the reference to the git hash from the commit message, since it's implied by the parent relationship anyways.

PR #151, opened 2022-06-29
Use fourcc value from the guest to set pixel format for VNC.

jordanhendricks:
### Summary

This PR uses the new interfaces added in https://github.com/oxidecomputer/rfb/pull/6 to set the pixel format for the VNC server based on the guest's fourcc. Fourcc is a 4-byte code that maps to a pixel format. Support for understanding these codes has been added into the rfb crate.

### Details

Previously, when booting a windows VM and viewing its framebuffer with VNC, one might see orange where the screen should be blue (such as the "Orange Screen of Death"). Here is an image of an orange screen that should be blue: 
ksnip_20220503-093009

The problem here was two-fold: first, the VNC server in the rfb crate didn't have a way for consumers to specify what the input pixel format is (which in this case, we get from the fourcc value of the guest). Second, the rfb protocol allows for clients to request an alternate pxiel format, and the rfb crate didn't have support for translating between formats. So the client would display pixels in the format it requested, by the pixels wouldn't actually be in that format. Both of these issues are addressed for our purposes in https://github.com/oxidecomputer/rfb/pull/6.

### Testing
Here's an example of the windows lock screen before -- note that the sky is orange (orange where there should be blue):
ksnip_20220429-200648

Now, when we connect to a windows VM, the colors look correct, regardless of the pixel format requested by the client.

Example from noVNC (which requests pixel format little-endian xBGR):
ksnip_20220629-125435

Example from tigervnc (which requests pixel format little-endian xRGB): 
ksnip_20220629-125742

### DIFF
--- a/propolis/src/hw/qemu/ramfb.rs
+++ b/propolis/src/hw/qemu/ramfb.rs
@@ -36,7 +36,7 @@ lazy_static! {
 
 fn fourcc_bytepp(fourcc: u32) -> Option<u32> {
     match fourcc {
-        // edk2 default - xRGB, 4 bytes per pixels
+        // The edk2 default: XR24, little-endian xRGB with 4 bytes per pixel.
         0x34325258 => Some(4),
         _ => None,
     }
@@ -72,9 +72,22 @@ impl Config {
         Some(())
     }
 
-    pub fn get_framebuffer_info(&self) -> (u64, u32, u32) {
-        (self.addr, self.width, self.height)
+    pub fn get_framebuffer_spec(&self) -> FramebufferSpec {
+        FramebufferSpec {
+            addr: self.addr,
+            width: self.width,
+            height: self.height,
+            fourcc: self.fourcc,
+        }
     }
+}
+
+#[derive(Clone, Copy)]
+pub struct FramebufferSpec {
+    pub addr: u64,
+    pub width: u32,
+    pub height: u32,
+    pub fourcc: u32,
 }
 
 type NotifyFn = Box<dyn Fn(&Config, bool) + Send + Sync + 'static>;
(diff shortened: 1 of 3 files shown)


propolis/src/hw/qemu/ramfb.rs
```diff
@@ -36,7 +36,7 @@ lazy_static! {
 
 fn fourcc_bytepp(fourcc: u32) -> Option<u32> {
     match fourcc {
-        // edk2 default - xRGB, 4 bytes per pixels
+        // The edk2 default: XR24, little-endian xRGB with 4 bytes per pixel.
         0x34325258 => Some(4),
```
gjcolombo, line 40: Can/should this be `rfb::pixel_formats::fourcc::FOURCC_XR24`?

server/src/lib/vnc.rs
```diff
@@ -65,12 +83,30 @@ impl PropolisVncServer {
...
-            let (addr, w, h) = config.get_framebuffer_info();
-            let fb = RamFb::new(addr, w, h);
-            self.initialize_framebuffer(fb).await;
+            let fb_spec = config.get_framebuffer_spec();
+            let fb = RamFb::new(fb_spec);
+            self.initialize_framebuffer(fb.clone()).await;
+
+            match fourcc::fourcc_to_pixel_format(fb.fourcc) {
+                Ok(pf) => {
+                    vnc_server.set_pixel_format(pf).await;
```
gjcolombo, line 100: Small nit: Does this routine require that the caller have exclusive access to `vnc_server`? By itself and without that requirement, it looks like there's a race here (if two callers supply different configs, the last one to call `initialize_framebuffer` won't necessarily be the last one to call `set_pixel_format`, so the two could get out of sync).

The only caller I can see locks the `vnc_server` it provides before calling this function, so I don't think there's a bug here so much as a small seam that could tear later. 

Would it make sense to make `vnc_server` a `&mut` reference to clarify that only one caller at a time can call this with the same server?

gjcolombo (approved):

jordanhendricks:
Merging for now because the rfb commit is merged and will break propolis otherwise. I'll address Greg's feedback in a follow up PR.

PR #159, opened 2022-07-08
Support viona export/import

pfmooney:
Not to be merged until [14755](https://www.illumos.org/issues/14755) (CR [2202](https://code.illumos.org/c/illumos-gate/+/2202)] has landed.


pfmooney:
I plan to merge this tomorrow now that 14755 has been integrated into illumos.  That'll allow for time for helios builds to run so folks can easily cross the flag day (since viona uses the new interfaces)

PR #160, opened 2022-07-11
Allow instance_serial_task to block after last client disconnects

### LINKED ISSUE #158: Ending a propolis-cli serial session causes its server to stop responding
### Repro steps

1. Create a TOML describing a simple VM (e.g. just an Alpine Linux image with a virtio-block backend)

From the Propolis target directory:

2. `pfexec ./propolis-server run $(VM_TOML) 127.0.0.1:9000`
3. `./propolis-cli -s 127.0.0.1 -p 9000 new -m 256 -c 2 vm1`
4. `./propolis-cli -s 127.0.0.1 -p 9000 state run`
5. `./propolis-cli -s 127.0.0.1 -p 9000 serial`
6. Wait for image to boot & interact with it a bit
7. Ctrl-C to exit the serial session
8. `./propolis-cli -s 127.0.0.1 -p 9000 serial`

Expected: The new CLI reconnects to the serial console.
Observed: Nothing happens. Other CLI commands (e.g. `propolis-cli -s 127.0.0.1 -p 9000 state stop` and `get`) all hang without connecting to the server.

This appears to have been introduced in 025ac85 (repros there but not on its parent bcf9f51). This also repros on 23a38f4 (HEAD at the time of writing).

### Additional observations

The first connection in step 5 produces some server logs about the serial API endpoint receiving a connection:

```
Jul 08 03:21:11.329 INFO accepted connection, remote_addr: 127.0.0.1:45487, local_addr: 127.0.0.1:9000
Jul 08 03:21:11.330 INFO request completed, response_code: 101, uri: /instance/serial, method: GET, req_id: dc1f442c-f629-4c77-828b-df53aaf816dd, remote_addr: 127.0.0.1:45487, local_addr: 127.0.0.1:9000
```

If you terminate the Propolis server after step 8 (while the CLI command is still processing), you get the following error:

```
Error: failed to create serial websocket stream

Caused by:
    0: IO error: Connection reset by peer (os error 131)
    1: Connection reset by peer (os error 131)
```

If instead I try the following:

1. Launch the server and VM as above
2. Run `propolis-cli serial`; call this client C1
3. Run `propolis-cli serial` again in another terminal; call this client C2
4. C1 and C2 mirror each other. Ctrl-C to kill client C1 and client C2 still works.
5. Run `propolis-cli serial` again to create client C3

I observe that C3 works properly and that C2/C3 mirror each other. Also, if I run other CLI commands while the serial session is active, they work normally--it's only when the serial console session goes down that the server stops responding.

My guess is that dropping the last serial connection is causing `instance_serial_task` to wedge in some way that prevents the server from servicing any other requests. I haven't done much to analyze this yet; `pstack` on an unresponsive server shows a tokio worker running the instance serial task, but there are lots of other seemingly idle workers, so it doesn't immediately seem to be an "oversubscribed the executor" sort of problem.

gjcolombo:
Each iteration of the Propolis server's serial task creates streams from its websocket stream/sink collections, passes these to `futures_util`'s `for_each_concurrent`, and then includes the resulting futures in its `select!` statement. If there are no active serial connections, these futures are trivially ready, so the serial task loops without blocking, which causes subsequent connections to the Propolis server to fail. Fix this by instead creating a terminated fused future to select on when these collections are empty.

In addition, reorder the serial task's `select!` block so that incoming data from the intermediate channel (the shared channel onto which clients' messages are pushed) is processed before new messages are read from the incoming streams. This prevents the same connection from being removed twice (without this the intermediate channel will fill with Close messages from the same client before the closure is processed, causing a double removal).

Tests:
- Verified that it's now possible to attach a serial client, detach it, and attach to the serial console again.
- Also verified that other CLI commands (e.g. `get`) work on an instance that has previously had a serial console attached.
- Verified that multiple serial clients work as expected (input on one client is reflected to the others, output from the guest is broadcast to all clients).

Closes #158.

### DIFF
--- a/server/src/lib/server.rs
+++ b/server/src/lib/server.rs
@@ -501,39 +501,48 @@ async fn instance_serial_task(
     let mut ws_streams: Vec<SplitStream<WebSocketStream<Upgraded>>> =
         Vec::new();
 
-    let (send_ch, mut recv_ch) = mpsc::channel(4);
+    let (send_ch, mut recv_ch) = mpsc::channel(1);
 
     loop {
-        let (uart_read, ws_send) = match &cur_output {
-            None => (
-                serial.read_source(&mut output, actx).fuse(),
-                Fuse::terminated(),
-            ),
-            Some(r) => (
-                Fuse::terminated(),
-                futures::stream::iter(
-                    ws_sinks
-                        .iter_mut()
-                        .zip(std::iter::repeat(Vec::from(&output[r.clone()]))),
-                )
-                .for_each_concurrent(4, |(ws, bin)| {
-                    ws.send(Message::binary(bin)).map(|_| ())
-                })
-                .fuse(),
-            ),
-        };
+        let (uart_read, ws_send) =
+            match &cur_output {
+                None => (
+                    serial.read_source(&mut output, actx).fuse(),
+                    Fuse::terminated(),
+                ),
+                Some(r) => (
+                    Fuse::terminated(),
+                    if !ws_sinks.is_empty() {
+                        futures::stream::iter(ws_sinks.iter_mut().zip(
+                            std::iter::repeat(Vec::from(&output[r.clone()])),
+                        ))
+                        .for_each_concurrent(4, |(ws, bin)| {
+                            ws.send(Message::binary(bin)).map(|_| ())
+                        })
+                        .fuse()
+                    } else {
+                        Fuse::terminated()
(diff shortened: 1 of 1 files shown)


luqmana:
cc @lifning

lifning (approved):

gjcolombo:
Thanks for reviewing, @lifning!

PR #161, opened 2022-07-11
Minor fixes to live migration protocol

gjcolombo:
This PR fixes a few small bugs in live migration:

- Make migration targets move to the correct "migrate destination" state after migration is finished so that they can be moved to the running state.
- Fix a URI disagreement between the Dropshot definitions in server.rs and the migration protocol initiation code in migrate/mod.rs.
- Change the VMM's `mem_bounds` routine to return the last GPA plus one as the upper bound on the address range. This prevents an assertion failure when constructing dirty bitmaps on the source.
- Iterate through GPAs correctly when computing dirty bitmaps on the source.
- Temporarily reorder migration steps on the source so that `pause` precedes `ram_push` instead of succeeding it. This ensures the source can't dirty pages once RAM is transferred. This lengthens blackout, but this process can be optimized to shorten it again. Assert at the end of source migration that there are no dirty pages left.

Tested by migrating an Alpine Linux VM with a virtio-block device and verifying that the target VM runs correctly and has an operable serial console and that the source doesn't assert.

### DIFF
--- a/bin/propolis-server/src/lib/migrate/mod.rs
+++ b/bin/propolis-server/src/lib/migrate/mod.rs
@@ -377,10 +377,8 @@ pub async fn dest_initiate(
 
     // TODO: https
     // TODO: We need to make sure the src_addr is a valid target
-    let src_migrate_url = format!(
-        "http://{}/instance/{}/migrate/start",
-        migrate_info.src_addr, migrate_info.src_uuid
-    );
+    let src_migrate_url =
+        format!("http://{}/instance/migrate/start", migrate_info.src_addr);
     info!(log, "Begin migration"; "src_migrate_url" => &src_migrate_url);
 
     let body = Body::from(
--- a/bin/propolis-server/src/lib/migrate/mod.rs
+++ b/bin/propolis-server/src/lib/migrate/mod.rs
@@ -375,7 +375,7 @@ pub async fn dest_initiate(
     // This should be a fresh propolis-server
     assert!(migrate_task.is_none());
 
-    // TODO: https
+    // TODO(#165): https
     // TODO: We need to make sure the src_addr is a valid target
     let src_migrate_url =
         format!("http://{}/instance/migrate/start", migrate_info.src_addr);
--- a/lib/propolis/src/vmm/machine.rs
+++ b/lib/propolis/src/vmm/machine.rs
@@ -487,7 +487,7 @@ impl<'a> MemCtx<'a> {
             .map
             .highest_addr(|entry| matches!(entry.kind, MapKind::SysMem(_, _)))?
             as u64;
-        Some(RangeInclusive::new(GuestAddr(lowest), GuestAddr(highest)))
+        Some(GuestAddr(lowest)..=GuestAddr(highest))
     }
 }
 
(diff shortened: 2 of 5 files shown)


propolis/src/vmm/machine.rs
```diff
@@ -476,7 +476,7 @@ impl<'a> MemCtx<'a> {
         }
     }
 
-    /// Returns the [lowest, highest] memory addresses in the space, inclusive.
+    /// Returns the [lowest, highest + 1) memory addresses in the space.
```
dancrossnyc, line 479: We've generally shied away from using clopen ranges because of overflow issues at the highest addresses, though I don't think that's much of a concern because the physical address space is constrained to be well within the range of a u64. Regardless, it strikes me that the original comment was just wrong; that range wasn't inclusive. One wonders whether another fix might be using an inclusive range?

gjcolombo: (For completeness, I also looked at changing `offer_ram` and its callees so that they expect `end` to be the last valid address instead of the one-past-the-end address. That also solves the expectation mismatch, but it creates a lot of code of the form `end = start + (page_count * page_size) - 1`, which I didn't love--it seems like most callers' natural inclination is to deal with addresses that are multiples of 4K, and making them all adjust downward by a byte seemed like it'd be a pebble in the shoes of future maintainers of this code. But I could be persuaded otherwise.)

bin/propolis-server/src/lib/migrate/mod.rs
```diff
@@ -377,10 +377,8 @@ pub async fn dest_initiate(
 
     // TODO: https
     // TODO: We need to make sure the src_addr is a valid target
-    let src_migrate_url = format!(
-        "http://{}/instance/{}/migrate/start",
-        migrate_info.src_addr, migrate_info.src_uuid
-    );
+    let src_migrate_url =
+        format!("http://{}/instance/migrate/start", migrate_info.src_addr);
```
dancrossnyc, line 381: I realize this was in the context of a formatting change to remove the instance ID, but do we want to hardcode http here? Perhaps it's something to address in a followup; maybe file a bug to audit protocol usage?

lib/propolis/src/vmm/machine.rs
```diff
@@ -486,7 +487,7 @@ impl<'a> MemCtx<'a> {
             .map
             .highest_addr(|entry| matches!(entry.kind, MapKind::SysMem(_, _)))?
             as u64;
-        Some(GuestAddr(lowest)..GuestAddr(highest))
+        Some(RangeInclusive::new(GuestAddr(lowest), GuestAddr(highest)))
```
dancrossnyc, line 490: Purely a question, but does the `..=` syntax not work here? It may require implementing some trait that's otherwise noisy, so this is perfectly good to go; I really am just curious.

dancrossnyc (approved):
This looks great. Thanks!

bin/propolis-server/src/lib/migrate/mod.rs
gjcolombo: There's a TODO for this nearby. I've filed #165 to track it.

lib/propolis/src/vmm/machine.rs
gjcolombo: Good call--it does work here, it just didn't occur to me to use it. Done.

PR #162, opened 2022-07-12
instance specs: define what it means to be "migration compatible"

gjcolombo:
Add to the instance spec module the (perhaps poorly-named) `MigrationCompatible` trait, which spec elements implement to allow two elements to decide if they describe components that are similar enough to permit migration from a VM bearing one to a VM bearing the other. This is a prerequisite to using instance specs in the live migration preamble to decide if a source and target are compatible.

Some compatibility checks are just straight-up checks for equality, but others are subtler, notably those for backends. Generally, two backends are compatible if they're the same kind of backend (e.g. two Crucible storage backends, two file-based storage backends, two virtio-net vNICs) irrespective of the backend _configuration_ (Crucible request contents, file paths, vNIC names).

Tests: new unit tests included with this change plus manual smoke tests; this PR just defines the new trait and implements it for spec elements, so it shouldn't change any existing behavior.


lib/propolis-client/src/instance_spec.rs
```diff
@@ -52,8 +52,52 @@ use serde::{Deserialize, Serialize};
...
+                Some(other_val) => {
+                    if !this_val.is_migration_compatible(other_val) {
+                        return false;
+                    }
+                }
+                None => return false,
+            }
+        }
+
+        true
```
luqmana, line 95: Since loops vs iter combinators came up in the space today :P

```suggestion
        self.iter().all(|(key, this_val)| {
            other.get(key).map_or(false, |other_val| {
                this_val.is_migration_compatible(other_val)
            })
        })
```

lib/propolis-client/src/instance_spec.rs
```diff
@@ -168,6 +218,14 @@ pub enum StorageBackendKind {
     InMemory,
 }
 
+impl MigrationCompatible for StorageBackendKind {
+    fn is_migration_compatible(&self, other: &Self) -> bool {
+        // Two storage backends are compatible if they have the same kind,
+        // irrespective of their configurations.
+        std::mem::discriminant(self) == std::mem::discriminant(other)
```
luqmana, line 225: Does this preclude Crucible (or other backends) from changing in an incompatible way? Or would that be handled by introducing a new `StorageBackendKind`?

lib/propolis-client/src/instance_spec.rs
gjcolombo: WHOA

dancrossnyc: Nifty!

luqmana (approved):

gjcolombo:
Will merge once the latest checks have completed. Thanks for the review!

PR #164, opened 2022-07-14
Move inner VNC server state behind a single lock.

jordanhendricks:
This is a follow up PR to https://github.com/oxidecomputer/propolis/pull/151 that addresses @gjcolombo's feedback about a possible race in state on the VNC server (see this comment: https://github.com/oxidecomputer/propolis/pull/151#discussion_r911525611). To address this, I've simplified this code a bit by putting all of the state on the vnc server behind a single lock, which I hope makes this code easier to reason about.


gjcolombo (approved):
LGTM--thanks for following up here!

PR #169, opened 2022-07-26
Use instance spec compatibility checks in migration sync phase

gjcolombo:
Stash an instance's spec in its server context. During a migration, have the source send its spec to the target, and have the target use `is_migration_compatible` to check the specs for compatibility.

Tested via:

- Manual migrations of Alpine Linux and WS2019 VMs
- Checked that migrations fail between servers with various incompatible device specifications
- cargo test

### DIFF
--- a/bin/propolis-server/src/lib/migrate/destination.rs
+++ b/bin/propolis-server/src/lib/migrate/destination.rs
@@ -4,6 +4,7 @@ use hyper::upgrade::Upgraded;
 use propolis::common::GuestAddr;
 use propolis::instance::MigrateRole;
 use propolis::migrate::{MigrateStateError, Migrator};
+use propolis_client::instance_spec::MigrationCompatible;
 use slog::{error, info, warn};
 use std::io;
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -86,12 +87,14 @@ impl DestinationProtocol {
...
-        if preamble.vm_descr.vcpus != vec![0u32, 1, 2, 3] {
+        info!(self.log(), "Destination read Preamble: {:?}", preamble);
+        if !preamble
+            .instance_spec
+            .is_migration_compatible(&self.mctx.instance_spec)
+        {
             error!(
                 self.log(),
-                "invalid CPU count in preamble ({:?})", preamble.vm_descr.vcpus
+                "Source and destination instance specs incompatible"
```
luqmana, line 97: Should we dump both instance specs here to help diagnose this? 
Not blocking on this PR but perhaps it might make sense in the future to adjust `is_migration_compatible` to return a `Result<(), ErrOfSomeKind>` instead of just a `bool`.

bin/propolis-server/src/lib/migrate/preamble.rs
```diff
@@ -1,75 +1,14 @@
...
-            pci: Vec::new(),
-            ports: Vec::new(),
-        }
-    }
-}
-
 #[derive(Deserialize, Serialize, Debug)]
 pub(crate) struct Preamble {
-    pub vm_descr: VmDescr,
+    pub instance_spec: InstanceSpec,
```
luqmana, line 6: 🎉 love the cleanup!

luqmana (approved):
Looks good, just a small nit in an error case. Thanks!

bin/propolis-server/src/lib/migrate/destination.rs
gjcolombo: Yeah, that's a good call. Filed #178 to track.

PR #179, opened 2022-07-27
PHD: add support for listing and filtering tests

### LINKED ISSUE #171: PHD: support listing and filtering test cases in command line arguments
PHD doesn't support any kind of conditional test execution or filtering today, which makes it impractical to try to reproduce failures in a specific test case. It also doesn't support listing of test cases in a way that would make it possible to use with a runner like [`cargo-nextest`](https://nexte.st/book/custom-test-harnesses.html) if we ever chose to do so.

gjcolombo:
Add the `phd-runner list` subcommand to enumerate tests and the `--include-filter` and `--exclude-filter` switches to aid in conditionally selecting tests to run or list.

Tested via ad hoc use of these switches.

Closes #171.


jordanhendricks (approved):
Sweet! Seems like a useful feature :+1:

gjcolombo:
Thanks for reviewing!

PR #185, opened 2022-08-02
Have instance spec compat checks return `Result`s

### LINKED ISSUE #178: Log better error messages on instance spec mismatch
After #169 merges, an instance spec mismatch between source and target will cause migration to fail. This will log a rather nondescript error message saying only that a mismatch occurred without indicating what the problem was. This should be expanded to point to the actual source of the issue.

gjcolombo:
Define two error types for use when comparing instance specs: the `SpecMismatchDetails` type contains enumerants that describe specific kinds of mismatches, and the `MigrationCompatibilityError` type wraps it with some additional context as to the fields that were being compared when an error was generated. Although most `SpecMismatchDetails` errors are rather specific, this scheme lets "generic" mismatches like "you had the wrong number of entries in a map" be returned alongside information about which specific map had the problem.

Tests:
- Manual live migration between incompatible VMs still fails and produces a more descriptive message than before
- Manual live migration between compatible VMs succeeds
- Manual PHD smoke tests

Closes #178.

### DIFF
--- a/bin/propolis-server/src/lib/migrate/destination.rs
+++ b/bin/propolis-server/src/lib/migrate/destination.rs
@@ -88,13 +88,13 @@ impl DestinationProtocol {
             }
         }?;
         info!(self.log(), "Destination read Preamble: {:?}", preamble);
-        if !preamble
+        if let Err(e) = preamble
             .instance_spec
             .is_migration_compatible(&self.mctx.instance_spec)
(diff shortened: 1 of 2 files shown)


luqmana (approved):

gjcolombo:
Thanks @luqmana for reviewing!

PR #187, opened 2022-08-05
Add auto-destruct interface

pfmooney:


pfmooney:
It should be ready for PHD and friends to consume once the underlying machine has been updated to bits which include the necessary illumos-gate change.

gjcolombo (approved):

luqmana:
Which were the upstream bits needed for this?

pfmooney:
> Which were the upstream bits needed for this?

illumos/illumos-gate@aa39f6d0fd4e491afca04b12f49a18ce955efc79

PR #188, opened 2022-08-05
Update crucible to include crucible-client-types crate

leftwo:
This updates the version of crucible to use the new `crucible-client-types` crate

Part of fixing https://github.com/oxidecomputer/omicron/issues/1537

### DIFF
--- a/bin/propolis-server/Cargo.toml
+++ b/bin/propolis-server/Cargo.toml
@@ -44,7 +44,7 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
 slog = "2.7"
-propolis = { path = "../../lib/propolis", features = ["crucible", "crucible-client-types", "oximeter"], default-features = false }
+propolis = { path = "../../lib/propolis", features = ["crucible-full", "oximeter"], default-features = false }
 propolis-client = { path = "../../lib/propolis-client" }
 rfb = { git = "https://github.com/oxidecomputer/rfb", branch = "main" }
 uuid = "1.0.0"
--- a/bin/propolis-server/Cargo.toml
+++ b/bin/propolis-server/Cargo.toml
@@ -44,7 +44,7 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
 slog = "2.7"
-propolis = { path = "../../lib/propolis", features = ["crucible", "oximeter"], default-features = false }
+propolis = { path = "../../lib/propolis", features = ["crucible", "crucible-client-types", "oximeter"], default-features = false }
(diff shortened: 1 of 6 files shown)


bin/propolis-server/Cargo.toml
```diff
@@ -44,7 +44,7 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
 slog = "2.7"
-propolis = { path = "../../lib/propolis", features = ["crucible", "oximeter"], default-features = false }
+propolis = { path = "../../lib/propolis", features = ["crucible", "crucible-client-types", "oximeter"], default-features = false }
```
pfmooney, line 47: If we add `crucible = ["crucible-client-types"]` to the `features` section of the propolis crates, does it handle this case for us automatically? (Since that dep is required to build crucible support in propolis anyways)

pfmooney:
Thanks for splitting those types out, btw.  Definitely a nice improvement

bin/propolis-server/Cargo.toml
leftwo: I tried adding a line like suggested above to `bin/propolis-server/Cargo.toml`, and it does not like it:

```
alan@atrium:propolis$ cargo build
error: failed to load manifest for workspace member `/home/alan/ws/propolis/bin/propolis-server`

Caused by:
  failed to parse manifest at `/home/alan/ws/propolis/bin/propolis-server/Cargo.toml`

Caused by:
  feature `crucible` includes `crucible-client-types` which is neither a dependency nor another feature
  ```

I'll poke at it a bit more and see if I can figure something out.  `crucible-client-types` is not really a `feature` of `crucible`, it's just another crate in the repo.  I also have an, ahem, learning opportunity here as I don't have a deep understanding of how this stuff works.

Comment (pfmooney) [reply bin/propolis-server/Cargo.toml:47]: That would go in `lib/propolis/Cargo.toml`, I think.

Comment (leftwo) [reply bin/propolis-server/Cargo.toml:47]: Same error for `lib/propolis/Cargo.toml`

However, I think the problem is a name conflict.
If I change the `[features]` section in `lib/propolis/Cargo.toml` to be `crucible-full`, then it works!

```
crucible-full = ["crucible", "crucible-client-types"]
```

Then, in `bin/propolis-server/Cargo.toml`:
```
propolis = { path = "../../lib/propolis", features = ["crucible-full", "oximeter"], default-features = false }
```

pfmooney (approved):

bin/propolis-server/Cargo.toml
pfmooney: Looks good.

PR #192, opened 2022-08-11
propolis-server: Simplify some error handling

pfmooney:


bin/propolis-server/src/lib/migrate/mod.rs
```diff
@@ -261,13 +261,15 @@ pub async fn source_start(
     ));
     info!(log, "Migration Source");
 
-    let mut context = rqctx.context().context.lock().await;
-    let context =
-        context.as_mut().ok_or_else(|| MigrateError::InstanceNotInitialized)?;
+    let inst_context = tokio::sync::MutexGuard::try_map(
```
luqmana, line 264: TIL about `MutexGuard::try_map` :open_mouth:

luqmana (approved):

bin/propolis-server/src/lib/migrate/mod.rs
pfmooney: It's super convenient!

Issue #176, opened 2022-07-27
PHD: break dependency on propolis-server


gjcolombo:
PHD depends on the propolis-server lib crate for its `Config` module, which PHD uses as a helper to write server config TOML files. This pulls almost the entire repo into PHD's dependency graph and accordingly drives up its build/link times. While not the end of the world, this is annoying. Possible fixes include:

- Refactor the `Config` module into a lower-dependency crate like `propolis-types`. Note that this is nontrivial because `Config` currently depends on a few Propolis library types.
- Use instance specs: add an API endpoint that takes a spec and initializes an instance from that. Instance specs are part of `propolis-client`, so this severs the `propolis-server` dependency entirely. It's also something we should consider doing anyway for the control plane's sake.
- Conjure a way to write config TOMLs that doesn't depend on the `Config` library. While doable this makes it easier for PHD and Propolis to get out of sync.

gjcolombo:
Resolved by https://github.com/oxidecomputer/propolis/commit/dacba04503c2cdbd7a2185f0852962f1e1d4587f.

PR #195, opened 2022-08-11
Split propolis-server config into separate crate

pfmooney:
When `phd-framework` was pulling in `propolis-server` for only the `Config` struct, it mean building the entirety of propolis-server _again_ because the phd profile differs from debug or release (using `panic = unwind`).  It'd be nice to split that config stuff out to avoid the extra building.

### DIFF
--- a/bin/propolis-server/Cargo.toml
+++ b/bin/propolis-server/Cargo.toml
@@ -46,6 +46,7 @@ serde_json = "1.0"
 slog = "2.7"
 propolis = { path = "../../lib/propolis", features = ["crucible-full", "oximeter"], default-features = false }
 propolis-client = { path = "../../lib/propolis-client" }
+propolis-server-config = { path = "../../crates/propolis-server-config" }
 rfb = { git = "https://github.com/oxidecomputer/rfb", branch = "main" }
 uuid = "1.0.0"
 base64 = "0.13"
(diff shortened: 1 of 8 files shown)


gjcolombo (approved):
Awesome, thanks for doing this! (This should close out #176.)

pfmooney:
Perhaps not the ideal way of doing it, but it'll definitely save a few minutes on builds.

gjcolombo:
IMO an even better solution would be to have a version of the `instance_ensure` API that accepts instance specs and some other non-TOML way of telling Propolis what bootrom to use. Then PHD could forgo building/writing the config TOML entirely. (Ideally Nexus would also be able to use that API, so it wouldn't be a PHD-only affordance; this has kinda-sorta been my plan for a while, but there are some details I need to pin down first.)

pfmooney:
> IMO an even better solution would be to have a version of the `instance_ensure` API that accepts instance specs and some other non-TOML way of telling Propolis what bootrom to use. Then PHD could forgo building/writing the config TOML entirely. (Ideally Nexus would also be able to use that API, so it wouldn't be a PHD-only affordance; this has kinda-sorta been my plan for a while, but there are some details I need to pin down first.)

Yeah, bootrom choice is absolutely something we'll need to specify via nexus, rather than in a local configuration.  That seems like reason enough to move forward with this quick fix, so we can work on doing The Right Thing™ for making the instance spec rich enough to include info like that.

pfmooney:
Merged in dacba04503c2cdbd7a2185f0852962f1e1d4587f

PR #197, opened 2022-08-12
propolis-server: reorganize the global Context structure and rename some of its members

gjcolombo:
Do a little housekeeping in the `propolis_server::server::Context` struct:

- Rename it to `DropshotEndpointContext` to make clear that this is the thing that gets passed to HTTP endpoint handlers
- Divide the context's members into `ServerObjects` and `StaticConfig` structs to distinguish the things the server creates/manages at runtime from initial server configuration that's needed to stand up new instances
- Rename some fields to (try to) make them clearer and more concise
- Add some doc comments
- Clean up unnecessary `pub` designations

No functionality should change in this PR, but I verified it passes PHD tests.

### DIFF
--- a/bin/propolis-server/src/lib/migrate/mod.rs
+++ b/bin/propolis-server/src/lib/migrate/mod.rs
@@ -18,7 +18,7 @@ use thiserror::Error;
 use tokio::{sync::RwLock, task::JoinHandle};
 use uuid::Uuid;
 
-use crate::server::Context;
+use crate::server::DropshotEndpointContext;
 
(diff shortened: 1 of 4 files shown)


jordanhendricks (approved):
I would like to get @pfmooney's feedback on this since he brought up this issue initially, but I like the new structure here (and I particularly appreciate the renaming of the Context object).

pfmooney (approved):

gjcolombo:
Sweet. Thanks folks!

Issue #155, opened 2022-07-06
Migration: creating a migration-target Crucible upstairs prevents (or may prevent) the source from pausing


gjcolombo:
@jmpesp and I were discussing live migration earlier and suspect the existing Crucible activation protocol and Propolis migration logic will combine to prevent migration of an instance with a Crucible backend. The troublesome sequence of operations is as follows:

1. Propolis P1 starts Crucible upstairs C1 which activates against a Crucible downstairs
2. A migration saga begins; Nexus creates Propolis P2 with the same device definitions as P1
3. When handling the instance ensure request that (AIUI) precedes migration, P2 creates Crucible upstairs C2, which activates against the same downstairs as C1
4. Activation obsoletes prior activations, so C1 is no longer connected to the downstairs, and all reads/writes will pend unless and until C1 reconnects

Then, when migration starts,

5. P1 asks all its devices and backends to quiesce so migration can continue
6. C1 handles the quiesce request by waiting for all outstanding I/Os to drain
7. But the I/Os won't complete because C2's activation from step 3 has invalidated C1's activation
8. So C1 never quiesces, so the migration never proceeds. Oops.

Note that we haven't verified this empirically yet--this is just what we think will happen based on code inspection and our understanding of the components and processes involved.

There are a couple possible ways to fix this:

1. Split the "entity creation" and "entity start" phases of the Propolis entity lifecycle such that backends only talk to external services when they're *starting*, as opposed to doing so when they're *created*. This solves the problem by deferring C2's activation from step 3 to a post-migration "start the target's entities" step.
2. Change the Crucible pause/migration discipline so that the Crucible backend migrates its in-flight operations instead of draining them. This solves the problem by allowing C1 to quiesce with I/Os outstanding; C2 then has to complete these I/Os as part of migration.

@jmpesp and I were leaning toward the first solution in our discussion today. I think the second one might work in principle, but would leave it to the Crucible experts to decide whether this sort of backend migration is possible for Crucible specifically.

leftwo:
(Hi, I'm on vacation this week, but thought i would add a comment anyway :)

I believe the initial design of storage migration was to be the #1 described above, the upstairs has a `connect to downstairs` phase, then an `activate` phase, and this `activation` was added for (in part) the purpose of migration.
There is also a `BlockOp::Deactivate` command that can be used when propolis wishes to cleanly disconnect an upstairs from its downstairs, though I'm not sure if this is available at the volume level.
 
[...]

PR #202, opened 2022-08-22
Add affordance to wait for pre-pause device quiesce to finish

### LINKED ISSUE #184: Panics in storage devices during source pause phase of LM
I've seen a couple of panics in storage devices while doing some ad hoc live migrations of VMs. This happens relatively rarely (it's not consistent at all and I haven't seen it in any of the PHD migration tests I've been developing this week).

This stack is in my notes from 26 July 2022:

```
   0: rust_begin_unwind
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/panicking.rs:142:14
   2: core::panicking::panic
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/panicking.rs:48:5
   3: <propolis::hw::nvme::PciNvme as propolis::inventory::Entity>::paused
             at /home/greg/propolis/lib/propolis/src/hw/nvme/mod.rs:943:9
   4: propolis_server::migrate::source::SourceProtocol::pause::{{closure}}
             at /home/greg/propolis/bin/propolis-server/src/lib/migrate/source.rs:206:29
```

And I just hit this one:

```
   0: rust_begin_unwind
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/panicking.rs:142:14
   2: core::panicking::panic
             at /rustc/a8314ef7d0ec7b75c336af2c9857bfaf43002bfc/library/core/src/panicking.rs:48:5
   3: propolis::block::Notifier::paused
             at /home/greg/propolis/lib/propolis/src/block/mod.rs:343:9
   4: <propolis::hw::virtio::block::PciVirtioBlock as propolis::inventory::Entity>::paused
             at /home/greg/propolis/lib/propolis/src/hw/virtio/block.rs:194:28
   5: propolis_server::migrate::source::SourceProtocol::pause::{{closure}}
             at /home/greg/propolis/bin/propolis-server/src/lib/migrate/source.rs:206:29
```

These are both failures to assert that a device has gotten its `pause` call by the time its `paused`. I took a short look around and initially suspect this race:

1. Source migration protocol's `pause` phase calls `self.mctx.instance.migrate_pause`
2. `migrate_pause` writes the "start pausing for migration" state to the instance and kicks the instance state driver
3. `migrate_pause` continues and starts calling `paused` on each device it captured in the inventory
4. The `paused` call for a storage device very occasionally happens before the state driver can call the `pause` routine for that device

If this theory is correct, inserting an artificial delay into the "pause devices" code in the state driver should make the problem much easier to reproduce.

gjcolombo:
During a migration, some devices expect to receive their `paused` calls from the migration task only after receiving a `pause` call from the instance state driver. `Instance::migrate_pause` directs the state driver to start pausing devices, but then returns immediately, allowing the migration task (`migrate_pause`'s caller) to race ahead of it and issue `paused` calls before devices are ready.

Add an affordance to `migrate_pause` that allows callers to supply a callback to invoke when the pause transition is completed. There are already some `Instance` members that are only relevant while migrating; take these and the new callback field and package them into a new `MigrateCtx` struct. Teach the migration task (and `standalone`'s snapshot code) to use this to wait for devices to get their pause notifications.

It likely makes some sense to have a more general "issue this state change and wait for it to finish" affordance than this one. I plan to think about that as part of a wholesale rethinking of the instance/migration state machines for #155.

Tested by adding an artificial delay to the state driver's transition callback path, verifying that this causes the race to repro 100% of the time, and verifying that the fix then mitigates the problem.

Fixes #184.

### DIFF
--- a/lib/propolis/src/instance.rs
+++ b/lib/propolis/src/instance.rs
@@ -199,8 +199,6 @@ pub enum TransitionError {
 type TransitionFunc =
     dyn Fn(State, Option<State>, &Inventory, &DispCtx) + Send + Sync + 'static;
 
-type PauseRequestsSentFunc = dyn FnOnce() -> () + Send + Sync;
-
 /// Context used when moving through migration states.
 #[derive(Default)]
 struct MigrateCtx {
@@ -208,10 +206,9 @@ struct MigrateCtx {
     /// must remain unblocked while quiescing for migration to proceed.
     task_id: Option<CtxId>,
 
-    /// A callback to invoke when all device pause requests have been sent (but
-    /// not necessarily completed; that is, an entity's `paused()` routine may
-    /// return an as-yet uncompleted future).
-    pause_requests_sent: Option<Box<PauseRequestsSentFunc>>,
+    /// This event is signaled when the state driver has finished sending pause
+    /// requests to all entities that receive them.
+    pause_requests_sent: Option<Arc<tokio::sync::Notify>>,
 
     /// The migration task writes to this channel to commit a transition to the
     /// migrate-paused state.
@@ -401,7 +398,9 @@ impl Instance {
         }
     }
 
-    /// Ask the instance to pause in prepartion for a migration.
+    /// Asks the instance to pause in preparation for a migration. Returns a
+    /// tokio Notify that the state driver signals when all entities have
+    /// been sent a pause request.
     ///
     /// This will ask the instance to transition to the Migrate Pause
     /// state. As part of that, we'll walk through the device inventory
@@ -414,8 +413,7 @@ impl Instance {
         &self,
         migrate_ctx_id: CtxId,
         pause_chan: std::sync::mpsc::Receiver<()>,
-        on_pause_reqs_sent: Option<Box<PauseRequestsSentFunc>>,
-    ) -> Result<(), TransitionError> {
+    ) -> Result<Arc<tokio::sync::Notify>, TransitionError> {
         let mut inner = self.inner.lock().unwrap();
 
         // Make sure the Instance is in the appropriate state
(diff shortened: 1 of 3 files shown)


lib/propolis/src/instance.rs
```diff
@@ -397,6 +414,7 @@ impl Instance {
         &self,
         migrate_ctx_id: CtxId,
         pause_chan: std::sync::mpsc::Receiver<()>,
+        on_pause_reqs_sent: Option<Box<PauseRequestsSentFunc>>,
     ) -> Result<(), TransitionError> {
```
luqmana, line 418: Sweet, thanks for looking into this! The only suggestion I have is just making `migrate_pause` directly return `tokio::sync::Notify` the caller can choose to wait on. The more generalized, callback based interface here is nice but the only callers we have don't really need it. Did you have another use case in mind?

gjcolombo: Funny you should ask :) I did exactly that in an earlier draft on Friday, but woke up this morning and felt a little squicky about having a non-`async` (and nominally environment-agnostic) routine start returning `tokio` primitives, so I switched it to something that lets the caller pick the notification discipline.

On the other hand, you're right that this code is only called in a couple of specialized places, and they already uphold the "I could easily `await` this" invariant. And it is simpler just to return the `Notify`. (And I'm about to revisit a bunch of this anyway to help deal with #155.) 

I'll change this over presently (I also want to rearrange things a bit so that the `notify` call/callback information lives with the other implicit state transition actions and not in the routine that otherwise just talks to entities).

gjcolombo: Should be fixed (and the notification site tidied up) in 6ab17d5.

luqmana (approved):

gjcolombo:
Thanks for reviewing, @luqmana!

PR #206, opened 2022-09-01
Dropshot-based websocket endpoint & Add Progenitor-based client

lifning:
### Use Dropshot 'channel' for serial console endpoint
This moves the websocket-related boilerplate into the code generated by Dropshot's `#[channel]` annotation macro.
(rel: https://github.com/oxidecomputer/dropshot/pull/403 )

### Generate an API client with Progenitor
This validates and/or generates an OpenAPI schema in a test, as done in omicron (with expectorate).
The generated client is currently opt-in via a feature flag, so as to mitigate breakages while migrating our other codebases from the handwritten client code.
(rel: https://github.com/oxidecomputer/progenitor/pull/183 )


lib/propolis-client/src/lib.rs
```diff
@@ -1,194 +1,116 @@
...
-
-impl Client {
-    pub fn new(address: SocketAddr, log: Logger) -> Client {
-        Client {
-            client: reqwest::Client::new(),
-            log: log.new(o!("propolis_client address" => address.to_string())),
-            address,
+#[cfg(feature = "generated")]
+pub mod generated;
+pub mod handmade;
```
pfmooney, line 13: From a docs standpoint, do we want the modules themselves to be visible, or simply exposed via the `pub use` below?

lifning: Added some more clarifying documentation around it and only made `handmade` `pub` when it's not globally `pub use`'d

Issue #191, opened 2022-08-09
Cleanly stopping an instance doesn't release its bhyve VM


gjcolombo:
This isn't severe for the reasons given in #154 (in production VMs get cleaned up by other means), but it's unintuitive (at least to me), so I'm filing an issue for it.

Repro steps:

1. `ls /dev/vmm`
2. Start a new Propolis server and set its state to running
3. ls `/dev/vmm`
4. Set the server's state to stopped
5. `ls /dev/vmm`

Expected: A new VM appears in step 2 and disappears in step 4.
Observed: The VM persists in step 4.

This is related to #154, though fixing this issue doesn't necessarily fix that one (ensuring that a stopped-and-destroyed instance destroys its bhyve VM helps frameworks do the right thing, but doesn't fix the case where a test framework launches a VM and doesn't explicitly stop it). 

---

I think (but haven't fully confirmed) that the root cause is an extra machine reference taken during instance setup. The Propolis server's global context structure contains a VNC server that holds a shared instance dispatch context that holds a reference to the instance's `Machine`, and AFAICT this reference doesn't get released when the instance stops. But I need to investigate further to make sure this theory is correct.

gjcolombo:
This seems to have regressed somewhat recently--I'm no longer seeing VM devices go away when a VM controller stops, though I note that tearing down the Propolis process itself (via control-C) at least gets rid of them.

This appears to repro at least as far back as fe0c4c579aa98d197571971d5af8f8dab1936806.

gjcolombo:
I can't repro this on a VM with no devices attached and on a VM with a file-backed disk. The VMs I was hitting this with yesterday had Crucible disks. I'll open a new issue for this since this is a different problem than what I recall this bug to have tracked.

PR #208, opened 2022-09-07
Rework Propolis task management and VM state machine

### LINKED ISSUE #170: Failure in the sync phase of LM leaves the source unable to start a new migration
Repro steps:

1. Pick up the changes in #169.
2. Start a source and target with mismatched instance specs (such that migration fails in the sync phase).
3. Try to migrate. This fails as expected with a `Source and destination instance specs incompatible` error message on the target.
4. Launch a new target with an instance spec that matches the source.
5. Try to migrate again.

Expected: The migration succeeds this time.

Observed: The source issues a 400 error (`INFO request completed, error_message_external: migration failed: a migration from the current instance is already in progress, error_message_internal: migration failed: a migration from the current instance is already in progress`) and the migration fails on the destination (`ERRO source instance failed to switch protocols: 400 Bad Request`).

gjcolombo:
This PR substantially reworks some Propolis internals to try to make room to fix some existing issues, sand off some complexity, and (hopefully) speed future development.

A note on the change history: this branch squashes the largest refactoring changes into individual commits, but keeps some individual bits of cleanup work in individual commits that can be removed during review if desired. My intention is to submit these changes (for history-readability purposes) as a pair of commits, one that reworks the main Propolis library and Propolis standalone and a second one that refactors the Propolis server on top of those changes.

# Major changes

These changes rework several fundamental Propolis constructs in ways that affect many consumers across the entire Propolis code base. Here are summaries of these large changes:

- propolis-lib: Remove the centralized dispatcher context and replace it with a library that allows consumers to create, pause, and resume individual groups of tasks. This fundamentally changes how instances pause and halt: instead of asking an instance to quiesce/stop all its dispatcher's tasks, the instance asks individual entities to pause/shut down, and they do so in whatever way works best for the task management style they're using. This helps to separate task management details from the inner workings of the instance state machine (e.g. special state machine affordances to keep migration tasks running while all other dispatcher tasks are paused).
- propolis-lib: Add explicit accessor handles that entities use to access memory and I/O space. Previously, entities that needed to access machine-wide resources did so through their dispatcher contexts. This removes the need for unit tests to synthesize a `DispCtx` to test routines that would previously have required one. The accessor model allows Propolis to implement more complex rules for deciding which machine resources an entity can access and when it can access them, paving the way for future work on, say, an emulated IOMMU.
- propolis-lib: Remove the instance state machine, allowing users of the library to decide how subtle of a state machine they need and what VM states make the most sense for them.
- propolis-lib: Modify the `Entity` trait to allow entities to distinguish between their first request to run in a particular incarnation of an instance and mere requests to resume from a paused state.
- propolis-server: Create a "VM controller" class that implements the server state machine. The controller receives signals from the server's external API and entities within the instance and pauses/resumes/stops the instance's devices as needed in response. Centralizing the state machine this way (hopefully) simplifies the graph of allowed state transitions and the synchronization needed to coordinate, say, an ongoing migration with incoming state change requests from the server API.
- propolis-server: Refactor so that the Dropshot endpoints in `server.rs` verify parameters and deal with conversion to/from HTTP request/response types, but otherwise delegate as much as possible to the VM controller or other server modules.

# Tests

Tested with PHD tests (including live migration variations) and ad hoc testing of migration, guest-directed reboot/shutdown, the serial console, and VNC.

# Related issues

Fixes #170. Provides a clearer path to fixing #155 and #191, but doesn't directly resolve them.


gjcolombo:
# Note to reviewers

The most important files to review in this change are the following:

- bin/propolis-server/src/lib/vm.rs
- bin/propolis-server/src/lib/server.rs
- bin/propolis-server/src/lib/migrate/*
- lib/propolis/src/accessors.rs
- lib/propolis/src/tasks.rs

Most of the rest of the files in this PR have few or no substantial functional changes. The bulk of the deltas outside these files are function signature and call site changes resulting from the removal of the Propolis lib's task dispatcher.

The high-level architectural changes in the key files are as follows:

### No more `DispCtx`

The task dispatcher is broken into two smaller components, one (in `accessors.rs`) that brokers device access to memory (and MSIs), and one (`tasks.rs`) that lets devices create logical tasks and ask them to quiesce, resume, or exit. In the old world, to access machine resources, you needed a `DispCtx`, which you obtained by being created as a task that was registered with a central task registry, which would then emit quiesce/resume/exit signals. Removing it simplifies many function signatures, call sites, and unit tests (this last especially, because many tests can now exercise device code without having to synthesize a fake instance and `DispCtx` first).

The long-term goal is to give devices handles to specific regions of memory/MSI space that the devices have to lock successfully in order to access the relevant resource. The handles are arranged hierarchically such that a given handle or any of its ancestors may forbid access to the resource. This will let us more easily model things like an IOMMU or runtime changes to PCI bus mastering. `accessors.rs` should be read/reviewed with these goals in mind, though in the short run the accessors mostly serve to replace the `DispCtx`-driven access model.

### No instance state machine in the main lib

Previously the library maintained its own instance state machine and let callers control it by setting a goal state for the state machine's worker thread to try to reach. This PR makes the Propolis library much less opinionated about instance states: now, the lib provides primitives to tell devices to start/quiesce/resume/halt, and it's up to the library's consumers to use those primitives to implement their own state machines.

### Request-queue-based state machine in propolis-server

The Propolis server state machine lives in `struct VmController` in `vm.rs`. Like the old library state machine, it drives all state transitions from a single worker thread, but instead of asking consumers to set goal states, its API lets consumers request specific *operations* or notify the controller of specific events, leaving it to the controller to decide how to twiddle the instance's devices and what logical states the VM should move through in response to each signal.

This more or less separates the notions of "the state the instance is in" and "the state of the devices in the instance": the state worker pauses/resumes instances to service requests instead of asking callers to request state transitions that have those effects. This (IMO) simplifies the API for callers and reduces the chance of bugs where, e.g., some in-flight operation requested goal state X expecting transitions A, B, and C, only to be undercut by a goal state set by intervening event Y.

### More direct interaction between live migration tasks and the state worker

The VM controller change significantly affects how live migrations work.

The server's migration tasks previously worked within the confines of the instance's "set-target-state" API. This API had special affordances to allow migration tasks to hold state transitions until their corresponding migration phases were complete. In this PR, the VM controller explicitly delegates to migration tasks when a migration is in progress by setting up channels that allow the migration task to ask the state worker to undertake actions on the tasks' behalf.

There are some gross `spawn_blocking`/`block_on` shenanigans in `migrate/mod.rs` that I needed to add to reckon properly with ownership of the upgraded HTTP connection and with the fact that the migration tasks are async while the state worker is not. I'd be especially grateful for suggestions about how to make this less gross (maybe I should just give in and make the state worker an async task?).

### Rearranged propolis-server

Finally, this change rearranges some propolis-server modules so that the Dropshot server module (`server.rs`) is concerned principally with receiving API requests, translating their parameters (where needed) into types consumed by the VM controller (or other high-level server tasks), and recasting those callees' responses back into HTTP errors.

There is one minor functional change in the server: it now tries much harder to drop the `VmController` and quiesce or shut down related services (serial tasks, VNC, stats collection) in response to a request to stop the VM. This is meant to make it easier to (eventually) deal with #191, though this PR doesn't itself fix that issue. To do this while maintaining existing API behavior (specifically, making `instance_get` report that a VM is `Destroyed` after it's been halted), the server has to maintain a very small amount of extra state to distinguish between "no instance was ever created" and "an instance was created but was since destroyed; here were its properties". I'm generally happy with how this code looks, but it merits mentioning as the one part of `server.rs` that isn't just a refactoring/rearranging of existing code.

bin/propolis-server/src/lib/vm.rs
```diff
@@ -0,0 +1,1383 @@
...
+            }
+        }
+    }
+
+    /// Asks to queue a request to start a destination migration task for this
+    /// VM. The migration will have the supplied `migration_id` and will obtain
+    /// its connection to the source by calling `upgrade_fn` to obtain a future
+    /// that yields the necessary connection.
+    ///
+    /// This routine fails if the VM has already been started or if a previous
```
leftwo, line 653: "if the VM has already been stared". Does this mean migrations can't happen if a VM is started, or does this mean that the routine will fail if a migration has already started?

dancrossnyc: If a destination VM has already started, then it is invalid as a target for a migration. So I'd say that yes, if a VM has started, then a migration into that VM cannot happen.

lib/propolis/src/block/crucible.rs
```diff
@@ -37,16 +37,24 @@ impl CrucibleBackend {
...
+        )
+        .map_err(map_crucible_error_to_io)
     }
 
     fn _create(
         gen: u64,
         request: VolumeConstructionRequest,
         read_only: bool,
         producer_registry: Option<ProducerRegistry>,
+        log: slog::Logger,
```
leftwo, line 57: Ohh, nice.  Maybe I (we, we being Crucible) can pass this log down to crucible and have it send message there instead of stdout like it does now.

lib/propolis/src/block/crucible.rs
```diff
@@ -57,23 +65,28 @@ impl CrucibleBackend {
...
             tokio::task::block_in_place(|| volume.get_block_size())?;
-
         let total_size = tokio::task::block_in_place(|| volume.total_size())?;
+        let sectors = total_size / block_size;
+
+        // TODO: make this tunable?
+        let worker_count = NonZeroUsize::new(8).unwrap();
 
-        be.sectors = total_size / be.block_size;
+        Ok(Arc::new_cyclic(|me| Self {
```
leftwo, line 77: I want to know more about this

leftwo: Ok, after some reading, is this to avoid the possibility of memory leaks if something fails during updates to the new `CrucibleBacked`?

lib/propolis/src/block/crucible.rs
```diff
@@ -109,34 +163,31 @@ impl block::Backend for CrucibleBackend {
...
+        self.driver.attach(dev)
     }
 }
 
 impl Entity for CrucibleBackend {
     fn type_name(&self) -> &'static str {
         "block-crucible"
     }
+    // TODO(#155): Move Crucible activation here. Also, this entity probably
+    // needs its own pause/resume routines akin to those in the other block
```
leftwo, line 186: Hmm.  There are some details here we have to discuss.  Specifically the block_size and volume_size are not known until after the volume is activated.  We may have to update the volume code to arrange the order here.

bin/propolis-server/src/lib/stats.rs
```diff
@@ -9,97 +12,97 @@ use oximeter::{
...
 }
 
-/// Setup Oximeter
-/// This starts a dropshot server, and then registers the PropStatOuter
-/// producer with Oximeter.
-/// Once registered, we return the server to the caller.
-/// By returning the server to the caller, we allow the ProducerRegister
-/// to be cloned and passed to any library that wishes to record metrics.
-pub async fn prop_oximeter(
+/// Launches and returns an Oximeter metrics server.
```
leftwo, line 89: Thanks for fixing all my low quality comments.

lib/propolis/src/block/crucible.rs
leftwo: https://github.com/oxidecomputer/crucible/issues/440

bin/propolis-server/src/lib/vm.rs
leftwo: Ah, so it's "if the destination VM has already been started".  I was not sure which VM we were talking about.

gjcolombo: In a0b434d I changed the comment around a bit (s/been started/begun to run) to clarify exactly which state transition this routine wants to check for.

bin/propolis-server/src/lib/vcpu_tasks.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn suspend_reset_event(&self, vcpu_id: i32);
+    fn suspend_triple_fault_event(&self, vcpu_id: i32);
+    fn unhandled_vm_exit(&self, vcpu_id: i32, exit: VmExitKind);
+    fn io_error_event(&self, vcpu_id: i32, error: std::io::Error);
+}
+
+impl VcpuTasks {
+    pub fn new(
+        instance: propolis::instance::InstanceGuard,
+        event_handler: Arc<dyn VcpuEventHandler>,
```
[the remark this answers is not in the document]
gjcolombo: Fixed in 87968b0. (For posterity, the original reason for this was largely symmetry with what we do for chipset events, which symmetry we will need if we ever move the vCPU task loops down into propolis-lib proper, but since we're not doing that in this PR we can just use the concrete type for now and reimplement this later if need be.)

bin/propolis-server/src/lib/vcpu_tasks.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub fn pause_all(&mut self) {
+        for task in self.tasks.iter_mut().map(|t| &mut t.0) {
+            task.hold().unwrap();
+        }
+
+        let task_futures: Vec<_> =
+            self.tasks.iter_mut().map(|t| t.0.held()).collect();
+        let results = self
+            .rt
+            .block_on(async { futures::future::join_all(task_futures).await });
```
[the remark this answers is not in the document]
gjcolombo: Fixed in 56c4f7e.

bin/propolis-server/src/lib/vcpu_tasks.rs
```diff
@@ -0,0 +1,212 @@
...
+                .map_err(|e| VcpuTaskError::BackingThreadSpawnFailed(e))?;
+            tasks.push((ctrl, thread));
+        }
+
+        Ok(Self { tasks, generation, rt: runtime })
+    }
+
+    pub fn pause_all(&mut self) {
+        for task in self.tasks.iter_mut().map(|t| &mut t.0) {
+            task.hold().unwrap();
```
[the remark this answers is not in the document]
gjcolombo: Filed #209 to track this.

bin/propolis-server/src/lib/vcpu_tasks.rs
```diff
@@ -0,0 +1,212 @@
...
+                        }
+                    },
+                    VmError::Suspended(suspend) => {
+                        // This vCPU will not successfully re-enter the guest
+                        // until the state worker does something about the
+                        // suspend condition, so hold the task until it does so.
+                        //
+                        // N.B. This usage assumes that it is safe for the VM
+                        //      controller to ask the task to hold again.
+                        task.force_hold();
```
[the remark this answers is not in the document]
gjcolombo: Oops. You can tell which part of the code I didn't exercise properly :) Should be fixed in 28ec23d.

lib/propolis/src/block/crucible.rs
gjcolombo: I agree, this'll need some further untangling. Currently, the storage devices (NVMe/virtio-block) expect to get this information when they're created, so if we want to defer obtaining this information until after activation, we'll also need to change the device init contract accordingly.

I think this is possible in principle (skimming the code suggests that the devices only care once they're actually handling guest requests, and those will only arrive once `start` has been called, since the VM controller only starts vCPUs after starting entities), but I'm not sure how hairy the refactored code would be. We'd probably need to try it and see.

lib/propolis/src/block/crucible.rs
gjcolombo: Peanut gallery comment: my read is that this lets us get rid of a heap allocation (the `Box<BackendProcessFn>` in `struct Driver`) by making the processing callback part of the `Backend` trait that this type already implements. The catch is that this creates a reference cycle (`CrucibleBackend -> Driver -> CrucibleBackend`) where there was none before (we had `CrucibleBackend -> BlockIO` and `CrucibleBackend -> Driver -> BackendProcessFn -> BlockIO`), so we need a weak reference to break it.

I think this simplification applies even if something later fails in the `CrucibleBackend` (since this cycle is set up in `CrucibleBackend::create` it applies no matter whether a subsequent call to something like `attach` succeeds or fails).

pfmooney: Yeah, I didn't want creation of the driver to be a separate step.

lib/propolis/src/block/crucible.rs
leftwo: I think taking our best guess on the Crucible side (as discussed in [issue 440](https://github.com/oxidecomputer/crucible/issues/440) ) is probably the easiest path forward here.

lib/propolis/src/tasks.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+}
+impl Future for Held<'_> {
+    type Output = ();
+
+    fn poll(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Self::Output> {
+        let mut guard = self.hdl.0.ctrl.lock().unwrap();
```
luqmana, line 202: We should not be blocking in any `poll` method:

From the [std docs](https://doc.rust-lang.org/stable/std/future/trait.Future.html#runtime-characteristics):
> An implementation of poll should strive to return quickly, and should not block.

In practise this might be fine but deadlocked tasks are a pain to debug.

lib/propolis/src/tasks.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+}
+impl Future for GetEvent<'_> {
+    type Output = Event;
+
+    fn poll(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Self::Output> {
+        let mut guard = self.hdl.0.ctrl.lock().unwrap();
```
luqmana, line 244: Similarly here

bin/propolis-server/src/lib/vcpu_tasks.rs
```diff
@@ -0,0 +1,201 @@
...
+                propolis::tasks::TaskHdl::new_held(Some(vcpu.barrier_fn()));
+            let task_log = log.new(slog::o!("vcpu" => vcpu.id));
+            let task_event_handler = event_handler.clone();
+            let task_gen = generation.clone();
+            let thread = std::thread::Builder::new()
+                .name(format!("vcpu-{}", vcpu.id))
+                .spawn(move || {
+                    Self::vcpu_loop(
+                        vcpu.as_ref(),
+                        &task,
```
luqmana, line 55: Can just pass this by value here.

luqmana (commented):
thanks for taking this on patrick and greg! looks mostly good. i did have one concern about possibly blocking in poll

lib/propolis/src/accessors.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+}
+
+struct NodeEntry<T> {
+    id: AccId,
+    tree: TreeBackref<T>,
+}
+struct NodeLocal<T> {
+    resource: Option<Arc<T>>,
+    // TODO: store local enable/disable state here for propagation
```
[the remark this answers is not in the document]
pfmooney: As noted in that little TODO, I do intend to store more state on the local node as this abstraction is fleshed out more.

lib/propolis/src/tasks.rs
pfmooney: I'm not especially concerned about this.  No significantly lengthy work is being done when the lock is being held, but using a traditional Mutex makes implementation of the synchronous side of this interface considerably easier.  I'm certainly not against heeding the guidance in `std`, but it is SHOULD rather than MUST here.  There are certainly other ways that async consumers may stumble over blocking operations (really anything touching the kernel vmm may block on in-kernel mutex during an ioctl)

jordanhendricks (commented):
excited for these changes! it's nice to see the dispatcher-related items ripped out of so many function signatures, and i like the new accessors interface.

just a few minor comments

bin/propolis-server/src/lib/vcpu_tasks.rs
gjcolombo: Good call. Should be fixed in 7ce5253.

bin/propolis-server/src/lib/server.rs
```diff
@@ -1,271 +1,309 @@
...
-
-    /// The currently active live migration task, if a migration is in progress.
-    pub migrate_task: Mutex<Option<migrate::MigrateTask>>,
-
-    /// The VNC server hosted in this process. Note that this server exists even
-    /// when no instance has been created yet (it serves a white screen);
-    /// creating an instance with `instance_ensure` hooks the inner
-    /// `PropolisVncServer` up to the new instance's framebuffer.
-    vnc_server: Arc<Mutex<VncServer<PropolisVncServer>>>,
+pub struct ServiceProviders {
```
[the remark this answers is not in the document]
gjcolombo: Good catch, thanks! Fixed in 551bc83.

lib/propolis/src/accessors.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    pub fn access(&self) -> Option<Guard<T>> {
+        self.0.guard()
+    }
+}
+
+pub type MemAccessor = Accessor<MemCtx>;
+
+// Keep the rest of VmmHdl hidden for the MSI accessor
+pub struct MsiAccessor(Accessor<VmmHdl>);
```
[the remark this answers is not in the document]
pfmooney: After chatting, it sounds like leaving them here will be fine for now.

lib/propolis/src/tasks.rs
```diff
@@ -0,0 +1,373 @@
...
+    fn request_exit(&self, mut guard: MutexGuard<Control>) {
+        guard.should_exit = true;
+        guard.wake_task();
+        self.cv.notify_all();
+        let _guard = self.notify_task(guard);
+    }
+
+    /// Call notifier function (if one exists) for task.
+    ///
+    /// This notification may access synchronously access resources which would
```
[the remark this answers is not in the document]
pfmooney: Done

lib/propolis/src/tasks.rs
pfmooney: I've added a comment to the mutex highlighting the care which must be taken

bin/propolis-server/src/lib/server.rs
```diff
@@ -1,271 +1,309 @@
...
+impl VmControllerState {
+    pub fn controller_mut(&mut self) -> Option<&mut Arc<VmController>> {
+        match self {
+            VmControllerState::NotCreated => None,
+            VmControllerState::Created(c) => Some(c),
+            VmControllerState::Destroyed(_) => None,
+        }
+    }
+
+    pub fn take_vm(&mut self) -> Option<Arc<VmController>> {
```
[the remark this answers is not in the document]
gjcolombo: Fixed in 551bc83 by renaming and adding some comments--let me know if I should take another swing at this. (I was probably too indexed on the fact that this replaced a call to `Option::as_mut`; hopefully the new name and docs clarify what's going on here.)

bin/propolis-server/src/lib/server.rs
```diff
@@ -276,261 +314,105 @@ async fn instance_ensure(
...
-            Ok(())
-        })
-        .map_err(|err| {
-            HttpError::for_internal_error(format!(
-                "Failed to initialize machine: {}",
-                err
-            ))
-        })?;
+        // Give the Propolis VNC adapter a back pointer to the VNC server to
+        // allow it to update pixel formats.
```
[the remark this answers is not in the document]
gjcolombo: Makes sense, thanks! Fixed in 551bc83.

bin/propolis-server/src/lib/server.rs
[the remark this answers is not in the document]
jordanhendricks: looks good! thanks.

lib/propolis/src/tasks.rs
```diff
@@ -0,0 +1,378 @@
...
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub enum Event {
+    Hold,
+    Exit,
+}
+
+pub struct TaskHdl(Arc<Inner>);
```
luqmana, line 120: Add some docs on `TaskHdl` vs `TaskCtrl` it's not clear at first glance the difference between the two especially as they have methods with the same names.

The next part shouldn't hold up the PR but is more for possible future changes: Is there any reason we can't just rename `Inner` to `Task` and just let the consumers use `Arc<Task>` or `Weak<Task>` as appropriate? To keep the same ergonomics we could keep all the current `TaskHdl` methods as-is and just change the self type they take to require the `Arc`. Unfortunately not so straightforward for `TaskCtrl` as `Weak` can't be used as a receiver type today. But either the upgrade could be pushed to the caller or just use static methods that take a `Weak` ref.

lib/propolis/src/util/sys.rs
```diff
@@ -7,8 +7,8 @@ use std::io::ErrorKind;
 // Deal with libc bits which vary enough between OSes to make cargo-check a pain
 
 #[cfg(target_os = "illumos")]
-pub fn ioctl<T>(fd: RawFd, cmd: i32, data: *mut T) -> Result<i32> {
-    let res = unsafe { libc::ioctl(fd, cmd, data) };
+pub unsafe fn ioctl<T>(fd: RawFd, cmd: i32, data: *mut T) -> Result<i32> {
```
luqmana, line 10: Update the not-illumos variant to match

lib/propolis/src/tasks.rs
pfmooney: > Add some docs on TaskHdl vs TaskCtrl it's not clear at first glance the difference between the two especially as they have methods with the same names.

Will do.

>  Is there any reason we can't just rename Inner to Task and just let the consumers use Arc<Task> or Weak<Task> as appropriate?

One aspect of the current implementation (and it wasn't always this way) is that there is only one "copy" each of the `TaskHdl` and `TaskCtrl`.  This simplifies storage of the Waker for the futures which we can initiate from either side.  It's part of the reason why the receivers for those future methods are `&mut` - so there can only be one at a time.  If we move to a naked `Arc`/`Weak`, then we lose that convenience.

lib/propolis/src/util/sys.rs
pfmooney: This is already done in the `psl` branch where I'm staging the squashed changes from this PR.

lib/propolis/src/tasks.rs
pfmooney: > Add some docs on TaskHdl vs TaskCtrl

I hope it's a little more clear with the added docs?

luqmana: Yes, thank you!

bin/propolis-server/src/lib/migrate/mod.rs
```diff
@@ -378,106 +284,83 @@ pub(crate) async fn dest_initiate(
...
-    let src_migrate_url =
-        format!("http://{}/instance/migrate/start", migrate_info.src_addr);
-    info!(log, "Begin migration"; "src_migrate_url" => &src_migrate_url);
-
-    let body = Body::from(
-        serde_json::to_string(&api::InstanceMigrateStartRequest {
-            migration_id,
+    let runtime_hdl = tokio::runtime::Handle::current();
+    tokio::runtime::Handle::current()
+        .spawn_blocking(move || -> Result<(), MigrateError> {
```
luqmana, line 289: Why does the dest side use `spawn_blocking` but not the source side?

luqmana (commented):
ok, a couple more comments. nothing too big. besides that LGTM

lib/propolis/src/vcpu.rs
```diff
@@ -3,29 +3,33 @@
...
-/// callback.
-pub type VcpuRunFunc = fn(&Vcpu, &mut SyncCtx);
-
 /// A handle to a virtual CPU.
 pub struct Vcpu {
     hdl: Arc<VmmHdl>,
-    id: i32,
+    pub id: i32,
+    pub bus_mmio: Arc<MmioBus>,
+    pub bus_pio: Arc<PioBus>,
```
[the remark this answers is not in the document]
pfmooney: The vCPU needs access to the buses to fulfill PIO/MMIO exits, and it seemed easier than trying to get a ref to the whole `Machine` in here.

lib/propolis/src/chardev/sock.rs
```diff
@@ -63,17 +61,15 @@ impl UDSock {
...
         self.sink_buf.attach(sink.as_ref());
         self.source_buf.attach(source.as_ref());
 
         let this = Arc::clone(self);
-        let actx = disp.async_ctx();
-        let task = tokio::spawn(async move {
-            let _ = this.run(sink, source, &actx).await;
+        let _task = tokio::spawn(async move {
+            let _ = this.run(sink, source).await;
+            todo!("get async task hdl");
```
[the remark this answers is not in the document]
pfmooney: Yeah, the sock stuff is mostly working, but there are still some rough edges to file down.

lib/propolis/src/block/in_memory.rs
```diff
@@ -71,45 +113,38 @@ impl block::Backend for InMemoryBackend {
...
 
 impl Entity for InMemoryBackend {
     fn type_name(&self) -> &'static str {
-        "in-memory"
+        "block-in-memory"
+    }
+    fn start(&self) {
+        self.driver.start();
     }
 }
```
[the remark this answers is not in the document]
pfmooney: Sounds like a good idea!

bin/propolis-server/src/lib/vcpu_tasks.rs
```diff
@@ -0,0 +1,201 @@
...
+
+impl VcpuTasks {
+    pub(crate) fn new(
+        instance: propolis::instance::InstanceGuard,
+        event_handler: Arc<super::vm::WorkerState>,
+        log: slog::Logger,
+    ) -> Result<Self, VcpuTaskError> {
+        let generation = Arc::new(AtomicUsize::new(0));
+        let mut tasks = Vec::new();
+        for vcpu in instance.machine().vcpus.iter().map(Arc::clone) {
```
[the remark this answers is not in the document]
pfmooney: Clippy was clean.  I prefer being a little bit more explicit about it being an Arc reference clone, but would cede that if the lint becomes more strict, or you think it'd be clearer that way

bin/propolis-server/src/lib/migrate/mod.rs
gjcolombo: The big difference between the two sides is that the destination has to kick off and block on a new HTTP request within this closure, but the source side doesn't. `spawn_blocking` is this code path's way of contending with the async-ness of that request.

`request_migration_into` currently needs to issue the request and wait for the response synchronously to ensure that the controller state doesn't change while the connection is being established. The trouble here is that the "get an upgraded connection future" closure can't just use `block_on` by itself, because while `request_migration_into` isn't `async`, its caller is, so trying to call `block_on` panics at runtime with a "starting a runtime from within a task" error. Using `spawn_blocking` to create a top-level extra-runtime work item (that this function can then `await`) gets around that problem.

I experimented a bit with making `request_migration_into` async, such that it can just `await` the upgrade future, but that doesn't work (at least not the way I've tried it), because the controller lock is held for writing at the point where `request_migration_into` wants to make the request; trying to `await` that request while holding that lock causes a cascade of `future returned by $FUNCTION is not Send` errors.

With all that said, it might be possible to pry this apart by making `request_migration_into`'s `upgrade_fn` into a future that outputs a `Result<OnUpgrade, MigrateError>` and stashing that in the `ExternalRequest::MigrateAsTarget` message. That also gets around the recursive-runtime problem by pushing everything to the state worker, which is already a free thread. I can give this a whirl but might be inclined to file an issue and try this in a follow-on change.

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -236,9 +238,9 @@ impl VirtQueue {
     pub fn pop_avail(&self, chain: &mut Chain, mem: &MemCtx) -> Option<u32> {
         assert!(chain.idx.is_none());
         let mut avail = self.avail.lock().unwrap();
-        let id = avail.read_next_avail(self.size, mem)?;
+        let id = avail.read_next_avail(self.size, &mem)?;
```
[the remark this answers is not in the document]
pfmooney: Fixed

lib/propolis/src/block/mod.rs
```diff
@@ -162,21 +173,27 @@ pub struct DeviceInfo {
 /// API to access a virtualized block device.
 pub trait Device: Send + Sync + 'static {
     /// Retreive the next request (if any)
-    fn next(&self, ctx: &DispCtx) -> Option<Request>;
+    fn next(&self) -> Option<Request>;
+
+    /// Complete processing of result
+    fn complete(&self, op: Operation, res: Result, payload: Box<dyn Any>);
```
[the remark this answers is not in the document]
pfmooney: Done (here, and in the consumers)

lib/propolis/src/block/mod.rs
```diff
@@ -346,121 +382,199 @@ impl Notifier {
...
+    /// # Panics
+    ///
+    /// This routine assumes that on entry, this notifier was previously asked
+    /// to pause and that the pause requestor waited for the notifier to pause
+    /// fully before resuming. To that end, this routine asserts that
+    ///
+    /// - this object has an active notification channel, and
+    /// - this object is not tracking any outstanding requests.
+    pub fn resume(&self) {
+        assert_eq!(self.outstanding.count.load(Ordering::Relaxed), 0);
```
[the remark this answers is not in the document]
pfmooney: Cleaned up

lib/propolis/src/block/mod.rs
```diff
@@ -346,121 +382,199 @@ impl Notifier {
...
+
+        // Spawn (held) worker tasks
+        let mut tasks = self.inner.task_ctrl.lock().unwrap();
+        for (i, ctrl_slot) in tasks.workers.iter_mut().enumerate() {
+            let (mut task, ctrl) = TaskHdl::new_held(Some(self.worker_waker()));
+
+            let worker_self = Arc::clone(&self.inner);
+            let worker_be = be.clone();
+            let _join = std::thread::Builder::new()
+                .name(format!("{} {i}", &self.name))
```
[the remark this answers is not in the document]
pfmooney: Done

lib/propolis/src/chardev/pollers.rs
```diff
@@ -563,236 +552,162 @@ impl Params {
...
-
     use super::*;
     use crate::chardev::*;
-    use crate::dispatch::DispCtx;
-    use crate::instance::{Instance, ReqState};
 
     use futures::FutureExt;
-    use tokio::runtime::Handle;
-
-    async fn async_inst_test<T, F>(task: T) -> Result<()>
```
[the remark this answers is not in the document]
pfmooney: Samesies

lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -620,8 +627,14 @@ impl PciNvme {
 
         let cur = state.ctrl.cc;
         if new.enabled() && !cur.enabled() {
+            let mem = self.mem_access();
+            if mem.is_none() {
+                return Err(NvmeError::MemoryInaccessible);
+            }
+            let mem = mem.unwrap();
```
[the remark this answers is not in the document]
pfmooney: Done

bin/propolis-server/src/lib/migrate/mod.rs
pfmooney: If it's not going to immediately blow up in our faces, a follow-up would be preferred.  I suspect we'll get a better feel for how we want to continue shaping these bits as we plug in more pieces.

bin/propolis-server/src/lib/vcpu_tasks.rs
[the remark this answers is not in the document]
luqmana: ah, don't think it'll complain. I kinda missed it was Arc::clone explicitly and not Clone::clone

luqmana (approved):

Issue #194, opened 2022-08-11
PHD: simplify the artifact TOML path & create a single canonical artifact list


gjcolombo:
@pfmooney and I discussed some enhancements to the artifact scheme that we think are righteous:

- Remove the `local_root` field from the artifact TOML and make it a PHD runner command-line parameter instead
- Merge the artifact TOMLs in the repo into a single canonical "this is what we test" artifact list (currently only one file, `sample-artifacts.toml`, is checked in, but after #193 lands there will be another one for Buildomat to use)

This way, there's a single checked-in canonical list of supported artifacts, but it expresses no opinion about where exactly the user should store them.

Advantages of doing this:

- New users of PHD should have a ready-to-use list of artifacts that's known to work with the framework's guest OS adapters. (The framework hard-codes various command prompt styles, username/password combinations, etc. on a per-guest-OS basis and assumes the user will guarantee that images in the artifact list behave as expected; having a list of known-good artifacts will make it harder to err here.) This list can be kept up-to-date as new images are qualified/produced/added to CI.
- It's more in keeping with the runner's philosophy about system configuration: TOMLs, temporary directories, and the like are generally customized on the command line and not hard-coded (in this case they're implicitly specified, but for most other things PHD prefers explicit parameters).

PR #216, opened 2022-09-23
Update recommended guest firmware image 

### LINKED ISSUE #214: Windows VMs often have an unreliable display
When interacting with a Windows VM over VNC, sometimes the screen will go black for a second or two and return to where the screen was before. After a few iterations of this, the VM will often kick the logged in user out back to the lock screen.

The lock screen has generally been stable for a long time in my experience -- specifically, I have never seen this behavior while the VM is locked and on this screen:
ksnip_20220922-093716

But once I'm into the password prompt screen:
ksnip_20220922-094609

sometimes the screen may start to become unresponsive, with a black screen shown as such:
ksnip_20220922-094210

after enough times of this happening, I've observed the VM sometimes kicks me back to the lock screen.

gjcolombo:
Update the recommended guest firmware image to a build that (among other things) indicates to the guest that the ACPI PM timer is 32 bits wide. Also update the package manifest to use this bootrom.

Collapse PHD's two artifact files (one 'sample' in its root directory and one for buildomat) into a single 'has been tested' artifact list, as suggested in #194. (There's a little more work to do to fully close that issue, but this helps.)

Tested with PHD tests; ran `cargo run -p propolis-package` and verified that the output directory had the expected bootrom and contained the expected files.

Closes #214.


README.md
```diff
@@ -143,10 +143,13 @@ are some options to get up-and-running quickly:
...
+The recommended prebuilt bootrom is available
+[here](https://oxide-omicron-build.s3.amazonaws.com/OVMF_CODE_20220922.fd).
+Rename this file to `OVMF_CODE.fd` to use it with the example configuration
+below.
+
+You can also [build your own OVMF image](https://wiki.ubuntu.com/UEFI/OVMF) or
+obtain one from a Linux installation by running:
 
 ```bash
 $ sudo apt-get install ovmf && dpkg -L ovmf | grep OVMF_CODE.fd
```
pfmooney, line 155: Considering the failure modes, I don't think we should be suggesting that folks use anything but the blessed ROMs, at least in the top-level README

pfmooney (commented):
Thanks for updating this stuff

README.md
gjcolombo: Good call. Pushed 523660f to tighten this language up.

PR #218, opened 2022-09-24
Support updated VMM destruction behavior

pfmooney:
An ioctl was added in illumos #14963 which allows a vmm instance to be destroyed through its own file descriptor, rather than needing to re-open vmmctl and request its destruction by name.  Additionally, that by-name destroy ioctl will now wait for the vmm resource to be closed, which would otherwise cause a deadlock if propolis tries to do the destroy with the VmmHdl holding the fd open.


gjcolombo (approved):

pfmooney:
Merged in b9da532d9057bcb2975c1a5211a93a1c071407de (which had a small comment tweak from this PR)

PR #219, opened 2022-09-28
Rework Crucible backend for async upstairs API

pfmooney:
This is a proposed change to consume the new Crucible upstairs API (oxidecomputer/crucible#449) which @jmpesp has been working on.

### DIFF
--- a/lib/propolis/src/block/crucible.rs
+++ b/lib/propolis/src/block/crucible.rs
@@ -1,30 +1,22 @@
 //! Implement a virtual block device backed by Crucible
 
-use std::io::{Error, ErrorKind, Result};
-use std::num::NonZeroUsize;
-use std::sync::{Arc, Weak};
+use std::io;
+use std::sync::Arc;
 
 use super::DeviceInfo;
 use crate::block;
 use crate::inventory::Entity;
-use crate::vmm::{MemCtx, SubMapping};
+use crate::vmm::MemCtx;
 
-use crucible::{
-    crucible_bail, BlockIO, Buffer, CrucibleError, SnapshotDetails, Volume,
-};
+use crucible::{BlockIO, Buffer, CrucibleError, SnapshotDetails, Volume};
 use crucible_client_types::VolumeConstructionRequest;
 use oximeter::types::ProducerRegistry;
+use thiserror::Error;
 use uuid::Uuid;
-
-/// Helper function, because Rust couldn't derive the types
-fn map_crucible_error_to_io(x: CrucibleError) -> std::io::Error {
-    x.into()
-}
 
 pub struct CrucibleBackend {
     block_io: Arc<dyn BlockIO + Send + Sync>,
-    driver: block::Driver,
-    log: slog::Logger,
+    scheduler: block::Scheduler,
 
     read_only: bool,
     block_size: u64,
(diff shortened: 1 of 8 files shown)


lib/propolis/src/block/mod.rs
```diff
@@ -58,7 +58,7 @@ pub struct Request {
     regions: Vec<GuestRegion>,
 
     /// Block device specific completion payload for this I/O request
-    payload: Option<Box<dyn Any + Send>>,
+    payload: Option<Box<dyn Any + Send + Sync>>,
```
luqmana, line 61:
```suggestion
    payload: Option<Box<BlockPayload>>,
```

lib/propolis/src/block/mod.rs
```diff
@@ -650,3 +650,260 @@ impl DriverInner {
...
+                avail = self.worker_idle.acquire() => {
+                    // We found an idle thread!
+                    avail.unwrap()
+                }
+            };
+
+            tokio::select! {
+                _event = task.get_event() => {
+                    // Take another lap if an event arrives while we are waiting
+                    // for a request to schedule to the worker
```
luqmana, line 889: Worth adding a note here that we're not handling the event explicitly here is fine since `get_event` is essentially level-triggered?

lib/propolis/src/block/mod.rs
```diff
@@ -650,3 +650,260 @@ impl DriverInner {
...
+            // Don't busy-loop on the device but just wait for it to wake us
+            // when the next request is available
+            self.wake.notified().await;
+        }
+    }
+
+    fn complete_req(&self, req: Request, res: Result) {
+        let bdev = self.bdev.get().expect("attached bdev");
+        req.complete(res, bdev.as_ref());
+        match self.active_req_count.fetch_sub(1, Ordering::SeqCst) {
```
luqmana, line 848: This (and the other use of `SeqCst` in `schedule_requests`) can be `AcqRel` right? `active_req_count` is the only atomic in question here so there's no other location we're trying to synchronize with such as to require the consistent total ordering `SeqCst` guarantees.

lib/propolis/src/block/crucible.rs
```diff
@@ -163,18 +96,44 @@ impl block::Backend for CrucibleBackend {
         }
     }
 
-    fn process(&self, req: &block::Request, mem: &MemCtx) -> block::Result {
-        match self.process_request(req, mem) {
-            Ok(_) => block::Result::Success,
-            Err(e) => {
-                slog::info!(self.log, "block IO error {:?}", req.op; "error" => e);
```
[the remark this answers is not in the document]
pfmooney: For now, probably leave it on the crucible side.  I think logging in general is something we'll want to revisit.

lib/propolis/src/block/mod.rs
pfmooney: Done

lib/propolis/src/block/mod.rs
pfmooney: Updated the comment with more detail on that subject

lib/propolis/src/block/mod.rs
pfmooney: Done

lib/propolis/src/block/crucible.rs
```diff
@@ -186,70 +145,94 @@ impl Entity for CrucibleBackend {
...
+    CopyError(usize, usize),
 
-    if nread as usize != len {
-        crucible_bail!(IoError, "nread != len! {} vs {}", nread, len);
-    }
+    #[error("IO Error")]
+    Io(#[from] io::Error),
 
-    Ok(())
+    #[error("Crucible Error")]
```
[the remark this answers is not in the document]
pfmooney: Done

lib/propolis/src/block/crucible.rs
```diff
@@ -37,51 +29,34 @@ impl CrucibleBackend {
...
+    ) -> Result<Arc<Self>, crucible::CrucibleError> {
+        let volume = Volume::construct(request, producer_registry).await?;
 
-        volume.activate(gen)?;
+        volume.activate(gen).await?;
+
+        let block_size = volume.get_block_size().await?;
+        let total_size = volume.total_size().await?;
 
         // After active negotiation, set sizes
```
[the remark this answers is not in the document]
pfmooney: Done

luqmana (approved):

PR #220, opened 2022-09-28
Split instance specs into 'device' and 'backend' modules

gjcolombo:
This is the first in a series of changes that simplifies instance specs and their supporting code. The goal is to make the migration preamble contain only the portions of an instance spec that are directly guest-visible (the "device" portions discussed in RFD 283 section 2.2.1), which reduces the effort needed to keep foreign types and large chunks of data out of the "backend" portion of the spec.

This PR starts this process by breaking `propolis-client/src/instance_spec.rs` into a module with `device` and `backend` submodules, defining separate `DeviceSpec` and `BackendSpec` types, and composing `struct InstanceSpec` from those types. Most of the line count in this PR is movement from `instance_spec.rs` into these files.

The main thing requiring attention is the new doc comment in `instance_spec/mod.rs`, which talks about the new device/backend distinction in a little more detail. There are otherwise no significant functional changes in this PR.


luqmana (approved):

gjcolombo:
You beat me to putting up the stacked PR :) Thanks @luqmana!

PR #221, opened 2022-09-30
Slim down the live migration preamble

gjcolombo:
This PR includes the following work:

1. Change the migration preamble to transmit a VM's device spec and the names of its backends, but not the backend types or their configuration.
2. Use this change to remove extra code to serialize and re-deserialize Crucible volume construction requests during instance setup.
3. Refactor instance spec error types to distinguish more clearly between "collection-level" errors (wrong number of items in a collection or a naming mismatch) and "element-level" errors (two devices are not compatible with each other).
4. Update doc comments in the device/backend modules to explain more about their versioning strategies and what types can be used in each portion of the spec.

The most important things to review in this change are the doc comments in (4) and the overall design direction implemented in (1). RFD 283 describes some of that direction, but below are some additional notes on what changes (1) and (2) do and why.

This change has one significant drawback: the migration preamble can no longer check to make sure that two backends with the same name also have the same kind. That is: suppose that there's a "StorageFoo" backend on both the source and target, but the source thinks it's a File backend and the target thinks it's an InMemory backend. Prior to this change, these instances would fail their migration compatibility check; now they will pass it. IMO, the benefits described below outweigh this cost, especially because the sync protocol gives backends other ways to check their compatibility with each other. If I have this tradeoff wrong, we might be able to bring this check back somewhat by writing separate `XyzBackendKind` enums that encode backend type information but no payloads and using those enums in the preamble.

---

Some additional motivation: Before this change, instance specs were used in two ways:

1. `instance_ensure` would convert its parameters into a spec and pass them to the VM controller to tell it how to initialize an instance.
2. The LM preamble includes a source VM's instance spec; the target uses its spec's `is_migration_compatible` method to decide if it can accept a migration from the source.

Usage (1) requires the spec to contain all the information needed to configure backends, even when this isn't relevant to the compatibility check in (2). This is important because that backend data can be very large (e.g. it might specify the entire contents of an in-memory cloud-init ISO) and may depend on external types whose contents can change in a way that might break data structure versioning. Declining to transmit this data (and to make migration targets deserialize it) makes the preamble less expensive to send and allows backend specs to contain types external to the instance spec crate.

This change makes use of these new properties by removing a bunch of extra logic that took incoming `VolumeConstructionRequest`s, serialized them to JSON to get them into an instance spec, and then immediately deserialized them to pass them back to Crucible. Exposing an `instance_ensure` API that takes an instance spec would also have simplified this, but it would have pushed the serialization work to Nexus. Now both sides can just use the `VolumeConstructionRequest` type natively (provided they agree on what it is).

### DIFF
--- a/lib/propolis-client/src/instance_spec/devices.rs
+++ b/lib/propolis-client/src/instance_spec/devices.rs
@@ -1,9 +1,32 @@
 //! The "device" portions of an instance spec: the parts of the spec that define
 //! VM properties that are visible to guest software.
+//!
+//! # Foreign types & versioning
+//!
+//! Propolis expects to be able to send [`DeviceSpec`] structures between
+//! instances to ensure that they describe the same guest-visible components.
+//! (This is part of the live migration protocol described in RFD 71.) This
+//! applies to Propolis version changes in both directions:
+//!
+//! - Propolis upgrade: Newer versions of Propolis must be able to deserialize
+//!   and interpret device specs constructed by earlier versions.
+//! - Propolis downgrade: Older versions of Propolis must be able to deserialize
+//!   and interpret device specs constructed by newer versions, even if the
+//!   newer versions contain fields not known to the older versions. This can be
+//!   done by explicitly versioning the device spec structure using enum
+//!   variants or by using optional/default-able fields that can be skipped with
+//!   the `skip_serializing_if` attribute.
+//!
+//! Because types defined outside this crate can change without heeding these
+//! requirements, device specs should use only types defined in the crate or
+//! built-in Rust types.
 
 use std::collections::BTreeMap;
 
-use super::{MigrationCompatible, PciPath, SpecKey, SpecMismatchDetails};
+use super::{
+    ElementCompatibilityError, MigrationCollection,
+    MigrationCompatibilityError, MigrationElement, PciPath, SpecKey,
+};
 use serde::{Deserialize, Serialize};
 
 //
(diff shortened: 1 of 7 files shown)


lib/propolis-client/src/instance_spec/mod.rs
```diff
@@ -148,40 +140,66 @@ pub enum SpecMismatchDetails {
...
+    CollectionMismatch(String, CollectionCompatibilityError),
+
+    #[error("Spec element {0} not compatible: {1}")]
+    ElementMismatch(String, ElementCompatibilityError),
 }
 
-/// Routines used to check whether two components are migration-compatible.
-trait MigrationCompatible {
+/// Implementors of this trait are individual devices or VMM components who can
+/// describe inconsistencies using a [`SpecElementMismatchDetails`] enumerant.
```
luqmana, line 170: nit: variant would be the rust term

lib/propolis-client/src/instance_spec/devices.rs
```diff
@@ -257,26 +282,46 @@ impl DeviceSpec {
     pub fn is_migration_compatible(
         &self,
         other: &Self,
-    ) -> Result<(), (String, SpecMismatchDetails)> {
-        self.board
-            .is_migration_compatible(&other.board)
-            .map_err(|e| ("board".to_string(), e))?;
+    ) -> Result<(), MigrationCompatibilityError> {
```
luqmana, line 285: It's a little unfortunate that we have 3 methods with this name with `&Self, &Self` as the args but differ in the return type. Would it be worth slightly renaming them? Or at the least I think the top-level ones could have slightly different ones?

luqmana (commented):
Thanks Greg! Love not needing to (de-)serialize the backend specs multiple times. I left some comments about naming. I did have some comments about some naming but not super big deal

lib/propolis-client/src/instance_spec/mod.rs
gjcolombo: Done in 783b37b.

lib/propolis-client/src/instance_spec/devices.rs
gjcolombo: 59ca882 implements some more descriptive names. I was hesitant to do this at first because having an `impl FooBarBaz` with routine names that included `foo_bar_baz` seemed a little strange, but I rather like how `Preamble::is_migration_compatible` looks now. Thanks for the nudge on this one!

luqmana (approved):

PR #224, opened 2022-10-07
Allow in-memory disks' initial contents to be specified in a StorageBackendSpec

gjcolombo:
#221 makes it much more acceptable for large chunks of data to live in an instance spec's `BackendSpec` members. Now that that has landed, move in-memory disk contents into the backend spec instead of dealing with them ad-hoc in propolis-server, and clean up the instance initialization code accordingly.

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -255,54 +255,6 @@ impl<'a> MachineInitializer<'a> {
         Ok(())
     }
 
-    fn initialize_persistent_storage_backend(
-        &self,
-        backend_spec: &StorageBackend,
-    ) -> Result<StorageBackendInstance, Error> {
-        Ok(match &backend_spec.kind {
-            StorageBackendKind::Crucible { gen, req } => {
-                info!(
-                    self.log,
-                    "Creating Crucible disk from request {:?}", req
-                );
-                let be = propolis::block::CrucibleBackend::create(
-                    *gen,
-                    req.clone(),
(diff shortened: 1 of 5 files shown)


bin/propolis-server/src/lib/initializer.rs
```diff
@@ -325,7 +325,17 @@ impl<'a> MachineInitializer<'a> {
                         );
                         StorageBackendInstance { be, child, crucible: None }
                     }
-                    StorageBackendKind::InMemory { bytes } => {
+                    StorageBackendKind::InMemory { base64 } => {
+                        let bytes = base64::decode(base64).map_err(|e| {
```
luqmana, line 329: Is the only change here just moving where we do the base64 decode? I'm a little inclined to leave `InMemory` holding bytes, perhaps decoding in `add_cloud_init_from_request`

gjcolombo: Yeah, that's all commit e634187 does.

This probably makes more sense in the context of other branches I have downstream of this one. My long-ish-term goal is to make `struct InstanceEnsureRequest` contain an instance spec instead of the `disks`/`nics`/`cloud_init_bytes` fields it has today (or at least have a version of the ensure API that takes a spec; one of my downstream changes implements such an API). The idea behind this commit is to make the spec look more like the existing request parameters (both to make it easier to teach Nexus about specs later and because (IIUC) large binary blobs are less expensive to transmit this way); we just won't reap any of the benefits of doing that in this particular PR.

luqmana (approved):

PR #225, opened 2022-10-12
Split instance spec builder into generic and server-specific pieces

gjcolombo:
Today, propolis-server has code to build an `InstanceSpec` out of the parameters in an `InstanceEnsureRequest` and the information in the server's configuration TOML. This code has two logical layers: one converts inputs to `InstanceSpec` components, and the other incrementally builds the spec, checking for errors at each step (e.g. duplicating a component name or PCI path).

Refactor so that the "incrementally build a spec" code lives in a spec builder in propolis-client (where the other instance spec code lives) and the "convert Propolis parameters to spec elements" logic calls it. This sets up other clients (PHD in the short term, Nexus in the long term) with an incremental spec builder that they can use to construct specs to send to Propolis.

Tested via PHD & ad hoc testing of a Windows Server 2022 VM.

### DIFF
--- a/bin/propolis-server/src/lib/spec.rs
+++ b/bin/propolis-server/src/lib/spec.rs
@@ -94,7 +94,7 @@ fn make_storage_backend_from_config(
                     .get("path")
                     .ok_or_else(|| {
                         ServerSpecBuilderError::ConfigTomlError(format!(
-                            "Invalid path for file backend {}",
+                            "Couldn't get path for file backend {}",
                             name
                         ))
                     })?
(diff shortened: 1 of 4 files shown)


bin/propolis-server/src/lib/spec.rs
```diff
@@ -14,22 +13,13 @@ use crate::config;
...
-    DeviceNameInUse(String),
-
-    #[error("A backend with name {0} already exists")]
-    BackendNameInUse(String),
-
-    #[error("A PCI device is already attached at {0:?}")]
-    PciPathInUse(PciPath),
+pub enum ServerSpecBuilderError {
+    #[error("Interior spec builder returned an error: {0}")]
+    InnerBuilderError(#[from] SpecBuilderError),
```
luqmana, line 18: `#[error(transparent)]` might be suitable here if we just want to forward the `SpecBuilderError`s

luqmana (approved):
Thanks for factoring that out. Just had one small comment

bin/propolis-server/src/lib/spec.rs
gjcolombo: Good call. Fixed in b243962.

gjcolombo:
Thanks as always for the review! Will merge once PHD checks are done.

PR #228, opened 2022-10-19
PHD: separate disk lifetimes from VM lifetimes

gjcolombo:
(N.B. This PR is stacked on #225.)

# Summary

Rework PHD's guest disk management so that test VMs create temporary copies of guest images and boot from those instead of booting directly from the images in the artifact store. This obsoletes the framework's ZFS snapshot fixture, so remove it. 

Also reorganize the VM configuration builder to accommodate this method of specifying disks. Along the way, make it more instance spec-driven to remove some getter routines and to make it easier to write functions that create migration target configs that refer to the same disks as their sources.

With this the runner no longer needs to reacquire and rehash guest images in between tests. This reduces CI execution times with the current test suite by a factor of 4 ([110.78 seconds](https://buildomat.eng.oxide.computer/wg/0/artefact/01GF98BHPREKZYWGJFF1T5BQJP/zLJm5QNvupkZZYtx2xvw9z8pz0S0hQVF2QVR12lSbK7o6ew9/01GF98C09CP2YKM0MTPQZBFAEH/01GF9995CXGW7WEM183AMY9BKG/phd-runner.log?format=x-bunyan) before, [28.61 seconds](https://buildomat.eng.oxide.computer/wg/0/artefact/01GFCF3FCG22G66TMSXS3PYQ5F/LcIDs7xK5291075BKhNW4uqRSc9Kpwajn3eFdkUkuSHcKGCv/01GFCF3RBT9GB9BVAXC7ZHVQ9M/01GFCFY0NXWZE76SM3Q3VWWHW5/phd-runner.log?format=x-bunyan) after).

More importantly, this creates a much more flexible interface for tests to create logical disks and attach them to VMs. This interface will feature in a future PR that will teach PHD to work with Crucible backends.

# Review notes

The most important thing to review in this PR is the framework interface for creating disks (i.e. the code in `phd-tests/framework/src/disk/*.rs`). Does the programming model in the doc comment for `DiskFactory::create_disk` seem reasonable? Do the resulting test ergonomics seem OK? Are there any obvious whoopsies in the new disk code? Most of the rest of the PR should just be refactoring to accommodate the new disk usage pattern.


jordanhendricks (approved):
This looks good to me! Thanks for your work on this.

The ergonomics make sense -- I appreciate the explanation in the comment above `create_disk` that explains why we might want the disk to live longer than a VM (or not).

gjcolombo:
Thanks for the review!

PR #231, opened 2022-10-27
Create an instance spec-driven instance initialization endpoint

gjcolombo:
(Updated 17 November 2022 to reflect the existence of RFD 344 and of some commits that I've recently pushed.)

Create Propolis server endpoints that can create an instance from a spec and yield the spec of a running (or previously running) instance. This realizes part of the design described in RFD 344. It also makes it much simpler to support Crucible disks in PHD.

RFD 344 discusses how Nexus and the sled agent will use these new APIs. Since that code doesn't exist yet, for the time being, implement the new instance spec APIs alongside the existing ensure/get APIs so that Omicron can pick up new Propolis builds without taking any breaking API changes. Once this PR lands and Omicron is ready to use the new endpoints, we can remove the old ones.

Other supporting changes in this PR:

- Switch instance specs' table structures from `BTreeMap` to `HashMap`. This makes it easier to convert handmade instance specs into their Progenitor-generated equivalents. (We use these containers just as associative arrays and don't care about BTreeMap's ordering guarantees.)
- Allow instance specs to produce types in an OpenAPI schema:
  - Derive `JsonSchema` for everything that needs it
  - Change `PciPath` from a tuple struct to an ordinary struct (schema generation doesn't support multi-element tuple structs)
  - Add some adapters to support From/Into conversions from handmade instance specs to generated specs
- Update PHD to use the instance spec endpoint when starting VMs. Remove the now-obsolete logic to turn instance specs back into config TOML files. (At this point, if instance specs allowed users to specify firmware images, the server wouldn't need to read config TOMLs at all...)

I am not at all sure that the handmade/generated type converters are the right approach to generating OpenAPI-compatible instance specs. A better approach might be to get the instance spec builder to produce specs of the generated type, but I hit some friction getting that to work (it looked like the generated types didn't derive some traits like `Default` that the builder was using). I can try to work this change into this PR if reviewers think it's important but am also happy to address the issue another day.


gjcolombo:
OK, I think this is actually ready for review at this point.

This is one of those "looks bigger than it really is" PRs, since most of the added lines are from the OpenAPI spec change (which will come back out when I merge the get/ensure endpoints back together). Beyond that, most of the remaining changes are more of the refactor-and-rearrange sort than the implement-brand-new-components sort. Nevertheless, it might be possible to break the PR up a bit if that makes it easier to review--let me know.

luqmana (approved):

Issue #209, opened 2022-09-12
Consider adding "failed" terminal VM state


gjcolombo:
Follow-up from the review of #208.

Fatal errors that are handled on the state worker (e.g. faults from a vCPU that halt the VM, errors from vCPU run ioctls) currently panic the host process. Instead, they could move the VM into a "faulted" terminal state, after which the server would wait for further instructions.

This would allow the VM owner or an operator to examine the VM and its state to try to diagnose the fault, though this would come at the cost of the resources needed to preserve the VM and any of its state while the VM is in the failed state.

Even if we don't want to keep that data, reporting guest-driven faults in an orderly manner will let us (or an operator) distinguish more clearly between, say, an assertion failure in Propolis (implying a code bug in the VMM) from an unrecoverable guest fault (possibly, albeit not absolutely, a problem with guest software).

luqmana:
So we do already have such a state in the API:

https://github.com/oxidecomputer/propolis/blob/9ee3e5657437f3c1cf2c4df9c7b48bc81e5b9b7f/lib/propolis-client/src/handmade/api.rs#L139

We just don't make use of it yet.

PR #235, opened 2022-10-31
Unify cleanup for external Stop request and guest-initiated shutdown.

luqmana:
We were only cleaning up the server-side resources if we got an
explicity external Stop request. But if the guest initiated a shutdown
we would correctly marked the VM as `Stopped` but never cleanup the
server's `ServiceProviders` nor ever mark it as `Destroyed`.

Instead this introduces a task that will trigger the cleanup when
the VM state worker exits.

### DIFF
--- a/bin/propolis-server/src/lib/server.rs
+++ b/bin/propolis-server/src/lib/server.rs
@@ -187,7 +187,7 @@ impl ServiceProviders {
 /// Context accessible from HTTP callbacks.
 pub struct DropshotEndpointContext {
     static_config: StaticConfig,
-    pub services: ServiceProviders,
+    pub services: Arc<ServiceProviders>,
     log: Logger,
 }
 
@@ -206,13 +206,13 @@ impl DropshotEndpointContext {
                 use_reservoir,
                 metrics: metric_config,
             },
-            services: ServiceProviders {
+            services: Arc::new(ServiceProviders {
                 vm: Mutex::new(VmControllerState::NotCreated),
                 serial_task: Mutex::new(None),
                 oximeter_server_task: Mutex::new(None),
                 oximeter_stats: Mutex::new(None),
                 vnc_server: Arc::new(Mutex::new(vnc_server)),
-            },
+            }),
             log,
         }
     }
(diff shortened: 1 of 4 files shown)


bin/propolis-server/src/lib/server.rs
```diff
@@ -458,6 +458,26 @@ async fn instance_ensure(
...
+        loop {
+            if let Err(_) = state_watcher.changed().await {
+                // The sender has been dropped (i.e. the VM has been destroyed).
+                // Nothing more for us to do so just bail.
+                break;
+            }
+
+            let state = state_watcher.borrow().state;
+            if state == api::InstanceState::Stopping {
+                // The VM has started Stopping, release the per-instance service providers
```
gjcolombo, line 474: Are we guaranteed to observe this transition? My understanding of the tokio `watch` is that you could have something like this happen:

- this task calls `state_watcher.changed` and awaits
- VM controller writes `Stopping` to the channel, satisfying the call to `changed`
- before the task gets to run again, the VM controller finishes halting the VM and writes `Stopped` to the channel
- this task calls `state_watcher.borrow()` and gets `Stopped`, so doesn't reach line 475

Assuming I'm not misreading the `watch::Receiver` docs, would it make sense to create a separate channel for "the guest shut itself down" messages?

gjcolombo (approved):
Looks great. Sorry about the bug--thanks for spending time on the fix!

bin/propolis-server/src/lib/server.rs
luqmana: Nice catch! I swapped it to use a one-shot channel that gets explicitly used once the state worker has finished.

luqmana:
> Looks great. Sorry about the bug--thanks for spending time on the fix!

You shouldn't apologize for that! Thank you for catching it. 😅

bin/propolis-server/src/lib/server.rs
```diff
@@ -458,6 +461,15 @@ async fn instance_ensure(
         }));
     }
 
+    let log = server_context.log.clone();
+    let services = Arc::clone(&server_context.services);
+    tokio::task::spawn(async move {
+        // Once the VmController has signaled that it is shutting down,
+        // we'll clean up the per-instance service providers as well.
+        let _ = stop_recv.await;
```
luqmana, line 469: So, this could return an `Err` meaning the Sender was dropped without sending. That could happen if the `vm_state_worker` thread panics. I don't think we ever actually mark the instance as `Failed` but here seems like maybe one place?

luqmana: Was going to open an issue, but there's already https://github.com/oxidecomputer/propolis/issues/209

PR #236, opened 2022-11-01
server: send close frames to serial websocket clients on instance stop

luqmana:
Try to gracefully inform any connected serial clients when a VM is stopped. Also remove the Drop impl on SerialTask as it would abort the serial task before it ever got the chance to handle the close message we send it.

This should get rid of the noise in the sled-agent log:

before:
>ERROR: SledAgent/InstanceManager/13728 on luq-helios: Reading TTY: Protocol(ResetWithoutClosingHandshake)

after:
>INFO: SledAgent/InstanceManager/29709 on luq-helios: Closing TTY connection: VM stopped (1001)

### DIFF
--- a/bin/propolis-server/src/lib/server.rs
+++ b/bin/propolis-server/src/lib/server.rs
@@ -172,10 +172,10 @@ impl ServiceProviders {
                         "strong_refs" => Arc::strong_count(&vm),
                         "weak_refs" => Arc::weak_count(&vm));
         }
-        if let Some(mut serial_task) = self.serial_task.lock().await.take() {
-            if let Some(close) = serial_task.close_ch.take() {
-                let _ = close.send(());
-            }
+        if let Some(serial_task) = self.serial_task.lock().await.take() {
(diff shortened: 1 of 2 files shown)


bin/propolis-server/src/lib/server.rs
```diff
@@ -629,15 +629,11 @@ async fn instance_serial(
...
+                error!(err_log, "Serial task failed: {}", e);
             }
         });
 
-        super::serial::SerialTask {
-            task,
-            close_ch: Some(close_ch),
-            websocks_ch,
-        }
+        super::serial::SerialTask { task, close_ch: close_ch, websocks_ch }
```
lifning, line 636: small, nonblocking: i think `close_ch: close_ch` can be simplified

lifning (approved):

bin/propolis-server/src/lib/server.rs
luqmana: right you are

Issue #233, opened 2022-10-28
Windows Server 2022 guests bugcheck after LM if an RDP session was alive on the source


gjcolombo:
This is a fun one. Repro steps:

1. Launch a Windows Server 2022 VM (latest virtio-net adapter, NVMe boot disk, netkvm.sys installed).
2. Remote Desktop to the VM.
3. With the connection still alive, live migrate the VM.

Expected: things function normally.
Observed: the system bugchecks in assorted strange and wonderful ways a few seconds after migrating.

Interestingly, having an active remote desktop session seems to be essential to cause this problem. The following conditions *don't* seem to cause it:

- Starting a command prompt over the SAC before migrating
- Having an open SSH session while migrating
- Logging in via VNC and then migrating

Connecting via RDP and then disconnecting seems to mitigate the problem, at least sometimes. _Signing out_ of the entire RDP session (as opposed to disconnecting) did not repro. I have in my notes that the bugcheck reproed once after merely _disconnecting_, but I wasn't able to reproduce that result just now; some more experimentation may be needed. (For reference, there are some RDP-specific kernel-mode drivers (`wudfrd` and `indirectkmd`) that get loaded while an RDP session is active and are stopped when no clients are connected.)

---

The bugchecks themselves feature several different codes, all carrying the vague scent of some kind of corrupted state. I've seen

* CRITICAL_PROCESS_DIED, in a couple of flavors:
  * Fatal RPC exception while trying to do an operation involving a token
  * Fatal exception while trying to execute some user-mode heap code (about which more below)
* CRITICAL_STRUCTURE_CORRUPTION, in both cases with arg4 set to 3 (processor GDT corruption)
* UNEXPECTED_KERNEL_MODE_TRAP, caused by a double fault

The "fatal exception while executing user-mode heap code" dump has better signal than the others. More on this one in a minute.

gjcolombo:
Trying to debug these sorts of things without symbols is always a little bit of a hassle, but in one case we've gotten lucky and have a little more to go on. Here is the faulting stack from one of the CRITICAL_PROCESS_DIED bugchecks:

```
1: kd> k
 # Child-SP          RetAddr           Call Site
00 fffffc8c`c8c91cb8 fffff805`3238f012 nt!KeBugCheckEx
01 fffffc8c`c8c91cc0 fffff805`3229375b nt!PspCatchCriticalBreak+0x11a
[...]

PR #239, opened 2022-11-03
propolis-server: make held vCPUs check for new events before re-entering the guest

gjcolombo:
This came up in #233. Without this, vCPUs on a migration source can run briefly before noticing that they're supposed to exit. This means that after a migration, the preserved vCPU state from the source VM (assuming there is some) can differ from what was sent to the target, which isn't great for debugging.


pfmooney (approved):
Ship it!

PR #240, opened 2022-11-03
Fix README to describe switches needed for LM

gjcolombo:
Follow-up from #233. Update the project README to include instructions for setting `vmm_allow_state_writes` and `gpt_track_dirty`, both of which are currently needed for (reliable) LM.


gjcolombo:
We discussed the verbiage change in 30eb5b8 offline, so I'm going to go ahead and merge this.

PR #243, opened 2022-11-07
Add SoftNPU and P9fs devices.

rcgoodfellow:
This PR adds two new devices to Propolis.

- A P4 ASIC emulator called SoftNPU.
- A Plan 9 filesystem device.

These devices have been in daily use through [Falcon](https://github.com/oxidecomputer/falcon) for quite some time now. This PR is an effort to

- Make these devices easier for folks to consume by being in mainline Propolis.
- Reduce maintenance overhead in keeping the `falcon` branch up to date.

A synopsis of these devices can be found in the main struct comments for [SoftNPU](https://github.com/oxidecomputer/propolis/blob/softnpu-pr/lib/propolis/src/hw/virtio/softnpu.rs#L45-L89) and [PciVirtio9pfs](https://github.com/oxidecomputer/propolis/blob/softnpu-pr/lib/propolis/src/hw/virtio/p9fs.rs#L41-L51).

Today these devices are used exclusively in development environments, and that is likely to remain true indefinitely for SoftNPU. SoftNPU supports loading shared library code given to it by programs running in the guest. This is not something most people would want anywhere near a production environment hypervisor. Due to these characteristics, both of these new devices are only compiled into Propolis if the `falcon` Cargo feature is present.

At the time of writing, this PR depends on a private `p4` repository. That repository will be going public soon. After that happens and after a bit more self-review, I'll move this PR out of the draft state.

### DIFF
--- a/bin/propolis-server/Cargo.toml
+++ b/bin/propolis-server/Cargo.toml
@@ -65,3 +65,4 @@ expectorate = "1.0.5"
 [features]
 default = ["dtrace-probes"]
 dtrace-probes = ["propolis/dtrace-probes", "dropshot/usdt-probes"]
+falcon = []
--- a/bin/propolis-server/Cargo.toml
+++ b/bin/propolis-server/Cargo.toml
@@ -65,4 +65,4 @@ expectorate = "1.0.5"
 [features]
 default = ["dtrace-probes"]
 dtrace-probes = ["propolis/dtrace-probes", "dropshot/usdt-probes"]
-falcon = []
+falcon = ["propolis/falcon"]
--- a/bin/propolis-server/src/lib/server.rs
+++ b/bin/propolis-server/src/lib/server.rs
@@ -271,6 +271,7 @@ fn instance_spec_from_request(
         instance_spec::SerialPortNumber::Com1,
         instance_spec::SerialPortNumber::Com2,
         instance_spec::SerialPortNumber::Com3,
+        #[cfg(not(feature = "falcon"))]
         instance_spec::SerialPortNumber::Com4,
     ] {
         spec_builder.add_serial_port(port)?;
--- a/bin/propolis-server/src/lib/server.rs
+++ b/bin/propolis-server/src/lib/server.rs
@@ -271,6 +271,7 @@ fn instance_spec_from_request(
         instance_spec::SerialPortNumber::Com1,
         instance_spec::SerialPortNumber::Com2,
         instance_spec::SerialPortNumber::Com3,
+        // SoftNpu uses this port for ASIC management.
         #[cfg(not(feature = "falcon"))]
         instance_spec::SerialPortNumber::Com4,
     ] {
--- a/lib/propolis-client/src/instance_spec/builder.rs
+++ b/lib/propolis-client/src/instance_spec/builder.rs
@@ -5,7 +5,7 @@ use thiserror::Error;
 
 use super::{Board, Chipset, InstanceSpec, SerialPort, SerialPortNumber};
 #[cfg(feature = "falcon")]
-use super::{P9fs, SoftNpuP9, SoftNpuPort, TfPort0};
+use super::{P9fs, SoftNpuP9, SoftNpuPort, SoftNpuPciPort};
 
 /// Errors that can arise while building an instance spec from component parts.
 #[derive(Debug, Error)]
@@ -168,12 +168,12 @@ impl SpecBuilder {
     }
 
     #[cfg(feature = "falcon")]
-    /// Sets softnpu tfport0
-    pub fn set_softnpu_tfport0(
+    /// Sets softnpu pci port
+    pub fn set_softnpu_pci_port(
         &mut self,
-        tfport0: TfPort0,
+        pci_port: SoftNpuPciPort,
     ) -> Result<&Self, SpecBuilderError> {
-        self.spec.devices.tfport0 = Some(tfport0);
+        self.spec.devices.softnpu_pci_port = Some(pci_port);
         Ok(self)
     }
 
--- a/lib/propolis-client/src/instance_spec/builder.rs
+++ b/lib/propolis-client/src/instance_spec/builder.rs
@@ -5,7 +5,7 @@ use thiserror::Error;
 
 use super::{Board, Chipset, InstanceSpec, SerialPort, SerialPortNumber};
 #[cfg(feature = "falcon")]
-use super::{P9fs, SoftNpuP9, SoftNpuPort, SoftNpuPciPort};
+use super::{P9fs, SoftNpuP9, SoftNpuPciPort, SoftNpuPort};
 
 /// Errors that can arise while building an instance spec from component parts.
 #[derive(Debug, Error)]
--- a/lib/propolis-client/src/instance_spec/builder.rs
+++ b/lib/propolis-client/src/instance_spec/builder.rs
@@ -5,7 +5,7 @@ use thiserror::Error;
 
 use super::{ Board, Chipset, InstanceSpec, SerialPort, SerialPortNumber };
 #[cfg(feature = "falcon")]
-use super::{ SoftNpuPort, SoftNpuP9, TfPort0 };
+use super::{ SoftNpuPort, SoftNpuP9, TfPort0, P9fs };
 
 /// Errors that can arise while building an instance spec from component parts.
 #[derive(Debug, Error)]
@@ -193,6 +193,15 @@ impl SpecBuilder {
         p9: SoftNpuP9,
     ) -> Result<&Self, SpecBuilderError> {
         self.spec.devices.softnpu_p9= Some(p9);
+        Ok(self)
+    }
+
+    #[cfg(feature = "falcon")]
+    pub fn set_p9fs(
+        &mut self,
+        p9fs: P9fs,
+    ) -> Result<&Self, SpecBuilderError> {
+        self.spec.devices.p9fs = Some(p9fs);
         Ok(self)
     }
 
--- a/lib/propolis-client/src/instance_spec/builder.rs
+++ b/lib/propolis-client/src/instance_spec/builder.rs
@@ -208,12 +208,14 @@ impl SpecBuilder {
         &mut self,
         p9: SoftNpuP9,
     ) -> Result<&Self, SpecBuilderError> {
+        self.register_pci_device(p9.pci_path)?;
         self.spec.devices.softnpu_p9 = Some(p9);
         Ok(self)
     }
 
     #[cfg(feature = "falcon")]
     pub fn set_p9fs(&mut self, p9fs: P9fs) -> Result<&Self, SpecBuilderError> {
+        self.register_pci_device(p9fs.pci_path)?;
         self.spec.devices.p9fs = Some(p9fs);
         Ok(self)
     }
--- a/lib/propolis-client/src/instance_spec/builder.rs
+++ b/lib/propolis-client/src/instance_spec/builder.rs
@@ -21,6 +21,9 @@ pub enum SpecBuilderError {
 
     #[error("Serial port {0:?} is already specified")]
     SerialPortInUse(SerialPortNumber),
+
+    #[error("SoftNpu port {0:?} is already specified")]
+    SoftNpuPortInUse(String),
 }
 
 /// A builder that constructs instance specs incrementally and catches basic
@@ -183,8 +186,11 @@ impl SpecBuilder {
         key: String,
         port: SoftNpuPort,
     ) -> Result<&Self, SpecBuilderError> {
-        self.spec.devices.softnpu_ports.insert(key, port);
-        Ok(self)
+        if self.spec.devices.softnpu_ports.insert(key, port.clone()).is_some() {
+            Err(SpecBuilderError::SoftNpuPortInUse(port.name.clone()))
+        } else {
+            Ok(self)
+        }
     }
 
     #[cfg(feature = "falcon")]
--- a/lib/propolis-client/src/instance_spec/builder.rs
+++ b/lib/propolis-client/src/instance_spec/builder.rs
@@ -3,9 +3,9 @@ use std::collections::BTreeSet;
 use propolis_types::PciPath;
 use thiserror::Error;
 
-use super::{ Board, Chipset, InstanceSpec, SerialPort, SerialPortNumber };
+use super::{Board, Chipset, InstanceSpec, SerialPort, SerialPortNumber};
 #[cfg(feature = "falcon")]
-use super::{ SoftNpuPort, SoftNpuP9, TfPort0, P9fs };
+use super::{P9fs, SoftNpuP9, SoftNpuPort, TfPort0};
 
 /// Errors that can arise while building an instance spec from component parts.
 #[derive(Debug, Error)]
@@ -192,15 +192,12 @@ impl SpecBuilder {
         &mut self,
         p9: SoftNpuP9,
     ) -> Result<&Self, SpecBuilderError> {
-        self.spec.devices.softnpu_p9= Some(p9);
+        self.spec.devices.softnpu_p9 = Some(p9);
         Ok(self)
     }
 
     #[cfg(feature = "falcon")]
-    pub fn set_p9fs(
-        &mut self,
-        p9fs: P9fs,
-    ) -> Result<&Self, SpecBuilderError> {
+    pub fn set_p9fs(&mut self, p9fs: P9fs) -> Result<&Self, SpecBuilderError> {
         self.spec.devices.p9fs = Some(p9fs);
         Ok(self)
     }
--- a/lib/propolis-client/src/instance_spec/builder.rs
+++ b/lib/propolis-client/src/instance_spec/builder.rs
@@ -3,7 +3,9 @@ use std::collections::BTreeSet;
 use propolis_types::PciPath;
 use thiserror::Error;
 
-use super::{Board, Chipset, InstanceSpec, SerialPort, SerialPortNumber};
+use super::{ Board, Chipset, InstanceSpec, SerialPort, SerialPortNumber };
+#[cfg(feature = "falcon")]
+use super::{ SoftNpuPort, SoftNpuP9, TfPort0 };
 
 /// Errors that can arise while building an instance spec from component parts.
 #[derive(Debug, Error)]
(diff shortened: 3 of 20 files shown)


lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -0,0 +1,1061 @@
...
+        }
+    }
+}
+
+impl Entity for SoftNpu {
+    fn type_name(&self) -> &'static str {
+        "softnpu"
+    }
+
+    fn start(&self) {
```
pfmooney, line 273: Given that only the `start` handler is implemented, is it safe to say that falcon guests (or at least those using SoftNPU) are never expected to reboot? (Or for that matter, cleanly halt)

lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -0,0 +1,1061 @@
...
+    virtio: Arc<PortVirtioState>,
+
+    /// UART for management from guest
+    uart: Arc<LpcUart>,
+
+    /// P9 file system endpoint for pre-compiled program transfer
+    pub p9fs: Arc<PciVirtio9pfs<SoftNpuP9Handler>>,
+
+    //TODO should be able to do this as a RwLock
+    pipeline: Arc<Mutex<Option<(Library, Box<dyn Pipeline>)>>>,
```
pfmooney, line 110: Given how `data_handles`, the (CPU port) virtio state, and pipeline are all used by the worker threads, does it make sense to keep extra copies here (where they are passed to said workers during spawn), or perhaps just hand off the `Arc<PciVirtioSoftNpuPort>`, into which those workers can reach for the bits they need?

lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -0,0 +1,1061 @@
...
+    fn handle_q0_req(&self, _vq: &Arc<VirtQueue>) {
+        // This is the queue that the virtio driver in the guest reads from and
+        // we write to. It seems that the correct way to handle a queue
+        // notification on this queue is to not handle it? If we vq.pop_avail to
+        // see what's in the queue, it's always zero data, and the act of doing
+        // a vq.pop_avail seems to drain the queue until it is unusable for
+        // writes, even if we do the corresponding push_used.
+
+        return;
+    }
```
pfmooney, line 392: The virtio abstraction(s) in propolis should probably grow a way to get/set `USED_F_NO_NOTIFY`.  The kernel side of viona does this, indicating to the guest that we do not want/need notifications for newly available descriptors in the RX ring.

lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -0,0 +1,1061 @@
...
+        }
+    }
+
+    /// Send a packet out the guest pci port using virtio.
+    fn send_packet_to_cpu_port<'a>(
+        pkt: &mut packet_out<'a>,
+        virtio: &Arc<PortVirtioState>,
+        _log: &Logger,
+    ) {
+        let mem = virtio.pci_state.acc_mem.access().unwrap();
```
pfmooney, line 606: We should probably be more graceful here if we cannot access guest memory through the pci state handle.  Such would be the case if the guest happened to reset the device and/or disable bus-mastering.  In that case, this handler can assume there is no free descriptor into which the packet can be placed.

lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -0,0 +1,1061 @@
...
+            let pl = match &mut *p {
+                Some(ref mut pl) => &mut pl.1,
+                None => continue, // no program is loaded
+            };
+
+            Self::process_external_packet(
+                index + 1,
+                pkt,
+                &data_handles,
+                virtio.clone(),
```
pfmooney, line 523: Any particular reason why the virtio refs are being cloned for every packet?

lib/propolis/Cargo.toml
```diff
@@ -47,6 +53,24 @@ git = "https://github.com/oxidecomputer/omicron"
 branch = "main"
 optional = true
 
+[dependencies.p4rs]
```
pfmooney, line 56: For ease of reading, can we include a comment here mentioning that these are associated w/ falcon too?

lib/propolis-client/src/instance_spec/devices.rs
```diff
@@ -276,6 +313,14 @@ pub struct DeviceSpec {
...
     pub serial_ports: BTreeMap<SpecKey, SerialPort>,
     pub pci_pci_bridges: BTreeMap<SpecKey, PciPciBridge>,
+    #[cfg(feature = "falcon")]
+    pub softnpu_pci_port: Option<SoftNpuPciPort>,
+    #[cfg(feature = "falcon")]
+    pub softnpu_ports: BTreeMap<SpecKey, SoftNpuPort>,
+    #[cfg(feature = "falcon")]
+    pub softnpu_p9: Option<SoftNpuP9>,
+    #[cfg(feature = "falcon")]
+    pub p9fs: Option<P9fs>,
```
pfmooney, line 323: I'll defer to @gjcolombo on the instance spec stuff here.

pfmooney (commented):
Thanks for putting this together.

lib/propolis/src/hw/virtio/softnpu.rs
pfmooney: In short: Ignoring notifications to vq0 is the right thing to do here.

bin/propolis-server/Cargo.toml
```diff
@@ -65,3 +65,4 @@ expectorate = "1.0.5"
 [features]
 default = ["dtrace-probes"]
 dtrace-probes = ["propolis/dtrace-probes", "dropshot/usdt-probes"]
+falcon = []
```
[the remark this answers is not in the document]
rcgoodfellow: Yeah, I think so. Seems I was getting away with this based on the fact that I was building everything all the time via `cargo build --features falcon` which applied the feature across the cargo projects.

lib/propolis-client/src/instance_spec/builder.rs
```diff
@@ -165,6 +170,44 @@ impl SpecBuilder {
         }
     }
 
+    #[cfg(feature = "falcon")]
+    /// Sets softnpu pci port
+    pub fn set_softnpu_pci_port(
+        &mut self,
+        pci_port: SoftNpuPciPort,
+    ) -> Result<&Self, SpecBuilderError> {
+        self.spec.devices.softnpu_pci_port = Some(pci_port);
```
gjcolombo, line 179: This should also call `self.register_pci_device` so that the builder knows the port is reserved (it will then catch attempts to use it again). Same thing for SoftNpuP9.

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -0,0 +1,1000 @@
...
+/// The design centers around a P9Handler trait that allows various different
+/// types of P9 devices to be implemented. This file includes a `HostFSHandler`
+/// implementation that allows mounting host filesystems in the guest.
+/// Currently filesystems can only be mounted as read-only. Another
+/// implementation is in the SoftNpu device that supports P4 program transfer
+/// via p9fs.
+pub struct PciVirtio9pfs<Handler: P9Handler> {
+    virtio_state: PciVirtioState,
+    pci_state: pci::DeviceState,
+    handler: Handler,
```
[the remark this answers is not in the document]
rcgoodfellow: SGTM. I've transitioned to a trait object in 1abd8d7.

lib/propolis-client/src/instance_spec/devices.rs
```diff
@@ -268,6 +268,43 @@ impl MigrationElement for PciPciBridge {
...
+}
+
+#[cfg(feature = "falcon")]
+#[derive(Clone, Deserialize, Serialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct SoftNpuPort {
+    /// The name of the SoftNpu port.
+    pub name: String,
+
+    /// The name of the associated VNIC.
```
gjcolombo, line 286: Minor nit: if this VNIC is similar to a `NetworkDevice` VNIC, i.e. it's some sort of host resource that's configured/managed outside of Propolis, then strictly speaking it should be defined over in the backend spec (and then this element can refer to that backend element by name the way a `NetworkDevice` does). I'm not sure if the existing virtio backend is fit for purpose here or if you'd need to define another backend type.

lib/propolis-client/src/instance_spec/devices.rs
```diff
@@ -276,6 +313,14 @@ pub struct DeviceSpec {
...
+    pub softnpu_pci_port: Option<SoftNpuPciPort>,
+    #[cfg(feature = "falcon")]
+    pub softnpu_ports: BTreeMap<SpecKey, SoftNpuPort>,
+    #[cfg(feature = "falcon")]
+    pub softnpu_p9: Option<SoftNpuP9>,
+    #[cfg(feature = "falcon")]
+    pub p9fs: Option<P9fs>,
 }
 
 impl DeviceSpec {
```
gjcolombo, line 326: I'm assuming P9 and SoftNpu devices aren't meant to live migrate for now? If they aren't, their `Entity` impls should indicate as much (see my comment over in softnpu.rs). If they are, the `can_migrate_devices_from` routine (which I cannot comment on directly even though it is LITERALLY ON THE NEXT LINE come on Github are you kidding me) should be updated to check these elements for compatibility.

If it existed, the doc comment for `can_migrate_devices_from` would explain that it returns `Ok(())` if `self` and `other` describe specs whose members are similar enough that if you initialized instance 1 from `self` and instance 2 from `other`, you could migrate from one to the other without breaking any guest abstractions (e.g. changing something's PCI path).

Again, my guess is that for now it's best for the components themselves just to say that they can't be migrated, but if you do want them to migrate these checks are important to add.

lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -0,0 +1,1061 @@
...
+            let pipeline = pipeline.clone();
+            let uart = uart.clone();
+            let log = log.clone();
+            handle_management_message(msg, pipeline, uart, radix, log.clone());
+            info!(log, "handled management message");
+        }
+    }
+}
+
+impl Entity for SoftNpu {
```
gjcolombo, line 268: Are SoftNpu and/or P9 devices expected to be migratable? If not they should also implement `migrate` and return `Migrator::NonMigratable`. (If they are then you probably need some instance spec changes; left a comment over in that module describing those.)

lib/propolis-client/src/instance_spec/devices.rs
```diff
@@ -268,6 +268,43 @@ impl MigrationElement for PciPciBridge {
...
+}
+
+#[cfg(feature = "falcon")]
+#[derive(Clone, Deserialize, Serialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct P9fs {
+    pub source: String,
+    pub target: String,
+    pub chunk_size: u32,
+    pub pci_path: PciPath,
```
gjcolombo, line 305: Nit: doc comments for these fields

lib/propolis-client/src/instance_spec/devices.rs
gjcolombo: Thanks! I've left a few comments.

lib/propolis/src/hw/virtio/softnpu.rs
rcgoodfellow: An interesting question. For the primary thing that SoftNPU currently emulates, a sidecar switch - the sidecar will indeed keep on going if the attached scrimlet reboots, halts, etc. So in that sense, I think we have the desired behavior here, where the guest (scrimlet) reboots and comes back to find the sidecar not having gone down with it.

If we start using SoftNpu for emulating something like a programmable NIC, that would be a different story.

lib/propolis-client/src/instance_spec/devices.rs
rcgoodfellow: I think this would be a different backend type than virtio. Virtio plugs into the VNIC with a viona data path, SoftNpu ports use DLPI.

lib/propolis/src/hw/virtio/softnpu.rs
rcgoodfellow: Nope

lib/propolis/Cargo.toml
rcgoodfellow: Done

lib/propolis/src/hw/virtio/softnpu.rs
rcgoodfellow: Now checking for none, kicking out a warning and returning.

lib/propolis-client/src/instance_spec/builder.rs
rcgoodfellow: Done

lib/propolis-client/src/instance_spec/devices.rs
rcgoodfellow: This field and the machinery around it have been updated to used a registered backend device.

lib/propolis-client/src/instance_spec/devices.rs
rcgoodfellow: Yeah, we don't need to worry about migration for these devices now. If P9fs support graduates from a Falcon feature to a product feature, we can worry about migration then. For now, Falcon does not migrate things ;)

lib/propolis/src/hw/virtio/softnpu.rs
rcgoodfellow: This is all tucked away in the `PciVirtioSoftnpuPort` now, copies have been removed from `SoftNpu`.

rcgoodfellow:
Thanks for the reviews @pfmooney and @gjcolombo!

I've made updates based on the feedback.

lib/propolis-client/src/instance_spec/devices.rs
gjcolombo: Works for me!

gjcolombo (approved):
LGTM for the instance spec and instance lifecycle pieces.

lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -0,0 +1,1042 @@
...
+            return self.handle_q0_req(vq);
+        }
+
+        let mem = match self.virtio_state.pci_state.acc_mem.access() {
+            Some(mem) => mem,
+            None => {
+                warn!(self.log, "failed to access virtio memory");
+                return;
+            }
+        };
```
pfmooney, line 315: I would probably just silently return like the `pop_avail` check below.

lib/propolis/src/hw/virtio/softnpu.rs
pfmooney: To be clear: I'm not suggesting that we should reset the emulated switch state, but as-is, we're ignoring a bunch of expected state management expected of an emulated (PCI) device if the guest undergoes reboot.

If you really do not expect a falcon guest (at least one with a SoftNPU device attached) to undergo reboot, then this is probably fine, but as-is, it is breaking some of the rules/expectations around how device emulation acts when the instance is walking through expected state transitions for a reboot or halt.

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -0,0 +1,986 @@
...
+                match id {
+                    P9fsReg::TagLen => {
+                        ro.write_u16(self.handler.target().len() as u16);
+                    }
+                    P9fsReg::Tag => {
+                        let mut bs = [0; 256];
+                        for (i, x) in self.handler.target().bytes().enumerate()
+                        {
+                            bs[i] = x as u8;
+                        }
```
pfmooney, line 92: I should have caught this earlier: this could probably be reduced to a `as_bytes()` call on the `&str`, rather than a buffer copying situation.  In both cases, some bounds checking is likely needed to avoid potential overflow.

bin/propolis-server/src/lib/server.rs
```diff
@@ -271,6 +271,8 @@ fn instance_spec_from_request(
         instance_spec::SerialPortNumber::Com1,
         instance_spec::SerialPortNumber::Com2,
         instance_spec::SerialPortNumber::Com3,
+        // SoftNpu uses this port for ASIC management.
+        #[cfg(not(feature = "falcon"))]
         instance_spec::SerialPortNumber::Com4,
```
pfmooney, line 276: Lingering question for @gjcolombo: Is there a future where we might want to attach serial ports to things other than a user console?  In such a case, where COM port attachment is described in the instance spec, we could probably avoid its omission here (in the falcon case), and just spell out that it's being used by the SoftNPU?

pfmooney: I also wonder if there's a device id we could "borrow" for a PCI-attached 16650 chip, which the SoftNPU could just attach at slot X function 1, where that exposed topology implied the connection between the UART and the NPU stuff.

gjcolombo: > Lingering question for @gjcolombo: Is there a future where we might want to attach serial ports to things other than a user console? In such a case, where COM port attachment is described in the instance spec, we could probably avoid its omission here (in the falcon case), and just spell out that it's being used by the SoftNPU?

I'd model that as a device/backend split: the device handles all the UART register emulation, and the bytes it receives/transmits are sent to/obtained from some backend, which could be a human-readable I/O channel of some kind, a SoftNPU device, etc. I'd have to look at the serial port code again (it's been a while) to see how big of a refactoring that would be.

This will also be sort of relevant when #231 lands, because it lets callers like Omicron and PHD pass instance specs directly to Propolis, and they won't take this path. But I don't think those consumers plan to use Falcon builds, if they do they probably won't attach a SoftNPU device, and if they do we can catch/resolve the conflict in Propolis (today I'm pretty sure it'll panic because two devices will try to map the same part of PIO space, but we can make that cleaner by checking the instance spec for a conflict). So I don't think we need to refactor anything right away or change anything in this PR (though I might want to touch up #231 a bit after this lands).

I've filed #259 to track the broader "serial ports should have devices and backends" idea.

lib/propolis/src/hw/virtio/p9fs.rs
rcgoodfellow: When I change this to 
```
P9fsReg::Tag => {
    ro.write_bytes(self.handler.target().as_bytes());
    ro.fill(0);
}
```

The result is
```
thread 'vcpu-0' panicked at 'source slice length (14) does not match destination slice length (256)', lib/propolis/src/common.rs:156:29
stack backtrace:
   0: rust_begin_unwind
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:142:14
   2: core::slice::<impl [T]>::copy_from_slice::len_mismatch_fail
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/slice/mod.rs:3239:13
   3: propolis::common::ReadOp::write_bytes
   4: propolis::util::regmap::RegMap<ID>::process
   5: <propolis::hw::virtio::p9fs::PciVirtio9pfs as propolis::hw::virtio::VirtioDevice>::cfg_rw
   6: propolis::util::regmap::RegMap<ID>::process
   7: propolis::hw::pci::device::<impl propolis::hw::pci::Endpoint for D>::bar_rw
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I've added a check to the current code to avoid overruning the buffer for tag strings longer than 256 bytes.

lib/propolis/src/hw/virtio/softnpu.rs
rcgoodfellow: done

lib/propolis/src/hw/virtio/softnpu.rs
pfmooney: I'd overlooked the fact that the `PciVirtioSoftNpuPort` device had wired up the appropriate events for ensuring a PciVirtio reset when the guest itself is rebooted.  While there are some improvements that could be made around how the dlpi recv thread(s) are being managed (they'll likely prevent clean propolis exit on instance halt), that is something which could be addressed in a follow-up.  Clearly the tools in place for using Falcon today already take that into account.

rcgoodfellow: I've added an issue to track this 
- https://github.com/oxidecomputer/propolis/issues/261

pfmooney (approved):
Thanks for putting this together.

PR #244, opened 2022-11-08
Cleanup VMM handle once Instance is dropped.

luqmana:
Fixes https://github.com/oxidecomputer/propolis/issues/191.

Update the VNC server to cleanup on Stop (depends on https://github.com/oxidecomputer/rfb/pull/9) and release its Instance ref.

Update `Instance`'s drop impl to cleanup the VMM handle.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,4 @@ panic = "unwind"
 #crucible-client-types = { path = "../crucible/crucible-client-types" }
 
 #[patch."http://github.com/oxidecomputer/rfb"]
-# Double / here is intentional to defeat cargo's git canonicalization and allow us to
-# patch a crate using the same source.
-#rfb = { git = "http://github.com/oxidecomputer//rfb", rev = "" }
+#rfb = { path = "../rfb" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,8 +33,7 @@ panic = "abort"
 inherits = "dev"
 panic = "unwind"
 
-[patch."http://github.com/oxidecomputer/rfb"]
+#[patch."http://github.com/oxidecomputer/rfb"]
 # Double / here is intentional to defeat cargo's git canonicalization and allow us to
 # patch a crate using the same source.
-# Temp until https://github.com/oxidecomputer/rfb/pull/9 lands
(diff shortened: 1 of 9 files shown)


lib/propolis/src/instance.rs
```diff
@@ -47,7 +47,14 @@ impl Drop for Instance {
...
         // call Instance::destroy().
-        self.0.get_mut().unwrap().inventory.clear()
+        self.0.get_mut().unwrap().inventory.clear();
+
+        // If Instance::destroy() was not called prior to drop, let's cleanup
+        // the underlying VMM resource.
+        if let Some(machine) = self.0.get_mut().unwrap().machine.take() {
+            let hdl = machine.destroy();
+            let _ = hdl.destroy();
+        }
```
pfmooney, line 57: Now that vmm instances can be configured for [auto-destruct](https://www.illumos.org/issues/14817), I wonder we if should just allow that to be configured rather than the explicit destroy on drop (in case there are times when we want to bail out w/o destroying the kernel vmm state)

On the flip side, we could just make that change if/when it's needed

luqmana: For right now I think holding off on auto-destruct is preferable. This change should only result in VMs that are explicitly stopped being cleaned up. If we get into a failed state or crash, or otherwise not exit cleanly, the VmmHandle should persist which could help for investigation.

pfmooney: Works for me!

Cargo.toml
```diff
@@ -32,3 +32,9 @@ panic = "abort"
 [profile.phd]
 inherits = "dev"
 panic = "unwind"
+
+[patch."http://github.com/oxidecomputer/rfb"]
+# Double / here is intentional to defeat cargo's git canonicalization and allow us to
+# patch a crate using the same source.
+# Temp until https://github.com/oxidecomputer/rfb/pull/9 lands
+rfb = { git = "http://github.com/oxidecomputer//rfb", branch = "luqmana/vnc-stop" }
```
pfmooney, line 40: Perhaps a good time to make the cargo dep specify an explicit rev like crucible?

luqmana: Yup, will do once the RFB PR lands.

luqmana:
Can't reproduce test failure locally, will attempt with 1 more retry in CI.

PR #246, opened 2022-11-08
PHD: improve server termination logging; fix test VM name

gjcolombo:
Make the Propolis server wrapper's `Drop` impl log that the server is being terminated, and make it wait for the actual demise of the server process before continuing. (This may help either resolve or make actionable the failure we're currently seeing in CI in #244, where one of the tests' Propolis servers is logging `Error: Failed to start server: error creating server listener: Address already in use (os error 125)` on instantiation, and is good practice in any case.)

Also fix the naming of the VM generated in `instance_reset_returns_to_running_test`.

### DIFF
--- a/.github/buildomat/jobs/phd-run.sh
+++ b/.github/buildomat/jobs/phd-run.sh
@@ -54,6 +54,8 @@ ls $runner
 ls $artifacts
 ls $propolis
 
+# Disable errexit so that we still upload logs on failure
+set +e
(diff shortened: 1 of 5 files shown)


luqmana:
Ah thanks, was just giving my own "wait for kill" change a try but this looks good. Also make sense to just pull https://github.com/oxidecomputer/propolis/pull/244/commits/cae354c14484b8cdda2c08d7c6e93da386819ead into this PR instead of mine?

luqmana (approved):
Thanks!

PR #247, opened 2022-11-08
Point to specific rfb rev.

luqmana:
Latest rfb branch requires https://github.com/oxidecomputer/propolis/pull/244 which hasn't landed. Update Cargo.toml to point to previous version that works with current propolis master.


luqmana:
Closing in favour of https://github.com/oxidecomputer/propolis/pull/246 which also addresses this.

PR #249, opened 2022-11-10
PHD: check for leaked vmm handles and have test harness try to gracefully cleanup on exit.

### LINKED ISSUE #154: Integration tests leak bhyve VMs
On an illumos host:

1. `pfexec cargo test integration`
2. `ls /dev/vmm`

Expected: No VMs are present.
Observed: Three VMs outlive the integration test.

This is a test-only problem in the sense that in production, Propolis VMMs will run in zones that will get cleaned up on instance stop, which will reclaim those VMs' resources, and we don't do this in tests. This is a moderately serious issue for the tests, though, since this effectively limits the number of VMs they can create in a single test run (if they create too many VMs, the host will OOM and new VM creation calls will block waiting for more memory).

luqmana:
Fixes #154

### DIFF
--- a/phd-tests/framework/src/test_vm/mod.rs
+++ b/phd-tests/framework/src/test_vm/mod.rs
@@ -51,6 +51,7 @@ enum VmState {
 /// and [`TestVm::wait_to_boot`] calls so they can begin interacting with the
 /// serial console.
 pub struct TestVm {
+    id: Uuid,
     rt: tokio::runtime::Runtime,
     client: Client,
     server: server::PropolisServer,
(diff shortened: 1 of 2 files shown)


phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -520,3 +522,33 @@ impl TestVm {
...
+
+        match self.get().map(|r| r.instance.state) {
+            Ok(InstanceState::Destroyed) => {
+                // Instance already destroyed, nothing to do.
+            }
+            Ok(_) => {
+                // Instance is up, best-effort attempt to let it clean up gracefully.
+                info!("Cleaning up Test VM on drop");
+                let _ = self.stop();
+                let _ = self.wait_for_state(
```
gjcolombo, line 543: I might `warn!` if this fails, just to make it a little easier to tell which test/test VM is responsible for the possible leakage.

gjcolombo (approved):
One minor suggestion for a tracing statement to add, but otherwise looks great to me. Thanks for the fix!

PR #251, opened 2022-11-12
server: Wait for Instance to be dropped before setting api state to Destroyed.

luqmana:
Previously, there was a bit of a race condition where the completion of the worker thread (and the subsequent dropping of WorkerStateInner) would set the API instance state to Destroyed, but the actual Instance itself might not've been destroyed yet.

Some PHD tests would thus sometimes leave left over VMM handles because the test framework immediately killed the propolis-server process when it saw the instance state get to Destroyed.

Also bail early if the VNC port is already used.

### DIFF
--- a/bin/propolis-server/src/lib/vm.rs
+++ b/bin/propolis-server/src/lib/vm.rs
@@ -163,6 +163,11 @@ pub(crate) struct VmObjects {
     /// A notification receiver to which the state worker publishes the most
     /// recent instance state and state generation.
     monitor_rx: tokio::sync::watch::Receiver<ApiMonitoredState>,
+
+    /// The sender side of the watcher that records the instance's externally
+    /// visible state, i.e., the state returned when a client invokes the
+    /// server's `get` API.
+    api_state: Arc<tokio::sync::watch::Sender<ApiMonitoredState>>,
 }
 
 /// A message sent from a live migration destination task to update the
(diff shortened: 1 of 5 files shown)


bin/propolis-server/src/lib/vm.rs
```diff
@@ -578,6 +571,60 @@ impl VmController {
...
+        // or destroyed. Normally, the existence of this structure implies the
+        // instance of at least one receiver, but at this point everything is
+        // being dropped, so this call to `send` is not safe to unwrap.
+        let gen = api_state.borrow().gen + 1;
+        let _ = api_state.send(ApiMonitoredState {
+            gen,
+            state: if failed {
+                ApiInstanceState::Failed
+            } else {
+                ApiInstanceState::Destroyed
```
gjcolombo, line 623: I could just be missing something, but I'm not sure this totally closes the race--we're publishing `Destroyed` here, but IIUC `self` is still in scope, so the instance still hasn't quite been dropped yet. (We'll also publish `Destroyed` even if we logged an error on line 608. Today I don't think this is much of a problem, since no one in the server module currently uses `VmController` references outside of the server's state lock, but I wonder if we should allow that--it feels a little weird to have this routine's correctness depend on the caller's synchronization scheme.)

Would it make sense to explicitly `destroy` the instance in this path instead (or turn `VmController::instance` into an `Mutex<Option<Arc<Instance>>>` to force everyone to reckon with the possibility that it's been destroyed)?

luqmana: `self` is out of scope at this point as `Arc::try_unwrap` takes it by value. (Though the `Err` case does mean there are other refs out there)

I went back and forth on just calling `Instance::destroy` manually but decided not to at least for now to not paper over leftover references. Sticking the instance behind an Option also meant a cascade of other changes which I suppose isn't the worst.

I'll go ponder on some approaches about how to lay things out so everything doesn't get too tangled.

PR #262, opened 2022-12-02
Update crucible version

sunshowers:
This is PR 2 of N to make sure that omicron doesn't get two separate
dependencies, one for progenitor with main and one for progenitor
without main.

For PR 1, see https://github.com/oxidecomputer/crucible/pull/552.

Without this fix, omicron ends up with two separate versions of
crucible, which causes build failures in omicron.

The update caused some breakage from
https://github.com/oxidecomputer/crucible/pull/516, which I've tried to
address (cc @leftwo to confirm that my changes look good).

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -292,13 +292,12 @@ impl<'a> MachineInitializer<'a> {
                 })?;
             let StorageBackendInstance { be: backend, child, crucible } =
                 match &backend_spec.kind {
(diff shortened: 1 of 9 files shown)


gjcolombo (approved):

sunshowers:
Thanks!

gjcolombo:
My pleasure--thanks for the changes! (I've made basically the same edits locally for some other work I'm doing and can confirm that all this seems to work fine with the Crucible integration tests I'm writing.)

PR #266, opened 2022-12-07
Store serial console history from instance boot.

lifning:
This change moves functionality from omicron's sled-agent into propolis-server to allow both querying serial console output history by byte range in GET requests, as well as allowing websocket clients to receive a backlog of stored data upon connecting, starting from any offset relative to either the first or the most recently outputted byte.

(This works with the existing implementation in omicron/sled-agent; the merge to remove its then-redundant copy should follow this one - https://github.com/oxidecomputer/omicron/pull/2022)

((~~[This failure in my unit test](https://github.com/oxidecomputer/propolis/actions/runs/3637686749/jobs/6139336587) seems to only happen on nightly - will try to figure out why tomorrow...~~ Fixed! Only was happening on nightlies new enough to be broken on Illumos per the rustix/nix `isatty` issue we ran into the other day. Seems to have originated from https://github.com/rust-lang/rust/commit/a1bf25e2bdbebad2f1f42118ba1d2c4a7ae4f7b0 changing how `VecDeque::drain` handles slices, if anyone's curious!))

### DIFF
--- a/bin/propolis-server/src/lib/server.rs
+++ b/bin/propolis-server/src/lib/server.rs
@@ -8,11 +8,13 @@
 use std::sync::Arc;
 use std::{collections::BTreeMap, net::SocketAddr};
 
+use crate::serial::history_buffer::SerialHistoryOffset;
 use dropshot::{
     channel, endpoint, ApiDescription, HttpError, HttpResponseCreated,
-    HttpResponseOk, HttpResponseUpdatedNoContent, Path, RequestContext,
+    HttpResponseOk, HttpResponseUpdatedNoContent, Path, Query, RequestContext,
     TypedBody, WebsocketConnection,
 };
+use futures::SinkExt;
 use hyper::{Body, Response};
 use oximeter::types::ProducerRegistry;
 use propolis_client::{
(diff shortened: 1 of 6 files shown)


bin/propolis-server/src/lib/server.rs
```diff
@@ -448,6 +450,30 @@ async fn instance_ensure_common(
...
+        let task = tokio::spawn(async move {
+            if let Err(e) = super::serial::instance_serial_task(
+                websocks_recv,
+                close_recv,
+                serial,
+                err_log.clone(),
+            )
+            .await
+            {
+                error!(err_log, "Spawning serial task failed: {}", e);
```
gjcolombo, line 469: extreme nit: I might drop the word "spawning" here; on my reading we've technically spawned the task but it failed for some other reason

lib/propolis-client/src/handmade/api.rs
```diff
@@ -186,6 +186,49 @@ pub struct Instance {
     pub nics: Vec<NetworkInterface>,
 }
 
+/// Request a specific range of an Instance's serial console output history.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq)]
+pub struct InstanceSerialConsoleHistoryRequest {
+    /// Character index in the serial buffer from which to read, counting the bytes output since
+    /// instance start. If this is not provided, `most_recent` must be provided, and if this *is*
```
gjcolombo, line 193: Can/should this be made into an option-of-enum to avoid the caller-supplied-both/caller-supplied-neither cases entirely? (Or will that run into issues with expressing this type in the generated API spec? ISTR hitting a problem along those lines once; it was with a different type but it made clear to me that I don't know all the rules here.)

gjcolombo (approved):
Thanks for putting this together! Everything LGTM; I had just a couple of small questions for which I've left comments.

lib/propolis-client/src/handmade/api.rs
lifning: i originally did that in my first pass of this feature in omicron, but i believe it shook out that rusty tagged enums is awkward for non-rust openapi consumers

gjcolombo: Makes sense, thanks!

PR #271, opened 2022-12-13
Wait until instance drop to publish InstanceState::Destroyed

### LINKED ISSUE #270: server: users of `instance_get` can't guarantee that an instance's bhyve VM is destroyed
When the server creates a VM controller, the controller creates a task that sends "instance stopped" notifications back to the upper layers of the server: https://github.com/oxidecomputer/propolis/blob/5ea6e6fd86bedce9a6a18f1f1b339550cbba5f60/bin/propolis-server/src/lib/vm.rs#L572-L584

This unwinds back to a call to `ServiceProviders::stop`, which tears down all of the upper-layer objects that the server manages: https://github.com/oxidecomputer/propolis/blob/5ea6e6fd86bedce9a6a18f1f1b339550cbba5f60/bin/propolis-server/src/lib/server.rs#L182-L192

The call to `take_controller` on line 186 changes the server's VM state to a "Destroyed" variant as follows: https://github.com/oxidecomputer/propolis/blob/5ea6e6fd86bedce9a6a18f1f1b339550cbba5f60/bin/propolis-server/src/lib/server.rs#L121-L143

Notice that lines 125 and 137 immediately change the server's externally-presented VM state to `InstanceState::Destroyed` even though the `Arc<VmController>` is still active. This creates a window in which `instance_get` will return `Destroyed` even though the underlying controller (and the instance and bhyve VMM to which it holds references) still has at least one active reference.

This currently causes flakiness in CI: PHD wants to make sure a test VM's bhyve handles are torn down before destroying any Propolis servers; to do that, it waits for Propolis to return the `Destroyed` state from a `get` call; but the race above means that `Destroyed` doesn't imply that the underlying VM is gone.

gjcolombo:
Ensure that instance get and monitor operations don't return an instance state of "Destroyed" until the relevant instance has been dropped and its bhyve VM handles are closed. To do that:

- Rearrange the VM controller so that it is the sole owner of an `Instance` that users can borrow, instead of allowing users to get their own references directly to the instance
- Publish the "Destroyed" state transition during `VmController` drop (when the instance is really going away) instead of when the state worker exits
- Fix `VmControllerState::take_controller` so that state queries for a stopped instance return what the VM controller most recently published instead of always returning Destroyed

Tested by running ~15 PHD iterations locally and verifying that no VMM handles were leaked.

Fixes #270 (hopefully).

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -65,7 +65,7 @@ pub fn build_instance(
     spec: &InstanceSpec,
     use_reservoir: bool,
     _log: slog::Logger,
-) -> Result<Arc<Instance>> {
+) -> Result<Instance> {
     let (lowmem, highmem) = get_spec_guest_ram_limits(spec);
     let create_opts = propolis::vmm::CreateOpts {
         force: true,
@@ -87,7 +87,7 @@ pub fn build_instance(
         builder = builder.add_mem_region(0x1_0000_0000, highmem, "highmem")?;
     }
 
-    Ok(Arc::new(Instance::create(builder.finalize()?)))
+    Ok(Instance::create(builder.finalize()?))
 }
 
 pub struct RegisteredChipset(Arc<I440Fx>, EntityID);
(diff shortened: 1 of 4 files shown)


luqmana (approved):
Thanks!

PR #274, opened 2022-12-14
SoftNpu early return on empty guest virtio buffers, bump dlpi crate, fix duplicate crate import issue

rcgoodfellow:
### SoftNpu early return on empty buffers
Break out of the guest packet processing loop early when there are no bytes to read. This has a non-trivial impact on latency reduction.

### Bump dlpi
Bump the dlpi dependency. The latest version is less chatty with logging and the bug previously referenced is no longer an issue since we are not using the async API exported by the dlpi create.

### Fix import issue
Set `publish = false` on the header check libraries. When importing propolis libraries, because both `viona-api` and `bhyve-api` have a sub-crate called `header-check`, cargo starts screaming. This is working around this [issue](https://github.com/rust-lang/cargo/issues/10752) with this [fix](https://github.com/rust-lang/cargo/pull/10767)

### Other
The changes in propolis-standalone are in response to clippy complaints.

### DIFF
--- a/crates/bhyve-api/header-check/Cargo.toml
+++ b/crates/bhyve-api/header-check/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.0.0"
 authors = ["Patrick Mooney <pmooney@oxide.computer>"]
 license = "MPL-2.0"
 build = "build.rs"
+publish = false
 
 [dependencies]
 bhyve_api = { path = ".." }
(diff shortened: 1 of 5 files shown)


crates/bhyve-api/header-check/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.0.0"
 authors = ["Patrick Mooney <pmooney@oxide.computer>"]
 license = "MPL-2.0"
 build = "build.rs"
+publish = false
```
pfmooney, line 7: Oh nice.  I didn't know about that

pfmooney (approved):

PR #279, opened 2022-12-20
"--mock" mode for propolis-server

lifning:
this builds on @bnaecker's [mock-server](https://github.com/oxidecomputer/propolis/tree/mock-server) branch from back in april, bringing it up to date with the changes to the API since then (e.g. no more UUID path params), and adding the serial console websocket (and serial console history-retrieving) API endpoints, backed with similar output as was used in the previous sled-agent-sim implementation.

### DIFF
--- a/bin/propolis-server/src/lib/lib.rs
+++ b/bin/propolis-server/src/lib/lib.rs
@@ -1,11 +1,21 @@
+// always present
 pub mod config;
-mod initializer;
-mod migrate;
 pub mod mock_server;
 mod serial;
-pub mod server;
+#[cfg_attr(feature = "mock-only", allow(unused))]
 mod spec;
+
+#[cfg(not(feature = "mock-only"))]
+mod initializer;
+#[cfg(not(feature = "mock-only"))]
+mod migrate;
+#[cfg(not(feature = "mock-only"))]
+pub mod server;
+#[cfg(not(feature = "mock-only"))]
 mod stats;
+#[cfg(not(feature = "mock-only"))]
 mod vcpu_tasks;
+#[cfg(not(feature = "mock-only"))]
 mod vm;
+#[cfg(not(feature = "mock-only"))]
 pub mod vnc;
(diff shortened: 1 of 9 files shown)


bin/propolis-server/src/lib/lib.rs
```diff
@@ -1,10 +1,21 @@
...
 pub mod server;
-mod spec;
+#[cfg(not(feature = "mock-only"))]
 mod stats;
+#[cfg(not(feature = "mock-only"))]
 mod vcpu_tasks;
+#[cfg(not(feature = "mock-only"))]
 mod vm;
+#[cfg(not(feature = "mock-only"))]
 pub mod vnc;
```
luqmana, line 21: We already have `cfg-if` in tree so might as well use it here to get rid of all the duplicate cfg's here.

bin/propolis-server/src/lib/mock_server.rs
```diff
@@ -0,0 +1,550 @@
...
+        .await
+        .map_err(|e| HttpError::for_bad_request(None, e.to_string()))?;
+
+    Ok(HttpResponseOk(api::InstanceSerialConsoleHistoryResponse {
+        data,
+        last_byte_offset: end as u64,
+    }))
+}
+
+// (ahem) mock *thou* art.
```
luqmana, line 400: 😂

bin/propolis-server/src/lib/lib.rs
```diff
@@ -1,10 +1,21 @@
+// always present
 pub mod config;
+pub mod mock_server;
```
luqmana, line 3: Should this also be behind the `mock-only` flag?

luqmana (commented):
Thanks for reviving this! Left a couple small comments. Only worry is how hard will it be to maintain as the non-mock server changes (not a blocker given the same could be said about the sled-agent-sim too).

bin/propolis-server/src/lib/lib.rs
lifning: i still wanted it being pulled in for unit tests, though i suppose i can do `#[cfg(any(test, feature = "mock-only"))]` instead

luqmana (approved):

PR #280, opened 2022-12-20
Use workspace dependencies

gjcolombo:
Switch to workspace dependencies for all standard build dependencies. (I left dev-only and build-only dependencies alone but confess I don't have a strong reason for having done this; happy to use workspace deps for these too if reviewers prefer it.) Bump MSRV to 1.64, where workspace dependencies were introduced.

Tested with cargo test/PHD.


luqmana:
I'd say convert the build and dev deps too. Having one place to keep versions in sync is really useful.

Issue #284, opened 2022-12-23
Updates to a VM controller's `lifecycle_stage` are race-prone


gjcolombo:
Passing `Run` to `VmController::put_state` is supposed to be a no-op if the instance is already active: https://github.com/oxidecomputer/propolis/blob/584eac234598b90e25f315a73cef1271e4eac398/bin/propolis-server/src/lib/vm.rs#L764-L790

The current synchronization between the controller and state worker creates a small window where the VM is already running but the lifecycle stage is not updated yet, allowing a `Start` request to be queued to (and--incorrectly--processed by) the state worker even though the VM is already running. This part of the state worker logic needs to be reworked a bit and (hopefully) simplified.

PR #285, opened 2022-12-23
Automatically start VMs after a successful migration in

### LINKED ISSUE #250: migrated VMs should resume automatically after successful migration
Currently, after a live migration completes successfully, the VM on the target host does not start running automatically. Instead, the VM must be set to running through the API. (The VM resume step of live migration is described in [RFD 71](https://github.com/oxidecomputer/rfd/tree/0071/rfd/0071#architectural-state-push).)

For example, this target reports that the migration has completed successfully:
```
Nov 11 17:01:24.064 INFO Destination Migration Successful, component: vm_controller           
Nov 11 17:01:24.064 INFO Migration source task exited: Ok(()), component: vm_state_worker   
```

While the VM has been migrated, it isn't yet running on the target:
```
# source
jordan@rodman:~/src/forks/propolis$ ./target/debug/propolis-cli -s 127.0.0.1 -p 8000 monitor                                                                                                                      
InstanceState: Destroyed  

# target
jordan@rodman:~/src/forks/propolis$ ./target/debug/propolis-cli -s 127.0.0.1 -p 8500 monitor                                                                                                                      
InstanceState: Migrating 
```

To get the VM running, one must set its state to run:
```
jordan@rodman:~/src/forks/propolis$ ./target/debug/propolis-cli -s 127.0.0.1 -p 8500 state run
Nov 11 17:01:39.097 INFO PUT request to http://127.0.0.1:8500/instance/state, propolis_client address: 127.0.0.1:8500

jordan@rodman:~/src/forks/propolis$ ./target/debug/propolis-cli -s 127.0.0.1 -p 8500 monitor
InstanceState: Running     
```

It seems a bit unintuitive that the resume step would need to be done explicitly. Further, as @gjcolombo pointed out, callers of the migration endpoint already have to deal with the fact that the VM can pause and resume without their knowledge, and have to handle this regardless; thus, if the fact that the VM resumes automatically on the target turns out to be a problem, it's likely something we're going to have to deal with on the source side as well.

gjcolombo:
Start instances automatically upon successfully migrating into them. Fix one small race that could cause a client's requests to run a VM to fail even though it could observe that a migration in had successfully completed. (There are still some other sore spots in this state machine that aren't fixed in this PR; those are now tracked by #284.)

Fixes #250.

### DIFF
--- a/bin/propolis-server/src/lib/vm/state_driver.rs
+++ b/bin/propolis-server/src/lib/vm/state_driver.rs
@@ -152,14 +152,15 @@ where
                     command_rx,
                 ) {
                     Ok(_) => {
-                        // Publish the "migrated" state before signaling that
-                        // migration is finished. Without this, there's a race
-                        // where a client can observe that migration is done
-                        // (because the migration state was written), but
-                        // further attempts to change the VM's state will fail
-                        // (because the lifecycle stage is what determines
-                        // whether a migration is "done" for the purposes of
-                        // accepting or rejecting a request to run the VM).
+                        // Set the lifecycle stage to "migrated" before
+                        // revealing to external clients (who might query for
+                        // the migration's status) that the migration is
+                        // finished (by setting the migration state). This
+                        // ensures that any client who observes the "finished"
+                        // migration state can change the VM's state further (as
+                        // opposed to seeing that the migration has finished but
+                        // having a call to e.g. stop the VM be rejected because
+                        // the lifecycle stage is wrong).
                         self.controller.set_lifecycle_stage(
                             LifecycleStage::NotStarted(StartupStage::Migrated),
                         );
@@ -168,9 +169,8 @@ where
                             ApiMigrationState::Finish,
                         );
 
-                        // The migration completed successfully, so immediately
-                        // push the VM into a 'running' state without waiting
-                        // for further input (to minimize blackout time).
+                        // Resume the VM immediately without waiting for a
+                        // request to enter the "running" state.
                         //
                         // TODO(#209) Transition to a "failed" state instead of
                         // expecting.
--- a/bin/propolis-server/src/lib/vm/state_driver.rs
+++ b/bin/propolis-server/src/lib/vm/state_driver.rs
@@ -509,7 +509,6 @@ mod tests {
         (vm_ctrl, vcpu_ctrl)
     }
 
-    /* TODO(#250)
     fn make_migration_target_mocks(
         migration_id: Uuid,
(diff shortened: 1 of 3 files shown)


bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -145,46 +145,53 @@ where
...
-                        Err(_) => LifecycleStage::NotStarted(
+                match self.migrate_as_target(
+                    migration_id,
+                    task,
+                    start_tx,
+                    command_rx,
+                ) {
+                    Ok(_) => {
+                        // Publish the "migrated" state before signaling that
+                        // migration is finished. Without this, there's a race
```
jordanhendricks, line 156: I'm having a bit of trouble parsing this sentence. It might help to spell out more concretely what "signaling that migration is finished" means here.

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -145,46 +145,53 @@ where
...
+                            LifecycleStage::NotStarted(StartupStage::Migrated),
+                        );
+                        self.controller.set_migration_state(
+                            migration_id,
+                            ApiMigrationState::Finish,
+                        );
+
+                        // The migration completed successfully, so immediately
+                        // push the VM into a 'running' state without waiting
+                        // for further input (to minimize blackout time).
```
jordanhendricks, line 173: what input would we be waiting on?

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -253,6 +260,25 @@ where
         }
     }
 
+    fn start_vm(&mut self, reset_required: bool) -> anyhow::Result<()> {
```
jordanhendricks, line 263: pulling this out into a function looks much cleaner than before -- thanks.

gjcolombo:
Thanks for the review! I've cleaned up the comments in 95d0b08.

> I am still wrapping my head around the ordering of when we update the lifecycle stage versus the API stage and preventing races. The existing comments are helpful, but maybe as more of the work you're doing lands we could add a block comment or docs to help describe the interactions here from a higher level.

Will do--I'd definitely like to get a theory statement into this module. I've been broadly planning to write one after making whatever changes we end up needing to support Nexus-driven live migrations. (I also think the whole state driver could be considerably simpler than it is now; I'll have another PR up shortly that takes a step or two in that direction.)

jordanhendricks:
The updated comments look great -- thanks!

PR #286, opened 2022-12-27
Migrate transmit/receive FIFOs for UART device

### LINKED ISSUE #282: transmit/receive FIFOs in UART device are not migrated
## Summary
While testing out inter-machine live migrations, I found that I was reliably not able to attach to the serial console on the destination VM. The VM otherwise appeared to have migrated successfully and started running, but the `serial` CLI command would hang with not output.

Ultimately, this is because the `rx_fifo` and `tx_fifo` buffers internal to the UART emulation are not currently migrated. I've confirmed that migrating the FIFOs fixes this issue and will post a PR.

## Observations while debugging
While this bug and its fix are relatively straightforward, I thought this failure mode was sort of interesting. So here's a summary of what I observed while debugging this, in case it's useful in the future:
- On the source, I could attach to the serial console before migrating.
- On the destination, after rebooting the migrated VM, I could attach to the console.
- In the propolis server serial code `select` loop that manages client connections and reading/writing data to the UART, no data was coming out of the UART from the guest. Data from the client connection was being written successfully to the UART.
- The VM appeared to be running -- I saw constant VM exits on all vcpus (BOGUS exit code), and INOUT exits on a single vcpu when typing into stdin of the `serial` command.
- Deeper in the stack, I saw that user data input to the `serial` command was making it to the UART device emulation, but nothing was coming out of the device from the guest in the emulation code.

The critical observations were these:
- The `thre_intr` (Transmitter Holding Register Empty Interrupt) value in the UART emulation was `false` on the destination. This interrupt is used to notify the host when it can write more data to the UART.
- There was nothing in the `tx_fifo` on the destination, despite the `thre_intr` state indicating that the host should not write more data. This is an invalid state.

These observations led me to find that there `tx_fifo` (and the `rx_fifo`) buffers are not migrated at all. Looking more concretely at what happened, I see that I was migrating the VM while it was still writing to the UART during boot. Thus there was data written to the `tx_fifo`, and the `thre_intr` field was cleared to false when the migration began. The interrupt data was migrated, but not the data itself, so the destination host gets stuck unless it's rebooted. This also explains why others had not seen this issue --- the console will work if the guest has been booted on the source for awhile and is no longer writing a lot of data to the UART.

jordanhendricks:
Closes #282 

Testing: I verified that the serial console now works reliably across live migrations.

### DIFF
--- a/lib/propolis/src/hw/uart/uart16550.rs
+++ b/lib/propolis/src/hw/uart/uart16550.rs
@@ -1,6 +1,22 @@
+use bits::*;
+use serde::{Deserialize, Serialize};
 use std::collections::VecDeque;
 
-use bits::*;
+/*
+ * 16550 UART
+ *
+ * Host -> Device Data Path:
+ * The host writes data to the UART via its Transmitter Holding Register (THR),
+ * which is backed by tx_fifo. After this data is received out of the THR, the
+ * UART will raise the Transmitter Holding Register Empty interrupt, which
+ * notifies the host that it can write more data.
+ */
+
+#[usdt::provider(provider = "propolis")]
+mod probes {
+    fn uart_reg_read(offset: u8, is_dlab: u8, val: u8) {}
+    fn uart_reg_write(offset: u8, is_dlab: u8, data: u8) {}
+}
 
 pub struct Uart {
     reg_intr_enable: u8,
@@ -15,7 +31,7 @@ pub struct Uart {
     reg_div_low: u8,
     reg_div_high: u8,
 
-    thre_intr: bool,
+    thre_intr: bool, // Transmitter Holding Register Empty interrupt
     intr_pin: bool,
 
     rx_fifo: Fifo,
@@ -44,7 +60,7 @@ impl Uart {
     }
     /// Read UART register
     pub fn reg_read(&mut self, offset: u8) -> u8 {
-        match (offset, self.is_dlab()) {
+        let val = match (offset, self.is_dlab()) {
             (REG_RHR, false) => {
                 if let Some(d) = self.rx_fifo.read() {
                     self.update_dr();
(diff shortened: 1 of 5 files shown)


lib/propolis/src/hw/uart/uart16550.rs
```diff
@@ -76,10 +92,22 @@ impl Uart {
             _ => {
                 panic!();
             }
-        }
+        };
+
+        probes::uart_reg_read!(|| {
+            let is_dlab = if self.is_dlab() { 1 } else { 0 };
```
luqmana, line 98: You can just cast a bool with `as`

lib/propolis/src/hw/uart/uart16550.rs
```diff
@@ -76,10 +92,22 @@ impl Uart {
...
+
+            (offset, is_dlab, val)
+        });
+
+        val
     }
     /// Write UART register
     pub fn reg_write(&mut self, offset: u8, data: u8) {
+        probes::uart_reg_write!(|| {
+            let is_dlab = if self.is_dlab() { 1 } else { 0 };
```
luqmana, line 108: Here too for the bool cast

lib/propolis/src/hw/uart/uart16550.rs
```diff
@@ -76,10 +92,22 @@ impl Uart {
             _ => {
                 panic!();
```
luqmana, line 93: Worth refactoring this a little bit so we also get the probe in this branch too? Ideally we don't panic but if we do, would be nice to know what caused it.

luqmana (commented):
Left a few comments. Thanks for finding and fixing that!

lib/propolis/src/hw/uart/uart16550.rs
jordanhendricks: Sure. Rather than refactoring to an option or something, I added the probe to that branch too -- assuming whoever is debugging is already tracing this probe, I figured it would be better to have a dummy value for the read byte there.

lib/propolis/src/hw/uart/uart16550.rs
jordanhendricks: done

lib/propolis/src/hw/uart/uart16550.rs
jordanhendricks: done

luqmana (approved):

PR #294, opened 2023-01-10
rearrange serial task probes to work around early exit

### LINKED ISSUE #292: propolis-server release builds' serial tasks exit unexpectedly
1. Check out 5c36674218c7a0a4551063d053462481f29e3ac1
2. `cargo build --release --bin propolis-server`
3. Start a PHD test run that uses this binary as the propolis-server binary

Expected: test passes
Observed: no serial console connection is established. Adding log lines to the beginning and end of `propolis_server::serial::instance_serial_task` shows that it returns almost immediately after the VM starts.

Additional data points:
- Test binaries built with `rustc 1.66.0 (69f9c33d7 2022-12-12)`, but this also repros under 1.65.0
- Repros when accessing the serial console directly via `propolis-cli`, through Omicron and the `oxide` CLI, and in PHD tests
- Doesn't repro with a debug `propolis-server`
- Modifying this bit of code: https://github.com/oxidecomputer/propolis/blob/5c36674218c7a0a4551063d053462481f29e3ac1/bin/propolis-server/src/lib/serial/mod.rs#L185-L197 in either of two ways suppresses the problem:
  - Adding a `slog::info!` logging statement below line 189 causes the issue not to repro (the statement need not refer to any variables)
  - Commenting out the USDT probe on line 189 causes the issue not to repro

This is especially impactful in Omicron deployments since they use release builds of propolis-server. (Our CI tests use debug builds, so we don't observe this issue in those runs, which is something else worth following up on.)

gjcolombo:
There is something magical about the release-build codegen of `propolis_server::serial::instance_serial_task` that causes the task to exit early, evidently via the "UART read channel was closed/returned no data from a successful read" clause of its main `select!` statement. This doesn't happen in dev builds, and the code works as intended if I remove the `serial_uart_read` probe from the relevant match arm or if I add other statements (e.g. logging statements) to that arm.

In the fullness of time, we should dig into exactly what's happening here. Since this only happens in release builds and seems to require the presence of the `serial_uart_read` probe (and *only* that probe) in the relevant match arm, this smells to me like some kind of interesting code generation and/or USDT corner case. In the immediate term, though, this breaks the serial console for any users of propolis-server release builds, which includes Omicron. Work around the problem by moving the relevant probe out of the match arm.

Also add a couple of logging statements (on entry/exit from the serial task) that I found useful while investigating this problem.

Tested: 
- Local PHD run that specifically used a release `propolis-server`
- Deployed a binary with the fix to a local Omicron install and verified that the serial console works again

Fixes #292, though I will file another (lower-priority) issue to track further investigation of what happened here.

### DIFF
--- a/bin/propolis-server/src/lib/serial/mod.rs
+++ b/bin/propolis-server/src/lib/serial/mod.rs
@@ -63,6 +63,7 @@ pub async fn instance_serial_task(
     serial: Arc<Serial<LpcUart>>,
     log: Logger,
 ) -> Result<(), SerialTaskError> {
+    info!(log, "Entered serial task");
     let mut output = [0u8; 1024];
     let mut cur_output: Option<Range<usize>> = None;
     let mut cur_input: Option<(Vec<u8>, usize)> = None;
@@ -160,13 +161,12 @@ pub async fn instance_serial_task(
 
             // Write bytes into the UART from the WS
             written = uart_write => {
+                probes::serial_uart_write!(|| { written.unwrap_or(0) });
                 match written {
                     Some(0) | None => {
-                        probes::serial_uart_write!(|| { 0 });
                         break;
                     }
                     Some(n) => {
-                        probes::serial_uart_write!(|| { n });
                         let (data, consumed) = cur_input.as_mut().unwrap();
                         *consumed += n;
                         if *consumed == data.len() {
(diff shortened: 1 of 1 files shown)


lifning (approved):
LGTM, i guess we might want to turn this into an issue in usdt with the commit here as context for whenever we have cycles to dig into this

gjcolombo:
I filed [usdt#86](https://github.com/oxidecomputer/usdt/issues/86) to track the follow-up investigation.

PR #303, opened 2023-01-23
Crucible deps

pfmooney:
With Cargo.lock committed now, we can switch the Crucible dependencies to track `main` .


leftwo (commented):
If propolis goes to "main" but Omicron still has a specific rev of crucible it uses, could we get into trouble?

pfmooney:
Going to keep this as-is for now, after discussing with Alan

PR #304, opened 2023-01-24
Use libnet crate for grabbing link info.

luqmana:
Avoids just shelling out to `dladm` and some vnic-specific methods to work with other link classes.


lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -56,9 +57,27 @@ impl PciVirtioViona {
...
+        let mtu = if let Some(mtu) = info.mtu {
+            Some(u16::try_from(mtu).map_err(|_| {
+                io::Error::new(
+                    io::ErrorKind::Other,
+                    format!("MTU too large: {mtu}"),
+                )
+            })?)
+        } else {
+            None
+        };
```
pfmooney, line 78: If we want to keep propolis able to build (albeit useless binaries) on non-illumos platforms, perhaps we can wrap the libnet stuff pulled out into a function in `cfg` blocks?

pfmooney (commented):
Thanks for getting this cleaned up.

pfmooney:
Since @rmustacc pointed out that the `libnet` crate is consuming a bunch of Private OS interfaces in order to expose its APIs, @luqmana and I chatted about exploring options which stick to public Committed interfaces for that data.  Either tweaking the info-gathering calls to `dladm`, and/or expanding `libdladm` upstream so it makes available the things we need.

Issue #308, opened 2023-02-07
thread 'vcpu-1' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidCompQueue(0)'


jmpesp:
saw the following:

```
thread 'vcpu-1' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidCompQueue(0)', lib/propolis/src/hw/nvme/mod.rs:306:44
stack backtrace:
   0:          0x1d01c4c - std::backtrace_rs::backtrace::libunwind::trace::h9b534fc6095f520f
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:          0x1d01c4c - std::backtrace_rs::backtrace::trace_unsynchronized::hfcd8a890f991a10d
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:          0x1d01c4c - std::sys_common::backtrace::_print_fmt::hdf432e5f0940fa9c
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/sys_common/backtrace.rs:65:5
   3:          0x1d01c4c - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h5693a803500ade80
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/sys_common/backtrace.rs:44:22
   4:          0x1d5937a - core::fmt::write::h94f582bdfeeeb1a6
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/fmt/mod.rs:1209:17
   5:          0x1cf2a34 - std::io::Write::write_fmt::hd3f00507505f204d
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/io/mod.rs:1682:15
   6:          0x1d01a20 - std::sys_common::backtrace::_print::h21d778336866aef0
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/sys_common/backtrace.rs:47:5
   7:          0x1d01a20 - std::sys_common::backtrace::print::h8238ed32ff302292
[...]

PR #309, opened 2023-02-07
NVMe: don't panic if guests tries a doorbell write for admin queues while controller is disabled

luqmana:
Addresses panic in #308 (should leave issue open to further investigate if there's a bug on the propolis side).

As was mentioned on the bug, we shouldn't panic just because the guest did a register write at an unexpected time. Updated the code here to spit out a warning instead.

Also removed another possible source of unintended panics. The queue doorbell registers are defined in the spec as 32-bit but the top 16-bits are reserved and should be 0. But there's nothing stopping a bad/malicious guest from doing otherwise. I had previously used `try_into().unwrap()` to go from a `u32` to a `u16` here which would've caused a panic with that kind of bad input. Change here is to do a simple `as` cast which truncates to the lower 16-bits instead.

### DIFF
--- a/lib/propolis/src/hw/nvme/mod.rs
+++ b/lib/propolis/src/hw/nvme/mod.rs
@@ -763,6 +763,17 @@ impl PciNvme {
                 // 32-bit register but ignore reserved top 16-bits
                 let val = wo.read_u32() as u16;
                 let state = self.state.lock().unwrap();
+
+                if !state.ctrl.cc.enabled() {
+                    slog::warn!(
+                        self.log,
+                        "Doorbell write while controller is disabled"
+                    );
+                    return Err(NvmeError::InvalidSubQueue(
+                        queue::ADMIN_QUEUE_ID,
+                    ));
+                }
+
                 let admin_sq = state.get_admin_sq();
                 admin_sq.notify_tail(val)?;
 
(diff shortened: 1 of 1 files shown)


lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -760,17 +760,41 @@ impl PciNvme {
...
+                    slog::warn!(
+                        self.log,
+                        "Doorbell write while controller is disabled"
+                    );
+                    return Err(NvmeError::InvalidSubQueue(
+                        queue::ADMIN_QUEUE_ID,
+                    ));
+                }
+
                 let admin_sq = state.get_admin_sq();
```
pfmooney, line 777: What about making `get_admin_sq` and `get_admin_cq` return an `Option<>` instead of checking the state bits manually?

luqmana: I'd like to keep the controller-enabled check cause those are slightly different conditions.

But good point on `get_admin_{cq,sq}`. In theory we shouldn't encounter a doorbell reg write while the controller is active but the admin queues haven't been created. But we shouldn't panic either. I switched them to just return a `Result` like the `get_{cq,sq}` they're wrapping.

pfmooney (approved):

PR #311, opened 2023-02-08
Replace migration transport layer with websockets

lifning:
...and (or to be more precise, *because of*) update to Dropshot 0.9.x to match Omicron main.

In Dropshot 0.9.x, an owned copy of the hyper::Request is no longer available to endpoint handlers through the RequestContext. This poses a problem for the custom wire protocol Propolis used for migration data because we can no longer negotiate a HTTP upgrade.

So instead, let's switch out the custom data framing protocol for the one provided in Dropshot `#[channel]` (that being the Websockets support I added a while back) while keeping most of the data format the same. In essence, this means just shoving it into a
`tungstenite::Message::Binary(vec![/* everything but the length, which is now handled for us */])`. The protocol version "negotiation" that was once done in the HTTP upgrade header is now done by exchanging `tungstenite::Message::Text(/* the version string */)` before handing the channel off.

I've also taken a couple small liberties as a matter of preference, since we're breaking compatibility anyhow:
- the API endpoints are now `/instance/migrate/{migration_id}/[start|status]`, which I believe is more idiomatic REST than having the `migration_id` be in the query parameters of one and the request body of the other, especially when it's the *only* parameter in both cases.
- the `MessageType` tag byte in the websocket message is at the end of the message, as it means the `u64`s we read from the beginning of the `Vec` are more likely to be nicely aligned in memory, in case that ever matters (i.e. we someday end up running on some platform where that's necessary).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,7 +106,7 @@ dependencies = [
 [[package]]
 name = "api_identity"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#3606be36c0a35c0bb111fbf813f503497601ba2f"
+source = "git+https://github.com/oxidecomputer/omicron?branch=main#bab91f6b5cba896ff391995d4b770b0ccc18e5c2"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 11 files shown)


gjcolombo (approved):
This looks really nice and simplifies a few things--thanks for the change!

PR #314, opened 2023-02-14
Util shuffle

pfmooney:
Dedicate a space to rarely used development utilities, and reshape some of the bhyve-api/viona-api crates to make their use easier in said utilities.


Cargo.toml
```diff
@@ -94,6 +97,7 @@ progenitor = { git = "https://github.com/oxidecomputer/progenitor", branch = "ma
 propolis = { path = "lib/propolis", default-features = false }
 propolis-client = { path = "lib/propolis-client" }
 propolis-server-config = { path = "crates/propolis-server-config" }
+propolis-standalone-config = { path = "crates/propolis-standalone-config" }
```
luqmana, line 100: It'd be nice to be able to just consolidate these and have a single config. The on-disk format doesn't necessarily have to change but maybe something like `Config::new_standalone`/`Config::new_server`? Or even just a simpler first step would be keeping both in a common crate `StandaloneConfig`/`ServerConfig`.

bin/propolis-utils/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "propolis-utils"
+version = "0.0.0"
+authors = ["Patrick Mooney <pmooney@oxide.computer>"]
+license = "MPL-2.0"
+edition = "2018"
```
luqmana, line 6: 2021 should be fine here, our current MSRV is 1.64 which is new enough (2021 edition was introduced with 1.56).

bin/propolis-utils/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This is a collection of assorted utilities which may be useful for certain
+development activities, but are otherwise not meant for general consumption.
+
+## Utilities
+
+- `savex`:  Extract data fields from a `propolis-standalone` instance which was
+  halted and saved with the `-s` flag.
+- `cpuid-gen`: Generated CPUID profile using the legacy emulated output from the
+  local host CPU, as filtered by the kernel VMM logic.
```
luqmana, line 11: :o cool addition!

bin/propolis-utils/src/bin/cpuid-gen.rs
```diff
@@ -0,0 +1,249 @@
...
+use clap::Parser;
+
+fn create_vm() -> anyhow::Result<VmmFd> {
+    let name = format!("cpuid-gen-{}", std::process::id());
+    let mut req = bhyve_api::vm_create_req::new(&name);
+
+    let ctl = VmmCtlFd::open()?;
+    let _ = unsafe { ctl.ioctl(bhyve_api::VMM_CREATE_VM, &mut req) }?;
+
+    let vm = VmmFd::open(&name).map_err(|e| {
```
luqmana, line 14: Could we extend the create request to also return a handle to the newly created vm (say with a new create flag)?

bin/propolis-utils/src/bin/cpuid-gen.rs
```diff
@@ -0,0 +1,249 @@
...
+        // Attempt to manually destroy the VM if we cannot open it
+        let mut destroy = bhyve_api::vm_destroy_req::new(&name);
+        let _ = unsafe { ctl.ioctl(bhyve_api::VMM_DESTROY_VM, &mut destroy) };
+        e
+    })?;
+
+    vm.ioctl_usize(bhyve_api::ioctls::VM_SET_AUTODESTRUCT, 1).map_err(|e| {
+        // Destroy instance (and faile) if auto-destruct cannot be set
```
luqmana, line 22: nit: don't need the bit in parens there?

bin/propolis-utils/src/bin/cpuid-gen.rs
```diff
@@ -0,0 +1,249 @@
...
+            0x1 => {
+                if data.ecx & CPU_FEAT_ECX_XSAVE != 0 {
+                    xsave_supported = true;
+                }
+                results.insert(CpuidKey::Leaf(eax), data);
+            }
+            0x7 => {
+                results.insert(CpuidKey::SubLeaf(eax, 0), data);
+
+                // TODO: handle more sub-leaf entries?
```
luqmana, line 144: Worth using the `x86` crate so we don't have to manually handle every case here? It does seem to support just providing a custom function instead of trying to execute cpuid itself: https://docs.rs/x86/latest/x86/cpuid/struct.CpuId.html#method.with_cpuid_fn

pfmooney: While some of the handling in that crate looks convenient, I'm not sure it's what we want here, since it seems less focused on manipulating the raw values?

.gitignore
```diff
@@ -1,5 +1,6 @@
 /target
 /crates/*/header-check/target
+/crates/*/header-check/Cargo.lock
```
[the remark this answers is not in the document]
pfmooney: I renamed them originally because I wanted them to be part of the workspace, but that was untenable since the header check `build.rs` requires a path to illumos sources lest it fail loudly.  This made rust-analyzer and friends complain about the workspace containing said failing builds.

Considering how the header-check stuff is mostly out of the way for most repo consumers, I think it's probably fine?

Cargo.toml
pfmooney: I've kept the standalone config stuff separate since we seem more likely to do weird stuff there than in propolis-server.  I'd really like to keep the option of experimentation there without having to worry about breaking something in -server.

bin/propolis-utils/README.md
pfmooney: Once CPUID handling gets wired up, I wanted a util to more easily generate profiles for hardware (such as Intel ;-) ) which differs from what we're shipping in production, but is still necessary for dev work.

bin/propolis-utils/src/bin/cpuid-gen.rs
pfmooney: As in part of the higher-level API of `bhyve_api` or the underlying ioctl itself?  I have considered adding it to the VM creation ioctl, but that'll probably wait until it can be part of [#15124](https://www.illumos.org/issues/15124)

bin/propolis-utils/src/bin/cpuid-gen.rs
pfmooney: Will clean that up

bin/propolis-utils/Cargo.toml
pfmooney: Will change that.  I mostly just cargo-culted when copying stuff over.

bin/propolis-utils/src/bin/cpuid-gen.rs
pfmooney: Filed [#15396](https://www.illumos.org/issues/15396) in regards to this.

luqmana (approved):

PR #317, opened 2023-02-17
softnpu: handle cpu/aux port

rcgoodfellow:
In `tfportd`/`tfport` based environments, a CPU/AUX port is now present. This device is not emulated by softnpu. However, we do need to be aware of and gracefully handle packets from this port, which has an index that is outside the range of normal softnpu ports.

All of the changes in the 2nd commit of this PR are taking care of clippy issues.

### DIFF
--- a/lib/propolis/src/hw/virtio/softnpu.rs
+++ b/lib/propolis/src/hw/virtio/softnpu.rs
@@ -582,7 +582,7 @@ impl PacketHandler {
         log: &Logger,
     ) {
         if usize::from(port) >= data_handles.len() {
-            error!(log, "port out of range {} > {}", port, data_handles.len());
+            error!(log, "port out of range {} >= {}", port, data_handles.len());
             return;
         }
         // get the dlpi handle for this port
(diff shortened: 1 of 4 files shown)


lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -569,12 +575,16 @@ impl PacketHandler {
...
-    fn send_packet_to_ext_port<'a>(
-        pkt: &mut packet_out<'a>,
+    fn send_packet_to_ext_port(
+        pkt: &mut packet_out<'_>,
         data_handles: &Vec<dlpi::DlpiHandle>,
         port: u16,
         log: &Logger,
     ) {
+        if usize::from(port) >= data_handles.len() {
+            error!(log, "port out of range {} > {}", port, data_handles.len());
```
luqmana, line 585:
```suggestion
            error!(log, "port out of range {} >= {}", port, data_handles.len());
```

nit: make the log output match the actual condition

luqmana (approved):

Issue #252, opened 2022-11-18
Migration sources shouldn't auto-resume after a failed migration


gjcolombo:
Today a migration failure automatically resumes the VM on the source: https://github.com/oxidecomputer/propolis/blob/5b25d1d2241405e85e2ab6806cbf13a1fdd45001/bin/propolis-server/src/lib/vm.rs#L1028-L1037

This is incorrect if the migration failed in a way that was destructive to the VM on the source. For example, if the target's Crucible backends connected downstairs and then the migration failed, the source can't resume because its Crucible connections will have been evicted.

The simplest solution to this is to mark the migration and instance as failed when this happens. A slightly more sophisticated solution is to let the control plane try to reason about what happened and recover if possible (it might need to send new backend configuration to the source first). An even more sophisticated solution is to automatically recover only if the target says that it's safe.

gjcolombo:
I think this bug is obsolete, at least for MVP, because the following condition won't apply:

> This is incorrect if the migration failed in a way that was destructive to the VM on the source.

Currently, the migration protocol doesn't work this way: the destination doesn't do anything destructive until it sees that migration has succeeded. The control plane-level migration procedure specified in RFD 361 relies on this property; it assumes that if migration fails, the migration source will simply resume (and send an appropriate state update) without any help from the control plane.

We can reopen this or change this behavior later but will need to decide what that implies for the control plane's migration logic.

PR #325, opened 2023-02-28
propolis-server: simplfy parts of the VM state driver

gjcolombo:
Today, an instance's VM controller owns the objects that record and publish an instance's externally visible instance state and its migration state, even though these statuses are logically owned by the VM state driver. This forces the state driver to call up to the VM controller every time it wants to update the externally-visible instance state.

Simplify matters as follows:

- Make the VM state driver own the sender side of the external state channel. Get rid of the `next_external` variable in the state driver loop; just publish to the channel directly when external state changes.
- Introduce a similar channel for migration state, also owned by the VM state driver. Add a bit of state to the VM controller to allow it to report on migrations that an external caller has requested but that the involved Propolises have not yet started.
- Remove the obsolete `StateDriverVmController` functions. The unit tests that set expectations on these mocks can now just read the states published by the state drivers under test.

Also remove the commented-out unit test for #252, since that issue is obsolete.

### DIFF
--- a/bin/propolis-server/src/lib/vm/mod.rs
+++ b/bin/propolis-server/src/lib/vm/mod.rs
@@ -169,6 +169,9 @@ pub(crate) struct VmObjects {
     /// A notification receiver to which the state worker publishes the most
     /// recent instance state and state generation.
     monitor_rx: tokio::sync::watch::Receiver<ApiMonitoredState>,
+
+    migrate_state_rx:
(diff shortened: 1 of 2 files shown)


jordanhendricks (approved):
Sorry for the delay in getting to this -- looks good, and I think this simplifies things a lot. Thanks!

gjcolombo:
Thanks for the review!

PR #327, opened 2023-03-02
Actually make use of bhyve and viona API versions

pfmooney:
Both bhyve (the kernel vmm portion) and viona expose version information as it relates to their respective ioctl APIs.  Although there are few (read: none) guarantees about one can depend on when it comes to inferring compatibility from said versions, it does provide some modicum of protection against weird behavior when expectations of those interfaces differ between propolis and the underlying OS.

### DIFF
--- a/bin/propolis-server/src/main.rs
+++ b/bin/propolis-server/src/main.rs
@@ -2,7 +2,7 @@
 #![cfg_attr(usdt_need_asm, feature(asm))]
 #![cfg_attr(all(target_os = "macos", usdt_need_asm_sym), feature(asm_sym))]
 
-use anyhow::anyhow;
+use anyhow::{anyhow, Context};
 use clap::Parser;
 use dropshot::{
     ConfigDropshot, ConfigLogging, ConfigLoggingLevel, HttpServerStarter,
@@ -87,25 +87,7 @@ async fn main() -> anyhow::Result<()> {
             )?;
 
             // Check that devices conform to expected API version
-            use propolis::api_version;
-            if let Err(e) = api_version::check() {
-                match e {
-                    api_version::Error::Io(io_err) => {
-                        slog::error!(log, "Error checking API versions");
-                        return Err(io_err.into());
-                    }
-                    api_version::Error::Mismatch(comp, act, exp) => {
-                        slog::error!(
-                            log,
-                            "Inadequate {} API version {} < {}",
-                            comp,
-                            act,
-                            exp
-                        );
-                        return Err(e.into());
-                    }
-                }
-            }
+            propolis::api_version::check().context("API version checks")?;
 
             let vnc_server = setup_vnc(&log, vnc_addr);
             let vnc_server_hdl = vnc_server.clone();
--- a/bin/propolis-server/src/main.rs
+++ b/bin/propolis-server/src/main.rs
@@ -87,13 +87,24 @@ async fn main() -> anyhow::Result<()> {
             )?;
 
             // Check that devices conform to expected API version
-            if let Err(e) = propolis::vmm::version::check() {
-                slog::error!(log, "Error checking VMM API version");
-                return Err(e.into());
-            }
-            if let Err(e) = propolis::hw::virtio::viona::version::check() {
-                slog::error!(log, "Error checking viona API version");
-                return Err(e.into());
+            use propolis::api_version;
+            if let Err(e) = api_version::check() {
+                match e {
+                    api_version::Error::Io(io_err) => {
+                        slog::error!(log, "Error checking API versions");
+                        return Err(io_err.into());
+                    }
+                    api_version::Error::Mismatch(comp, act, exp) => {
+                        slog::error!(
+                            log,
+                            "Inadequate {} API version {} < {}",
+                            comp,
+                            act,
+                            exp
+                        );
+                        return Err(e.into());
+                    }
+                }
             }
 
             let vnc_server = setup_vnc(&log, vnc_addr);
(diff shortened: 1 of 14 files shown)


bin/propolis-server/src/main.rs
```diff
@@ -86,6 +86,16 @@ async fn main() -> anyhow::Result<()> {
                 |error| anyhow!("failed to create logger: {}", error),
             )?;
 
+            // Check that devices conform to expected API version
+            if let Err(e) = propolis::vmm::version::check() {
+                slog::error!(log, "Error checking VMM API version");
+                return Err(e.into());
+            }
+            if let Err(e) = propolis::hw::virtio::viona::version::check() {
```
luqmana, line 94: Perhaps merge these checks into a single method the main propolis lib exports? Having every consumer explicitly check for sub component versions (like viona) seems like we're overly leaking some implementation details.

pfmooney: The versioning for those two components are independent, so the consumer may want to make differing decisions about what's required.  (I almost made the propolis-standalone version check to see if there was even a viona device in the config in order to trigger the check)

luqmana: Hmm, perhaps something like:

```rust
pub enum VersionError {
    Io(std::io::Error),
    VmmMismatch(u32, u32),
    ComponentMismatch(&'static str, u32, u32),
}

fn version_check() -> Result<(), VersionError> {
    vmm_version::check()?;
    viona_version::check()?;
    Ok(())
}
```

With appropriate from impls for vmm/viona version errors, e.g.:
```rust
impl From<viona_version::VersionError> for VersionError {
    fn from(err: viona_version::VersionError) -> Self {
        use viona_version::VersionError as VionaVerErr;
        match err {
            VionaVerErr::Io(e) => VersionError::Io(e),
            VionaVerErr::Mismatch(a, b) => VersionError::ComponentMismatch("viona", a, b),
        }
    }
}
```

So server can just fail on error from `version_check` but standalone can check for viona (either just the string or replace with another enum if you want) specific errors and ignore them if it wants.

pfmooney: How's the reworked stuff look?

bin/propolis-standalone/src/main.rs
```diff
@@ -863,6 +863,32 @@ fn setup_instance(
...
+            // user decide if they want to quit
+            slog::error!(
+                log,
+                "{} API version mismatch {} != {}",
+                comp,
+                act,
+                exp
+            );
+        }
+        _ => {}
```
luqmana, line 886: No other variants right? I'd just remove to make sure any new additions don't silently pass here.

bin/propolis-standalone/src/main.rs
```diff
@@ -887,6 +913,12 @@ fn main() -> anyhow::Result<()> {
 
     let (log, _log_async_guard) = build_log();
 
+    // Check that vmm and viona device version match what we expect
+    if let Err(e) = api_version_checks(&log) {
+        slog::error!(&log, "IO error during API version checks");
+        return Err(e.into());
+    }
```
luqmana, line 920: The error's already going to be outputted to stderr so I'd just skip the extra log here:
```suggestion
    api_version_checks(&log).context("API version checks")?;
```

bin/propolis-server/src/main.rs
```diff
@@ -86,6 +86,27 @@ async fn main() -> anyhow::Result<()> {
...
+                        slog::error!(
+                            log,
+                            "Inadequate {} API version {} < {}",
+                            comp,
+                            act,
+                            exp
+                        );
+                        return Err(e.into());
+                    }
+                }
```
luqmana, line 107: You already have a Display impl so can just use that for the log too:
```suggestion
                slog::error!(log, "API version check failed: {}", e);
                return Err(e.into());
```

luqmana: Or alternatively just skip the log altogether and do something like mentioned in the standalone case.

crates/bhyve-api/src/lib.rs
```diff
@@ -151,3 +163,43 @@ unsafe fn ioctl(
 ) -> Result<i32> {
     Err(Error::new(ErrorKind::Other, "illumos required"))
 }
+
+/// Convenience constants to provide some documentation on what changes have
```
luqmana, line 167: Thanks for the great summary!

crates/viona-api/sys/src/lib.rs
```diff
@@ -65,5 +65,10 @@ mod structs {
     }
 }
 
+/// This is the viona interface version which viona_api expects to operate
+/// against.  All constants and structs definied by the crate are done so in
```
luqmana, line 69: nit: `defined`

luqmana (approved):
Thanks, a couple small comments but otherwise lgtm

bin/propolis-standalone/src/main.rs
pfmooney: Filled out w/ the Ok case so new error cases will result in attention being required

bin/propolis-standalone/src/main.rs
pfmooney: Done

bin/propolis-server/src/main.rs
pfmooney: Yeah, I just went to the `context()` case

crates/viona-api/sys/src/lib.rs
pfmooney: Done

crates/bhyve-api/src/lib.rs
pfmooney: NP!  It's for our future selves.

pfmooney:
Merged in 50343abd8b1987acb2f242fa64eb2dd3cef94382

PR #330, opened 2023-03-08
Make migration status available after migration out

### LINKED ISSUE #288: Can't query migration status on a migration source after a successful migration
propolis-server configures itself to tear down its current VM controller when the VM contained within it stops: https://github.com/oxidecomputer/propolis/blob/9af863c1b00e5c5083f9321f01fd1dcba174e0fd/bin/propolis-server/src/lib/server.rs#L479-L484

It is possible to query instance status even after this because `ServiceProviders::stop` and `instance_get_common` do some extra work to hold onto a VM's properties and status channel even after the controller goes away.

No such affordance exists for live migration, though, so querying migrate status for an instance that stopped due to a successful live migration will fail. There should be such an affordance.

gjcolombo:
Use a similar technique to what we use for instance states: when a server drops its VM controller, clone the receiver side of the migration state channel and use the clone to answer subsequent migration state queries.

Tested with ad hoc live migrations & new PHD tests.

Fixes #288.

### DIFF
--- a/bin/propolis-server/src/lib/server.rs
+++ b/bin/propolis-server/src/lib/server.rs
@@ -9,6 +9,7 @@ use std::convert::TryFrom;
 use std::sync::Arc;
 use std::{collections::BTreeMap, net::SocketAddr};
 
+use crate::migrate::MigrateError;
 use crate::serial::history_buffer::SerialHistoryOffset;
 use dropshot::{
     channel, endpoint, ApiDescription, HttpError, HttpResponseCreated,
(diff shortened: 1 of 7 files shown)


bin/propolis-server/src/lib/server.rs
```diff
@@ -102,8 +104,11 @@ pub enum VmControllerState {
         /// outgoing controller can publish new state changes even after the
         /// server has dropped its reference to it (its state worker may
         /// continue running for a time).
-        watcher:
+        state_watcher:
             tokio::sync::watch::Receiver<api::InstanceStateMonitorResponse>,
+
+        migrate_state_watcher:
```
jordanhendricks, line 110: nit: doc comment here too?

jordanhendricks (approved):
thank you for fixing this! definitely an annoying papercut as someone running a lot of migrations.

one small comment but otherwise LGTM

bin/propolis-server/src/lib/server.rs
gjcolombo: Good call. Done in 9166b92.

gjcolombo:
Thanks as always for the review! Will merge once the CI run is done.

PR #332, opened 2023-03-08
Remove MigrateStart instance state request

### LINKED ISSUE #331: `InstanceStateRequested::MigrateStart` has irreversible side effects
Requesting this state for an instance sets a flag that says it's OK to migrate the instance out. If the flag's not set, prospective migration sources reject migration requests from prospective destinations. There's no way to unset the flag once it's set (short of destroying the entire VM). This presents a problem in Omicron: Nexus has to ask to set the flag as part of its migration saga, but there's no affordance to clear the flag if the relevant saga step is undone.

Some possible ways to solve this problem:

1. Add an affordance to clear the "migration out OK" flag.
2. Get rid of the `MigrateStart` request entirely and accept any request to migrate out that arrives while the VM is in a valid state.
3. Remove the "migration out OK" flag, but keep the `MigrateStart` request and have the request succeed iff the VM can be migrated from. This allows the migration procedure to fail fast (notably, without starting a destination Propolis at all) if the source is not in a state that permits migration. Note, though, that this information can go out of date as soon as Nexus queries it.

I favor option 2 since it requires the least code, and I don't think the benefits of option 3 are all that substantial, but I could be convinced otherwise.

Needed for MVP; Nexus's migration saga's undo logic can't be written correctly without resolving this tension.

gjcolombo:
This request sets a flag in the VM controller that lets a VM act as a migration source if its state otherwise permits it. There's no way to clear the flag, which makes it tough to write an undo action for this in the Nexus live migration saga that (presumably) would issue this call.

Since the request has no other side effects, the VM controller already rejects other invalid requests to migrate out, and sled agent doesn't presently use this request type, just remove it entirely.

Tested with an ad hoc migration & PHD tests.

Fixes #331. See that issue for discussion of alternative approaches.

### DIFF
--- a/bin/propolis-cli/src/main.rs
+++ b/bin/propolis-cli/src/main.rs
@@ -441,14 +441,6 @@ async fn migrate_instance(
         cloud_init_bytes: None,
     };
 
-    // Get the source instance ready
-    src_client
-        .instance_state_put(InstanceStateRequested::MigrateStart)
-        .await
-        .with_context(|| {
-        anyhow!("failed to place src instance in migrate start state")
-    })?;
-
     // Initiate the migration via the destination instance
     let migration_id = dst_client
         .instance_ensure(&request)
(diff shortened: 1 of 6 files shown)


luqmana (approved):

gjcolombo:
Thanks for reviewing!

PR #342, opened 2023-03-12
Add comment about matching crucible version

leftwo:
Added some language to indicate that the crucible version should match what propolis was compiled with.


jordanhendricks (approved):
thanks!

PR #343, opened 2023-03-13
server: simplify logic for accepting/rejecting external API requests

### LINKED ISSUE #284: Updates to a VM controller's `lifecycle_stage` are race-prone

### LINKED ISSUE #341: Requests to migrate into an instance are not idempotent
https://github.com/oxidecomputer/propolis/blob/d9434e54f8695b2c0e2f4f47c98c11095479e0e1/bin/propolis-server/src/lib/vm/mod.rs#L779-L800

The first request to migrate into an instance sets the lifecycle stage to "MigratePending" and so blocks future requests to become a migration target. The migration saga will generally assume these requests are idempotent, since it has a step that issues them.

I have some local changes drafted that should help address this.

gjcolombo:
**Reviewer note:** This might be a little more digestible when reviewed as a series of commits.

---

Simplify the propolis-server VM controller's logic for deciding what external API requests to accept or reject. The commit descriptions in this PR's commits explain the changes in more detail, but to summarize:

Currently, the propolis-server VM controller maintains some high-level state (a "lifecycle stage" and accompanying flags) and matches on that state to decide how to handle incoming external API requests. This code has some idempotency bugs (#284, #341), is hard to read, and requires a bunch of extra upcalls in the state driver -> VM controller interface (because the state can change both when requests are queued and when they are completed).

To fix this, introduce an "external request queue" module that serves the same function as the old lifecycle stage code. Instead of keeping track of VM states, the queue receives requests and VM state changes as inputs and stores as state the dispositions it should give new external requests as they come in. The queue is directly shared by the VM controller and state driver so that the latter doesn't have to call up to the former to get at the queue.

Also rearrange some state driver code for readability (move logic out of the top-level dispatch loop into callees).

After this some follow-up work is required to make sure we publish all the right states at the right times (and that publishing the "Failed" state has the correct semantics), especially when completing migrations.

Fixes #284. Fixes #341.

### DIFF
--- a/bin/propolis-server/src/lib/vm/request_queue.rs
+++ b/bin/propolis-server/src/lib/vm/request_queue.rs
@@ -264,11 +264,11 @@ impl ExternalRequestQueue {
             // Requests to reboot don't affect whether operations can be
             // performed.
             ExternalRequest::Reboot => {
-                assert!(matches!(
-                    self.allowed.migrate_as_target,
-                    Disposition::Deny(_)
-                ));
                 assert!(matches!(self.allowed.start, Disposition::Ignore));
+                assert!(!matches!(
+                    self.allowed.migrate_as_target,
+                    Disposition::Enqueue
+                ));
             }
 
             // Queueing a request to stop an instance disables any other
--- a/bin/propolis-server/src/lib/vm/request_queue.rs
+++ b/bin/propolis-server/src/lib/vm/request_queue.rs
@@ -1,4 +1,21 @@
-#![allow(dead_code)]
+//! Handles requests to change a Propolis server's state via the external API.
+//!
+//! An instance accepts or rejects requests to change state based on a
+//! combination of its current state and its knowledge of the requests it has
+//! previously queued but not processed yet. The latter knowledge is used to
+//! reject requests that will never be fulfilled (because they're preceded by an
+//! action that will forbid them; consider rebooting after stopping) or that may
+//! need be to redirected to a migration target.
+//!
+//! The queue maintains a disposition for each kind of request that can be sent
+//! to it, which allows that request to be enqueued, denied, or silently ignored
+//! (for idempotency purposes). These dispositions can change as new requests
+//! are queued. The queue also provides callbacks to the VM state driver that
+//! allow the driver to advise the queue of state changes that further affect
+//! what requests should be accepted.
+//!
+//! Users who want to share a queue must wrap it in the synchronization objects
+//! of their choice.
 
 use std::collections::VecDeque;
 
@@ -93,20 +110,27 @@ pub enum RequestDeniedReason {
     InstanceFailed,
 }
 
+/// The possible methods of handling a request to queue a state change.
 #[derive(Copy, Clone, Debug)]
-pub enum RequestDisposition {
+enum RequestDisposition {
+    /// Put the state change on the queue.
     Enqueue,
+
+    /// Drop the state change silently.
     Ignore,
+
+    /// Deny the request to change state.
     Deny(RequestDeniedReason),
 }
 
+/// The current disposition for each kind of incoming request.
 #[derive(Debug)]
-pub struct AllowedRequests {
-    pub migrate_as_target: RequestDisposition,
-    pub start: RequestDisposition,
-    pub migrate_as_source: RequestDisposition,
-    pub reboot: RequestDisposition,
-    pub stop: RequestDisposition,
+struct AllowedRequests {
+    migrate_as_target: RequestDisposition,
+    start: RequestDisposition,
+    migrate_as_source: RequestDisposition,
+    reboot: RequestDisposition,
+    stop: RequestDisposition,
 }
 
 #[derive(Debug)]
@@ -117,6 +141,7 @@ pub struct ExternalRequestQueue {
 }
 
 impl ExternalRequestQueue {
+    /// Creates a new queue that logs to the supplied logger.
     pub fn new(log: Logger) -> Self {
         Self {
             queue: VecDeque::new(),
@@ -135,14 +160,17 @@ impl ExternalRequestQueue {
         }
     }
 
+    /// Pops the request at the front of the queue.
     pub fn pop_front(&mut self) -> Option<ExternalRequest> {
         self.queue.pop_front()
     }
 
+    /// Indicates whether the queue is empty.
     pub fn is_empty(&self) -> bool {
         self.queue.is_empty()
     }
 
+    /// Asks to place the supplied request on the queue.
     pub fn try_queue(
         &mut self,
         request: ExternalRequest,
@@ -249,6 +277,8 @@ impl ExternalRequestQueue {
         Ok(())
     }
 
+    /// Notifies the queue that its instance is now running. This allows
+    /// requests to reboot and migrate out of the instance.
     pub fn notify_instance_running(&mut self) {
         info!(
             self.log,
@@ -258,6 +288,8 @@ impl ExternalRequestQueue {
         self.allowed.reboot = RequestDisposition::Enqueue;
     }
 
+    /// Notifies the queue that its instance has stopped. This blocks all new
+    /// requests except requests to stop, which are ignored.
     pub fn notify_instance_stopped(&mut self) {
         info!(self.log, "Instance is stopped, denying all actions");
         self.allowed.migrate_as_target =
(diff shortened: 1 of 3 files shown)


bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -0,0 +1,358 @@
...
+    #[error(
+        "The requested operation cannot be performed on a migration source"
+    )]
+    InvalidRequestForMigrationSource,
+
+    #[error("The instance is preparing to stop")]
+    HaltPending,
+
+    #[error("The instance failed to start or halted due to a failure")]
+    InstanceFailed,
```
jordanhendricks, line 110: a bit nitty, but I think I would prefer if these errors are a bit less verbose, such as removing The/A at the beginning

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -0,0 +1,358 @@
...
+}
+
+/// The possible methods of handling a request to queue a state change.
+#[derive(Copy, Clone, Debug)]
+enum RequestDisposition {
+    /// Put the state change on the queue.
+    Enqueue,
+
+    /// Drop the state change silently.
```
jordanhendricks, line 119: I know this is mentioned elsewhere, but I would mention this disposition exists for idempotency purposes here.

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -0,0 +1,358 @@
...
+        };
+
+        info!(&self.log, "queuing external request";
+              "request" => ?request,
+              "disposition" => ?disposition);
+
+        // At this point the request will be queued. Queuing some requests
+        // logically forecloses on other kinds of requests. Update the
+        // dispositions of these requests, then queue the request.
+        match request {
```
jordanhendricks, line 210: When reading this piece of this method, I found myself wanting a table helping me translate the request and the set of allowed states after we enqueue it. This was particularly true where the resulting disposition of a request type would be implicit.

What would you think about matching on the request and returning from the `match` the AllowedRequests struct that would be associated with it? That might make it easier to grok reading this, since you can walk up to the function and quickly glean what the resulting dispositions will be.

For example, for the `MigrateAsTarget` case, something like:
```
let allowed_requests = match request {
    ExternalRequest::MigrateAsTarget { .. } => {
        AllowedRequests {
            migrate_as_target: Disposition::Ignore,
            start: Disposition::Ignore,
            migrate_as_source: Disposition::Deny(DenyReason::MigrateTargetInProgress),
            reboot: Disposition::Deny(DenyReason::MigrateTargetInProgress),
            stop: Disposition::Enqueue,
        }
    },

// ... handle other request types

};
```

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -0,0 +1,358 @@
...
+        info!(self.log, "Instance has failed, denying all actions");
+        self.allowed.migrate_as_target =
+            RequestDisposition::Deny(RequestDeniedReason::InstanceFailed);
+        self.allowed.start =
+            RequestDisposition::Deny(RequestDeniedReason::InstanceFailed);
+        self.allowed.migrate_as_source =
+            RequestDisposition::Deny(RequestDeniedReason::InstanceFailed);
+        self.allowed.reboot =
+            RequestDisposition::Deny(RequestDeniedReason::InstanceFailed);
+    }
```
jordanhendricks, line 331: We could also do something similar here -- return a new AllowedRequests struct to wholesale replace the existing one.

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -0,0 +1,358 @@
...
+
+    /// Indicates whether the queue would allow a request to migrate into this
+    /// instance. This can be used to avoid setting up migration tasks for
+    /// requests that will ultimately be denied.
+    pub fn migrate_as_target_allowed(&self) -> Result<(), RequestDeniedReason> {
+        match self.allowed.migrate_as_target {
+            RequestDisposition::Enqueue => Ok(()),
+            RequestDisposition::Ignore => {
+                panic!("requests to migrate as target should not be ignored")
```
jordanhendricks, line 340: this feels more like an `assert!`?

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    /// Indicates whether the queue would allow a request to migrate out of this
+    /// instance. This can be used to avoid setting up migration tasks for
+    /// requests that will ultimately be denied.
+    pub fn migrate_as_source_allowed(&self) -> Result<(), RequestDeniedReason> {
+        match self.allowed.migrate_as_source {
+            RequestDisposition::Enqueue => Ok(()),
+            RequestDisposition::Ignore => {
+                panic!("requests to migrate as source should not be ignored")
```
jordanhendricks, line 353: same here

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -124,17 +128,9 @@ where
...
         loop {
-            let event = self.controller.wait_for_next_event();
-            let outcome = self.handle_event(event, &mut next_lifecycle);
-
-            if let Some(next_lifecycle) = next_lifecycle.take() {
-                self.controller.set_lifecycle_stage(next_lifecycle);
-            }
-
+            let event = self.shared_state.wait_for_next_event();
+            let outcome = self.handle_event(event);
```
jordanhendricks, line 133: This pattern is definitely much cleaner than what was here before -- it looks nice!

jordanhendricks (commented):
This looks pretty good! I really like how much simpler this makes the event handling to follow. A couple review comments for now.

gjcolombo:
Thanks @jordanhendricks! Commit 20c2a32 should address all the feedback above. I also tacked on some unit tests in 318fbc4 (including one covering some behavior that was incorrect in the original change...).

bin/propolis-server/src/lib/vm/request_queue.rs
gjcolombo: I really like the way this worked out--thanks for suggesting it!

gjcolombo:
Hmm, that vmm handle leak in the PHD run is new... Will look into it.

gjcolombo:
> Hmm, that vmm handle leak in the PHD run is new... Will look into it.

This is a regression in the last batch of changes. Working on a fix.

gjcolombo:
The bug was a copy-and-paste error. Transitioning to a Failed state (which we now do in this PR when migration in fails) used to preserve the last disposition for requests to stop, which (in the relevant PHD tests) was Enqueue. 20c2a32 broke this and changed it to Ignore. e0d8da2 fixes the problem and adds a unit test. With any luck the PHD run will come out clean this time.

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -0,0 +1,549 @@
...
+    /// instance. This can be used to avoid setting up migration tasks for
+    /// requests that will ultimately be denied.
+    ///
+    /// # Return value
+    ///
+    /// - `Ok(true)` if the request will be queued.
+    /// - `Ok(false)` if the request is allowed for idempotency reasons but will
+    ///   not be queued.
+    /// - `Err` if the request is forbidden.
+    pub fn migrate_as_target_will_enqueue(
```
jordanhendricks, line 245: Thanks for this -- I like this naming better than the original version.

jordanhendricks (approved):

bin/propolis-server/src/lib/vm/request_queue.rs
jordanhendricks: glad to hear it! I like the way the new version looks, too.

PR #344, opened 2023-03-14
add crucible to propolis-standalone config toml

faithanalog:
this adds basic support for configuring propolis-standalone to use Crucible. I'm not super familiar with propolis internals; let me know if anything needs adjusting.

There's currently an issue with it where if there's a connection problem between crucible upstairs/downstairs, upstairs will get stuck in a reconnection loop, blocking the VM from exiting in response to Ctrl-C or shutdown. It's definitely something just blocking on waiting for all the async jobs to complete, which they never do. Things exit normally if the connection doesn't have any problems before the exit.

 i don't like that it does that, but I'm not really sure how to fix it. for the main thing i want to use this in, it doesnt matter and i can just `kill -9`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,72 @@ which acts as a serial port. One such tool for accessing this serial port is
 [sercons](https://github.com/jclulow/vmware-sercons), though others (such as
 "screen") would also work.
 
+propolis-standalone also supports defining crucible-backed storage devices,
+thought it is somewhat inconvenient to do so without scripting. Currently you can only use a single region per block device. Read the comments in this TOML example for more details:
+
+```toml
+[block_dev.some_datastore]
+type = "crucible"
+
+# === REQUIRED OPTIONS ===
+# these MUST match the region configuration downstairs
+block_size = 512
+blocks_per_extent = 262144
+extent_count = 32
+
+# Array of the SocketAddrs of the Downstairs instances. There must be three
+# of these, or propolis-standalone will panic.
+targets = [
+  "127.0.0.1:3810",
+  "127.0.0.1:3820",
+  "127.0.0.1:3830",
+]
+
+# Generation number used when connecting to Downstairs. This must
+# monotonically increase with each successive connection to the Downstairs,
+# which means that you need to bump this number every time you restart
+# your VM. Kind of annoying, maybe we can get a better way to pass it in.
+# Anyway, if you don't want to read-modify-write this value, a hack you
+# could do is set this to the current number of seconds since the epoch.
+# This'll always work, except for if the system time goes backwards, which
+# it can definitely do! So, you know. Be careful.
+generation = 1
+# === END REQUIRED OPTIONS ===
+
+
+# === OPTIONAL OPTIONS ===
+# This should be a UUID. It can be anything, really. When unset, defaults
+# to a random UUIDv4
+# upstairs_id = "e4396bd0-ede1-48d7-ac14-3d2094dfba5b"
+
+# When true, some random amount of IO requests will synthetically "fail".
+# This is useful when testing IO behavior under Bad Conditions.
+# Defaults to false.
+# lossy = false
+
+# YYY what does this do?
+# flush_timeout = <number>
+
+# Base64'd encryption key used to encrypt data at rest. Keys are 256 bits.
+# Note that the region must have already been created with encryption
+# enabled for this to work. That may change later though.
+# encryption_key = ""
+
+# These three values are pem files for TLS encryption of data between
+# propolis and the downstairs.
+# cert_pem = ""
+# key_pem = ""
+# root_cert_pem = ""
+
+# Specifies the SocketAddr of the crucible control interface. YYY I'm not
+# really sure what this actually does, but it defaults to Nothing
+# control_addr = ""
+
+# When true, the device will be read-only. Defaults to false
+# read_only = false
+# === END OPTIONAL OPTIONS ===
+```
+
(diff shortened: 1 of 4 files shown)


bin/propolis-standalone/Cargo.toml
```diff
@@ -17,14 +17,17 @@ libc.workspace = true
 toml.workspace = true
 tokio = { workspace = true, features = ["io-util", "rt-multi-thread"] }
 serde = { workspace = true, features = ["derive"] }
-propolis.workspace = true
+propolis = { workspace = true, features = ["crucible-full"], default-features = false }
```
pfmooney, line 20: I'd like to keep crucible as an optional dependency for those not using it in their standalone work.

bin/propolis-standalone/src/config.rs
```diff
@@ -43,6 +44,133 @@ pub fn block_backend(
             let creg = ChildRegister::new(&be, Some(path.to_string()));
             (be, creg)
         }
+        "crucible" => {
+            info!(log, "Building a crucible VolumeConstructionRequest from options {:?}", be.options);
```
pfmooney, line 48: Assuming the crucible feature is made optional, move this into a function gated with a `cfg()` conditional?

jordanhendricks:
> There's currently an issue with it where if there's a connection problem between crucible upstairs/downstairs, upstairs will get stuck in a reconnection loop, blocking the VM from exiting in response to Ctrl-C or shutdown.

You might have already seen this, but this sounds like https://github.com/oxidecomputer/crucible/issues/660 to me. I think the final verdict there was that this is caused by a crucible version mismatch. It would be interesting to know if that wasn't the case here.

README.md
```diff
@@ -157,6 +157,72 @@ which acts as a serial port. One such tool for accessing this serial port is
 [sercons](https://github.com/jclulow/vmware-sercons), though others (such as
 "screen") would also work.
 
+propolis-standalone also supports defining crucible-backed storage devices,
+thought it is somewhat inconvenient to do so without scripting. Currently you can only use a single region per block device. Read the comments in this TOML example for more details:
```
jordanhendricks, line 161: probably worth linking to relevant crucible docs here in case any of documented values here change

jordanhendricks (commented):
I'm curious if you explored how easy it would be to add support for a crucible backend in propolis-server. Since propolis-server is what we're running in prod, and the TOML file is not the way propolis-server will be configured in prod, we might have to a bit more careful about it, but I think that's true for all things in the TOML file.

If it seems like too much of a lift to add this to propolis-server as well, I would like to add least add a pointer to the readme on instructions getting going there (docs/migrate-with-crucible.md)

bin/propolis-standalone/Cargo.toml
faithanalog: ah, my brain missed the feature flags at the bottom of the toml. can do

faithanalog:
> > There's currently an issue with it where if there's a connection problem between crucible upstairs/downstairs, upstairs will get stuck in a reconnection loop, blocking the VM from exiting in response to Ctrl-C or shutdown.
> 
> You might have already seen this, but this sounds like [oxidecomputer/crucible#660](https://github.com/oxidecomputer/crucible/issues/660) to me. I think the final verdict there was that this is caused by a crucible version mismatch. It would be interesting to know if that wasn't the case here.

yeah i was experiencing it with specifically a version mismatch

faithanalog:
> > > There's currently an issue with it where if there's a connection problem between crucible upstairs/downstairs, upstairs will get stuck in a reconnection loop, blocking the VM from exiting in response to Ctrl-C or shutdown.
> > 
> > 
> > You might have already seen this, but this sounds like [oxidecomputer/crucible#660](https://github.com/oxidecomputer/crucible/issues/660) to me. I think the final verdict there was that this is caused by a crucible version mismatch. It would be interesting to know if that wasn't the case here.
> 
> yeah i was experiencing it with specifically a version mismatch

actually i also get this if i just never turn on downstairs. upstairs tries to connect forever so propolis-standalone never fully shuts down.

README.md
```diff
@@ -157,6 +157,87 @@ which acts as a serial port. One such tool for accessing this serial port is
 [sercons](https://github.com/jclulow/vmware-sercons), though others (such as
 "screen") would also work.
 
+propolis-standalone also supports defining crucible-backed storage devices in
+the TOML config. It is somewhat inconvenient to do this without scripting,
+because `generation` must monotonically increase with each successive connection
```
pfmooney, line 162: Maybe nest the crucible details under a section header or something?

bin/propolis-standalone/Cargo.toml
```diff
@@ -17,16 +17,19 @@ libc.workspace = true
 toml.workspace = true
 tokio = { workspace = true, features = ["io-util", "rt-multi-thread"] }
 serde = { workspace = true, features = ["derive"] }
-propolis.workspace = true
+propolis = { workspace = true }
+crucible-client-types.workspace = true
```
pfmooney, line 21: Shouldn't this be an optional dep, pulled in by the `crucible` feature?

pfmooney (approved):
Looks fine to me, if it's working for you as desired.

The config parsing is rather brittle (with all the unwraps and whatnot) if the user provides an imperfect configuration, but given that propolis-standalone is a developer playground, that's par for the course.  Having the example in the README should be helpful on that front.

Please squash your commits down prior to integration.

pfmooney:
Integrated via #486

Issue #340, opened 2023-03-10
panicked at 'vCPU 3: Unhandled VM exit: Paging(4294967292, 2)'


jmpesp:
In testing for https://github.com/oxidecomputer/propolis/issues/333, I built a new helios image using the tools at helios-engvm, and booted that this morning. Propolis had a different panic:

```
Mar 10 14:43:33.848 INFO rdmsr, msr: 3221291039, vcpu: 0, component: vcpu_tasks
Mar 10 14:43:33.849 INFO wrmsr, value: 70368744177664, msr: 3221291039, vcpu: 0, component: vcpu_tasks
thread 'vcpu-3' panicked at 'vCPU 3: Unhandled VM exit: Paging(4294967292, 2)', bin/propolis-server/src/lib/vm/mod.rs:436:9
stack backtrace:
   0:          0x1cddb9c - std::backtrace_rs::backtrace::libunwind::trace::h9b534fc6095f520f
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:          0x1cddb9c - std::backtrace_rs::backtrace::trace_unsynchronized::hfcd8a890f991a10d
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:          0x1cddb9c - std::sys_common::backtrace::_print_fmt::hdf432e5f0940fa9c
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/sys_common/backtrace.rs:65:5
   3:          0x1cddb9c - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h5693a803500ade80
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/sys_common/backtrace.rs:44:22
   4:          0x1d352ca - core::fmt::write::h94f582bdfeeeb1a6
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/fmt/mod.rs:1209:17
   5:          0x1cce984 - std::io::Write::write_fmt::hd3f00507505f204d
                               at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/std/src/io/mod.rs:1682:15
[...]

PR #346, opened 2023-03-22
Change state correctly after migration finishes; make use of the Failed state

### LINKED ISSUE #209: Consider adding "failed" terminal VM state

### LINKED ISSUE #323: Fail & destroy the VM controller when migration in fails
Today, a migration target remains in the "Migrating" state when a migration in fails, and has a lifecycle stage that prevents a subsequent migration from being attempted with the same VM controller. VMs in this state should self-destruct instead.

This is required to handle cases where migration fails because the source crashed. In this case, Nexus needs to be able to figure out that the instance has failed. But it's unsafe for the target to tell Nexus this: if migration failed because the source/target connection was interrupted, but the source is otherwise alive, then the instance is still running. So the safest course is just to terminate the target: if the source is still alive, it will notice that migration failed and push an appropriate update to Nexus; if the source is gone, then neither the source nor the target will be responsive, and Nexus can observe this by checking both VMMs for a heartbeat.

### LINKED ISSUE #326: Failed migration out doesn't return VM controller to the Running state
`StateDriver::migrate_as_source` ensures that VMs resume on migration failure, but no one actually publishes a transition back to the "Running" state when this happens. Not only is this misleading, sled agent will want to see this transition so it can determine that migration failed and that the source needs to tell Nexus about that.

gjcolombo:
Ensure that after a live migration in or out, the state driver transitions the VM to some sensible state. #343 fixed some of this by making migration targets go to the Failed state after failing to migrate in; this PR aims to finish the job for all migration outcomes.

Turn the Failed state into a terminal state that destroys the underlying VM so that the control plane doesn't have to coordinate requests to stop such a VM. Make the VM controller's `Drop` impl preserve the Failed state so that it's distinct from Destroyed (sled agent/Nexus treat Destroyed as Stopped, but they shouldn't do this for Failed).

Tweak the migration code in the state driver to be crisper about who's responsible for publishing migration state updates. Previously, the state driver published some of these; now it assumes the migration tasks always do.

Fixes #209. Fixes #323. Fixes #326.


gjcolombo:
Strictly speaking, this PR implements a slightly different definition of "failed" than that described in #209. This PR uses it more or less to mean "we couldn't even start the VM," but #209 talks about having more of a "the guest has become sad" state where we might want to allow an operator/VM user to examine the state of the VM more closely (see, for example, #340). I would be inclined to treat these as separate conditions though ("Failed" vs "GuestFailed" or something like that).

I'm a little concerned that destroying the VM on a startup failure (without requiring a request to stop) is too aggressive of an approach. If others are similarly concerned I can try weakening that a bit.

gjcolombo:
I believe Buildomat is taking a nap, so the PHD run for these changes is on hold. I'll wait until it's back before merging anything here.

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -798,13 +830,99 @@ mod tests {
...
+        // is not directly processing the queue here, the test has to issue this
+        // call itself.
+        assert!(matches!(
+            driver.driver.get_migration_state(),
+            Some((_, ApiMigrationState::Finish))
+        ));
+        assert_eq!(
+            driver.driver.get_migration_state().unwrap().0,
+            migration_id
+        );
```
luqmana, line 840: You can fold these together with a guard on the `matches!`:

```suggestion
        assert!(matches!(
            driver.driver.get_migration_state(),
            Some((id, ApiMigrationState::Finish)) if id == migration_id
        ));
```

luqmana (commented):
>I'm a little concerned that destroying the VM on a startup failure (without requiring a request to stop) is too aggressive of an approach. If others are similarly concerned I can try weakening that a bit.

I'm inclined to agree, I think we should we should mark it as `Failed` but leave the vmm state available. Sled Agent or whatever other consumer can choose to further clean it up.

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: Good call. Done in 3a2c071.

gjcolombo:
> I'm inclined to agree, I think we should we should mark it as Failed but leave the vmm state available. Sled Agent or whatever other consumer can choose to further clean it up.

I gave this a bit more thought. It still seems a bit odd to me to preserve the `Instance` in cases where we didn't actually try to start it (e.g. because of a failure in the migration protocol), but I've come around to thinking it's valuable to have in cases where we fail to start the VM and want to have the option of writing a dump and debugging it. So I changed this back in c724f0b.

We could also split the difference here and return `HandleEventOutcome::Exit` from migration protocol failures and `HandleEventOutcome::Continue` from VM start failures. I didn't do that for now just to keep things consistent. If it continues to be a big pain point for migration testing, we can always revisit (my WIP sled agent changes don't currently care about the behavior here so we have time to settle on what it should be). CC @jordanhendricks on this point since it came up in the discussion on #323.

gjcolombo:
Thanks for the review!

Issue #337, opened 2023-03-09
implement live migration support for guest timing data


jordanhendricks:
Ongoing work for illumos #[15467](https://www.illumos.org/issues/15467) will introduce kernel interfaces allowing the guest TSC and device timers to continue working across live migrations. As a part of the broader TSC/time migration work, propolis-server needs to consume these  interfaces and migrate the associated data appropriately.

Issue #324, opened 2023-02-27
Investigate/profile live migration performance


gjcolombo:
A short while ago I did an ad-hoc live migration for the first time in a while, and the memory transfer phase "felt" (how scientific!) a lot slower than it did the last time I tried one of these (which was probably a month or two ago).

In the short run, we should do some ad hoc tests to see if there are any obvious bottlenecks. (I'd like in particular to compare a few recent builds to see if I'm just imagining a speed difference.)

In the long run, we need better tools to assess blackout times and how they change over time. That probably includes

- Collecting information about at least the following:
  - how long a VM's vCPUs are paused in general
  - how long individual expensive phases of migration take, especially memory transfer and target device startup
- Surfacing this information so that we can get at it in test automation
- Writing automated tests that track migration performance over time
  - ex.: boot a guest; have it dirty as much memory as possible given its size (to try to get a consistent number of dirty pages to transfer); migrate it; gather migration metrics

These various steps should get their own issues; for now I'm just filing this to keep track of this work before I get distracted and forget about it.

gjcolombo:
I'm going to spend some time looking into this today/this week.

PR #347, opened 2023-03-22
Add coarse instrumentation for measuring migration protocol times

jordanhendricks:
Wrap migration steps with USDT probes such that we can measure how long each takes. Add a D script to propolis-utils to consume these probes.

Also includes some cleanup of the live migration protocol, including:
- remove unused arch state protocol step
- remove start/end protocol phases, which only log that the migration has begun or ended. Move these log messages up to the original caller.
- lower the logging level of some noisy live migration messages


jordanhendricks:
### Sample D script output

#### Source

Verbose output:
```
$ pfexec ./scripts/live-migration-times.d 939 v
tracing live migration protocol times for pid 939...

MigrateSync  BEGIN      timestamp=16139671421368
MigrateSync  END        timestamp=16139671885314
Pause        BEGIN      timestamp=16139671902136
Pause        END        timestamp=16139672493713
RamPush      BEGIN      timestamp=16139672507879
RamPush      END        timestamp=16167408932572
DeviceState  BEGIN      timestamp=16167408946669
DeviceState  END        timestamp=16167426488264
RamPull      BEGIN      timestamp=16167426500778
RamPull      END        timestamp=16167427288936
Finish       BEGIN      timestamp=16167427299897
Finish       END        timestamp=16167451756998
^C


STEP                    TIME ELAPSED (usec)
MigrateSync                             463
Pause                                   591
RamPush                            27736424
DeviceState                           17541
RamPull                                 788
Finish                             24457101

TOTAL                              27780267
```

Non-verbose output (default):
```
$ pfexec ./scripts/live-migration-times.d 939
tracing live migration protocol times for pid 939...

^C


STEP                    TIME ELAPSED (usec)
MigrateSync                             463
Pause                                   591
RamPush                            27736424
DeviceState                           17541
RamPull                                 788
Finish                             24457101

TOTAL                              27780267
```

#### Destination propolis-server
```
$ pfexec ./scripts/live-migration-times.d 937 v
tracing live migration protocol times for pid 937...                                 
  
MigrateSync  BEGIN      timestamp=16139671554780
MigrateSync  END        timestamp=16139671865928
RamPush      BEGIN      timestamp=16139671882689
RamPush      END        timestamp=16167408909268
DeviceState  BEGIN      timestamp=16167408924237
DeviceState  END        timestamp=16167426451174
RamPull      BEGIN      timestamp=16167426459320
RamPull      END        timestamp=16167427241226
Finish       BEGIN      timestamp=16167427247889
Finish       END        timestamp=16167427386300
^C


STEP                    TIME ELAPSED (usec)
MigrateSync                             311
RamPush                            27737026
DeviceState                           17526
RamPull                                 781
Finish                               138411

TOTAL                              27755784
```

Non-verbose output (default):
```
$ pfexec ./scripts/live-migration-times.d 937
tracing live migration protocol times for pid 937...                                 
  
^C


STEP                    TIME ELAPSED (usec)
MigrateSync                             311
RamPush                            27737026
DeviceState                           17526
RamPull                                 781
Finish                               138411

TOTAL                              27755784
```

bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -81,20 +82,41 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> DestinationProtocol<T> {
...
-        self.arch_state().await?;
-        self.ram_pull().await?;
-        self.finish().await?;
-        self.end();
-        Ok(())
+    async fn run_phase(
+        &mut self,
+        step: MigratePhase,
+    ) -> Result<(), MigrateError> {
+        probes::migrate_phase_begin!(|| { format!("{:?}", step) });
```
gjcolombo, line 89: I might be inclined to `impl Display for MigratePhase` (instead of or in addition to deriving `Debug`) so that we directly control how the phase names are formatted instead of relying on the derive's implementation details (hard to foresee them changing, but *technically* they're [not stable](https://doc.rust-lang.org/std/fmt/trait.Debug.html#stability)...).

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -129,7 +148,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
 
         loop {
             let m = self.read_msg().await?;
-            info!(self.log(), "ram_push: source xfer phase recvd {:?}", m);
+            trace!(self.log(), "ram_push: source xfer phase recvd {:?}", m);
             match m {
                 codec::Message::MemDone => break,
                 codec::Message::MemFetch(start, end, bits) => {
```
gjcolombo, line 154: (it has now been 0 reviews since I got mad about Github not letting me comment on arbitrary lines of a file in a change)

I think it'd also be good to have a probe somewhere around here (or in `xfer_ram`) that indicates the number of pages we transferred. This should let us compute data transfer rates in the script, which are important context when evaluating whether the `RamPush` phase is taking a long time.

scripts/live-migration-times.d
```diff
@@ -0,0 +1,139 @@
...
+		this->vcpu_pause += this->d_fin;
+	}
+	
+	printf("\n");
+
+	/* Print vCPU pause time */
+	if (this->total !=0 && this->d_pause != 0) {
+		printf("%-25s %30d\n", "MAX vCPU PAUSE TIME",
+		    this->vcpu_pause / 1000);
+	}
```
gjcolombo, line 133: This is maybe more of a "file an issue and fix in another PR" sort of thing, but I think this figure should come with a warning label. IIUC it will indicate the total time that elapsed between when the source paused and when the source decided migration was done, but will (necessarily, since the source and destination are different processes) exclude the time between when the destination completed migration and the time it resumed the CPUs. This latter period can be significant if we have to do things like repair Crucible volumes.

Probably the thing to do here is to add some probes to the VM state driver's `start_vm` routine that describe how long it takes to start entities and when we finally get around to starting vCPUs. But it's probably also easier to wait to do that until all the other dust in the state driver has settled.

gjcolombo: An additional thought: even if we do this, it seems like it might be mildly annoying to have to stitch together traces from multiple hosts to get all the available information about migration performance. Would it be useful to have the source (optionally) transmit performance statistics to the target as part of the protocol so that the target can report everything in one go? (The additional server-state phase in #345 would be a natural place for this, I think.)

gjcolombo (commented):
Thanks for these changes! Things generally LGTM, but I have a couple of comments about probes I think we'd want to add to get a complete picture of how a migration has performed.

scripts/live-migration-times.d
jordanhendricks: >  I think this figure should come with a warning label. IIUC it will indicate the total time that elapsed between when the source paused and when the source decided migration was done, but will (necessarily, since the source and destination are different processes) exclude the time between when the destination completed migration and the time it resumed the CPUs.

Yeah, this is a good point, and one I had thought about, both for the vCPU pause time and the script overall, since it's only tracking these specific pieces of the live migration process (though the `RamPush` phase obviously is the bottleneck for the whole process, generally). I added the vCPU bit here because I thought it might at least be nice to have the script do some amount of math for you to get a sense of how long the guest was paused.

For now, I'm happy to remove the vCPU pause message all together and we can file a new issue to add some probes to the state driver, as you suggested. Then we can add that in when we have better instrumentation there. Thoughts on that?

As for stitching together traces from multiple hosts, that wasn't something I was trying to address yet here -- I was more intending for a development tool that could perhaps be run in CI and could assist me with verifying that the adjustments we are making to the guest TSC as a part of the TSC migration work (#337 and related illumos work, [15467)](https://www.illumos.org/issues/15467). I also haven't looked at #345 yet, so I can't speak to that yet. Maybe we could file a new issue for how to stitch together this data for multiple hosts, to track along with the other work laid out in #324?

bin/propolis-server/src/lib/migrate/destination.rs
jordanhendricks: Good thought -- I also felt like this was cheating a bit but wasn't super focused on the stability of the script. Implementing `Display` is easy to add though, so I can do that.

bin/propolis-server/src/lib/migrate/source.rs
jordanhendricks: Sure, that's something I've wanted to. I'll take a stab at doing so in this PR.

scripts/live-migration-times.d
gjcolombo: > For now, I'm happy to remove the vCPU pause message all together and we can file a new issue to add some probes to the state driver, as you suggested. 

> Maybe we could file a new issue for how to stitch together this data for multiple hosts, to track along with the other work laid out in https://github.com/oxidecomputer/propolis/issues/324?

Both these approaches sound good to me. I think it's probably also fine to leave the "time spent paused" message in there more or less as-is, or with a comment in the D script describing its limitations. It's up to you; these are (as you mentioned) developer-facing scripts to begin with, so I'm not concerned about e.g. an operator being led astray here.

gjcolombo (approved):
Ship it!

scripts/live-migration-times.d
jordanhendricks: after trying to write up the limitations of the vCPU pause message, I ended up deciding it was more trouble than it was worth to have that information in there -- all it's really doing is subtracting out the `Sync` phase time anyway.

PR #351, opened 2023-03-24
migration: make source/target mutually acknowledge completion

### LINKED ISSUE #350: Migration source/target pair can disagree about who should run the guest
The end-of-migration protocol (in the Finish phase) currently proceeds as follows:

1. Target sends "OK" message to source, checking for errors
2. Source waits for "OK" message from target, checking for errors
3. Source sends "OK" message to target, ignoring errors
4. Target waits for "OK" message from source, ignoring errors

This seems to me to allow the following sequence of events:

1. Target sends "OK" to source; this succeeds
2. Source fails to receive "OK" from target; this error causes the source's migration to unwind
3. Source never sends a reciprocal "OK" to the target (because it unwound already)
4. Target fails to read a message from the source (makes sense since none was ever sent); this error is ignored and migration completes on the target

Now both the source and target will try to run the instance. I think requiring steps 3 and 4 to succeed fixes this problem: the target can't run until it's received the final "OK" message from the source, which implies that the source successfully sent that last message, and we can guarantee that migration succeeds on the source once that last message is sent.

gjcolombo:
In the current migration protocol's Finish phase, if the destination successfully sends its final "OK" message, it will think migration has succeeded irrespective of whether the source received or acknowledged that message. This can create a scenario where both source and target think they're supposed to be running a VM (the target because it thinks migration succeeded, the source because it thinks migration failed).

Fix this by changing this phase so that (a) the source succeeds only if it successfully sends a final "OK" message, and (b) the target succeeds only if it successfully receives that message. This creates the possibility that *no one* will run the VM (source sent the last handoff, target didn't receive it), but that's not much different from the VM successfully migrating and then immediately crashing (and is probably better than having two identical instances fighting over the same backends, etc.).

Also change this phase to publish the "Finish" state only after passing the point of no return. This lets callers use the Finish migration state to mean "migration completed successfully" without having to worry about a later transition to the Error state.

Fixes #350.

### DIFF
--- a/bin/propolis-server/src/lib/migrate/destination.rs
+++ b/bin/propolis-server/src/lib/migrate/destination.rs
@@ -326,8 +326,6 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> DestinationProtocol<T> {
     }
 
     async fn finish(&mut self) -> Result<(), MigrateError> {
-        self.update_state(MigrationState::Finish).await;
-
         // Tell the source this destination is ready to run the VM.
         self.send_msg(codec::Message::Okay).await?;
 
@@ -336,6 +334,10 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> DestinationProtocol<T> {
         // to be sure the source hasn't decided the migration has failed and
         // that it should resume the VM.
         self.read_ok().await?;
+
+        // Now that control is definitely being transferred, publish that the
+        // migration has succeeded.
+        self.update_state(MigrationState::Finish).await;
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -326,9 +326,18 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> DestinationProtocol<T> {
     }
 
     async fn finish(&mut self) -> Result<(), MigrateError> {
-        self.update_state(MigrationState::Finish).await;
+        // Tell the source this destination is ready to run the VM.
         self.send_msg(codec::Message::Okay).await?;
-        let _ = self.read_ok().await; // A failure here is ok.
+
+        // Wait for the source to acknowledge that it's handing control to this
+        // destination. If this acknolwedgement doesn't arrive, there's no way
```
jordanhendricks, line 333: "acknowledgement"

jordanhendricks (approved):
great find! LGTM

bin/propolis-server/src/lib/migrate/destination.rs
gjcolombo: Whoops. Thanks! Fixed in c783903.

PR #352, opened 2023-03-24
Expose API for reservoir manipulation

pfmooney:
This includes a rust clone of `rsrvrctl` as an example of the API usage.

### DIFF
--- a/bin/propolis-utils/src/bin/rsrvrctl.rs
+++ b/bin/propolis-utils/src/bin/rsrvrctl.rs
@@ -1,0 +1,101 @@
+use bhyve_api::{ReservoirError, VmmCtlFd};
+use clap::Parser;
+
+#[derive(clap::Parser, Debug)]
+struct Opts {
+    #[clap(subcommand)]
+    cmd: Command,
+}
+
+#[derive(clap::Subcommand, Debug)]
+enum Command {
+    /// Add to reservoir capacity
+    Add {
+        /// Size to add (MiB)
+        sz: usize,
+        /// Chunk size (MiB)
+        chunk: Option<usize>,
+    },
+    /// Remove from reservoir capacity
+    Remove {
+        /// Size to remove (MiB)
+        sz: usize,
+        /// Chunk size (MiB)
+        chunk: Option<usize>,
+    },
+    /// Set reservoir capacity
+    Set {
+        /// Target size (MiB)
+        sz: usize,
+        /// Chunk size (MiB)
+        chunk: Option<usize>,
+    },
+    /// Query current reservoir information
+    Query,
+}
+
+const MB: usize = 1024 * 1024;
+
+fn do_resize(
+    ctl: &VmmCtlFd,
+    sz_bytes: usize,
+    chunk_mb: usize,
+) -> std::io::Result<()> {
+    loop {
+        match ctl.reservoir_resize(sz_bytes, chunk_mb * MB) {
+            Err(ReservoirError::Interrupted(sz)) => {
+                println!("Reservoir size: {}MiB", sz / MB);
+            }
+            Err(ReservoirError::Io(e)) => return Err(e),
+            Ok(_) => return Ok(()),
+        }
+    }
+}
+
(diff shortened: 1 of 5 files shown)


bin/propolis-utils/src/bin/rsrvrctl.rs
```diff
@@ -0,0 +1,96 @@
...
+        Command::Set { sz, chunk } => {
+            do_resize(&ctl, sz * MB, chunk.unwrap_or(0))?;
+        }
+        Command::Query => {
+            let sz = ctl.reservoir_query()?;
+            println!(
+                "Free KiB:\t\t\t{}
+Allocated KiB:\t\t\t{}
+Transient Allocated KiB:\t{}
+Size limit KiB:\t\t\t{}",
```
luqmana, line 86: You can line this up in the source too by adding `\` at the end to ignore any leading whitespace on the next line.
```suggestion
                "Free KiB:\t\t\t{}\n\
                 Allocated KiB:\t\t\t{}\n\
                 Transient Allocated KiB:\t{}\n\
                 Size limit KiB:\t\t\t{}",
```

bin/propolis-utils/src/bin/rsrvrctl.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    let ctl = VmmCtlFd::open()?;
+
+    let size_total =
+        |q: bhyve_api::vmm_resv_query| q.vrq_free_sz + q.vrq_alloc_sz;
+
+    match opts.cmd {
+        Command::Add { sz, chunk } => {
+            let cur_sz = ctl.reservoir_query().map(size_total)?;
+            do_resize(&ctl, cur_sz + sz * MB, chunk.unwrap_or(0))?;
```
luqmana, line 66: should this be a `saturating_add` to match the remove case?

pfmooney: Yeah, good catch.

smklein:
FYI, I'm still working through it, but https://github.com/oxidecomputer/omicron/pull/2684 seems related

pfmooney:
> FYI, I'm still working through it, but [oxidecomputer/omicron#2684](https://github.com/oxidecomputer/omicron/pull/2684) seems related

Yeah, I tagged you here so you'd be aware of the API addition so it could be used elsewhere (like in oxidecomputer/omicron#2684)

smklein:
> > FYI, I'm still working through it, but [oxidecomputer/omicron#2684](https://github.com/oxidecomputer/omicron/pull/2684) seems related
> 
> Yeah, I tagged you here so you'd be aware of the API addition so it could be used elsewhere (like in [oxidecomputer/omicron#2684](https://github.com/oxidecomputer/omicron/pull/2684))

Just to clarify:

- This PR adds a binary to `propolis-utils`, which, AFAICT, are developer tools which I should not depend on from Omicron.
- It also adds an interface around an ioctl in the `bhyve-api` crate, which we don't *currently* depend on directly from Omicron.

IMO this is fine; it's great for developers using Propolis, and I'm happy with this PR, but I was planning on just calling the `rsrvrctl` CLI tool directly during initialization from Omicron. Does that seem reasonable to you?

pfmooney:
>     * This PR adds a binary to `propolis-utils`, which, AFAICT, are developer tools which I should not depend on from Omicron.
>     * It also adds an interface around an ioctl in the `bhyve-api` crate, which we don't _currently_ depend on directly from Omicron.

Correct, the `rsrvrctl` added to `propolis-utils` here wasn't meant to be used by Omicron, but rather serve as an example of how one might consume the API added to `bhyve-api`.


>  I was planning on just calling the `rsrvrctl` CLI tool directly during initialization from Omicron. Does that seem reasonable to you?

The interface offered by the in-gate `rsrvrctl` is not Committed or anything (although neither is the ioctl API exposed via `bhyve-api`).  I added the interface to those ioctls so they could be used by sled-agent for this task.  It seems to me like an easier way of achieving it, and being able to handle any errors along the way.

PR #358, opened 2023-03-27
`propolis-standalone` doesn't need usdt-related features

ahl:
Even where supported, the use of the feature(s) in the library is sufficient.

This fixed a build error on macOS


pfmooney (approved):
You mean `propolis-standalone` in your description, not `-server`, right?

ahl:
d'oheth

Issue #357, opened 2023-03-27
want policy for handling time drift between machines in migration


jordanhendricks:
Part of ongoing work for #337 

One aspect of handling time across live migrations is accounting for the delta between reading the timing data on the source side and writing out the time data on the destination side. In this step on the destination, the current implementation of #337 makes calculations based on the wall clock times of both machines. In my testing so far, I have observed negative migration deltas (<2000 usecs) between the source read and the target receipt of the data.

One extremely load-bearing assumption in this method is the idea that NTP is working properly, and both machines' wall clocks are synchronized. The lab machine cluster I'm using uses a public NTP server (see also: https://github.com/oxidecomputer/meta/issues/146), and in production we will have a local NTP server with presumably a much tighter window for the wall clocks.

Even with a smaller window, we still need to consider this case. I think we will want some sort of reasonable bound on an acceptable window (perhaps clamping the delta to 0 for perceived negative deltas), and ideally some sort of way to tune this value. I'm not sure exactly what such a tunable would look like -- perhaps a new API endpoint?

jordanhendricks:
Note: for #359, I opted to clamp negative migrate deltas, should they occur, to 0. We still may want to consider a broader policy here, but given that our implementation assumes a working ntp, I think this is pretty low-priority for now.

PR #359, opened 2023-03-27
Migrate guest time-related data

### LINKED ISSUE #337: implement live migration support for guest timing data

jordanhendricks:
This PR makes use of the new bhyve VMM timing data interfaces introduced in the accompanying bhyve change ([CR link](https://code.illumos.org/c/illumos-gate/+/2711)). See RFD 358 for additional background.

Closes #337.

**Overview**

This change adds the following functionality for guests that are migrated:
- the guest TSC continues functioning at the same frequency, adjusted for live migration time
- device timers continue functioning properly

To do so, propolis-server makes use of the new bhyve VMM timing data interfaces. On the source side, we read the timing data before the lengthier parts of migration begin. Toward the end, we send this data over. The destination side calculates migration time based on the source and destination wall clocks, and adjusts the guest TSC based on migration time, and adjusts the `boot_hrtime` (the value used for device timers) based on the guest uptime, migration time, and the hrtime of its new host.

A load-bearing assumption in our adjustment calculations is that the wall clocks of the source and destination hosts are synchronized, which is an assumption we are making in other parts of the product.

**Bhyve VMM timing data interface details**

On a read, the timing data interface provides the following:
- current guest TSC
- guest TSC frequency
- guest `boot_hrtime` (hrtime of the host when the guest was booted)
- host wall clock
- host hrtime (high res clock)

Callers writing the data provide the guest-related fields, as well as the hrtime and wall clock time of the host when when the caller sent the request. This allows bhyve to make additional adjustments to account for latency between the userspace write and kernel-side receipt of the data.

**Timing Adjustment D script**
This PR adds a couple of probes to where we adjust the timing data to help validate the data were are seeing is correct.

Some example output of this script, run for a VM with no load on on my otherwise not-busy test machine:
```
$ pfexec ./scripts/time-adjustments.d 527
tracing pid 527...
timing data imported; press CTRL+C for summary
^C
GUEST FREQ (Hz)        UPTIME (usec)    TOTAL TIME (usec)
     2649924837              1431678              2818798

EVENT           DURATION (usec)            GUEST TSC          BOOT HRTIME
Migration               2818767           3793847219        3899256607629
[adjustment]                              7469522214          -6033009351
Propolis                     12          11263369433        3893223598278
[adjustment]                                   82044                31221
Kernel                       17          11263451477        3893223629499

```

**Notes for reviewers**

- There are some TODOs in here, and most of them (except the one about writing test) I wouldn't mind feedback on.
- I still need to write some automated tests and write up a bunch of testing notes, but I wanted to start getting eyes on this.

### DIFF
--- a/bin/propolis-server/src/lib/migrate/destination.rs
+++ b/bin/propolis-server/src/lib/migrate/destination.rs
@@ -5,6 +5,7 @@ use propolis::inventory::Entity;
 use propolis::migrate::{
     MigrateCtx, MigrateStateError, Migrator, PayloadOffer, PayloadOffers,
 };
+use propolis::vmm;
 use slog::{error, info, trace, warn};
 use std::convert::TryInto;
 use std::io;
@@ -108,6 +109,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> DestinationProtocol<T> {
                 self.ram_push(&step).await
             }
             MigratePhase::DeviceState => self.device_state().await,
+            MigratePhase::TimeData => self.time_data().await,
             MigratePhase::RamPull => self.ram_pull().await,
             MigratePhase::ServerState => self.server_state().await,
             MigratePhase::Finish => self.finish().await,
@@ -129,6 +131,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> DestinationProtocol<T> {
         // pre- and post-pause steps.
         self.run_phase(MigratePhase::RamPushPrePause).await?;
         self.run_phase(MigratePhase::RamPushPostPause).await?;
+
+        // Import of the time data *must* be done before we import device
+        // state: the proper functioning of device timers depends on an adjusted
+        // boot_hrtime.
+        self.run_phase(MigratePhase::TimeData).await?;
         self.run_phase(MigratePhase::DeviceState).await?;
         self.run_phase(MigratePhase::RamPull).await?;
         self.run_phase(MigratePhase::ServerState).await?;
@@ -321,6 +328,110 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> DestinationProtocol<T> {
                 self.import_device(&target, &device, &migrate_ctx)?;
             }
         }
+        self.send_msg(codec::Message::Okay).await
+    }
+
+    // Get the guest time data from the source, make updates to it based on the
+    // new host, and write the data out to bhvye.
+    async fn time_data(&mut self) -> Result<(), MigrateError> {
+        // Read time data sent by the source and deserialize
+        let raw: String = match self.read_msg().await? {
+            codec::Message::Serialized(encoded) => encoded,
+            msg => {
+                error!(self.log(), "time data: unexpected message: {msg:?}");
+                return Err(MigrateError::UnexpectedMessage);
+            }
+        };
+        info!(self.log(), "VMM Time Data: {:?}", raw);
+        let time_data_src: vmm::time::VmTimeData = ron::from_str(&raw)
+            .map_err(|e| {
+                MigrateError::TimeData(format!(
+                    "VMM Time Data deserialization error: {}",
+                    e
+                ))
+            })?;
+        probes::migrate_time_data_before!(|| {
+            (
+                time_data_src.guest_freq,
+                time_data_src.guest_tsc,
+                time_data_src.boot_hrtime,
+            )
+        });
+
+        // Take a snapshot of the host hrtime/wall clock time, then adjust
+        // time data appropriately.
+        let vmm_hdl = {
+            let instance_guard = self.vm_controller.instance().lock();
+            &instance_guard.machine().hdl.clone()
+        };
+        let (dst_hrt, dst_wc) = vmm::time::host_time_snapshot(vmm_hdl)
+            .map_err(|e| {
+                MigrateError::TimeData(format!(
+                    "could not read host time: {}",
+                    e
+                ))
+            })?;
+        let (time_data_dst, adjust) =
+            vmm::time::adjust_time_data(time_data_src, dst_hrt, dst_wc)
+                .map_err(|e| {
+                    MigrateError::TimeData(format!(
+                        "could not adjust VMM Time Data: {}",
+                        e
+                    ))
+                })?;
+
+        // In case import fails, log adjustments made to time data and fire
+        // dtrace probe first
+        if adjust.migrate_delta_negative {
+            warn!(
+                self.log(),
+                "Found negative wall clock delta between target import \
+                and source export:\n\
+                - source wall clock time: {:?}\n\
+                - target wall clock time: {:?}\n",
+                time_data_src.wall_clock(),
+                dst_wc
+            );
+        }
+        info!(
+            self.log(),
+            "Time data adjustments:\n\
+            - guest TSC freq: {} Hz = {} GHz\n\
+            - guest uptime ns: {:?}\n\
+            - migration time delta: {:?}\n\
+            - guest_tsc adjustment = {} + {} = {}\n\
+            - boot_hrtime adjustment = {} ---> {} - {} = {}\n\
+            - dest highres clock time: {}\n\
+            - dest wall clock time: {:?}",
+            time_data_dst.guest_freq,
+            time_data_dst.guest_freq as f64 / vmm::time::NS_PER_SEC as f64,
+            adjust.guest_uptime_ns,
+            adjust.migrate_delta,
+            time_data_src.guest_tsc,
+            adjust.guest_tsc_delta,
+            time_data_dst.guest_tsc,
+            time_data_src.boot_hrtime,
+            dst_hrt,
+            adjust.boot_hrtime_delta,
+            time_data_dst.boot_hrtime,
+            dst_hrt,
+            dst_wc
+        );
+        probes::migrate_time_data_after!(|| {
+            (
+                time_data_dst.guest_freq,
+                time_data_dst.guest_tsc,
+                time_data_dst.boot_hrtime,
+                adjust.guest_uptime_ns,
+                adjust.migrate_delta.as_nanos() as u64,
+                adjust.migrate_delta_negative,
+            )
+        });
+
+        // Import the adjusted time data
+        vmm::time::import_time_data(vmm_hdl, time_data_dst).map_err(|e| {
+            MigrateError::TimeData(format!("VMM Time Data import error: {}", e))
+        })?;
 
         self.send_msg(codec::Message::Okay).await
     }
--- a/bin/propolis-server/src/lib/migrate/source.rs
+++ b/bin/propolis-server/src/lib/migrate/source.rs
@@ -4,6 +4,7 @@ use propolis::inventory::Order;
 use propolis::migrate::{
     MigrateCtx, MigrateStateError, Migrator, PayloadOutputs,
 };
+use propolis::vmm;
 use slog::{error, info, trace};
 use std::convert::TryInto;
 use std::io;
@@ -112,6 +113,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
             MigratePhase::RamPushPrePause | MigratePhase::RamPushPostPause => {
                 self.ram_push(&step).await
             }
+            MigratePhase::TimeData => self.time_data().await,
             MigratePhase::DeviceState => self.device_state().await,
             MigratePhase::RamPull => self.ram_pull().await,
             MigratePhase::ServerState => self.server_state().await,
@@ -130,6 +132,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
         self.run_phase(MigratePhase::RamPushPrePause).await?;
         self.run_phase(MigratePhase::Pause).await?;
         self.run_phase(MigratePhase::RamPushPostPause).await?;
+        self.run_phase(MigratePhase::TimeData).await?;
         self.run_phase(MigratePhase::DeviceState).await?;
         self.run_phase(MigratePhase::RamPull).await?;
         self.run_phase(MigratePhase::ServerState).await?;
@@ -386,6 +389,29 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
         .await?;
 
         self.send_msg(codec::Message::Okay).await?;
+        self.read_ok().await
+    }
+
+    // Read and send over the time data
+    async fn time_data(&mut self) -> Result<(), MigrateError> {
+        let vmm_hdl = {
+            let instance_guard = self.vm_controller.instance().lock();
+            &instance_guard.machine().hdl.clone()
+        };
+        let vm_time_data =
+            vmm::time::export_time_data(vmm_hdl).map_err(|e| {
+                MigrateError::TimeData(format!(
+                    "VMM Time Data export error: {}",
+                    e
+                ))
+            })?;
+        info!(self.log(), "VMM Time Data: {:#?}", vm_time_data);
+
+        let time_data_serialized = ron::ser::to_string(&vm_time_data)
+            .map_err(codec::ProtocolError::from)?;
+        info!(self.log(), "VMM Time Data: {:#?}", time_data_serialized);
+        self.send_msg(codec::Message::Serialized(time_data_serialized)).await?;
+
         self.read_ok().await
     }
 
--- a/lib/propolis/src/vmm/hdl.rs
+++ b/lib/propolis/src/vmm/hdl.rs
@@ -17,7 +17,6 @@ use std::time::Duration;
 use crate::common::PAGE_SIZE;
 use crate::vmm::mem::Prot;
 
-#[derive(Default, Copy, Clone)]
 /// Configurable options for VMM instance creation
 ///
 /// # Options:
@@ -26,6 +25,7 @@ use crate::vmm::mem::Prot;
 /// - `use_reservoir`: Allocate guest memory (only) from the VMM reservoir.  If
 /// this is enabled, and memory in excess of what is available from the
 /// reservoir is requested, creation of that guest memory resource will fail.
+#[derive(Default, Copy, Clone)]
 pub struct CreateOpts {
     pub force: bool,
     pub use_reservoir: bool,
@@ -418,7 +418,7 @@ impl VmmHdl {
 #[cfg(test)]
 impl VmmHdl {
     /// Build a VmmHdl instance suitable for unit tests, but nothing else, since
-    /// it will not be backed by any real vmm reousrces.
+    /// it will not be backed by any real vmm resources.
     pub(crate) fn new_test(mem_size: usize) -> Result<Self> {
         use tempfile::tempfile;
         let fp = tempfile()?;
--- a/lib/propolis/src/vmm/time.rs
+++ b/lib/propolis/src/vmm/time.rs
@@ -1,0 +1,788 @@
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
+use thiserror::Error;
+
+use super::VmmHdl;
+
+pub const NS_PER_SEC: u64 = 1_000_000_000;
+pub const SEC_PER_DAY: u64 = 24 * 60 * 60;
+
+/// Representation of guest time data state
+///
+/// This is serialized/deserialized as part of the migration protocol
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+pub struct VmTimeData {
+    /// guest TSC frequency (hz)
+    pub guest_freq: u64,
+
+    /// current guest TSC
+    pub guest_tsc: u64,
+
+    /// monotonic host clock (ns)
+    pub hrtime: i64,
+
+    /// wall clock host clock (sec)
+    pub hres_sec: u64,
+
+    /// wall clock host clock (ns)
+    pub hres_ns: u64,
+
+    /// guest boot_hrtime (can be negative)
+    pub boot_hrtime: i64,
+}
+
+/// A collection of data about adjustments made to the time data to enable
+/// richer log messages
+#[derive(Debug)]
+pub struct VmTimeDataAdjustments {
+    pub guest_uptime_ns: u64,
+    pub migrate_delta: Duration,
+    pub migrate_delta_negative: bool,
+    pub guest_tsc_delta: u64,
+    pub boot_hrtime_delta: u64,
+}
+
+impl VmTimeData {
+    pub fn wall_clock(&self) -> Duration {
+        Duration::new(self.hres_sec, self.hres_ns as u32)
+    }
+}
+
+impl From<bhyve_api::vdi_time_info_v1> for VmTimeData {
+    fn from(raw: bhyve_api::vdi_time_info_v1) -> Self {
+        Self {
+            guest_freq: raw.vt_guest_freq,
+            guest_tsc: raw.vt_guest_tsc,
+            hrtime: raw.vt_hrtime,
+            hres_sec: raw.vt_hres_sec,
+            hres_ns: raw.vt_hres_ns,
+            boot_hrtime: raw.vt_boot_hrtime,
+        }
+    }
+}
+impl From<VmTimeData> for bhyve_api::vdi_time_info_v1 {
+    fn from(info: VmTimeData) -> Self {
+        bhyve_api::vdi_time_info_v1 {
+            vt_guest_freq: info.guest_freq,
+            vt_guest_tsc: info.guest_tsc,
+            vt_hrtime: info.hrtime,
+            vt_hres_sec: info.hres_sec,
+            vt_hres_ns: info.hres_ns,
+            vt_boot_hrtime: info.boot_hrtime,
+        }
+    }
+}
+
+pub fn import_time_data(
+    hdl: &VmmHdl,
+    time_info: VmTimeData,
+) -> std::io::Result<()> {
+    let raw = bhyve_api::vdi_time_info_v1::from(time_info);
+    crate::vmm::data::write(hdl, -1, bhyve_api::VDC_VMM_TIME, 1, raw)?;
+
+    Ok(())
+}
+
+pub fn export_time_data(hdl: &VmmHdl) -> std::io::Result<VmTimeData> {
+    let time_info: bhyve_api::vdi_time_info_v1 =
+        crate::vmm::data::read(hdl, -1, bhyve_api::VDC_VMM_TIME, 1)?;
+
+    Ok(VmTimeData::from(time_info))
+}
+
+/// Returns the current host hrtime and wall clock time
+//
+// The current host hrtime and wall clock time are exposed via the VMM time data
+// interface.
+//
+// The kernel side of the interface disables interrupts while it takes the clock
+// readings; in the absence of a function to translate between the two clock
+// values, this is a best effort way to read the hrtime and wall clock times as
+// close to as possible at the same point in time. Thus fishing this data out of
+// the VMM time data read payload is strictly better than calling
+// clock_gettime(3c) twice from userspace.
+pub fn host_time_snapshot(hdl: &VmmHdl) -> std::io::Result<(i64, Duration)> {
+    let ti = export_time_data(hdl)?;
+    let wc = Duration::new(ti.hres_sec, ti.hres_ns as u32);
+    let hrt = ti.hrtime;
+
+    Ok((hrt, wc))
+}
+
+/// Given an input representation of guest time data on a source host, and a
+/// current host hrtime and wallclock time on the target host, output an
+/// "adjusted" view of the guest time data. This data can be imported to bhyve
+/// to allow guest time (namely, the guest TSC and its device timers) to allow
+/// the guest's sense of time to function properly on the target.
+//  See comments inline for more details about how we calculate a new guest TSC
+//  and boot_hrtime.
+pub fn adjust_time_data(
+    src: VmTimeData,
+    dst_hrt: i64,
+    dst_wc: Duration,
+) -> Result<(VmTimeData, VmTimeDataAdjustments), TimeAdjustError> {
+    // Basic validation: there is no reason system hrtime should be negative,
+    // and other calculations assume that, so validate that first
+    if dst_hrt < 0 {
+        return Err(TimeAdjustError::Hrtime { hrtime: dst_hrt });
+    }
+
+    // Find delta between export on source and import on target using wall clock
+    // time. This delta is used for adjusting the TSC and boot_hrtime.
+    //
+    // We expect to be operating on machines with well-synchronized wall
+    // clocks, so using wall clock time is a useful shorthand for observing how
+    // much time has passed. If for some reason we see a negative delta (see
+    // also: #357), clamp the delta to 0.
+    //
+    // migrate_delta = target wall clock - source wall clock
+    let (migrate_delta, migrate_delta_negative) =
+        match dst_wc.checked_sub(src.wall_clock()) {
+            Some(d) => (d, false),
+            None => (Duration::from_secs(0), true),
+        };
+
+    if migrate_delta.as_nanos() > (i64::MAX as u128) {
+        // migrate delta won't fit in hrtime calculations
+        return Err(TimeAdjustError::InvalidMigrateDelta {
+            src_wc: src.wall_clock(),
+            dst_wc,
+        });
+    }
+    let migrate_delta_ns = migrate_delta.as_nanos() as i64;
+    assert!(migrate_delta_ns >= 0, "migrate delta cannot be negative");
+
+    // Find a new boot_hrtime for the guest
+    //
+    // Device timers are scheduled based on hrtime of the host: For example, a
+    // timer that should fire after 1 second is scheduled as: host hrtime + 1s.
+    //
+    // When devices are exported for migration, time values are normalized
+    // against the guest's boot_hrtime on export, and de-normalized against
+    // boot_hrtime on import. The boot_hrtime of the guest is set to the hrtime
+    // of when the guest booted. Because this value is used on import to fix up
+    // timer values, it is critical to set this value prior to importing device
+    // state such that existing timers are normalized correctly. As with booting
+    // a guest, on the target, it should be set to the hrtime of the host when
+    // the guest would have booted, had it booted on that target.
+    //
+    // An example may be helpful. Consider a guest that has 5 days of uptime,
+    // booted on a host with 30 days of uptime. Suppose that guest is migrated
+    // with a device timer that should fire 1 second in the future.
+    //
+    // +=================================================================+
+    // | hrtime (source) | guest hrtime | boot_hrtime  | timer value     |
+    // +-----------------------------------------------------------------+
+    // | 30 days         | 5 days       |(30 - 5) days | src hrtime + 1s |
+    // |                 |              | 25 days      | 30 days + 1s    |
+    // +=================================================================+
+    //
+    // Suppose the guest is then migrated to a host with 100 days of uptime.
+    // On migration, the existing timer is normalized before export by
+    // subtracting out boot_hrtime:
+    //       normalized = timer - boot_hrtime
+    //                  = (30 days + 1 sec) - 25 days
+    //                  = 5 days + 1 sec
+    //
+    // When the timer is imported, it is denormalized by adding back in
+    // the new boot_hrtime. The timer should still fire 1 second from the
+    // current hrtime of the host. The target hrtime is 100 days, so the timer
+    // should fire at 100 days + 1 sec.
+    //
+    // Working backwards to get the new boot_hrtime, we have:
+    //
+    //       denormalized = normalized + boot_hrtime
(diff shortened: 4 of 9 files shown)


lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,348 @@
...
+/// uptime = hrtime - boot_hrtime
+pub fn calc_guest_uptime(
+    src_hrt: u64,
+    boot_hrtime: Hrtime,
+) -> Result<Duration, TimeAdjustError> {
+    let src_hrt_ns: i128 = src_hrt as i128;
+    let boot_hrt_ns: i128 = boot_hrtime as i128;
+    let uptime: u128;
+
+    match src_hrt_ns.checked_sub(boot_hrt_ns) {
```
jordanhendricks, line 176: Re-reading through and noticed this check is wrong -- will fix in a followup.

scripts/time-adjustments.d
```diff
@@ -0,0 +1,141 @@
...
+	this->krn_d_tsc = krn_adj_gtsc - usr_adj_gtsc;
+	this->krn_d_bhrt = krn_adj_bhrt - usr_adj_bhrt;
+
+	printf("%15s %20s %20s\n",
+	    "GUEST FREQ (Hz)", "UPTIME (usec)", "TOTAL TIME (usec)");
+	printf("%15lu %20lu %20lu\n", start_gf,vm_uptime / 1000,
+	    this->total / 1000);
+	printf("\n");
+
+	printf("%-10s %20s %20s %20s\n",
```
jordanhendricks, line 128: Noticed that the output ordering here is a bit confusing -- will fix in a follow up

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,348 @@
...
+            return Err(TimeAdjustError::RealTimeClock(e.to_string()));
+        }
+    }
+
+    Ok(HostTime { hrtime, wall_clock })
+}
+
+/// Find the perceived wall clock time difference between when timing data
+/// was read on the source versus the current time on the destination.
+// TODO(#357): Right now we throw up our hands if the delta is negative.
```
gjcolombo, line 149: Could we avoid this problem by trying to measure elapsed migration time from the point of view of just one of the actors involved?

That is, suppose the Pause phase included some kind of mutual acknowledgement in the protocol:

- Source sends an "I'm pausing" message
- Target records the time it received this message and immediately sends an "OK to pause" message
- Only then does the source start pausing

(This could also be an "I read timing data"/"OK you read the timing data" message pair if that's what we want; I'll probably work it out as I keep reviewing this.)

This can produce a "total migration time" estimate on the target that's always *at least* the amount of time the VM was actually paused. It might be an overestimate because it also includes the time it took for the source to receive the "OK to pause now" message and actually start pausing things. We could possibly correct for that further by having the *source* measure some length of time (e.g. from when it sent its "am I OK to pause?" message to when it actually paused things) and then send that to the target to use to correct its view of the total migration time.

Anyway: I haven't totally worked out the arithmetic here, so I could be totally wrong about all this, but I'd be interested in seeing if there's an approach that doesn't rely on real time synchronization between the two machines involved.

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,348 @@
...
+        tsc: u64,
+
+        /// calculated TSC adjustment
+        adjust: u64,
+    },
+}
+
+/// Returns a non-atomic snapshot of the current hrtime and wall clock time.
+// TODO: We may want to put a bound in here to make sure the two
+// values are close enough, and we may want that to be tunable. What's the
```
gjcolombo, line 119: To clarify: close enough to what? IIUC these two values aren't directly comparable (since CLOCK_REALTIME uses the epoch as a base but CLOCK_HIGHRES has an arbitrary base), though I suppose you could do a high-low-high kind of thing where you read one, then read the other, then read the first one again and make sure the delta from the first reading is less than X (and you can retry if it's not).

gjcolombo (commented):
Thanks for putting this together! Left a few initial thoughts (hopefully nothing too off-base with the bhyve changes, which I've skimmed but not looked at terribly closely yet).

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -60,6 +60,18 @@ struct SourceProtocol<T: AsyncRead + AsyncWrite + Unpin + Send> {
 
     /// Transport to the destination Instance.
     conn: WebSocketStream<T>,
+
+    // TODO: A hacky way for now to hang onto the VMM timing data between
+    // migration phases.
+    //
+    // We want to read the VMM timing data as soon as we can after we pause the
```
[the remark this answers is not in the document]
pfmooney: > We want to read the VMM timing data as soon as we can after we pause the
> source, and make adjustments on the destination as close as we can to
> the end of the migration.

With the way things are structured, I'm not sure that's necessary for accuracy?

As Greg notes, we could change some of the ordering around when userspace device entities are paused vs the vCPUs.  I imagine we'll end up evolving those events with more time (such as adding a stage when disk/network IO is throttled, but the devices themselves aren't paused).

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -60,6 +60,18 @@ struct SourceProtocol<T: AsyncRead + AsyncWrite + Unpin + Send> {
...
+    // source, and make adjustments on the destination as close as we can to
+    // the end of the migration.
+    //
+    // This lets us hang on to the data between export in the pause phase,
```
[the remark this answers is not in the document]
pfmooney: > The destination has to compute the total amount of time the VM spent paused in any case

I'm not sure it so much has to calculate how much time the VM was paused, but rather how much time passed between when the reading of timing (TSC, etc) information was taken, and when its adjusted value(s) are being loaded on the destination.  The destination VM could remain paused for additional time past that, but it would appear no different than a case where vCPU threads are not being scheduled for some reason.

Contextually, it is important that the timing information be loaded first, before other device state, since the adjusted `boot_hrtime` value is used to offset the other timer values in the emulated device state.

lib/propolis/src/vmm/hdl.rs
```diff
@@ -492,20 +662,52 @@ pub mod migrate {
         }
     }
 
+    impl From<vdi_timing_info_v1> for TimingInfoV1 {
+        fn from(raw: vdi_timing_info_v1) -> Self {
+            Self {
+                guest_freq: raw.vt_guest_freq,
+                guest_tsc: raw.vt_guest_tsc,
+                hrtime: raw.vt_hrtime as u64,
+                // TODO: u64 to u32 cast
```
pfmooney, line 671: Why the TODO?

lib/propolis/src/vmm/hdl.rs
```diff
@@ -404,21 +414,148 @@ impl VmmHdl {
...
+    /// when these adjustments were made
+    /// - (`hrestime`)[TimingInfoV1::hrestime]: the host hrestime (wall clock)
+    /// when these adjustments were made
+    ///
+    /// See the [`time_adjust`] module for more details about how the adjustment
+    /// calculations here are performed.
+    fn adjust_timing_data(
+        &self,
+        timing_data: &mut TimingInfoV1,
+    ) -> std::result::Result<(), MigrateStateError> {
```
pfmooney, line 481: Might make it easier to reason about this function (and/or instrument it with logging) if it took a TimingInfoV1 argument, and emitted one through its Result type, rather than a mutable reference.  It'd be a pain to do in C (returning structs and all), but it seems more rust-like in this context. *shrug*

lib/propolis/src/vmm/time_adjust.rs
pfmooney: Using the wallclock difference (and assuming the respective clocks on either side are adequately synced via NTP) seemed like a simple way to get this off the ground.  Ultimately we want some means of judging the passage of time (from some global observer) between the point at which the timing measurement (TSC, boot_hrtime, etc) is taken on the source, and when its adjusted values are to be loaded on the destination.

It's been suggested that we could do more sophisticated things here in the future, including something like a little mini implementation of the NTP algo between the two propolis instances, but cheating with wallclock in the short term seemed like a reasonable starting point.

lib/propolis/src/vmm/hdl.rs
```diff
@@ -477,6 +629,24 @@ pub mod migrate {
...
+        pub guest_freq: u64,
+
+        /// guest TSC
+        pub guest_tsc: u64,
+
+        /// monotonic host clock (ns)
+        pub hrtime: u64,
+
+        /// wall clock host clock
+        pub hrestime: Duration,
```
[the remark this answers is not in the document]
jordanhendricks: I liked `Duration` because there are a bunch of existing functions for converting between it and nanoseconds, and for doing checked addition/subtraction to deal with overflow. Would you suggest something different?

bin/propolis-server/src/lib/migrate/source.rs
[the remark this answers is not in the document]
jordanhendricks: > How deeply do we care about which protocol phase sends this data?

As Patrick said, it needs to be sent before the device state is imported (which this current change does *not* do), but also we are trying to push most of the adjustments made into userspace, which implies we need to collect the data before any lengthier pieces of migration, where possible. Perhaps I was a bit too constraining when describing this in terms of when we pause the vCPU, but it's been my mental model for thinking about it. I can change some of the comments here to reflect that.

I had avoided reintroducing the ArchState phase in the hopes that it would be cleaner given that we split up the collecting of time data and the sending of it. But perhaps that isn't true, as it might be harder to grok that this data must be imported before the device state is imported.

I could add the ArchState phase back as such: Pause (with time data collection) -> RamPush -> ArchState (time data send) -> DeviceState...

How does that sound?

lib/propolis/src/vmm/time_adjust.rs
jordanhendricks: FWIW, the problem I ran into with a negative wall clock delta was in an earlier prototype where the time data was sent immediately after collecting it. In practice (if not for longer delay between reading and sending, but hopefully also because ntp is better behaved than on the lab machines I was on), my hope is that this won't matter, and I think I'm comfortable with leaving this behavior for now. When we begin to optimize things such that migrations are faster, we may want to reconsider it (and I'm happy to leave #357 as a placeholder for that).

lib/propolis/src/vmm/hdl.rs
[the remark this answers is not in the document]
jordanhendricks: (If you think it would be better to use sec/ns as explicit fields in the serde type and convert internally to a Duration, that seems fine too)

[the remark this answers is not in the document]
pfmooney: I would prefer explicit fields, since the internal fields of `Duration` are considered private.

As for internal calculations, using `Duration` for convenience and safety sounds ideal.

bin/propolis-server/src/lib/migrate/source.rs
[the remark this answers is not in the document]
jordanhendricks: In 803081, I added:
- a TimeDataRead step on the source side
- a TimeData phase for both sides to exchange the data

lib/propolis/src/vmm/time_adjust.rs
jordanhendricks: reworked this to just do another read of the VMM_TIME data (which exposes the wall clock / hrtime) in 803081

crates/time/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+    }
+}
+
+/// Get the system wall clock time (not monotonic)
+pub fn get_real_time() -> Result<Duration> {
+    match sys::clock_gettime(sys::CLOCK_REALTIME) {
+        Ok(ts) => Ok(Duration::new(ts.tv_sec as u64, ts.tv_nsec as u32)),
+        Err(e) => Err(e),
+    }
+}
```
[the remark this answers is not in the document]
jordanhendricks: I ripped this whole thing out.

lib/propolis/src/vmm/hdl.rs
```diff
@@ -121,6 +130,7 @@ pub struct VmmHdl {
     pub(super) inner: bhyve_api::VmmFd,
     destroyed: AtomicBool,
     name: String,
+    log: Option<Logger>,
```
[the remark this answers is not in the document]
jordanhendricks: done (`803081`)

lib/propolis/src/vmm/hdl.rs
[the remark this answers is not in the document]
jordanhendricks: done (`803081`)

lib/propolis/src/vmm/hdl.rs
jordanhendricks: done (`803081`)

lib/propolis/src/vmm/hdl.rs
jordanhendricks: (this is gone now)

jordanhendricks:
Got through most of the review feedback in `803081`, and changed the bhyve-interfaces and naming conventions to reflect the current state of the bhyve bits. I added a new phase for exchanging the time data. I want to do a bit more refactoring there to only import the time data pieces (and ignore the arch entries).

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,550 @@
...
+    if let Some(v) = delta_ns.checked_mul(guest_hz as u128) {
+        upper = v;
+    } else {
+        return Err(TimeAdjustError::TscAdjustOverflow {
+            desc: "migrate_delta * guest_hz".to_string(),
+            migrate_delta,
+            guest_hz,
+            tsc_adjust,
+        });
+    }
```
pfmooney, line 241:
```rust
let upper = delta_ns.checked_mul(guest_hz as u128)
    .ok_or_else(|| TimeAdjsutError::....)?;
```

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    tsc_adjust = upper / NS_PER_SEC;
```
pfmooney, line 243: Use `checked_div`, since we're being so careful with math everywhere else?

crates/bhyve-api/sys/src/vmm_data.rs
```diff
@@ -189,16 +190,23 @@ impl Default for vdi_rtc_v1 {
     }
 }
 
+// VDC_VMM_TIME v1 interface
+
+#[repr(C)]
+#[derive(Copy, Clone, Default, Serialize, Deserialize)]
```
pfmooney, line 196: A few of the other devices are sending their vmm_data structs over the wire as the migration payload out of laziness (coming work will eliminate that).  Since that's not the case for the time data (since you have your own migrate type), it should be fine to omit `Serialize` and `Deserialize` here.

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -245,6 +261,16 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
         }
     }
 
+    async fn time_data_read(&mut self) -> Result<(), MigrateError> {
+        let instance_guard = self.vm_controller.instance().lock();
+        let vmm_hdl = &instance_guard.machine().hdl;
+        let raw = vmm_hdl.export_vm()?;
```
gjcolombo, line 267: There's a subtlety in here that I think would be good to fix: `export_vm` returns a `MigrateStateError`, which has a `From<std::io:Error>` impl that converts all errors into `MigrateStateError::ImportFailed`, which is confusing since we're exporting here. (This hasn't been a problem before because entity state export is infallible.)

I'm not sure of the best way to fix this. We could remove the `From` impl but that would break all the uses of `?` in the import routines. Maybe we need a `StateExportError` enum for `export_vm` to use?

jordanhendricks:
cdbe3ef is a medium-sized refactor and addresses a hodge podge of other review feedback. Specifically:
- 627be132 ripped out the `migrate` module for `VmmHdl` and suggested an alternate way of handling some of the global VM data; I did something similar here by creating a `VmTimeData` struct to represent the abstract notion of guest time data. This is what gets serialized/deserialized in the protocol, and there are explicit functions to export/import this state that we can build versioning into as needed.
- The logic in the former `time_adjust` module was moved to live with the `VmTimeData` state in `lib/propolis/src/vmm/time.rs`; basically all of the guest time -related utilities (import/export, adjustment calculations) there in that module now.
- The logic around the destination side reading the source time data, adjusting it, and importing it now lives with the rest of the protocol code, calling into the `time` module for making adjustments to it and importing it. I think this makes a bit more sense, as it is the consumer of the import/export interfaces, and it makes more sense to me that the migration code would be responsible for this (as opposed to the `import` function, as things were structured before).

bin/propolis-server/src/lib/migrate/source.rs
jordanhendricks: Yeah, this is a great point. I took a stab at making error handling a bit better in cdbe3ef; there is a now a `MigrateError::TimeData` type that we convert all TimeData{Read} phase errors into. Let me know what you think.

crates/bhyve-api/sys/src/vmm_data.rs
jordanhendricks: done in cdbe3ef

lib/propolis/src/vmm/time_adjust.rs
jordanhendricks: done in cdbe3ef

lib/propolis/src/vmm/time_adjust.rs
jordanhendricks: done in cdbe3ef

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    // Note: TryFrom<u128> for u64 is currently unstable
+    if uptime > u64::MAX as u128 {
+        return Err(TimeAdjustError::GuestUptimeOverflow {
+            src_hrt: src_hrt_ns,
+            boot_hrtime: boot_hrt_ns,
+        });
+    }
+
+    Ok(Duration::from_nanos(uptime as u64))
```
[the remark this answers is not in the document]
jordanhendricks: This just falls out of the way the calling code is structured -- the boot_hrtime delta is the uptime + the migrate delta. The migrate delta comes from the wall clock, which is a Duration.

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,550 @@
...
+/// boot_hrtime_delta = uptime + migrate_delta
+pub fn calc_boot_hrtime_delta(
+    uptime: Duration,
+    migrate_delta: Duration,
+) -> Result<Duration, TimeAdjustError> {
+    match uptime.checked_add(migrate_delta) {
+        Some(v) => Ok(v),
+        None => {
+            Err(TimeAdjustError::TimeDeltaOverflow { uptime, migrate_delta })
+        }
```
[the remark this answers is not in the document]
jordanhendricks: done in cdbe3ef

lib/propolis/src/vmm/hdl.rs
```diff
@@ -404,24 +409,156 @@ impl VmmHdl {
...
+    ) -> std::result::Result<TimeInfoV1, MigrateStateError> {
+        info!(log, "Adjusting time data for guest: {:#?}", time_data);
+        probes::adj_time_begin!(|| (
+            time_data.guest_freq,
+            time_data.guest_tsc,
+            time_data.boot_hrtime,
+        ));
+
+        // Take a snapshot of time on this host.
+        let (dst_hrtime, dst_wc) = migrate::vmm_time_snapshot(self).unwrap();
```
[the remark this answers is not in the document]
jordanhendricks: this call was moved into the destination migration protocol, and its error is handled there as of cdbe3ef

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    /// Invalid calculated boot_hrtime
+    #[error(
+        "guest boot_hrtime cannot be represented:\
+            desc=\"{desc}\", total_delta={total_delta:?},\
+            dst_hrtime={dst_hrtime:?}"
+    )]
+    BootHrtimeOverflow {
+        /// error description
+        desc: String,
```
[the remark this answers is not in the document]
jordanhendricks: done in cdbe3ef

lib/propolis/src/vmm/hdl.rs
```diff
@@ -492,34 +664,80 @@ pub mod migrate {
...
+            vmm::data::write(hdl, -1, bhyve_api::VDC_VMM_TIME, 1, time_info)?;
+
+            Ok(())
+        }
+    }
+
     impl BhyveVmV1 {
         pub(super) fn read(hdl: &VmmHdl) -> io::Result<Self> {
             let arch_entries: Vec<bhyve_api::vdi_field_entry_v1> =
                 vmm::data::read_many(hdl, -1, bhyve_api::VDC_VMM_ARCH, 1)?;
```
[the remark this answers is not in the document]
jordanhendricks: this was ripped out as of 627be13

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    Ok(tsc_adjust as u64)
+}
+
+/// Calculate the new guest TSC, given an adjustment.
+pub fn calc_guest_tsc(tsc: u64, adjust: u64) -> Result<u64, TimeAdjustError> {
+    match tsc.checked_add(adjust) {
+        Some(new_tsc) => Ok(new_tsc),
+        None => Err(TimeAdjustError::GuestTscOverflow { tsc, adjust }),
+    }
```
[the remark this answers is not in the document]
jordanhendricks: good point; done in cdbe3ef

lib/propolis/src/vmm/hdl.rs
```diff
@@ -404,24 +409,156 @@ impl VmmHdl {
...
+    /// - (`hrestime`)[TimeInfoV1::hrestime]: the host hrestime (wall clock)
+    /// when these adjustments were made
+    ///
+    /// See the [`time_adjust`] module for more details about how the adjustment
+    /// calculations here are performed.
+    fn adjust_time_data(
+        &self,
+        time_data: TimeInfoV1,
+        log: &Logger,
+    ) -> std::result::Result<TimeInfoV1, MigrateStateError> {
```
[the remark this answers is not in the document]
jordanhendricks: update as of cdbe3ef:

Long ago, I ended up on the structure of hiding the various calculations behind function calls after first inlining everything and finding it a bit hard to follow. Having function names allows provides a semantic foothold for what each calculation is doing and makes them easy to unit test, as you point out. I took another stab at inlining the calculations in this round of work on the PR, and I still felt similarly -- the function with checked operations and custom errors just gets really unwieldy, and I find it easier to reason about what's happening with each step if those bits are pulled out.

I did like the idea of moving the host time call out of this function, so I did that.

Moving logging out of the adjustment function I think is a bit harder, as I've found it useful in testing and debugging to log the specific adjustments made in detail as it is now. The probes could be moved out, I suppose, but I think this is fine as is?

jordanhendricks:
I have a couple of other small review comments to wrap up still, but wanted to point out, the PHD live migration tests passed again :tada:

bin/propolis-server/src/lib/migrate/source.rs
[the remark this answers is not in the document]
jordanhendricks: this is all restructured as of cdbe3ef, though I still kept around a field to hang on to time data state between export and send in the protocol

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -61,6 +62,10 @@ struct SourceProtocol<T: AsyncRead + AsyncWrite + Unpin + Send> {
 
     /// Transport to the destination Instance.
     conn: WebSocketStream<T>,
+
+    /// Saved VM Time Data, which is read earlier in the protocol (TimeDataRead
+    /// phase) than it is sent (TimeData phase)
+    time_data: Option<propolis::vmm::time::VmTimeData>,
```
pfmooney, line 68: Chatted some about why this was implemented as two different phases.  If they can be consolidated into a single phase, we probably don't need to keep this around.

pfmooney (commented):
Will take a more thorough pass when you've pushed pending updates

jordanhendricks:
Per recent discussion about landing this ASAP, I squashed my commits and force-pushed after rebasing with master to minimize merge conflicts with other recent changes in master.

Additional notes:
- I removed the change that turned PHD migration tests back on
- This change is clippy clean

Testing notes:
- I ran a handful of physical migrations back and forth between machines. Everything migrated successfully
- `cargo test` passes on my linux laptop and helios dev machine

I tried to run the PHD tests locally but hit an unrelated problem; working on sorting that out.

Miscellaneous output for posterity:

**live-migration-times.d** (source)
```
jordan@maxwell ~/propolis/scripts $ pfexec ./live-migration-times.d 1533 v                             
tracing live migration protocol times for pid 1533...                                                  
                                                                                                       
PHASE                                        TIMESTAMP                                                 
Sync         BEGIN                      17437355109672
Sync         END                        17437355230924
RamPushPrePause BEGIN                      17437355249524      
RamPushPrePause END                        17446758097054
RamPushPostPause BEGIN                      17446758114444                                             
RamPushPostPause END                        17446960120090                                             
TimeData     BEGIN                      17446960141050
TimeData     END                        17446960297073 
DeviceState  BEGIN                      17446960307583                                                 
DeviceState  END                        17447021682763                                                 
RamPull      BEGIN                      17447021689593
RamPull      END                        17447079199443                                                 
ServerState  BEGIN                      17447079210023         
ServerState  END                        17447143771862                    
Finish       BEGIN                      17447143778262
Finish       END                        17447199125099
^C                                                                                                     
                                                                                                       
                                                                                                       
PHASE                           TIME ELAPSED (usec)                                                    
Sync                                            121                                                    
RamPushPrePause                             9402847                                                    
RamPushPostPause                             202005                                                    
TimeData                                        156                                                    
DeviceState                                   61375                                                    
RamPull                                       57509                                                    
ServerState                                   64561                                                    
Finish                                        55346                                                    
TOTAL                                       9843924     
```

**time-adjustments.d** (dest)
```
jordan@yuban ~/propolis/scripts $ pfexec ./time-adjustments.d 2160
tracing pid 2160...
time data imported; press CTRL+C for summary
^C
GUEST FREQ (Hz)        UPTIME (usec)    TOTAL TIME (usec)
     2649958945            879497608                  148

EVENT           DURATION (usec)            GUEST TSC          BOOT HRTIME
Migration                   107        2330632597459       17148534696948
[adjustment]                                  285087          -9235619709
Propolis                     22        2330632882546       17139299077239
[adjustment]                                   84400                    0
Kernel                       18        2330632966946       17139299077239
```

example log messages of time adjustments (dest):
```
May 12 00:18:33.582 INFO VMM Time Data: "(guest_freq:2649958945,guest_tsc:2330632597459,hrtime:18028032
305249,hres_sec:1683850713,hres_ns:581965192,boot_hrtime:17148534696948)", component: vm_controller    
May 12 00:18:33.582 INFO Time data adjustments:                                                        
- guest TSC freq: 2649958945 Hz = 2.649958945 GHz                                                      
- guest uptime ns: 879497608301                                                                        
- migration time delta: 107.582µs                                                                      
- guest_tsc adjustment = 2330632597459 + 285087 = 2330632882546                                        
- boot_hrtime adjustment = 17148534696948 ---> 18018796793122 - 879497715883 = 17139299077239          
- dest highres clock time: 18018796793122                                                              
- dest wall clock time: 1683850713.582072774s, component: vm_controller  
```

lib/propolis/src/vmm/time.rs
```diff
@@ -0,0 +1,720 @@
...
+        /// guest TSC frequency (hz)
+        guest_hz: u64,
+
+        /// calculated TSC adjustment
+        tsc_adjust: u128,
+    },
+
+    /// Failed to read host hrtime/wall clock time
+    #[error("could not read host time: {0}")]
+    HostTimeRead(String),
```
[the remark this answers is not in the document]
jordanhendricks: this error is removed

lib/propolis/src/vmm/time.rs
```diff
@@ -0,0 +1,720 @@
...
+// before we switched them to use a local ntp server. We may want to consider
+// guard rails here (both positive and negative).
+fn calc_migrate_delta(
+    src_wc: Duration,
+    dst_wc: Duration,
+) -> Result<Duration, TimeAdjustError> {
+    match dst_wc.checked_sub(src_wc) {
+        Some(d) => Ok(d),
+        None => Err(TimeAdjustError::InvalidMigrateDelta { src_wc, dst_wc }),
+    }
```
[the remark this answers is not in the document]
jordanhendricks: this code is pretty different from when you reviewed it, but I believe I got through all of these

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,550 @@
...
+    if (adjusted_bhrt < i64::MIN as i128) || (adjusted_bhrt > i64::MAX as i128)
+    {
+        return Err(TimeAdjustError::BootHrtimeOverflow {
+            desc: "boot_hrtime to i64".to_string(),
+            total_delta,
+            dst_hrtime,
+        });
+    }
+
+    Ok(adjusted_bhrt as Hrtime)
```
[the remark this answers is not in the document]
jordanhendricks: done

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,550 @@
...
+/// the target host has a smaller uptime than the guest. This is okay: hrtime_t
+/// is signed, and the boot_hrtime is used by bhyve as a normalization value for
+/// device timers.
+///
+pub fn calc_boot_hrtime(
+    total_delta: Duration,
+    dst_hrtime: Duration,
+) -> Result<Hrtime, TimeAdjustError> {
+    let dst_hrt_ns: i128 = dst_hrtime.as_nanos() as i128;
+    let adjusted_bhrt: i128;
```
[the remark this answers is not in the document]
jordanhendricks: removed

lib/propolis/src/vmm/time_adjust.rs
```diff
@@ -0,0 +1,550 @@
...
+/// boot_hrtime and the hrtime of the host.
+///
+/// uptime = hrtime - boot_hrtime
+pub fn calc_guest_uptime(
+    src_hrt: u64,
+    boot_hrtime: Hrtime,
+) -> Result<Duration, TimeAdjustError> {
+    // convert input to 128-bits so we can check for overflow
+    let src_hrt_ns: i128 = src_hrt as i128;
+    let boot_hrt_ns: i128 = boot_hrtime as i128;
```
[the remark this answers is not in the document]
jordanhendricks: removed

lib/propolis/src/vmm/hdl.rs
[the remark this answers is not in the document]
jordanhendricks: I ended up doing a more "pure" function as originally suggested, and inlined most of the calculations.

bin/propolis-server/src/lib/migrate/source.rs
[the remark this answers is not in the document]
jordanhendricks: as of d09938fb, there is only one time data phase of the protocol

I realized through review and discussion that having two phases more closely mapped to my initial model of how to correct the guest TSC (namely, measure how long migration takes and add that to the guest TSC). But it doesn't actually matter for correctness when we export/import the data, since the guest TSC is calculated from host hrtime. As long as we import the time data before the device state, we should be in good shape.

bin/propolis-server/src/lib/migrate/source.rs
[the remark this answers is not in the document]
jordanhendricks: see: https://github.com/oxidecomputer/propolis/pull/359#discussion_r1191799079

I consolidated down to one protocol phase

lib/propolis/src/vmm/time_adjust.rs
jordanhendricks: In d09938fb, which has a much tighter window between export and import than previous versions of this PR, I clamped the migrate delta to 0 if for some reason we see a negative delta.

bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -321,6 +328,114 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> DestinationProtocol<T> {
...
+            .map_err(codec::ProtocolError::from)?;
+        let deserializer =
+            &mut <dyn erased_serde::Deserializer>::erase(&mut deserializer);
+        let time_data_src: vmm::time::VmTimeData =
+            erased_serde::deserialize(deserializer).map_err(|e| {
+                MigrateError::TimeData(format!(
+                    "VMM Time Data deserialization error: {}",
+                    e
+                ))
+            })?;
```
gjcolombo, line 356: Since the destination type is known here, I think this can just be written as `let time_data_src: VmTimeData = ron::from_str(&raw)` (with an appropriate error mapping)--no need to do any of the type erasure/boxing.

scripts/live-migration-times.d
```diff
@@ -138,6 +142,10 @@ dtrace:::END
 		printf("%-16s %29d\n", this->rpush_post, this->d_rpush_post / 1000);
 		this->total += this->d_rpush_post;
 	}
+	if (this->d_td != 0) {
+		printf("%-15s %30d\n", this->td, this->d_td / 1000);
```
gjcolombo, line 146: nit: same width specifiers as the surrounding prints? (If the output aligns properly, then no worries; I just remember having to mess around with this to get all the headers to line up right when I added the post-pause phase.)

lib/propolis/src/vmm/time.rs
```diff
@@ -0,0 +1,788 @@
...
+/// to allow guest time (namely, the guest TSC and its device timers) to allow
+/// the guest's sense of time to function properly on the target.
+//  See comments inline for more details about how we calculate a new guest TSC
+//  and boot_hrtime.
+pub fn adjust_time_data(
+    src: VmTimeData,
+    dst_hrt: i64,
+    dst_wc: Duration,
+) -> Result<(VmTimeData, VmTimeDataAdjustments), TimeAdjustError> {
+    // Basic validation: there is no reason system hrtime should be negative,
```
gjcolombo, line 124: Optional/nit: would it be valid to check this when the timestamp is fetched and convert it to an unsigned value there, or do we use it in signed form in other places?

lib/propolis/src/vmm/time.rs
```diff
@@ -0,0 +1,788 @@
...
+    // +=====================================================================+
+    //
+    // NB: It is possible for boot_hrtime to be negative; this occurs if a
+    // guest has a longer uptime than its host (an expected common case for
+    // migration). This is okay: hrtime is a signed value, and the normalization
+    // maths still work with negative values.
+    //
+
+    // guest_uptime_ns  = source hrtime - source boot_hrtime
+    let guest_uptime_ns: i64 = src
```
gjcolombo, line 217: I'm having trouble squaring this with the comment on line 181. If the exported timer value is normalized, hasn't the source boot_hrtime been subtracted from src.hrtime already?

gjcolombo (approved):
A couple of small bits of cleanup, but generally LGTM. Thanks for putting the work into this!

lib/propolis/src/vmm/time.rs
jordanhendricks: We chatted about this offline -- there are no timers being normalized/denormalized here; this is part of the calculation of the new boot_hrtime. This code and comment could make what is happening here more clear.

scripts/live-migration-times.d
jordanhendricks: will fix this (and my own D script nits https://github.com/oxidecomputer/propolis/pull/359#discussion_r1152605234) in a follow on PR

bin/propolis-server/src/lib/migrate/destination.rs
jordanhendricks: fixed in 505caa0

lib/propolis/src/vmm/time.rs
jordanhendricks: chatted about this offline -- I did use unsigned values for hrtime values that really should never be negative (such as uptime) in earlier versions of writing this code. I found that it was kind of a pain to go back and forth between signed and unsigned, since there are more constrained overflow checks for signed values (and thus I would have to convert any unsigned values back to signed anyway, or otherwise handle that for overflow checks). For my own sanity, I chose to keep all hrtime values signed throughout. Maybe there's a nicer way to write this code, but I would rather not do this for now.

PR #366, opened 2023-04-06
lib: demote rdmsr/wrmsr exit messages to debug level

### LINKED ISSUE #313: slog messages for MSR exits are loud
Propolis logs at info level on RDMSR and WRMSR exits (see `propolis::vcpu_process` in the main lib). These exits happen frequently and don't give us any real information (we already know we're not handling these currently). 

The messages should be reduced to debug/trace level or replaced with DTrace probes, both to make logs more legible and to reduce the chance of slog dropping messages.

gjcolombo:
We already know we're not handling these in Propolis, and they make our logs much less legible (and more likely to overflow/drop messages, though slog overflow can still occur and needs to be dealt with outside this change).

Tested by running a local propolis-server VM.

Fixes #313.

### DIFF
--- a/lib/propolis/src/lib.rs
+++ b/lib/propolis/src/lib.rs
@@ -106,14 +106,14 @@ pub fn vcpu_process(
                 .map_err(|_| VmError::Unhandled(exit)),
         },
         VmExitKind::Rdmsr(msr) => {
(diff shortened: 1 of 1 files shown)


jordanhendricks (approved):
thanks!

PR #367, opened 2023-04-07
Support illumos#15143 data and vCPU pause/resume

### LINKED ISSUE #291: propolis-server should use pause/resume ioctls
While  dd9a9d28f5f137b3d3dbcd8d0494a98ee8f3d5c3 added the interfaces to bhyve-api for the pause/resume functionality added in [illumos#14568](https://www.illumos.org/issues/14568) and 9af863c1b00e5c5083f9321f01fd1dcba174e0fd wired it into `propolis-standalone`, `propolis-server` still need to be updated so it properly pauses and resumes kernel vmm state when doing a migration.

pfmooney:
With the integration of illumos#15143, additional pending interrupt and exception state for guest vCPUs is exposed to userspace.  In order to take a consistent snapshot of said state, as well as the guest devices emulated in-kernel, propolis-server needs to issue VM_PAUSE requests prior to exporting or importing state, and VM_RESUME requests when starting a VM after a migration.

Importation of the "new" vCPU state is guarded with a version check for now, so that propolis continues to work on hosts which have not yet updated to bits featuring illumos#15143.

Closes #291 as well.

### DIFF
--- a/bin/propolis-server/src/lib/vm/state_driver.rs
+++ b/bin/propolis-server/src/lib/vm/state_driver.rs
@@ -274,6 +274,10 @@ where
                     self.get_instance_state(),
                     ApiInstanceState::Migrating
                 );
+                // Signal the kernel VMM to resume devices which are handled by
+                // the in-kernel emulation.  They were kept paused for
+                // consistency while migration state was loaded.
+                self.controller.resume_vm();
             }
         }
 
@@ -346,6 +350,10 @@ where
         // to start so that reset doesn't overwrite any state written by
         // migration.
         self.reset_vcpus();
+
+        // Place the VM in a paused state so we can load emulated device state
+        // in a consistent manner
+        self.controller.pause_vm();
 
         start_tx.send(()).unwrap();
         loop {
@@ -437,6 +445,7 @@ where
                         ));
 
                         if self.paused {
+                            self.controller.resume_vm();
                             self.controller.resume_entities();
                             self.vcpu_tasks.resume_all();
                             self.paused = false;
@@ -455,6 +464,7 @@ where
                     MigrateSourceCommand::Pause => {
                         self.vcpu_tasks.pause_all();
                         self.controller.pause_entities();
+                        self.controller.pause_vm();
                         self.paused = true;
                         response_tx
                             .blocking_send(MigrateSourceResponse::Pause(Ok(())))
@@ -810,6 +820,8 @@ mod tests {
         vm_ctrl.expect_halt_entities().times(1).returning(|| ());
         vm_ctrl.expect_resume_entities().never();
         vcpu_ctrl.expect_resume_all().never();
+        vm_ctrl.expect_pause_vm().times(1).returning(|| ());
+        vm_ctrl.expect_resume_vm().never();
 
         let mut driver = make_state_driver(test_objects);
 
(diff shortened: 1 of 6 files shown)


lib/propolis/src/vcpu.rs
```diff
@@ -351,12 +355,96 @@ pub mod migrate {
...
+                )?;
+
+            // Load all of the pending interrupt/exception state
+            //
+            // If illumos#15143 support is missing, none of these fields will be
+            // present, so the values will remain false/zeroed.  Such an outcome
+            // is fine for now.
+            let (
+                mut pending_nmi,
+                mut pending_extintr,
```
gjcolombo, line 375: turbo-nit: should this be `pending_extint` to match the struct field?

crates/bhyve-api/src/lib.rs
```diff
@@ -193,19 +194,46 @@ impl AsRawFd for VmmFd {
...
+
+    if VERSION_CACHE.load(Ordering::Acquire) == 0 {
+        let newval = match do_query() {
+            Ok(x) => x as i64,
+            Err(e) => -(e.raw_os_error().unwrap_or(libc::ENOENT) as i64),
+        };
+        let _ = VERSION_CACHE.compare_exchange(
+            0,
+            newval,
+            Ordering::Acquire,
```
gjcolombo, line 221: nit: should this store have release semantics? I'm out of practice with barrier semantics, so I might have this all wrong, but the Rust docs for `atomic::Ordering` suggest that `Acquire` semantics only provide additional ordering when used with stores that have `Release` semantics. (Or, since we're only dealing with the one memory location, could/should the whole thing just be `Relaxed` everywhere?)

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -893,6 +905,11 @@ mod tests {
         vcpu_ctrl.expect_pause_all().times(1).returning(|| ());
         vcpu_ctrl.expect_resume_all().times(1).returning(|| ());
 
+        // VMM will be paused once prior to exporting state, and then resumed
+        // afterwards when the migration fails.
+        vm_ctrl.expect_pause_vm().times(1).returning(|| ());
+        vm_ctrl.expect_resume_vm().times(1).returning(|| ());
```
gjcolombo, line 911: It might be worthwhile to add a `mockall::Sequence` or two in here to enforce that the `pause_vm`/`resume_vm` calls are ordered correctly with respect to the calls to deal with entities and vCPUs. (Arguably this test could always have had one to make sure we pause things before resuming them, but I think the value's much higher when we need to verify ordering amongst several pause or resume operations.)

gjcolombo (approved):
LGTM, just a couple of nits and one possible test improvement. Thanks for putting this together!

lib/propolis/src/vcpu.rs
pfmooney: Yeah, I changed the name of the migration data field at some point to match the kernel names, but missed this.

(also, lol @ "turbo-nit")

bin/propolis-server/src/lib/vm/state_driver.rs
pfmooney: Good idea.  I've added the sequences and checked that the tests still pass.

crates/bhyve-api/src/lib.rs
pfmooney: Yeah, I think Relaxed is adequate here for now.  I get a little paranoid about memory ordering being too loose, and rust doesn't make cmpxchg any clearer with how it's structured and documented.

PR #368, opened 2023-04-12
Include buildomat-produced firmware in Propolis zone images

### LINKED ISSUE #362: want better process for packaging bootrom images
Today the Propolis zone image includes a bootrom that's built from [oxidecomputer/edk2](https://github.com/oxidecomputer/edk2) on a local workstation and posted to our S3 bucket. This is probably not ideal from a provenance perspective. It would instead be nicer to build firmware images in a Buildomat job on our edk2 repo and package those.

Work needed here:

- Create buildomat job that produces a bootrom as an artifact
- Teach propolis-package to pick up these bootroms

gjcolombo:
Move to the latest omicron-zone-package rev to pick up support for including Buildomat artifacts in packages, then use that to include a Buildomat-produced firmware image in Propolis zones. (This PR uses an explicit rev of the packaging tools because a new version of the crate hasn't been cut yet.)

Also update PHD to use this firmware image. PHD previously assumed that all artifacts were available from a single common remote artifact root, but now they aren't (the Alpine image used for testing still lives in S3). Lightly refactor artifact management to remove this restriction.

Tests: PHD; manually examined the `propolis-server.tar.gz` image produced by `propolis-package` and verified that all the files are in the right places.

Fixes #362.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ libloading = "0.7"
 mockall = "0.11"
 num_enum = "0.5"
 omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-omicron-zone-package = "0.7.1"
+omicron-zone-package = { git = "https://github.com/oxidecomputer/omicron-package", rev = "a425676edcdefe5a40f77d436f74836bb906b302" }
 once_cell = "1.13"
 oximeter-producer = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
 oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ libloading = "0.7"
 mockall = "0.11"
 num_enum = "0.5"
 omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -82,7 +82,7 @@ libloading = "0.7"
 mockall = "0.11"
 num_enum = "0.5"
 omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-omicron-zone-package = "0.7.1"
+omicron-zone-package = { git = "https://github.com/oxidecomputer/omicron-package", rev = "a425676edcdefe5a40f77d436f74836bb906b302" }
```
luqmana, line 85: Opened https://github.com/oxidecomputer/omicron-package/pull/53 to bump version. Should be published soon.

phd-tests/framework/src/artifacts.rs
```diff
@@ -394,8 +381,6 @@ mod test {
     #[test]
     fn verify_raw_toml() {
         let raw = r#"
-            remote_root = "https://10.0.0.255"
-
             [guest_images.alpine]
             guest_os_kind = "alpine"
             metadata.relative_local_path = "alpine.raw"
```
luqmana, line 386: Can't comment directly (thanks Github) but looks like this test is still using `relative_remote_path`?

luqmana (approved):

PR #369, opened 2023-04-20
bump Crucible rev

gjcolombo:
Tested: local PHD run including Crucible tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,13 +144,13 @@ dependencies = [
 
 [[package]]
 name = "async-recursion"
-version = "1.0.2"
+version = "1.0.4"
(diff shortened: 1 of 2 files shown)


gjcolombo:
The CI failure here is a rustc segfault encountered while building the `dladm` crate. This appears to have been introduced by moving to rustc 1.69.0. Moving to the latest compatible version of `num_enum` (which has some codegen improvements that reduce its stack consumption) seems to resolve the issue. Will have that change pushed shortly.

leftwo (approved):
If this passes your PHD tests, (and the image builds in CI) then you have my rubber stamp :)

smklein (approved):
Thanks for the bump!

gjcolombo:
Another successful hour at the software factory!

PR #375, opened 2023-04-28
Temporarily disable migration tests in CI

gjcolombo:
This prevents PHD runs from failing for some in-flight PRs that depend on host OS changes that haven't integrated yet.

These can be reenabled once all the relevant changes have landed and new master images have been deployed to the Buildomat runners.

### DIFF
--- a/.github/buildomat/jobs/phd-run.sh
+++ b/.github/buildomat/jobs/phd-run.sh
@@ -56,13 +56,17 @@ ls $propolis
 
 # Disable errexit so that we still upload logs on failure
 set +e
+
+# TODO: reenable migration tests once all relevant OS changes have landed and
+# been deployed.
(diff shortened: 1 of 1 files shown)


pfmooney (approved):
I should have reached for the filter, rather than "off".

Thanks for pointing this out (and adding it to phd in the first place).

gjcolombo:
No worries--I hadn't advertised it much, and you had to know it was there before you could reach for it :)

There's also an `--include-filter` for running a specific subset of tests. Test cases are named after the fully-qualified paths to the test routines, so you can filter using module names/paths (what we're ultimately doing here). Multiple filters of either kind should also be supported, though I don't actually recall what the runner will do if you both include and exclude something--probably just exclude it, though it should probably complain instead.

Issue #378, opened 2023-04-28
Guest double fault in live migration testing


gjcolombo:
Repro scenario: repeatedly live migrate a Debian VM around a multi-machine Omicron cluster. Propolis is from a private branch that's based on 7ff360db76ed1941040a9521d3a7881719f98de3 and has #359 merged into it.

This works most of the time, but eventually the guest panics:

```
[  145.267125] traps: PANIC: double fault, error_code: 0x0
[  145.267125] double fault: 0000 [#1] SMP NOPTI
[  145.267126] CPU: 0 PID: 500 Comm: sshd Not tainted 5.10.0-18-amd64 #1 Debian 5.10.140-1
[  145.267126] RIP: 0033:0x55e1662a9c40
[  145.267126] RSP: 002b:00007ffcbb248608 EFLAGS: 00010202
[  145.267127] RAX: 000000000000002f RBX: 000055e1676ed460 RCX: 0000000000000001
[  145.267127] RDX: 0000000000004000 RSI: 000000000000005e RDI: 000055e1676fbce0
[  145.267127] RBP: 0000000000000000 R08: 0000000000000000 R09: 0000000000001000
[  145.267127] R10: 0000000000000000 R11: 0000000000000246 R12: 000055e1676fbce0
[  145.267128] R13: 00007ffcbb248628 R14: 000055e1676fd530 R15: 000000000000002f
[  145.267128] FS:  00007fd426513900 GS:  ffff927e7bc00000
[  145.267128] Modules linked in: dm_mod nls_ascii nls_cp437 vfat fat ghash_clmulni_intel aesni_intel libaes crypto_simd serio_raw cryptd evdev glue_helper pcspkr efi_pstore button rapl drm fuse configfs efivarfs ip_tables x_tables autofs4 ext4 crc16 mbcache jbd2 crc32c_generic virtio_net net_failover virtio_blk failover nvme nvme_core t10_pi crc_t10dif crc32_pclmul crct10dif_generic crc32c_intel virtio_pci crct10dif_pclmul crct10dif_common virtio_ring virtio i2c_piix4
[  145.267135] traps: PANIC: double fault, error_code: 0x0
[  145.279653] ---[ end trace 60c7607a5d74dead ]---
[  145.279653] double fault: 0000 [#2] SMP NOPTI
[  145.279654] CPU: 1 PID: 3437 Comm: cmatrix Tainted: G      D           5.10.0-18-amd64 #1 Debian 5.10.140-1
[...]

PR #383, opened 2023-05-04
Fix missing selectors for LDTR and TR

pfmooney:
The vCPU migration payloads for the segment registers was not properly saving/restoring the selector state for the LDTR and TR.


luqmana (approved):
Nice catch!

pfmooney:
This should have practically no impact on modern (64-bit) guests/software, but we should do the right thing regardless.  @gjcolombo took it for a spin during testing for #378, where it seemed to have little impact (positive or negative)

gjcolombo (approved):

PR #386, opened 2023-05-05
migration: reduce page transfers during pause

gjcolombo:
Send all pages from the source to the target once before pausing, then pause, then send only those pages that were modified after being sent the first time. This drastically reduces the number of pages that have to be sent while the VM is not running.

This change sends *all* pages in the pre-pause RAM transfer page, irrespective of whether they're dirty. This is a low-subtlety way of handling the following case:

1. Source reads its dirty page bitmap
2. Bhyve marks the previously-dirty pages as clean
3. Source sends the pages to the target
4. The target crashes and migration fails

To migrate correctly on a second attempt, the source needs to send the new target all the pages it sent in step 2 (in addition to any other modified pages). It'd be more efficient to remember exactly which pages need to be sent, but for now, this change just sends all guest pages once, since the set of required pages is guaranteed to be a subset of the set of all pages. Note that it's still important to query the dirty bitmap in this phase (so that the pages will only be transferred while paused if they were dirtied after being sent the first time).

Tested: PHD incl. migration variations; 200 migrations of a guest running a memory stress mix.

### DIFF
--- a/bin/propolis-server/src/lib/migrate/mod.rs
+++ b/bin/propolis-server/src/lib/migrate/mod.rs
@@ -405,5 +405,6 @@ impl<'a> Iterator for PageIter<'a> {
 mod probes {
     fn migrate_phase_begin(step_desc: &str) {}
     fn migrate_phase_end(step_desc: &str) {}
-    fn migrate_xfer_ram_page(addr: u64, size: u64, paused: u8) {}
+    fn migrate_xfer_ram_region(pages: u64, size: u64, paused: u8) {}
+    fn migrate_xfer_ram_page(addr: u64, size: u64) {}
 }
--- a/bin/propolis-server/src/lib/migrate/mod.rs
+++ b/bin/propolis-server/src/lib/migrate/mod.rs
@@ -57,6 +57,7 @@ pub enum MigrateRole {
     Destination,
 }
 
+// N.B. Keep in sync with scripts/live-migration-times.d.
 #[derive(Debug)]
 enum MigratePhase {
     MigrateSync,
@@ -404,5 +405,5 @@ impl<'a> Iterator for PageIter<'a> {
 mod probes {
     fn migrate_phase_begin(step_desc: &str) {}
     fn migrate_phase_end(step_desc: &str) {}
-    fn migrate_xfer_ram_page(addr: u64, size: u64) {}
+    fn migrate_xfer_ram_page(addr: u64, size: u64, paused: u8) {}
 }
--- a/bin/propolis-server/src/lib/migrate/mod.rs
+++ b/bin/propolis-server/src/lib/migrate/mod.rs
@@ -57,10 +57,12 @@ pub enum MigrateRole {
     Destination,
 }
 
+#[derive(Debug)]
 enum MigratePhase {
     MigrateSync,
     Pause,
-    RamPush,
+    RamPushPrePause,
+    RamPushPostPause,
     DeviceState,
     RamPull,
     ServerState,
@@ -72,7 +74,8 @@ impl std::fmt::Display for MigratePhase {
         let s = match self {
             MigratePhase::MigrateSync => "Sync",
             MigratePhase::Pause => "Pause",
-            MigratePhase::RamPush => "RamPush",
+            MigratePhase::RamPushPrePause => "RamPushPrePause",
+            MigratePhase::RamPushPostPause => "RamPushPostPause",
             MigratePhase::DeviceState => "DeviceState",
             MigratePhase::RamPull => "RamPull",
             MigratePhase::ServerState => "ServerState",
--- a/bin/propolis-server/src/lib/migrate/source.rs
+++ b/bin/propolis-server/src/lib/migrate/source.rs
@@ -168,9 +168,9 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
             vmm_ram_range
         );
 
-        // TODO: Ideally, both the pre-pause and post-pause phases would offer
-        // just dirty pages. To do this safely, the source must remember all
-        // the pages it has ever offered to any target so that they can be
+        // TODO(#387): Ideally, both the pre-pause and post-pause phases would
+        // offer just dirty pages. To do this safely, the source must remember
+        // all the pages it has ever offered to any target so that they can be
         // re-offered if migration fails and is later retried.
         //
         // Offering all pages before pausing guarantees that all modified pages
--- a/bin/propolis-server/src/lib/migrate/source.rs
+++ b/bin/propolis-server/src/lib/migrate/source.rs
@@ -20,6 +20,16 @@ use crate::migrate::{
 };
 use crate::vm::{MigrateSourceCommand, MigrateSourceResponse, VmController};
 
+/// Specifies which pages should be offered during a RAM transfer phase.
+#[derive(Debug)]
+enum RamOfferDiscipline {
+    /// Offer all pages irrespective of whether they are dirty.
+    OfferAll,
+
+    /// Offer only pages that the hypervisor are marked as dirty.
+    OfferDirty,
+}
+
 pub async fn migrate<T: AsyncRead + AsyncWrite + Unpin + Send>(
     vm_controller: Arc<VmController>,
     command_tx: tokio::sync::mpsc::Sender<MigrateSourceCommand>,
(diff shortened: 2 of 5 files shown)


bin/propolis-server/src/lib/migrate/mod.rs
```diff
@@ -57,10 +57,12 @@ pub enum MigrateRole {
     Destination,
 }
 
+#[derive(Debug)]
```
jordanhendricks, line 60: do you mind fixing up scripts/live-migration-times.md to reflect the new phases?

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -137,17 +147,47 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
...
             self.log(),
-            "ram_push: got query for range {:?}, vm range {:?}",
+            "ram_push ({:?}): got query for range {:?}, vm range {:?}",
+            phase,
             req_ram_range,
             vmm_ram_range
         );
-        self.offer_ram(vmm_ram_range, req_ram_range).await?;
+
+        // TODO: Ideally, both the pre-pause and post-pause phases would offer
```
jordanhendricks, line 171: File an issue for this?

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -202,10 +243,23 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
 
         let step = bits.len() * 8 * 4096;
         for gpa in (start_gpa..end_gpa).step_by(step) {
+            // Call bhyve to get the dirty page mask irrespective of the offer
```
jordanhendricks, line 246: "Call bhyve" ==> "Tell bhyve" maybe?

bin/propolis-server/src/lib/migrate/source.rs
gjcolombo: Filed #387 and updated the comment in 46f8f99.

bin/propolis-server/src/lib/migrate/mod.rs
gjcolombo: Done in 0a387ff, 8bbd69d, and 071df1e (took a few tries to get everything arranged in a way I was happy with).

I thought briefly that I would want to have pause/resume probes in the VM state driver to help with this, so I added those in e11dacc. I didn't end up using the probes, but the other cleanup in that commit seems useful to me, so I kept it in this sequence.

bin/propolis-server/src/lib/migrate/source.rs
gjcolombo: Reworded in 90ed9b4.

bin/propolis-server/src/lib/migrate/source.rs
jordanhendricks: thanks!

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -159,12 +199,27 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
...
                     // do not because of the way access is implemented, we
                     // should probably disallow it at the protocol level.
                     self.xfer_ram(start, end, &bits).await?;
+                    probes::migrate_xfer_ram_region!(|| {
+                        use bitvec::prelude::{BitSlice, Lsb0};
+                        let bits = BitSlice::<_, Lsb0>::from_slice(&bits);
+                        let pages = bits.count_ones() as u64;
+                        (
+                            pages,
+                            pages * 4096,
```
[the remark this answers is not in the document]
pfmooney: https://github.com/oxidecomputer/propolis/blob/a54fe33f4b0d77d9a0a5bafccef5c1822fdf26e1/lib/propolis/src/common.rs#L359

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -137,17 +147,47 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
...
+            phase,
             req_ram_range,
             vmm_ram_range
         );
-        self.offer_ram(vmm_ram_range, req_ram_range).await?;
+
+        // TODO(#387): Ideally, both the pre-pause and post-pause phases would
+        // offer just dirty pages. To do this safely, the source must remember
+        // all the pages it has ever offered to any target so that they can be
+        // re-offered if migration fails and is later retried.
```
pfmooney, line 174: We should probably just mark all pages as dirty in the case of a failed migration?

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -504,6 +508,26 @@ where
...
     }
 
+    fn pause(&mut self, pause_kernel: PauseKernelVmm) {
+        assert!(!self.paused);
+        probes::state_driver_pause!(|| ());
+        self.vcpu_tasks.pause_all();
+        self.controller.pause_entities();
+        if pause_kernel == PauseKernelVmm::Yes {
+            self.controller.pause_vm();
+        }
```
pfmooney, line 518: Is there a reason not to issue the `VM_PAUSE`?  Sure it's somewhat meaningless for a halt (or similar), but skipping it here leaves room for trouble later if one makes a `pause(PauseKernelVmm:No)` call, followed by a `resume()` call later.

bin/propolis-server/src/lib/migrate/source.rs
pfmooney: Filed [illumos#15639](https://www.illumos.org/issues/15639) for this

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: Nah, this was just rote-equivalence translation from the old code to the new. Cleaned up in 53f1a2e.

scripts/live-migration-times.d
```diff
@@ -21,14 +21,28 @@
...
  *   a key for tracking the phase deltas. If those names change, or phases are
  *   added/removed, this script will break.
+ * - When calculating RAM transfer rates, the script assumes that the majority
+ *   of the time Propolis spends in a RAM copy phase is spent actually copying
+ *   guest memory and that the cost of other tasks in the phase (entering it,
+ *   leaving it, logging messages, Propolis control flow, etc.) is negligible
+ *   by comparison. If a RAM transfer phase copies very little memory (e.g.
+ *   because the VM is mostly idle and there's nothing to copy in the post-
+ *   pause phase), this assumption will not hold and the script will report
+ *   unusually low transfer rates.
```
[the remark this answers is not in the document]
gjcolombo: Reworded in d2afece.

bin/propolis-server/src/lib/migrate/source.rs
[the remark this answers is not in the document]
gjcolombo: Cleaned up in e3d19da and 074ef50. (Amusingly, we also use the constant 4,096 as the length of our page bitmaps, but I think I've managed to substitute `PAGE_SIZE` only in the place where we actually mean the page size and not the bitmap length.)

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -137,17 +147,47 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
         self.read_ok().await
     }
 
-    async fn ram_push(&mut self) -> Result<(), MigrateError> {
-        self.update_state(MigrationState::RamPush).await;
+    async fn ram_push(
+        &mut self,
+        phase: &MigratePhase,
+    ) -> Result<(), MigrateError> {
+        // Only publish the "RAM push" phase before pausing (so that the
```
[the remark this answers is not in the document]
gjcolombo: I wanted to have avoid having the publicly-visible migration phase go from `RamPush` to `Pause` back to `RamPush` because that looked to me like the migration was going backwards. It turns out that the API already defines a separate `RamPushDirty` state for this, so we can just use that; fixed in 7d08ab7.

scripts/live-migration-times.d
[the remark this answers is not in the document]
jordanhendricks: looks great! thanks

bin/propolis-server/src/lib/migrate/source.rs
gjcolombo: I think the ideal would be to have a "read but don't clear" discipline for getting dirty bits (as 15639 describes). That way we still don't have to transfer pages that have absolutely never been touched by anyone just because a prior migration failed.

That said, this seems like a useful middle ground if it turns out that migrating all pages before pausing is intractably expensive: we can just have the VM controller store the offer discipline to use, change it from "offer dirty" to "offer all" on migration failure, and then pump it into the migration context as a parameter when a new migration begins.

jordanhendricks (approved):

PR #403, opened 2023-05-15
Crucible tell Nexus when a scrub is done

leftwo:
When an instance is started with a volume that contains a read-only-parent, crucible upstairs (after a slight delay) will now start scrubbing that read-only-parent and copying the data to the underlying volume.  When that scrub has completed, crucible will notify Nexus that the scrub is done and it can remove that read-only-parent from the Volume.

Crucible has to know what the address the propolis server is running at, and from that can determine the Nexus endpoint.


gjcolombo:
Do scrub operations initiated by the pantry also have to notify Nexus of their success/failure in this way, or do those get communicated separately? I see the pantry code returns a job ID as part of its `ScrubResponse`, but I'm not turning up anything in Nexus that uses that today.

lib/propolis/src/block/crucible.rs
```diff
@@ -29,11 +82,26 @@ impl CrucibleBackend {
...
         producer_registry: Option<ProducerRegistry>,
+        my_address: SocketAddr,
+        log: slog::Logger,
     ) -> io::Result<Arc<Self>> {
+        // If provided an address for propolis server, figure out if it
+        // is IPv6 and if so, store that address for possible later use.
+        let address = match my_address {
+            SocketAddr::V6(my_address) => Some(*my_address.ip()),
+            SocketAddr::V4(_) => None,
+        };
```
gjcolombo, line 93: I'd be inclined to phrase this a little more explicitly as "should this backend try to contact Nexus after scrubbing completes, and if so, how should it obtain Nexus's address?" It seems legitimate to me to have a Propolis server that's serving with an IPv6 address but that's not part of an Omicron deployment (e.g. we might want to do this in CI someday to enable multi-machine tests with links fast enough not to limit live migration speeds). This is probably harmless in this change (it'll just produce spurious errors in `remove_read_only_parent`), but I think it's worthwhile to have this path be explicit about what it's configuring.

The exact method by which I'd do this probably depends on whether the pantry also has to notify Nexus when scrubbing attempts complete. If it does, then maybe the "scrub notification config" information should go into the VCR so that the `Volume` can do the notifying directly. If the pantry is going to do something else, or there are other reasons not to put this in the VCR, we could instead put this configuration into the `Crucible` variant of `InstanceSpec::StorageBackendKind` (and then have sled agent make sure to specify it while other instance spec clients like PHD leave it blank).

lib/propolis/src/block/crucible.rs
```diff
@@ -61,6 +172,52 @@ impl CrucibleBackend {
...
+                        error!(
+                            log,
+                            "Failed removal of read only parent: {}", e,
+                        );
+                    }
+                }
+            }
+            Err(e) => {
+                error!(log, "Failed to connect to Nexus: {}", e);
+            }
```
gjcolombo, line 217: If the failure to resolve Nexus's address was transient, does this need to be retried?

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -301,10 +303,39 @@ impl<'a> MachineInitializer<'a> {
...
+                                id,
+                                block_size: _,
+                                url: _,
+                            } => id.to_string(),
+                            VolumeConstructionRequest::Region {
+                                block_size: _,
+                                blocks_per_extent: _,
+                                extent_count: _,
+                                opts: _,
+                                gen: _,
```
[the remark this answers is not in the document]
leftwo: Ah, yes, I recently discovered that and it's soooooo much nicer.  Will update.

lib/propolis/src/block/crucible.rs
```diff
@@ -42,9 +110,52 @@ impl CrucibleBackend {
...
+        my_address: Option<Ipv6Addr>,
+        log: slog::Logger,
     ) -> Result<Arc<Self>, crucible::CrucibleError> {
+        // Construct the volume.
         let volume = Volume::construct(request, producer_registry).await?;
 
+        // Decide if we need to scrub this volume or not.
+        if volume.has_read_only_parent() {
+            let vclone = volume.clone();
+            tokio::spawn(async move {
```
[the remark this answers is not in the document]
leftwo: Crucible upstairs drives the scrub (and this task specifically).
Inside the `scrub` method, the upstairs will track where the scrub has covered and IOs that arrive will be directed according to where the scrub has reached.

leftwo:
> Do scrub operations initiated by the pantry also have to notify Nexus of their success/failure in this way, or do those get communicated separately? I see the pantry code returns a job ID as part of its `ScrubResponse`, but I'm not turning up anything in Nexus that uses that today.

The pantry side scrub is not yet enabled (https://github.com/oxidecomputer/crucible/issues/590), so you won't find anything there :)

lib/propolis/src/block/crucible.rs
leftwo: I can update the comment to better reflect your wording above, as this is indeed for Nexus notification.

The pantry may/will have to support a scrub, but it depends on how long pantry instances are going to be around.

Another piece to consider here is that crucible or even other parts of propolis may want to send other messages/errors/warnings to Nexus, and how might we fit those into this framework as well?  If we include
the Nexus contact info in the VCR, then Propolis might not be able to use it.

lib/propolis/src/block/crucible.rs
leftwo: I'm happy to put a loop around this, but then we need to decide on a retry policy.
Try once every minute, forever?
Try once every minute for five times then give up?

It's not ideal if we can't scrub the volume, but it's also not fatal.  The next time this instance
is started we will try the scrub again.

That being said, I'm happy to take input here to improve this.

gjcolombo: > It's not ideal if we can't scrub the volume, but it's also not fatal. The next time this instance is started we will try the scrub again.

If this is the case (i.e., nothing will get irreparably out of sync or become unusable because Propolis failed to send a "scrub finished" notification to Nexus), then I think we can just leave this as-is (though it might be good to add a comment to this effect).

lib/propolis/src/block/crucible.rs
gjcolombo: > Another piece to consider here is that crucible or even other parts of propolis may want to send other messages/errors/warnings to Nexus, and how might we fit those into this framework as well? If we include
the Nexus contact info in the VCR, then Propolis might not be able to use it.

In this case, I'd be inclined to treat the Nexus client like the Oximeter producer registry (server gets static configuration parameters that define whether and how to construct a Nexus client; during ensure, create a client if needed, then pass it to the VM controller to hand down to any objects that are interested in it). To be honest, I kind of like the idea of doing that anyway, just to keep all of the "external services that individual entities might or might not care to talk to" abstractions in the same place.

IMO that approach would also make it more palatable to have a `--create-nexus-client` command line option that more clearly describes whether the server's creator intends for the server to be able to talk to Nexus, irrespective of the server address the creator chose.

leftwo: So, currently, inside `crucible.rs` we do this:
```
    async fn _create(
        read_only: bool,
        producer_registry: Option<ProducerRegistry>,
        my_address: Option<Ipv6Addr>,
        log: slog::Logger,
    ) -> Result<Arc<Self>, crucible::CrucibleError> {
```
The idea would be to do this step, the translation to the actual nexus client, earlier in propolis, then just pass the resulting client info in with the  `create()`?

bin/propolis-server/src/lib/initializer.rs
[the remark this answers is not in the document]
leftwo: Updated in latest commit

bin/propolis-server/src/main.rs
```diff
@@ -91,6 +92,7 @@ async fn run_server(
         use_reservoir,
         log.new(slog::o!()),
         config_metrics,
+        propolis_addr,
```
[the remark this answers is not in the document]
leftwo: Done

lib/propolis/src/block/crucible.rs
```diff
@@ -11,9 +11,62 @@ use crate::vmm::MemCtx;
...
+
+    pub async fn get_ip(&self) -> Result<Ipv6Addr, ResolveError> {
+        self.inner.resolver.lookup_ipv6(ServiceName::Nexus).await
+    }
+
+    pub async fn get(&self) -> Result<NexusClient, ResolveError> {
+        let address = self.get_ip().await?;
+
+        Ok(NexusClient::new(
+            &format!("http://[{}]:{}", address, NEXUS_INTERNAL_PORT),
```
luqmana, line 64: I would use `lookup_socket_v6` instead of `lookup_ipv6` so you don't have to hardcode the port here.

bin/propolis-server/src/lib/server.rs
```diff
@@ -69,6 +69,9 @@ pub struct StaticConfig {
     /// The configuration to use when setting up this server's Oximeter
     /// endpoint.
     metrics: Option<MetricsEndpointConfig>,
+
+    /// Address used by propolis server.
+    my_addr: SocketAddr,
```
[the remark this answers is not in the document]
leftwo: Done

lib/propolis/src/block/crucible.rs
leftwo: I've updated the PR, the latest changes should have this implemented as (I thought) you suggested.

lib/propolis/src/block/crucible.rs
leftwo: Things have moved around a bit, make sure this still makes sense to you (comment wise)

lib/propolis/src/block/crucible.rs
leftwo: Nice, and done.

leftwo:
I tested this latest changes and things are still working as expected, this is ready for another review.

lib/propolis/Cargo.toml
```diff
@@ -33,6 +33,9 @@ once_cell.workspace = true
 crucible-client-types = { workspace = true, optional = true }
 crucible = { workspace = true, optional = true }
 oximeter = { workspace = true, optional = true }
+nexus-client.workspace = true
+internal-dns.workspace = true
+omicron-common.workspace = true
```
pfmooney, line 38: These should be optional and gated on `crucible-full` like the other bits.

pfmooney, line 38: I didn't see references to `internal_dns` or `omicron_common` added to propolis-lib.  Did I miss them, or can these deps be elided?

bin/propolis-server/src/lib/server.rs
```diff
@@ -34,6 +34,11 @@ use tokio_tungstenite::WebSocketStream;
 use crate::spec::{ServerSpecBuilder, ServerSpecBuilderError};
 use crate::vm::VmController;
 use crate::vnc::PropolisVncServer;
+use internal_dns::resolver::{ResolveError, Resolver};
+use internal_dns::ServiceName;
+pub use nexus_client::Client as NexusClient;
+use std::net::Ipv6Addr;
+use std::net::SocketAddrV6;
```
pfmooney, line 41: style nit: please group these with the `std` imports in the group above.

bin/propolis-server/src/lib/server.rs
```diff
@@ -332,6 +337,79 @@ async fn register_oximeter(
...
+// Return that endpoint if successful.
+async fn find_local_nexus_client(
+    local_addr: SocketAddr,
+    log: Logger,
+) -> Option<NexusClient> {
+    // At the moment, we only support converting an IPv6 address into a
+    // Nexus endpoint.
+    let address = match local_addr {
+        SocketAddr::V6(my_address) => *my_address.ip(),
+        SocketAddr::V4(_) => return None,
```
pfmooney, line 393: Perhaps worthwhile emitting a `warn` message here as well, to match the other cases where we won't be emitting a usable `NexusClient`?

lib/propolis/Cargo.toml
leftwo: Done.

leftwo: They have been elided.
There has to be some rust tool that will do this for you, right?

bin/propolis-server/src/lib/server.rs
leftwo: Done

bin/propolis-server/src/lib/server.rs
leftwo: Done.

leftwo (commented):
Patrick, I believe my last commit should address all (ALL!) your isues.

lib/propolis/Cargo.toml
pfmooney: Looks like `cargo-udeps` is the ticket for that, but I'm also not certain about how well it handles cases like workspace dependencies.

bin/propolis-server/src/lib/server.rs
```diff
@@ -337,6 +337,82 @@ async fn register_oximeter(
...
+/// the DNS lookup until accessed.
+///
+/// Without the assistance of OS-level DNS lookups, the [`NexusClient`]
+/// interface requires knowledge of the target service IP address.
+/// For some services, like Nexus, this can be painful, as the IP address
+/// may not have even been allocated when the Sled Agent starts.
+///
+/// This structure allows clients to access the client on-demand, performing
+/// the DNS lookup only once it is actually needed.
+struct Inner {
```
gjcolombo, line 350: naming nit: this is a pretty wide-ranging module with lots of distinct structs, so I'd go with `LazyNexusClientInner` for this.

Even better, I think, would be to move this from `omicron-nexus` to `nexus-client` or `omicron-common` so we can just reuse that code, but that's not needed for this change. (It might be worth filing an Omicron issue for, though.)

gjcolombo (approved):
Thanks for reorganizing the Nexus client logic--I think the new code will be much more flexible in the future. LGTM with one small comment about struct naming in `server.rs`.

bin/propolis-server/src/lib/server.rs
leftwo: I have done the re-name.

For the larger issue of moving this, you mean move the methods/functions here into Omicron proper?  Or put them in something local to this repo but in a common area?

gjcolombo: The former. My last comment was wrong. There's an identical-looking `LazyNexusClient` in the sled agent (`omicron/sled-agent/src/nexus.rs`; I misread the path and thought it was in `omicron-nexus` but that's in `omicron-sled-agent`). It'd be nice to consolidate these. Practically speaking, I'd do that by moving sled agent's lazy client into a smaller common crate that sled agent and Propolis (and anyone else who wants it) can link to, and I suspect that common crate should live in the Omicron repo. Any details beyond that are TBD :)

PR #405, opened 2023-05-16
server: add RFD 71-style protocol negotiation

gjcolombo:
Add support for migration protocol negotiation as described in RFD 71. This gives us some leeway to adjust the current migration protocol (e.g. by adding or removing phases or changing phases' semantics) in future releases without having to define a completely different migration API endpoint. (Nothing in this PR prevents us from creating a "`/v2/migrate`" API in the future; the idea here is to reserve that sort of revision for changes that completely replace the back-and-forth IPC protocol with e.g. a less synchronous RPC-driven model.)

RFD 71 describes the following negotiation process:

- When starting migration, the destination sends the source a comma-delimited list of protocol specifiers of the form `propolis-migrate-<encoding>/<version>`.
- The source selects a compatible protocol with the highest version. (The RFD is ambiguous as to whether versions have to be unique or whether it's permitted to have two protocols with the same version but different encodings, so this change allows the latter.)
- The source sends the protocol string back to the destination, and the two sides proceed to use the selected protocol.

To do this:

- Add a module that lists the supplied protocol options and add routines that convert protocol options to and from specifiers and that select a protocol to use given a specifier list.
- Plumb protocol selections through the migration module to the `source::migrate` and `destination::migrate` functions that create a protocol object that actually executes the migration. These functions can now match on this `migrate::protocol::Protocol` and create an appropriately versioned protocol object.


bin/propolis-server/src/lib/migrate/protocol.rs
```diff
@@ -0,0 +1,323 @@
...
+    let mut s = String::new();
+    while let Some(p) = iter.next() {
+        s.push_str(&p.offer_string());
+        if iter.peek().is_some() {
+            s.push(DELIMITER);
+        }
+    }
+
+    s
+}
```
pfmooney, line 198: All the `BTreeSet` machinery seems a bit much for while we have only one (or two) valid versions, but I suppose that since it's written (and tested), there's no harm in leaving it as-is

bin/propolis-server/src/lib/migrate/protocol.rs
```diff
@@ -0,0 +1,323 @@
...
+            )
+        })?;
+
+        Ok(ProtocolParts { encoding, version })
+    }
+}
+
+/// The set of protocol variants supported by this migration library, expressed
+/// as a set of parts.
+static PROTOCOL_PARTS: Lazy<BTreeSet<ProtocolParts>> = Lazy::new(|| {
```
pfmooney, line 173: Is `lazy_static` (as we've used elsewhere) inadequate for this?

bin/propolis-server/src/lib/migrate/protocol.rs
```diff
@@ -0,0 +1,323 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+struct ProtocolParts {
+    version: u32,
+    encoding: Encoding,
+}
+
+impl ProtocolParts {
+    fn offer_string(&self) -> String {
+        format!(
+            "{PREFIX}{}{ENCODING_VERSION_SEPARATOR}{}",
```
pfmooney, line 131: I'll be honest, the name binding throws me for a loop, but I'm not sure it'd be any clearer with those passed as args.

pfmooney (commented):
Thanks for taking a swing at this.  It'll be great to have something established in the protocol to get us out of a jam later, even if adding/removing/tweaking phases is likely to mean a lot of duplication (or complication).

As you point out, none of this forecloses on a v2 endpoint, should we decide to go that route eventually.

bin/propolis-server/src/lib/migrate/mod.rs
```diff
@@ -101,9 +76,12 @@ pub enum MigrateError {
     #[error("couldn't establish migration connection to source instance")]
     Initiate,
 
+    #[error("failed to parse the offered protocol list ({0}): {1}")]
+    ProtocolParse(String, protocol::ProtocolParseError),
```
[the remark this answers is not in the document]
gjcolombo: The rest of the enum uses strings for this sort of thing, so I think it makes sense to be consistent with that. Done in 4e5d99e.

bin/propolis-server/src/lib/migrate/protocol.rs
gjcolombo: I think what's weird about it is that it interleaves interpolated references and plain old arguments. Doing one or the other looks less silly. Fixed in 17dbf47.

bin/propolis-server/src/lib/migrate/protocol.rs
gjcolombo: Nah. Fixed in 17dbf47.

bin/propolis-server/src/lib/migrate/protocol.rs
gjcolombo: I reached for them to avoid having to worry about sorting collections in the right places (since `select_compatible_protocol` needs to iterate over the offer in descending version order), but now that the code's baked a little more I think we can just use plain Vecs. Tidied this as up as part of 17dbf47.

PR #410, opened 2023-05-18
propolis-server should use reservoir on prod gear

### LINKED ISSUE #408: propolis-server running in a zone complains that it could not query VMM reservoir
On startup, propolis-server queries the system VMM reservoir and [determines](https://github.com/oxidecomputer/propolis/blob/3e1bfb0f248ee5f9ef58607b957d6eb2bc20d1c4/bin/propolis-server/src/lib/config.rs#L67-L96) whether it should tell bhyve to pull the guest memory from the reservoir. While looking at a propolis-server log in a full omicron deployment (that is, propolis-server was running in a zone), I noticed that the log message that prints when this action occurs failed with `EPERM`:

```
May 17 10:28:54.434 ERRO could not query reservoir Os { code: 1, kind: PermissionDenied, message: "Not owner" }
```

Today, only the global zone can query the VMM reservoir, so it is expected that this would fail. There are a couple of things to fix here:
- since a zone can use memory from the reservoir, it should also be able to query it
- in production, we should not query the reservoir on startup; this is done today to allow propolis-server to work on developer machines. We should compile this behavior out for production binaries.

pfmooney:
Add "omicron-build" feature to propolis-server which compiles out the developer-friendly vmm-reservoir check, opting to use the reservoir unconditionally on the assumption that Oxide machines will be appropriately configured for such duty.

Fix #408

### DIFF
--- a/.github/buildomat/jobs/image.sh
+++ b/.github/buildomat/jobs/image.sh
@@ -27,7 +27,10 @@ cargo --version
 rustc --version
 
 banner build
-ptime -m cargo build --release --verbose -p propolis-server
+
+# Enable the "omicron-build" feature to indicate this is an artifact destined
+# for production use on an appropriately configured Oxide machine
+ptime -m cargo build --release --verbose -p propolis-server --features omicron-build
 
 banner image
 ptime -m cargo run -p propolis-package
--- a/bin/propolis-server/Cargo.toml
+++ b/bin/propolis-server/Cargo.toml
@@ -53,7 +53,7 @@ slog-async.workspace = true
 slog-bunyan.workspace = true
 slog-dtrace.workspace = true
 slog-term.workspace = true
-propolis.workspace = true
+propolis = { workspace = true, features = ["crucible-full", "oximeter"] }
 propolis-client = { workspace = true, features = ["generated"] }
 propolis-server-config.workspace = true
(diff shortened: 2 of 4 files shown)


gjcolombo (approved):
LGTM. In the future, it'd be great to use the "prod" configuration in CI if we can. I filed #412 for that, but I don't think we need to block on it right now.

jordanhendricks:
I think we need to wait to merge this until oxidecomputer/omicron#2650 until is done, or at least wait to update the propolis dep in omicron after this is merged. I think the latter will be easier to miss than just not merging this yet.

pfmooney:
> I think we need to wait to merge this until [oxidecomputer/omicron#2650](https://github.com/oxidecomputer/omicron/issues/2650) until is done, or at least wait to update the propolis dep in omicron after this is merged. I think the latter will be easier to miss than just not merging this yet.

Yep.  I was just rebasing since it had fallen behind.

gjcolombo:
When we make the change to pick this up in Omicron, we'll also need to change `omicron/smf/sled-agent/non-gimlet/config.toml` (and `gimlet-standalone/config.toml`) to specify a nonzero reservoir size, or Omicron dev clusters won't be able to start instances (unless they pick up a Propolis image that isn't built with this feature).

Issue #391, opened 2023-05-08
serial console can hang if user enters data after migration begins


jordanhendricks:
@jmpesp and I have both observed the serial console sometimes hanging after a migration. I am able to reliably reproduce such a hang by spamming the console with characters (such as `Enter`) after kicking off migration.

After reproducing this several times, I noticed the following things about the device state is migrated:
- `rx_fifo` of com1 always has data; `rx_fifo` is the receive buffer from the guest perspective, so it contains data typed from a user at the other end with the propolis-cli `serial` command, for example
- the level of IRQ4 of `atpic`, is always 1; IRQ4 maps to an interrupt for com1/com4

Example `com1` and `atpic` device state from such a migration:
```
    Device {                                                                                                           
        instance_name: "lpc-bhyve-atpic",                                                                              
        payload: "(data:(va_chip:((vac_icw_state:0,vac_status:33,vac_reg_irr:27,vac_reg_isr:0,vac_reg_imr:255,vac_irq_base:32,vac_lowprio:7,vac_elc:0,vac_level:(0,0,0,0,1,0,0,0)),(vac_icw_state:0,vac_status:33,vac_reg_irr:16,vac_reg_isr:0,vac_reg_imr:255,vac_irq_base:40,vac_lowprio:7,vac_elc:0,vac_level:(0,0,0,0,0,0,0,0)))))",                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
    },         

    Device {                                                                                           
        instance_name: "lpc-uart-com1",                                                                
        payload: "(uart_state:(intr_enable:15,intr_status:4,line_ctrl:19,line_status:97,modem_ctrl:11,m
odem_status:0,scratch:90,div_low:1,div_high:0,thre_state:false,rx_fifo:(len:1,buf:[13]),tx_fifo:(len:1,
buf:[])))",                                                                                            
    },    
```

After looking at this a bit more, I identified a couple of problems:
- the UART `intr_pin` state isn't migrated at all (#390)
- it is possible for the UART device to receive data from a typing user after migration starts (#389)

Since I could reproduce this easily, I experimented a bit to see what exactly was causing the stuck state. I tried migrating `intr_pin` and syncing it upon import; this resulted in the same hang and device state as above. I implemented a poor eng's pause/resume for the UART, and this did fix the hang, but the serial console connection died after migration; I was able to reconnect though, and the guest console was responsive as suspected.

[...]

PR #411, opened 2023-05-18
Implement a cheap pause/resume for LPC UART device

jordanhendricks:
In #391, either the guest or the console user can induce the serial console to be in a state such that it is never usable again, even after a guest reboot. This PR implements a simple pause/resume on the LPC UART that keeps those hangs from reproducing, as a mitigation to prevent bricking the console completely. In the fullness of time, we probably want a more thoughtful pause/resume functionality for this device.

On current master, both of the following situations reliably produce a hang after migration:
- typing a lot of input at the serial console during migration
- causing the guest to print a lot of output to the console during migration

with this change, I've tested that I can no longer reproduce the above hangs.

### DIFF
--- a/lib/propolis/src/hw/uart/lpc.rs
+++ b/lib/propolis/src/hw/uart/lpc.rs
@@ -15,6 +15,10 @@ struct UartState {
     uart: Uart,
     irq_pin: Box<dyn IntrPin>,
     auto_discard: bool,
+
+    // In the absence of better interfaces for chardev save/restore behavior,
+    // allow the device to be coarsely paused (dropping all reads and writes).
+    paused: bool,
 }
 
 impl UartState {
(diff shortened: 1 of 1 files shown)


lib/propolis/src/hw/uart/lpc.rs
```diff
@@ -15,6 +15,7 @@ struct UartState {
     uart: Uart,
     irq_pin: Box<dyn IntrPin>,
     auto_discard: bool,
+    paused: bool,
```
pfmooney, line 18: Would be nice to add a little context in a comment here about why we have `paused` (to paper over chardev shortcomings around save/restore), so it's clear if/when we circle back to fix that stuff.

jordanhendricks: added something in e7fe716

pfmooney (approved):

Issue #377, opened 2023-04-28
Re-enable migration tests in CI


gjcolombo:
#375 disabled these to reduce noise while we merge a bunch of migration bits into Propolis and bhyve. Once everything has landed we should turn the tests back on.

PR #413, opened 2023-05-18
Re-enable PHD live migration tests

jordanhendricks:
fixes https://github.com/oxidecomputer/propolis/issues/377


gjcolombo:
phd-run is mad because there are extant VMMs at the end of the run. This appears to be because a Propolis panicked mid-run. I can repro locally and filed #415 to track. I'll try to get a fix up quickly to unblock this.

gjcolombo (approved):

PR #419, opened 2023-05-20
Various NVMe emulation fixes

citrus-it:
With these fixes, and a separate change to the illumos nvme driver to make it happier at being under a PCI root complex, I can successfully boot an illumos guest under propolis with a working NVMe drive.

```
kayak-r151047# uname -a
SunOS unknown 5.11 omnios-master-42fe848aa52 i86pc i386 i86pc

kayak-r151047# diskinfo
TYPE    DISK                    VID      PID              SIZE          RMV SSD 
-       c5t0d0                  Virtio   Block Device        0.57 GiB   no  no  
NVME    c6t1d0                  NVMe     -                 300.00 GiB   no  yes

kayak-r151047# zpool create bob c6t1d0
kayak-r151047# zpool list
NAME   SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
bob    298G   146K   298G        -         -     0%     0%  1.00x    ONLINE  -
```

### DIFF
--- a/lib/propolis/src/hw/nvme/admin.rs
+++ b/lib/propolis/src/hw/nvme/admin.rs
@@ -258,11 +258,12 @@ impl NvmeCtrl {
                 // `ncqa`/`nsqa` are 0-based values so subtract 1
                 cmds::Completion::success_val((ncqa - 1) << 16 | (nsqa - 1))
             }
-            // NVMe 1.0e Figure 66 Identify - Identify Controller Data Structure
-            // "If a volatile write cache is present, then the host may ...
-            //  control whether it is enabled with Set Features specifying the
-            //  Volatile Write Cache feature identifier."
             cmds::FeatureIdent::VolatileWriteCache => {
+                // NVMe 1.0e Figure 66 Identify - Identify Controller Data
+                // Structure "If a volatile write cache [VWC] is present, then
+                // the host may ... control whether it is enabled with Set
+                // Features specifying the Volatile Write Cache feature
+                // identifier."
                 cmds::Completion::success()
             }
             cmds::FeatureIdent::Reserved
--- a/lib/propolis/src/hw/nvme/admin.rs
+++ b/lib/propolis/src/hw/nvme/admin.rs
@@ -258,13 +258,19 @@ impl NvmeCtrl {
                 // `ncqa`/`nsqa` are 0-based values so subtract 1
                 cmds::Completion::success_val((ncqa - 1) << 16 | (nsqa - 1))
             }
+            // NVMe 1.0e Figure 66 Identify - Identify Controller Data Structure
+            // "If a volatile write cache is present, then the host may ...
+            //  control whether it is enabled with Set Features specifying the
+            //  Volatile Write Cache feature identifier."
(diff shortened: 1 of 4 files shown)


lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -763,15 +765,26 @@ impl Completion {
...
+        Self {
+            dw0: 0,
+            status: Self::status_field(StatusCodeType::Generic, status, 1),
         }
     }
 
     /// Helper method to combine StatusCodeType and status code
-    fn status_field(sct: StatusCodeType, sc: u8) -> u16 {
-        (sc as u16) << 1 | ((sct as u8) as u16) << 9
+    fn status_field(sct: StatusCodeType, sc: u8, dnr: u8) -> u16 {
```
pfmooney, line 785: Since DNR is just a flag, we should probably pass it as a `bool`

lib/propolis/src/hw/nvme/admin.rs
```diff
@@ -258,13 +258,19 @@ impl NvmeCtrl {
                 // `ncqa`/`nsqa` are 0-based values so subtract 1
                 cmds::Completion::success_val((ncqa - 1) << 16 | (nsqa - 1))
             }
+            // NVMe 1.0e Figure 66 Identify - Identify Controller Data Structure
+            // "If a volatile write cache is present, then the host may ...
+            //  control whether it is enabled with Set Features specifying the
+            //  Volatile Write Cache feature identifier."
```
pfmooney, line 264: style nit: perhaps move the comment into the match case body, to match what you did for `AsyncEventReq`?

lib/propolis/src/hw/nvme/bits.rs
```diff
@@ -867,7 +867,7 @@ pub struct IdentifyController {
     /// See NVMe 1.0e Section 4.2, Figure 8 Command Format - Admin and NVM Vendor Specific Commands (Optional)
     pub avscc: u8,
     /// Reserved
-    pub _resv2: [u8; 246],
+    pub _resv2: [u8; 247],
```
pfmooney, line 870: Would a `sizeof` check on the struct have caught this?  If so, does it make sense adding a unit test with that?

pfmooney (commented):
Thanks for putting this together

lib/propolis/src/hw/nvme/bits.rs
citrus-it: There was one, and it didn't catch it because of the final alignment padding. Adding `packed(1)` to the struct definition makes the test fail on master, and pass on this branch:

```
running 1 test
test hw::nvme::bits::test::entry_sizing ... FAILED

failures:

---- hw::nvme::bits::test::entry_sizing stdout ----
thread 'hw::nvme::bits::test::entry_sizing' panicked at 'assertion failed: `(left == right)`
  left: `4095`,
 right: `4096`', lib/propolis/src/hw/nvme/bits.rs:1145:9
```

I'll add this to all of these structs.

pfmooney (approved):

PR #431, opened 2023-06-06
server: improve HTTP error codes

gjcolombo:
Change some of the HTTP status codes the server emits to hew more closely to the error code semantics in RFC 9110:

- Operations that look up an active VM controller in the server's VM services block, either by hand or through the `vm()` helper function, and don't find one now return 404 Not Found (was 500 Internal Server Error)
- Attempting to ensure an instance with a conflicting ID or properties now returns 409 Conflict (was 500 Internal Server Error)
- Monitoring state for an instance that was destroyed now returns 410 Gone (was 503 Server Unavailable)
- VM controller errors of the "you can't do that because the VM is in the wrong state" class now translate to 403 Forbidden (was 400 Bad Request)

The first of these fixes is especially important when interoperating with sled agent and Nexus. If Nexus asks sled agent to change a Propolis VM's state, and this call fails, sled agent returns the Propolis error to Nexus unmodified. Nexus then tries to reason about the instance's health using the level of the returned error code: 400-level errors are treated as requests that were rejected by a healthy instance, but 500-level errors are treated as fatal, permanent conditions that should move an instance to the Failed state.

Tests: updated PHD tests; ran these changes in an Omicron deployment (with some additional sled agent fixes needed to allow Propolis error statuses to be propagated correctly) and verified that they cleaned up some of the transient instance failures I had previously seen.


pfmooney (approved):

pfmooney:
> Thanks for the review!

Thanks for fixing those result codes.  This looks much better for consumers.

PR #438, opened 2023-06-15
[propolis] make InstanceSerialConsoleHelper::recv cancel-safe

sunshowers:
Alternative to #435.


sunshowers:
This is incorrect because as pointed out on #435, we need to not send characters on the old `ws_stream` once migration has started.

sunshowers:
Per a discussion with @lifning, abandoning this in favor of #435. That's for two reasons:

1. We need to not send characters on the old `ws_stream` once migration is complete, and this gets us farther away from that goal, not closer to it.
2. By the time we receive a migration message, the migration has already completed, so the remaining part (cancel-unsafe) is relatively quick.

PR #442, opened 2023-06-20
Drive vCPUs into consistent state before quiesce

### LINKED ISSUE #429: vCPUs must be in proper state to be exported
As described in [illumos#15728](https://www.illumos.org/issues/15728), there are times when vCPU state is not adequately consistent to be safely migrated when in the middle of instruction emulation in userspace.  Propolis will need to make use of the functionality in the 15728 fix in order to ensure its vCPUs are in a proper state before the instance is considered quiescent and safe for export.

pfmooney:
On platforms with support for the exit-when-consistent flag on VM-entry, ensure that vCPU state is consistent prior to allowing the task which drives it to enter a held (quiesced) state.

Fixes #429


pfmooney:
This uses the mechanism offered by the yet-to-be-integrated [illumos#15728](https://www.illumos.org/issues/15728) ([CR 2908](https://code.illumos.org/c/illumos-gate/+/2908))

gjcolombo (approved):

PR #445, opened 2023-06-21
[propolis-client] ensure that InstanceSerialConsoleHelper is Send

sunshowers:
We regressed on this in #435. Ensure that it's Send, and statically
assert it.


leftwo (approved):
Thanks for the update

PR #451, opened 2023-06-27
Clean up PCI device classes

pfmooney:
A few magic numbers were lingering in the PIIX4 device definitions, in addition to the PIIX3PM revision ID causing dmesg noise when interrogated by modern Linux kernels.


pfmooney:
The dmesg noise about the PM-timer is as follows:
```
[    0.682298] * Found PM-Timer Bug on the chipset. Due to workarounds for a bug,
[    0.682298] * this clock source is slow. Consider trying other clock sources
```

luqmana (approved):

PR #452, opened 2023-06-27
Bump crucible rev to latest

jmpesp:
Pick up the following PRs:

- Set open file resource limit to the max
- Update Rust crate ringbuffer to 0.14
- DTrace meet cmon
- Widen assert values to u128 to deal with u64::MAX
- Change size_to_validate from usize to u64
- Turn on live-repair test in CI
- Increase flush_timeout for some tests, collect cores
- Update to latest dropshot

Also ran the following:

    cargo update -p dropshot

### DIFF
--- a/bin/propolis-server/src/lib/stats.rs
+++ b/bin/propolis-server/src/lib/stats.rs
@@ -3,7 +3,10 @@
 // Copyright 2022 Oxide Computer Company
 
 use anyhow::anyhow;
-use dropshot::{ConfigDropshot, ConfigLogging, ConfigLoggingLevel};
+use dropshot::ConfigDropshot;
+use dropshot::ConfigLogging;
+use dropshot::ConfigLoggingLevel;
+use dropshot::HandlerTaskMode;
 use omicron_common::api::internal::nexus::ProducerEndpoint;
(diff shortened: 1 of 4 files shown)


bin/propolis-server/src/lib/stats.rs
```diff
@@ -3,7 +3,10 @@
 // Copyright 2022 Oxide Computer Company
 
 use anyhow::anyhow;
-use dropshot::{ConfigDropshot, ConfigLogging, ConfigLoggingLevel};
+use dropshot::ConfigDropshot;
+use dropshot::ConfigLogging;
+use dropshot::ConfigLoggingLevel;
+use dropshot::HandlerTaskMode;
```
pfmooney, line 9: Did rustfmt convert this from the parenthetical style?

pfmooney: Do you mind keeping the old style, here and in `main.rs` too?

jmpesp: Not at all, reverted in 88d7a53

pfmooney (approved):
Thanks.  Ship it

PR #454, opened 2023-06-29
Make the propolis zone self-assembling

smklein:
Uses a method script in the Propolis zone to make it self-assembling.

This should, along with a corresponding Omicron change, avoid the need to `zlogin` to the Propolis zone at any point.

Related to: https://github.com/oxidecomputer/omicron/pull/3456

### DIFF
--- a/packaging/smf/propolis-server/manifest.xml
+++ b/packaging/smf/propolis-server/manifest.xml
@@ -4,6 +4,7 @@
 <service_bundle type='manifest' name='propolis-server'>
 
 <service name='system/illumos/propolis-server' type='service' version='1'>
+  <create_default_instance enabled='true' />
   <dependency name='network' grouping='require_all' restart_on='none'
     type='service'>
   <service_fmri value='svc:/milestone/network:default' />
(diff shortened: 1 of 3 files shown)


packaging/smf/propolis-server/manifest.xml
```diff
@@ -15,16 +15,16 @@
...
   <property_group name='config' type='application'>
-    <propval name='server_addr' type='astring' value='unknown' />
-  </property_group>
-
-  <property_group name='config' type='application'>
-    <propval name='metric_addr' type='astring' value='unknown' />
+    <propval name='datalink' type='astring' value='unknown' />
+    <propval name='gateway' type='astring' value='unknown' />
+    <propval name='listen_addr' type='astring' value='127.0.0.1' />
+    <propval name='listen_port' type='astring' value='17000' />
```
luqmana, line 26: Worth changing to the default `12400` propolis-server uses?

smklein: Done

jordanhendricks (approved):

PR #461, opened 2023-07-13
Switch to building on heliosv2

citrus-it:
This is a pre-requisite for moving the rack software on top of heliosv2. Once this is integrated, we can point the omicron pins at the CI artefacts built with the newer helios version.

There is unfortunately no `lab-2.0` buildomat target yet, as @jordanhendricks has noted below; that should appear next week and the PHD tests will start running again.

### DIFF
--- a/.github/buildomat/jobs/falcon-build.sh
+++ b/.github/buildomat/jobs/falcon-build.sh
@@ -2,7 +2,7 @@
 #:
 #: name = "falcon"
 #: variety = "basic"
-#: target = "helios-latest"
+#: target = "helios-2.0"
 #: rust_toolchain = "stable"
 #: output_rules = [
 #:   "/work/release/*",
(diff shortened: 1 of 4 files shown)


jordanhendricks:
The PHD test failures are because we couldn't start up a propolis-server. That failed with:
```
ld.so.1: propolis-server: fatal: libssl.so.3: open failed: No such file or directory
```

I ran a buildomat job against the lab target to check what was in `/etc/os-release`. Looks like we need to update the lab machines to have helios v2:
```
SunOS sock 5.11 helios-1.0.21616 i86pc i386 i86pc
NAME="Helios"
PRETTY_NAME="Oxide Helios 1"
CPE_NAME="cpe:/o:oxide:helios:1:0"
ID=helios
VERSION=1
VERSION_ID=1
BUILD_ID=1.0.2020.10.28
HOME_URL="https://oxide.computer/"
SUPPORT_URL="https://oxide.computer/"
STATE CHANGE: running -> completed
BUG_REPORT_URL="https://github.com/oxidecomputer/helios/issues/new"
```

pfmooney (approved):
We can tolerate the phd-test failures for a short while if this helps unblock other work.

PR #464, opened 2023-07-14
PHD: add Ubuntu 22.04 guest adapter

gjcolombo:
This requires an Ubuntu 22.04 image that has been run through cloud-init such that the user/password combination in the adapter can be used to log in via the serial console. I've uploaded an image to catacomb that has this property.

Tested via a local run with the new guest OS type.

### DIFF
--- a/phd-tests/framework/src/guest_os/mod.rs
+++ b/phd-tests/framework/src/guest_os/mod.rs
@@ -7,6 +7,7 @@ use serde::{Deserialize, Serialize};
 
 mod alpine;
 mod debian11_nocloud;
+mod ubuntu22_04;
 
 /// An entry in a sequence of interactions with the guest's command prompt.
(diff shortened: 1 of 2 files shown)


gjcolombo:
The adapter works as written, but the test image on catacomb needs a little more configuration to really shine:

- the image on catacomb is 10 GiB, which takes a comparatively long time to hash/copy around; it would be nice if it were smaller
- the image's `ubuntu` account can elevate with `sudo` if it needs to run something as root, but using `sudo` from other guest adapters like `Debian11NoCloud` makes the guest sad; it would be nice to resolve this tension one way or another

Probably not a blocker but we'd need to get this resolved on the way to our glorious future in which CI periodically runs tests against a wide array of guests.

jordanhendricks (approved):

PR #469, opened 2023-07-18
Make license headers consistent (and checked)

pfmooney:


leftwo (commented):
Is there a way you can check the repo for licenses outside of github?  
Like from the command line?

pfmooney:
> Is there a way you can check the repo for licenses outside of github? Like from the command line?

Maybe?  The skywalking-eyes tool from the github action can be downloaded and run locally.  I just started with what Hubris had put in place.

pfmooney:
There, now there's an extremely crude xtask for it.

leftwo (approved):
Thanks for adding the local xtask for this!

PR #480, opened 2023-08-02
Elide test (and doctest) steps where not required

pfmooney:


lib/propolis-client/Cargo.toml
```diff
@@ -3,7 +3,8 @@ name = "propolis-client"
 version = "0.1.0"
 edition = "2021"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+doctest = false
```
gjcolombo, line 7: I'd vote to leave doctests enabled for the client lib, since it's the major user-facing crate of this bunch. (I'm not sure how many interesting doctests it has right now, but #472 will add a couple of examples to the instance spec library, and it was helpful to know that the examples in the module documentation at least compiled while I was writing it.)

gjcolombo (approved):
I have one suggested tweak, but beyond that LGTM.

For my understanding (since I've not played with these options much), is the motivation here just to reduce the build time/run time/output spew from running `cargo test`? Do we get additional benefits from eliding these?

pfmooney:
> I have one suggested tweak, but beyond that LGTM.
> 
> For my understanding (since I've not played with these options much), is the motivation here just to reduce the build time/run time/output spew from running `cargo test`?

Yeah, I wanted to skip building crates (for test) which lack any tests to be run.

lib/propolis-client/Cargo.toml
pfmooney: How do you actually get those doctests to be successfully run?  When I build with the `generated` feature, the only one which seems to ship doctests, they are all ignored:
```shell
$ cargo test -p propolis-client --doc -F generated
    Finished test [unoptimized + debuginfo] target(s) in 0.23s
   Doc-tests propolis-client

running 11 tests
test src/generated.rs - generated::Client::instance_ensure (line 9) ... ignored
test src/generated.rs - generated::Client::instance_get (line 9) ... ignored
test src/generated.rs - generated::Client::instance_issue_crucible_snapshot_request (line 11) ... ignored
test src/generated.rs - generated::Client::instance_issue_nmi (line 11) ... ignored
test src/generated.rs - generated::Client::instance_migrate_status (line 9) ... ignored
test src/generated.rs - generated::Client::instance_serial (line 12) ... ignored
test src/generated.rs - generated::Client::instance_serial_history_get (line 13) ... ignored
test src/generated.rs - generated::Client::instance_spec_ensure (line 9) ... ignored
test src/generated.rs - generated::Client::instance_spec_get (line 9) ... ignored
test src/generated.rs - generated::Client::instance_state_monitor (line 9) ... ignored
test src/generated.rs - generated::Client::instance_state_put (line 9) ... ignored

test result: ok. 0 passed; 0 failed; 11 ignored; 0 measured; 0 filtered out; finished in 0.01s

```

gjcolombo:
Makes sense to me, thanks!

lib/propolis-client/Cargo.toml
gjcolombo: Huh, I guess there aren't any in there yet besides the generated ones. In that case, let's merge as-is, and I'll take care of making sure the right tests are running when I next rebase #472 (since I know that will add some).

pfmooney: Sounds good.  I definitely didn't want to disable any existing (and functional) tests with this.

PR #483, opened 2023-08-07
Make crucible task count tunable, with higher default

pfmooney:


lib/propolis/src/block/crucible.rs
```diff
@@ -29,45 +30,54 @@ pub struct CrucibleBackend {
     read_only: bool,
     block_size: u64,
     sectors: u64,
+
+    task_count: NonZeroUsize,
+}
+
+/// If [CreateOptions] does not specify a `task_count`, this value will be used.
+pub const DEFAULT_TASK_COUNT: NonZeroUsize =
+    unsafe { NonZeroUsize::new_unchecked(64) };
```
jordanhendricks, line 39: why is this a NonZeroUsize? Worth a comment?

jordanhendricks (commented):
For the eventual commit/issue description, I think I would like to be a bit more specific about what "tunable" means -- it seems we are making creation of the crucible backend more flexible with a tunable worker task number as input, but we aren't yet exposing any capabilities for changing that beyond the default.

pfmooney:
Closing this out.  If a similar need arises in the future, we can deal with it then.

PR #484, opened 2023-08-09
try reenabling PHD jobs

gjcolombo:
Switch the PHD run job to the `lab-2.0-opte-0.23` target per the discussion in oxidecomputer/meta#245 (this is the correct target to get a baremetal Helios 2.0 environment in the lab).


jordanhendricks (approved):
thanks for pushing this through!

Issue #470, opened 2023-07-18
want support for hot replacing a Crucible downstairs


gjcolombo:
This is a specific case of #205. We discussed it at the 20 April 2023 hypervisor huddle.

Crucible allows a faulted downstairs to be hot-replaced: https://github.com/oxidecomputer/crucible/blob/179e68b8718d24673893dac46a8ab4a9249ec318/upstairs/src/volume.rs#L815. Propolis needs a way to invoke this upstairs function when needed.

The general plan we discussed to wire this up to the control plane involved the following (with some additional details tentatively filled in by yours truly):

- Propolis server will have an "update Crucible downstairs" endpoint that takes a disk ID and a `VolumeConstructionRequest` (or similar, e.g. a new Crucible storage backend element from the instance spec module)
- When invoked, the endpoint takes a 'state change hold' on the current VM's controller that prevents certain state changes from occurring. The list of such changes is TBD; presumably we don't want to live migrate while a configuration change is in progress; is instance stop/reboot still allowed?
- Propolis server looks in the VM controller for
  - a storage backend in the instance description with the correct name (derived from the disk ID)
  - an entry in the disk-ID-to-Crucible-backend map with the correct disk ID
- If both of these exist, the server passes the old and new volume construction requests to the Crucible backend of interest
- The backend compares the requests to see which regions have one or more new addresses
- The backend issues the appropriate `replace_downstairs` calls and returns a result
- If the replacement succeeded, Propolis server updates the relevant instance spec element to reflect the changes that were applied
- Propolis API call returns

Details we need to clarify:

- What instance lifecycle operations are and aren't allowed while a downstairs replacement is in progress?
- What happens if a downstairs replacement request arrives when we're already doing something that's mutually exclusive with that replacement?
  - Queuing the request and completing it asynchronously is challenging because the replacement may happen in the context of a saga, and (IME) it's challenging for a saga to poll for the completion of a particular asynchronous operation.
- Along similar lines: are downstairs replacements fast enough that we can reasonably try to complete them synchronously? What happens if such a request times out?

leftwo:
> Are downstairs replacements fast enough that we can reasonably try to complete them synchronously?

Yes and no.  The "replace a downstairs" interface is asynchronous.  You can wait and verify that the request
is valid from the upstairs and that it will now begin the replacement process internally, as that should happen
without much delay.  The total time of the replacement will be dependent on the size of the volume.

[...]

PR #485, opened 2023-08-10
Update propolis with new Crucible Volume change

leftwo:
New crucible, which has updated Volume structure with Logger field. New VolumeConstructionRequest format which has sub-structures for all the fields.

This also adds support for using a new VCR for replacement
This enables downstairs replacement in a running Volume, and can also be used for migration to validate a VCR.

### DIFF
--- a/lib/propolis-client/src/handmade/api.rs
+++ b/lib/propolis-client/src/handmade/api.rs
@@ -16,7 +16,9 @@ use std::net::SocketAddr;
 use uuid::Uuid;
 
 // Re-export types that are of a public struct
-pub use crucible_client_types::VolumeConstructionRequest;
+pub use crucible_client_types::{
+    VcrFile, VcrRegion, VcrUrl, VcrVolume, VolumeConstructionRequest,
+};
 
 #[derive(Clone, Deserialize, Serialize, JsonSchema)]
 pub struct InstanceNameParams {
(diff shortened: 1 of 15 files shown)


lib/propolis-client/src/handmade/api.rs
```diff
@@ -16,7 +16,9 @@ use std::net::SocketAddr;
 use uuid::Uuid;
 
 // Re-export types that are of a public struct
-pub use crucible_client_types::VolumeConstructionRequest;
+pub use crucible_client_types::{
+    VcrFile, VcrRegion, VcrUrl, VcrVolume, VolumeConstructionRequest,
```
leftwo, line 20: I'm not 100% sure this was the correct place to export these, but it worked :)

leftwo:
This adds some of the support required in https://github.com/oxidecomputer/propolis/issues/470

lib/propolis-client/src/handmade/api.rs
leftwo: This change is no longer part of the PR

leftwo:
This change has been mostly redacted.

All we have here now is an update to the Volume structure to include a `Logger` field, and support for the
few places that had wanted that logger as an arg.

gjcolombo (approved):

PR #488, opened 2023-08-14
propolis-standalone: Update expected crucible opts

leftwo:
The crucible opts have changed.
Propolis-standalone needs to use the new opts.


pfmooney (approved):
Ship it

PR #490, opened 2023-08-14
Add more to standalone-with-crucible

leftwo:
A bit more polish.  
Perfect and getting better every day.


jordanhendricks (approved):
I ran through these instructions yesterday and have a list of things I want to add to the docs. But I see no reason not to merge this now!

Issue #205, opened 2022-08-25
Allow callers to modify a running instance's devices and backends


gjcolombo:
This is needed to support hotplug but might also be used to update backend parameters at runtime (e.g. sending Crucible a replacement VolumeConstructionRequest).

This roughly corresponds to this TODO in `instance_ensure`:

https://github.com/oxidecomputer/propolis/blob/d4bc251898acd96ef3609904fb44931dcc8bea20/bin/propolis-server/src/lib/server.rs#L259-L267

gjcolombo:
With the advent of #470 and #495 this is going to become increasingly important. At a minimum, we need to design a mechanism to ensure that a VM's set of entities remains stable during a migration.

luqmana:
I'd actually not limit this to just `propolis-server` but make it a part of the core `propolis` lib.

There's definitely some design work to do but I'm a bit inclined to have the inventory just store `Arc<RwLock<dyn Entity>>`'s. The general idea being, most access can be via shared refs but occasionally we'd like exclusive access to a particular entity. This does come with some difficulties in straddling async/sync contexts as well as ensuring fairness (readers shouldn't starve out writers).

pfmooney:
>have the inventory just store `Arc<RwLock<dyn Entity>>`'s. 

That would be a pretty substantial change when it comes to the `dyn` types in propolis generally.  

> The general idea being, most access can be via shared refs but occasionally we'd like exclusive access to a particular entity.

What operations are you thinking would use such exclusive access to an entity?

luqmana:
Yea, definitely not a simple tweak. Mostly just been thinking about modeling block backend/device interactions and allowing re-configuring a backend.

PR #495, opened 2023-08-17
new API endpoint for propolis-server to replace a crucible downstairs

leftwo:
A new propolis-server API was added to support the replacement of a downstairs via a new VolumeConstructionRequest.  To do this replacement, we need the current VCR to go along with the new VCR, and we can get that from the InstanceSpec.

As part of this change we have moved the VmObjects InstanceSpec behind a Mutex, as a successful replacement of a crucible backends VCR should be followed by a replacement of the VCR in the instance spec.

Added a new vcr option to crucible-cli which can call this API

This also includes an updated Crucible, as we need the latest version that does not require a Logger with calls to target replace.


bin/propolis-server/src/lib/server.rs
```diff
@@ -953,6 +955,77 @@ async fn instance_issue_crucible_snapshot_request(
...
+    let request = request.into_inner();
+    let new_vcr_json = request.vcr_json;
+    let disk_name = request.name;
+    let log = rqctx.log.clone();
+
+    // Get the instance spec for storage backend from the disk name.  We use
+    // the VCR stored there to send to crucible along with the new VCR we want
+    // to replace it.
+    let vm_controller = rqctx.context().vm().await?;
+    let mut spec = vm_controller.instance_spec().await;
```
gjcolombo, line 978: I think we need a better story for mutating a VM this way, but don't want to hold up progress on the replacement workflow while we work out exactly what that should look like. #205 is tracking the general "change a VM's configuration at runtime" story, so I suggest we refer to that here, and I'll bump its priority up accordingly.

```suggestion

    // TODO(#205): Mutating a VM's configuration should be a first-class
    // operation in the VM controller that synchronizes with ongoing migrations
    // and other attempts to mutate the VM. For the time being, use the instance
    // spec lock to exclude other concurrent attempts to reconfigure this
    // backend.
    let mut spec = vm_controller.instance_spec().await;
```

bin/propolis-server/src/lib/server.rs
```diff
@@ -953,6 +955,77 @@ async fn instance_issue_crucible_snapshot_request(
...
+    // Our replacement request was accepted.  We now need to update the
+    // spec stored in propolis so it matches what the downstairs now has.
+    let new_storage_backend: StorageBackendV0 =
+        StorageBackendV0::Crucible(CrucibleStorageBackend {
+            readonly,
+            request_json: new_vcr_json,
+        });
+    v0_spec.backends.storage_backends.insert(disk_name, new_storage_backend);
+
+    slog::info!(log, "Replaced the VCR in backend of {:?}", path_params.id);
```
gjcolombo, line 1024: nit: I might push this kind of operational log down into `vcr_replace`, but it's up to you.

bin/propolis-server/src/lib/server.rs
```diff
@@ -953,6 +955,77 @@ async fn instance_issue_crucible_snapshot_request(
...
+    })?;
+
+    slog::info!(
+        log,
+        "{:?} {:?} replace {:?} with {:?}",
+        disk_name,
+        path_params.id,
+        old_vcr_json,
+        new_vcr_json,
+    );
```
gjcolombo, line 1005: Optional: I've begun evangelizing for putting values into log messages using KV pairs instead of interleaving them into messages:

```suggestion
    slog::info!(
        log,
        "replacing downstairs for Crucible disk";
        "disk_name" => %disk_name,
        "volume_id" => %path_params.id,
        "old_vcr" => %old_vcr_json,
        "new_vcr" => %new_vcr_json,
    );
```

When examining logs you can then filter on the specific fields you're interested in instead of having to search for them in the message body; e.g. you can do `looker -c 'r.volume_id?.contains("<guid>")' and find any record that has a structured volume ID set in it irrespective of how its message was formatted.

Then again, `looker -c 'r.msg.contains("<guid>")` is not *that* much more onerous, so this is your call.

leftwo (commented):
I updated the docs too, now that the VCR is expected to be a string.

bin/propolis-server/src/lib/server.rs
gjcolombo: Makes sense to me. I suspect we'll want something similar in whatever we ultimately do for #205.

gjcolombo (approved):
LGTM once we've got a pointer to #205 and a note that `instance_spec()` returns a lock guard in `instance_issue_crucible_vcr_request`. (We touched base on this offline.)

Thanks for driving this one home!

bin/propolis-server/src/lib/server.rs
leftwo: I took the log line out, but I'll remember that trick for future logging.

bin/propolis-server/src/lib/server.rs
leftwo: Got it, I finally figured out what github was showing me here.

PR #500, opened 2023-08-18
Add MemAsync block backend

pfmooney:
The MemAsync block backend is intended for use in performance testing block device emulation.  It uses a region of anonymous memory as "storage" to keep IOP costs low.


pfmooney:
Merged in ff10efaec1b4c60db957a783160a903489bbd641

PR #501, opened 2023-08-21
merge VCR replacement with versioned instance specs

gjcolombo:
Pick up #472 into #495 and fix the build.

The main differences from #472 are the following:

- `InstanceSpec` is now `VersionedInstanceSpec`; some destructuring is needed to work with `InstanceSpecV0` where previously it sufficed work with `InstanceSpec`.
- The new Crucible backend spec stores `VolumeConstructionRequest`s in serialized form. This is so that Crucible can handle VCR versioning however it wants without having to worry about Propolis's instance spec versioning rules.

To deal with the second change, I updated the VCR replacement endpoint to take the new VCR in serialized form. In the long run I think this will end up being more efficient for Omicron: Nexus can fetch a serialized VCR and pass it in string form to sled agent, which can pass it in that form to Propolis, which can pass it to the Crucible backend, which can pass it to the Crucible upstairs, which can deserialize it. Without this, at a minimum, sled agent will have to deserialize the VCR to hand it to the propolis-client API, which will promptly serialize it to send it to propolis-server, which will immediately deserialize it again.

This change passes `cargo test` but I haven't explicitly retested the replacement API.


bin/propolis-server/src/lib/server.rs
```diff
@@ -975,23 +976,15 @@ async fn instance_issue_crucible_vcr_request(
...
-                StorageBackendKind::Crucible { req } => (readonly, req),
-                x => {
-                    let s = format!(
-                        "Invalid StorageBackendKind for VCR replacement: {:?}",
-                        x
-                    );
-                    return Err(HttpError::for_not_found(Some(s.clone()), s));
-                }
-            }
+            (readonly, &bes.request_json)
```
leftwo, line 985: Could we just return `(bes.readonly, &bes.request_json)` and avoid having a `readonly` local here?

leftwo (approved):
So, I got a little lost here trying to figure out what pieces were changed to make my original PR compatable, and what parts were stuff that we are merging from main.  I think I found
the parts were that happened, and I don't see any issue with those changes.  Maybe after we merge this I'll take a second look at the diffs again to see if anything stands out, but
I'm pretty sure I won't have any issue.

We can merge this whenever you are comfortable with it landing, I know you have
pending comments on my PR and we don't need to merge this till you are done with
those.

bin/propolis-server/src/lib/server.rs
gjcolombo: Oops. Fixed in 03e3d99.

gjcolombo:
We discussed this in DMs. All the interesting bits are in the post-merge commits in this diff: https://github.com/oxidecomputer/propolis/pull/501/files/f1d8ae325793bc7329d95f87dd7bdf3fe99534e8..ec06f6611b25f0b9e5447ae1003872c1ce65c2fc. There are some resolved conflicts in the merge commit itself, but they're trivial (renaming `InstanceSpec` to `VersionedInstanceSpec` when needed).

Hopefully merging this won't make #495 explode.

PR #502, opened 2023-08-22
Clean up and restructure CQE handling in NVMe

pfmooney:
This attempts to consolidate some of the shared logic between NVMe completion queues and submission queues.  Additionally, addresses a theoretical issue in setting the correct phase bit in CQEs when threads race to push completions as the queue rolls over.

This is work I drafted while investigating #427.  It isn't especially high priority, but I wanted to get it cleaned up and posted, rather than stagnating in a branch.

### DIFF
--- a/lib/propolis/src/hw/nvme/cmds.rs
+++ b/lib/propolis/src/hw/nvme/cmds.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use super::bits::{self, RawSubmission, StatusCodeType};
+use super::bits::{self, StatusCodeType, SubmissionQueueEntry};
 use super::queue::{QueueCreateErr, QueueId};
 use crate::block;
 use crate::common::*;
@@ -17,10 +17,10 @@ mod probes {
     fn nvme_prp_error(err: &'static str) {}
 }
 
-/// Errors that may be encounted during command parsing.
+/// Errors that may be encountered during command parsing.
 #[derive(Debug, Error)]
 pub enum ParseErr {
-    /// Encounted a fused operation which we don't currently support.
+    /// We do not currently support fused operations
     #[error("Fused ops not supported")]
     Fused,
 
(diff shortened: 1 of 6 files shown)


lib/propolis/src/hw/nvme/queue.rs
```diff
@@ -551,63 +593,55 @@ impl CompQueue {
...
-    /// Returns the corresponding [`GuestAddr`] for a given entry in
-    /// the Completion Queue.
-    fn entry_addr(&self, idx: u16) -> GuestAddr {
-        let res = self.base.0
-            + idx as u64 * std::mem::size_of::<RawCompletion>() as u64;
-        GuestAddr(res)
+        // The only definite indicator that a CQE has become valid is the phase
+        // bit being toggled.  Since the interface for writing to guest memory
+        // cannot ensure that the other bits of the CQE are written before the
+        // phase bit, we must proceed carefully:
```
luqmana, line 628: oof, nice catch!

luqmana (approved):
Thanks for cleaning this up!

lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -53,12 +53,12 @@ pub enum AdminCmd {
     /// Asynchronous Event Request Command
     AsyncEventReq,
     /// An unknown admin command
-    Unknown(RawSubmission),
+    Unknown(SubmissionQueueEntry),
 }
 
 impl AdminCmd {
     /// Triy to parse an `AdminCmd` out of a raw Submission Entry.
```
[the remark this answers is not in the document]
pfmooney: Done

lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -448,12 +448,12 @@ pub enum NvmCmd {
     /// Read data and metadata
     Read(ReadCmd),
     /// An unknown NVM command
-    Unknown(RawSubmission),
+    Unknown(SubmissionQueueEntry),
 }
 
 impl NvmCmd {
     /// Triy to parse an `NvmCmd` out of a raw Submission Entry.
```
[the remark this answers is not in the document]
pfmooney: Done

lib/propolis/src/hw/nvme/queue.rs
pfmooney: Thanks.

Was it also clear how the old system of querying the phase bit for the CQ outside the lock could result in the wrong phase being set for an entry if two race during completion?  I _think_ I got that right, but there also wasn't a good place to make a comment about it.

pfmooney:
I took a wider spin through that file with `set spell` on to find a few more typos.

lib/propolis/src/hw/nvme/queue.rs
luqmana: mmm, that was definitely an oversight in the previous handling.

luqmana (approved):

Issue #492, opened 2023-08-17
virtio-block needs `VIRTIO_BLK_F_FLUSH` / `VIRTIO_BLK_T_FLUSH` support.


luqmana:
While investigating seemingly much worse performance for the `nvme` device compared to `virtio-block`, @rmustacc pointed out we set the [Volatile Write Cache bit for `nvme` devices](https://github.com/oxidecomputer/propolis/blob/803cb36d9cc1ca64cb38665173a7be5c06887580/lib/propolis/src/hw/nvme/mod.rs#L525-L526) but not the similar flush capability for `virtio-block`.

As a quick test, I tried clearing the `vwc` bit and rerunning `pgbench` with an `nvme` device and got similar (if not very slightly better) results compared to `virtio-block` (both using the file backend). Given the fact the file backend also doesn't use any sync flags on open, the speedup we were seeing on `virtio-block` makes sense: turns out it's faster to just assume things are synchronous (when in actually they're not) and never call flush/fsync.

The issue is [we don't](https://github.com/oxidecomputer/propolis/blob/803cb36d9cc1ca64cb38665173a7be5c06887580/lib/propolis/src/hw/virtio/block.rs#L163-L174) ever try to negotiate the `VIRTIO_BLK_F_FLUSH` feature today. Per the VIRTIO spec[^1]:

>An implementation that does not offer `VIRTIO_BLK_F_FLUSH` and does not commit completed writes will not be resilient to data loss in case of crashes.

In addition to advertising and trying to negotiate `VIRTIO_BLK_F_FLUSH`, we then subsequently need to support `VIRTIO_BLK_T_FLUSH` commands and forward appropriately to the backend.

Note: implementation wise we can also just choose to always commit writes even without flush support:

>If `VIRTIO_BLK_F_FLUSH` was not offered by the device, the device MAY also commit writes to persistent device backend storage before reporting their completion.

But this relies on better support on the backend's side as well.

[^1]: [5.2.6.2 Device Requirements: Device Operation](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-2520002)
[^2]: note at this point adding flush support would seemingly regress `virtio-block` performance but that's just a reflection of no longer trading in reliability. Regardless, there's separate ongoing work to improve the performance overall.

PR #504, opened 2023-08-23
vioblk: add flush support

luqmana:
Partially addresses #492. Still need more work to support actually negotiating features and allowing reconfiguring block backends. But this bring us to par with the nvme device today.

### DIFF
--- a/lib/propolis/src/block/mod.rs
+++ b/lib/propolis/src/block/mod.rs
@@ -57,9 +57,6 @@ pub type BlockPayload = dyn Any + Send + Sync + 'static;
 
 /// Block device operation request
 pub struct Request {
-    /// A device specific value identifying this request.
-    id: u32,
-
     /// The type of operation requested by the block device
     op: Operation,
 
@@ -79,44 +76,35 @@ pub struct Request {
 }
 impl Request {
     pub fn new_read(
-        id: u32,
         off: usize,
         regions: Vec<GuestRegion>,
         payload: Box<BlockPayload>,
     ) -> Self {
         let op = Operation::Read(off);
-        Self { id, op, regions, payload: Some(payload), outstanding: None }
+        Self { op, regions, payload: Some(payload), outstanding: None }
     }
 
     pub fn new_write(
-        id: u32,
         off: usize,
         regions: Vec<GuestRegion>,
         payload: Box<BlockPayload>,
     ) -> Self {
         let op = Operation::Write(off);
-        Self { id, op, regions, payload: Some(payload), outstanding: None }
+        Self { op, regions, payload: Some(payload), outstanding: None }
     }
 
     pub fn new_flush(
-        id: u32,
         off: usize,
         len: usize,
         payload: Box<BlockPayload>,
     ) -> Self {
         let op = Operation::Flush(off, len);
         Self {
-            id,
             op,
             regions: Vec::new(),
             payload: Some(payload),
             outstanding: None,
         }
-    }
-
-    /// Device specific request ID.
-    pub fn id(&self) -> u32 {
-        self.id
     }
 
     /// Type of operation being issued.
@@ -154,7 +142,7 @@ impl Request {
     /// Indiciate disposition of completed request
     pub fn complete(mut self, res: Result, dev: &dyn Device) {
         let payload = self.payload.take().unwrap();
-        dev.complete(self.id, self.op, res, payload);
+        dev.complete(self.op, res, payload);
 
         // Update the outstanding I/O count
         self.outstanding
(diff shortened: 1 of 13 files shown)


lib/propolis/src/block/mod.rs
```diff
@@ -57,6 +57,9 @@ pub type BlockPayload = dyn Any + Send + Sync + 'static;
 
 /// Block device operation request
 pub struct Request {
+    /// A device specific value identifying this request.
+    id: u32,
+
```
pfmooney, line 62: Why not leave this as part of the payload?  Then NVMe could use struct members rather than having to pack it into a u32

lib/propolis/src/hw/virtio/block.rs
```diff
@@ -76,6 +90,15 @@ impl PciVirtioBlock {
         }
     }
 
+    /// Return a combined queue + request id for associating an I/O
+    /// request with its subsequent completion.
+    ///
+    /// Note: this is only used for debugging purposes.
+    fn next_req_id(&self) -> u32 {
+        // We only support a single queue today
+        (0 << 16) | self.next_request_id.fetch_add(1, Ordering::Relaxed) as u32
```
pfmooney, line 99: We could pull the index out of the `Chain` and use that as the ID

lib/propolis/src/block/mod.rs
luqmana: done

lib/propolis/src/hw/virtio/block.rs
luqmana: i did try that and found the space of ids is much smaller that way (e.g. booting alpine would issue some ~1300 reads but reuse `idx` 0-8). Not a huge issue, but figured if a trace drops some events you might end up correlating the wrong enqueue/complete probes.

lib/propolis/src/block/mod.rs
```diff
@@ -42,13 +42,13 @@ pub enum Operation {
     Read(ByteOffset),
     /// Write to offset
     Write(ByteOffset),
-    /// Flush buffer(s) for [offset, offset + len)
-    Flush(ByteOffset, ByteLen),
```
pfmooney, line 46: Sensible clean-up until we have a need for the extra detail

lib/propolis/src/hw/virtio/block.rs
pfmooney: Ah, so this is just the descriptor ID.  We could have `VirtQueue::pop_avail` (and `VqAvail::read_next_avail`) emit the index in the avail ring in addition to the descriptor ID.  Then you'd get a range as large as virtqueue, and be able to correlate that back to memory contents if you paused the VM to inspect it.

We could also leave as-written.  What do you think?

luqmana: Updated to use the avail idx

lib/propolis/src/hw/nvme/requests.rs
```diff
@@ -90,11 +87,21 @@ impl PciNvme {
                 let cmd = NvmCmd::parse(sub);
 
                 match cmd {
-                    Ok(NvmCmd::Write(_)) if !state.binfo.writable => {
+                    Ok(NvmCmd::Write(cmd)) if !state.binfo.writable => {
+                        let off = state.nlb_to_size(cmd.slba as usize) as u64;
+                        let size = state.nlb_to_size(cmd.nlb as usize) as u64;
+                        probes::nvme_write_enqueue!(|| (
+                            qid, idx, cid, off, size
+                        ));
```
pfmooney, line 95: Good to simulate the enqueue/complete for the known failure

lib/propolis/src/block/mod.rs
```diff
@@ -42,13 +42,13 @@ pub enum Operation {
...
+    /// Flush buffer(s)
+    Flush,
 }
 
 #[derive(Copy, Clone, Debug)]
 pub enum Result {
-    Success,
+    Success = 0,
```
pfmooney, line 51: Nice.  Glad it cleaned up the resnum cast later.

lib/propolis/src/hw/nvme/queue.rs
```diff
@@ -712,10 +712,14 @@ pub struct ProtoPermit {
 
     /// The Submission Queue for which this entry is reserved.
     sq: Weak<SubQueue>,
+
+    /// The Submission Queue's ID (stored separately to avoid going through
+    /// the Weak ref).
```
pfmooney, line 717: Good idea to cache this

lib/propolis/src/hw/virtio/block.rs
```diff
@@ -192,13 +234,14 @@ impl block::Device for PciVirtioBlock {
...
-        _op: block::Operation,
+        op: block::Operation,
         res: block::Result,
         payload: Box<block::BlockPayload>,
     ) {
-        let mut chain: Box<Chain> =
+        let mut payload: Box<CompletionPayload> =
             payload.downcast().expect("payload must be correct type");
-        self.complete_req(res, &mut chain);
+        let CompletionPayload { rid, ref mut chain } = *payload;
```
pfmooney, line 243: I'm sad that the stabilization of `Box::into_inner()` seems to have stalled.

pfmooney (approved):

lib/propolis/src/hw/virtio/block.rs
luqmana: Yea, I had initially reached for that until I realized 😓

PR #506, opened 2023-08-23
Improve error message when /dev/vmmctl not present

### LINKED ISSUE #505: errror on failure to `open` vmm controller fd could include more context
In `create_vm`, the `open` for `/dev/vmmctl` simply returns the IO error on failure: https://github.com/oxidecomputer/propolis/blob/f9bff8368bf8625319a99d7c09d8aa09c112e0e5/lib/propolis/src/vmm/hdl.rs#L49

On machines in which SVM is disabled in the BIOS, this `open` call will fail, but the error message isn't super helpful (`error 2 file not found`).

This error message could at least include a bit more context about what happened.

arjenroodselaar:
When `/dev/vmmctl` is not present both `propolis-standalone` and `propolis-server` exit during the API version check without providing the user much context as to why. This diff logs the fact that the device node is not there so the user has a pointer where to start looking for the problem. The resulting output looks as follows:
```
arjen@helios:~/propolis/bin/propolis-server$ cargo run -- run server.cfg 127.0.0.1:3000
Aug 23 22:45:55.319 ERRO Failed to open /dev/vmmctl
Error: API version checks

Caused by:
    0: IO Error
    1: No such file or directory (os error 2)
```
```
arjen@helios:...opolis/bin/propolis-standalone$ cargo run -- vm.cfg
Aug 23 22:45:24.783 ERRO Failed to open /dev/vmmctl
Error: API version checks

Caused by:
    No such file or directory (os error 2)
```
The logging with context and nested types make this less clean than I was shooting for, but it's functional. Happy to take a different approach though.

Fixes #505.

### DIFF
--- a/bin/propolis-server/src/main.rs
+++ b/bin/propolis-server/src/main.rs
@@ -73,8 +73,18 @@ async fn run_server(
     vnc_addr: SocketAddr,
     log: slog::Logger,
 ) -> anyhow::Result<()> {
+    use propolis::api_version;
+
     // Check that devices conform to expected API version
-    propolis::api_version::check().context("API version checks")?;
+    if let Err(e) = api_version::check() {
+        if let api_version::Error::Io(ioe) = &e {
+            if ioe.kind() == std::io::ErrorKind::NotFound {
+                slog::error!(log, "Failed to open /dev/vmmctl");
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


pfmooney (approved):

jordanhendricks (approved):
Thanks!

PR #519, opened 2023-08-30
Make storage backend config more flexible and consistent

pfmooney:
With the integration of #504, certain lab workloads exhibited pathological performance due to the added flushes.  It would be nice to be able to elide said flushes when desired.

Additionally, making handling of such general configuration items (block size, read-only status, flush elision) consistent across the backends would be a welcomed improvement.

### DIFF
--- a/lib/propolis/src/block/crucible.rs
+++ b/lib/propolis/src/block/crucible.rs
@@ -26,15 +26,14 @@ pub struct CrucibleBackend {
     block_io: Arc<Volume>,
     scheduler: block::Scheduler,
 
-    read_only: bool,
-    block_size: u64,
-    sectors: u64,
+    info: block::DeviceInfo,
+    skip_flush: bool,
 }
 
 impl CrucibleBackend {
     pub fn create(
         request: VolumeConstructionRequest,
-        read_only: bool,
+        opts: block::BackendOpts,
         producer_registry: Option<ProducerRegistry>,
         nexus_client: Option<NexusClient>,
         log: slog::Logger,
@@ -43,7 +42,7 @@ impl CrucibleBackend {
         rt.block_on(async move {
             CrucibleBackend::_create(
                 request,
-                read_only,
+                opts,
                 producer_registry,
                 nexus_client,
                 log,
@@ -55,7 +54,7 @@ impl CrucibleBackend {
 
     async fn _create(
         request: VolumeConstructionRequest,
-        read_only: bool,
+        opts: block::BackendOpts,
         producer_registry: Option<ProducerRegistry>,
         nexus_client: Option<NexusClient>,
         log: slog::Logger,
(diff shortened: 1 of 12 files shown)


lib/propolis/src/block/mem_async.rs
```diff
@@ -59,17 +45,19 @@ impl MemAsyncBackend {
...
             workers,
             scheduler: block::Scheduler::new(),
 
-            read_only,
-            block_size,
-            sectors: size / block_size,
+            info: block::DeviceInfo {
+                block_size,
+                total_size: size / block_size as u64,
+                writable: opts.read_only.unwrap_or(false),
```
luqmana, line 59:
```suggestion
                writable: !opts.read_only.unwrap_or(false),
```

Mind updating `DeviceInfo::writable` to `read_only` while you're here to be consistent everywhere

lib/propolis/src/block/in_memory.rs
```diff
@@ -59,9 +46,11 @@ impl InMemoryBackend {
                 NonZeroUsize::new(1).unwrap(),
             ),
 
-            read_only,
-            block_size,
-            sectors: len / block_size,
+            info: block::DeviceInfo {
+                block_size,
+                total_size: len as u64 / block_size as u64,
+                writable: opts.read_only.unwrap_or(false),
```
luqmana, line 52:
```suggestion
                writable: !opts.read_only.unwrap_or(false),
```

lib/propolis/src/block/file.rs
```diff
@@ -45,10 +44,19 @@ impl FileBackend {
...
+            )),
+            (Some(ro), false) => Ok(ro),
+            (_, file_ro) => Ok(file_ro),
+        }?;
 
         let fp = OpenOptions::new().read(true).write(!read_only).open(p)?;
-        let len = fp.metadata().unwrap().len() as usize;
+        let len = fp.metadata().unwrap().len();
+        // TODO: attempt to query blocksize from underlying file/zvol
+        let block_size = opts.block_size.unwrap_or(512);
```
luqmana, line 59: Don't wanna use `DEFAULT_BLOCK_SIZE` here too?

lib/propolis/src/block/crucible.rs
```diff
@@ -116,9 +115,12 @@ impl CrucibleBackend {
             block_io: Arc::new(volume),
             scheduler: block::Scheduler::new(),
 
-            read_only,
-            block_size,
-            sectors,
+            info: block::DeviceInfo {
+                block_size: block_size as u32,
+                total_size: sectors,
+                writable: opts.read_only.unwrap_or(false),
```
luqmana, line 121:
```suggestion
                writable: !opts.read_only.unwrap_or(false),
```

lib/propolis/src/block/file.rs
pfmooney: Waffled on it, so.... done.

lib/propolis/src/block/mem_async.rs
pfmooney: > Mind updating `DeviceInfo::writable` to `read_only`

Should be done.

luqmana (approved):

PR #524, opened 2023-09-15
Wire up bits for CPUID customization

pfmooney:
This is a first cut at making CPUID information customizable, rather than using the in-kernel vmm logic which applies limited transformations to the data exposed by the host CPU.  Basic plumbing in propolis-lib is provided, which is wired into propolis-standalone.  The same configuration syntax is defined for propolis-server, but it is going unused for now until some internal structure is hammered out.

### DIFF
--- a/bin/propolis-standalone/src/config.rs
+++ b/bin/propolis-standalone/src/config.rs
@@ -11,11 +11,12 @@ use anyhow::Context;
 use serde::Deserialize;
 
 use propolis::block;
+use propolis::cpuid;
 use propolis::hw::pci::Bdf;
 use propolis::inventory::ChildRegister;
 
-use propolis_standalone_config::Device;
 pub use propolis_standalone_config::{Config, SnapshotTag};
+use propolis_standalone_config::{CpuVendor, CpuidEntry, Device};
 
 #[derive(Deserialize)]
 struct FileConfig {
@@ -116,6 +117,34 @@ pub fn parse_bdf(v: &str) -> Option<Bdf> {
         Bdf::new(fields[0], fields[1], fields[2])
     } else {
         None
+    }
+}
+
+pub fn parse_cpuid(config: &Config) -> anyhow::Result<Option<cpuid::Set>> {
+    if let Some(profile) = config.cpuid_profile() {
+        let vendor = match profile.vendor {
+            CpuVendor::Amd => cpuid::VendorKind::Amd,
+            CpuVendor::Intel => cpuid::VendorKind::Intel,
+        };
+        let mut set = cpuid::Set::new(vendor);
+        let entries: Vec<CpuidEntry> = profile.try_into()?;
+        for entry in entries {
+            let conflict = set.insert(
+                cpuid::Ident(entry.func, entry.idx),
+                cpuid::Entry::from(entry.values),
+            );
+
+            if conflict.is_some() {
+                anyhow::bail!(
+                    "conflicing entry at func:{:#?} idx:{:#?}",
+                    entry.func,
+                    entry.idx
+                )
+            }
+        }
+        Ok(Some(set))
+    } else {
+        Ok(None)
     }
 }
 
--- a/crates/bhyve-api/sys/src/structs.rs
+++ b/crates/bhyve-api/sys/src/structs.rs
@@ -392,16 +392,57 @@ impl Default for vm_data_xfer {
 pub const VCE_FLAG_MATCH_INDEX: u32 = 1 << 0;
 
 #[repr(C)]
-#[derive(Copy, Clone)]
-struct vcpu_cpuid_entry {
-    vce_function: u32,
-    vce_index: u32,
-    vce_flags: u32,
-    vce_eax: u32,
-    vce_ebx: u32,
-    vce_ecx: u32,
-    vce_edx: u32,
-    _pad: u32,
+#[derive(Copy, Clone, Default)]
+pub struct vcpu_cpuid_entry {
+    pub vce_function: u32,
+    pub vce_index: u32,
+    pub vce_flags: u32,
+    pub vce_eax: u32,
+    pub vce_ebx: u32,
+    pub vce_ecx: u32,
+    pub vce_edx: u32,
+    pub _pad: u32,
+}
+impl vcpu_cpuid_entry {
+    fn match_idx(&self) -> bool {
+        self.vce_flags & VCE_FLAG_MATCH_INDEX != 0
+    }
+    /// Order entries for proper cpuid evaluation by the kernel VMM.
+    ///
+    /// Bhyve expects that cpuid entries are sorted by function, and then index,
+    /// from least to greatest.  Entries which must match on index should come
+    /// before (less-than) those that do not, so the former can take precedence
+    /// in matching.
+    ///
+    /// This function is provided so that a list of entries can be easily sorted
+    /// prior to loading them into the kernel VMM.
+    ///
+    /// ```
+    /// let mut entries: Vec<vcpu_cpuid_entry> = vec![
+    ///     // entries loaded here
+    /// ];
+    /// entries.sort_by(vcpu_cpuid_entry::eval_sort);
+    /// let config = vm_vcpu_cpuid_config {
+    ///     vvcc_cpuid: 0,
+    ///     vvcc_flags: 0,
+    ///     vvcc_nent: entries.len(),
+    ///     vvcc_entries: &mut entries,
+    /// };
+    /// // perform ioctl(VM_SET_CPUID, &config) ...
+    /// ```
+    pub fn eval_sort(a: &Self, b: &Self) -> std::cmp::Ordering {
+        use std::cmp::Ordering;
+
+        match a.vce_function.cmp(&b.vce_function) {
+            Ordering::Equal => match (a.match_idx(), b.match_idx()) {
+                (true, false) => Ordering::Less,
+                (false, true) => Ordering::Greater,
+                (true, true) | (false, false) => a.vce_index.cmp(&b.vce_index),
+            },
+
+            ord => ord,
+        }
+    }
 }
 
 /// Use legacy hard-coded cpuid masking tables applied to the host CPU
@@ -420,6 +461,17 @@ pub struct vm_vcpu_cpuid_config {
     pub vvcc_nent: u32,
     pub _pad: u32,
     pub vvcc_entries: *mut c_void,
+}
+impl Default for vm_vcpu_cpuid_config {
+    fn default() -> Self {
+        Self {
+            vvcc_vcpuid: 0,
+            vvcc_flags: 0,
+            vvcc_nent: 0,
+            _pad: 0,
+            vvcc_entries: std::ptr::null_mut(),
+        }
+    }
 }
 
 #[repr(C)]
--- a/crates/cpuid-profile-config/src/lib.rs
+++ b/crates/cpuid-profile-config/src/lib.rs
@@ -1,0 +1,85 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::collections::BTreeMap;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize, Copy, Clone, Debug, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum CpuVendor {
+    Amd,
+    Intel,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct CpuidProfile {
+    pub vendor: CpuVendor,
+    #[serde(flatten, default)]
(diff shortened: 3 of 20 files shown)


lib/propolis/src/cpuid.rs
```diff
@@ -0,0 +1,413 @@
...
+            inout("eax") ident.0 => res.eax,
+            inout("ecx") ident.1.unwrap_or(0) => res.ecx,
+            out("edx") res.edx,
+        );
+    }
+    res
+}
+#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
+pub fn host_query(_ident: Ident) -> Entry {
+    panic("this is not going to work on non-x86")
```
gjcolombo, line 412: s/panic/panic!

(propolis on AARCH64 when)

lib/propolis/src/cpuid.rs
```diff
@@ -0,0 +1,413 @@
...
+
+/// Specialize a set of cpuid leafs for provided attributes.
+///
+/// This includes things such as a CPU topology (cores/threads/etc), a given
+/// vCPU ID (APIC, core/thread ID, etc), or other info tidbits.
+#[derive(Default)]
+pub struct Specializer {
+    has_smt: bool,
+    num_vcpu: Option<NonZeroU8>,
+    vcpuid: Option<i32>,
```
gjcolombo, line 159: Do we ever expect these to be None? I'd almost be inclined to make `num_vcpu` a required parameter to `new` and `vcpuid` a required parameter to `execute`.

lib/propolis/src/cpuid.rs
```diff
@@ -0,0 +1,413 @@
...
+    vendor_kind: Option<VendorKind>,
+    cpu_topo: BTreeSet<TopoKind>,
+    do_cache_topo: bool,
+}
+impl Specializer {
+    pub fn new() -> Self {
+        Self::default()
```
gjcolombo, line 166: Should `has_smt` default to true or false? (RFD 314's discussion of leaf 0xB proposes enabling SMT in the guest by default.)

lib/propolis/src/cpuid.rs
```diff
@@ -0,0 +1,413 @@
...
+                    vals.eax |= (visible_count - 1) << 14;
+                }
+            }
+        }
+    }
+    fn fix_cpu_topo(&self, set: &mut Set) {
+        for topo in enum_iterator::all::<TopoKind>() {
+            // Nuke any existing info in order to potentially override it
+            let leaf = topo as u32;
+            set.remove_all(leaf);
```
gjcolombo, line 283: This strikes me as a bit odd. If I write a config file that has explicit settings for leaf B/1F/8..1E, but then the specializer declines to pass a value for `num_vcpu`, then the entries I put in the config file will be silently dropped.

Should it be an error to `fix_cpu_topo` for a particular leaf if the original `set` contains that leaf and this function doesn't have enough information to specialize it?

lib/propolis/src/cpuid.rs
```diff
@@ -0,0 +1,413 @@
...
+                    continue;
+                }
+            };
+
+            if self.cpu_topo.contains(&topo) {
+                match topo {
+                    TopoKind::StdB => {
+                        // Queries with invalid ecx will get all-zeroes
+                        set.insert(Ident(leaf, None), Entry::zero());
+                        if self.has_smt {
```
gjcolombo, line 298: Related to the above: if I want to test a guest OS with a zany set of outputs from a topology leaf, and I write a config file accordingly, and Propolis runs this logic, my settings will be silently overwritten. Should this return an error instead?

lib/propolis/src/cpuid.rs
```diff
@@ -0,0 +1,413 @@
...
+                        // TODO: add 0x1f topo info
+                    }
+                    TopoKind::Ext1E => {
+                        let id = self.vcpuid.unwrap_or(0) as u32;
+                        let compute_id = if self.has_smt { id / 2 } else { id };
+                        set.insert(
+                            Ident(leaf, None),
+                            Entry {
+                                eax: id,
+                                ebx: 0x100 | id,
```
gjcolombo, line 342: If I'm reading things right, `ebx` bit 8 should be cleared if `has_smt` is false.

lib/propolis/src/cpuid.rs
```diff
@@ -0,0 +1,413 @@
...
+                    let visible_count = match (vals.eax & 0b11100000 >> 5) {
+                        0b001 | 0b010 => {
+                            // L1/L2 shared by SMT siblings
+                            if self.has_smt {
+                                2
+                            } else {
+                                1
+                            }
+                        }
+                        0b011 => {
```
gjcolombo, line 261: This is fine for now but is a little different from the RFD 314 proposal here (I've leaned toward exposing the underlying CCX membership to guests so that guest schedulers can cooperate with whatever CCX-aware scheduling we happen to implement). We should file an issue to close on this.

As with the B/1F/8..1E settings, this is a place where I think it might make sense to honor explicit settings from the caller if they happen to exist (so that the control plane can define distinct virtual platforms that expose different cache topologies and apply them at different times to different instances without having to upgrade/downgrade any Propolis binaries).

bin/propolis-standalone/src/config.rs
```diff
@@ -119,6 +120,27 @@ pub fn parse_bdf(v: &str) -> Option<Bdf> {
...
+pub fn parse_cpuid(config: &Config) -> anyhow::Result<Option<cpuid::Set>> {
+    if let Some(profile) = config.cpuid_profile() {
+        let vendor = match profile.vendor {
+            CpuVendor::Amd => cpuid::VendorKind::Amd,
+            CpuVendor::Intel => cpuid::VendorKind::Intel,
+        };
+        let mut set = cpuid::Set::new(vendor);
+        let entries: Vec<CpuidEntry> = profile.try_into()?;
+        for entry in entries {
+            // TODO: balk at conflicts?
```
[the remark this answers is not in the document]
pfmooney: Done

crates/bhyve-api/sys/src/structs.rs
```diff
@@ -392,16 +392,57 @@ impl Default for vm_data_xfer {
...
+}
+impl vcpu_cpuid_entry {
+    fn match_idx(&self) -> bool {
+        self.vce_flags & VCE_FLAG_MATCH_INDEX != 0
+    }
+    /// Order entries for proper cpuid evaluation by the kernel VMM.
+    ///
+    /// Bhyve expects that cpuid entries are sorted by function, and then index,
+    /// from least to greatest.  Entries which must match on index should come
+    /// before (less-than) those do not, so the former can take precedence in
```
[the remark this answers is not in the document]
pfmooney: Done

crates/cpuid-profile-config/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    /// Values (eax, ebx, ecx, edx) for `cpuid` leaf
+    pub values: [u32; 4],
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum CpuidParseError {
+    #[error("Unable to parse leaf {0}: {1}")]
+    Leaf(String, std::num::ParseIntError),
+    #[error("Unable to values: {0}")]
```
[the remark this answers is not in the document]
pfmooney: Done

lib/propolis/src/cpuid.rs
pfmooney: Those sort of policy decisions are left to the propolis-lib consumer

lib/propolis/src/cpuid.rs
pfmooney: Yeah

lib/propolis/src/cpuid.rs
pfmooney: Done

lib/propolis/src/cpuid.rs
pfmooney: I'm still curious to see how the interface (and usage) gets fleshed out, so I'm going to leave it for now.  We can circle back to making it more strict in the future.

lib/propolis/src/cpuid.rs
pfmooney: Until we have better core/thread-aware scheduling, I think going as far as CCX topology seems a bit... much?  I thought I had this close to matching what 314 spelled out.

lib/propolis/src/cpuid.rs
pfmooney: What's a bit unfortunate here is that with the builder pattern, we could get the `num_vcpu` after the topology fix-up as been requested.  I'll spell out that the user shouldn't omit num_vcpus in the comment at least.

lib/propolis/src/cpuid.rs
pfmooney: I've tweaked the leaf clearing behavior.  LMK if that's fine for now.

lib/propolis/src/vcpu.rs
```diff
@@ -139,6 +140,106 @@ impl Vcpu {
...
+        let intel_fallback =
+            config.vvcc_flags & bhyve_api::VCC_FLAG_INTEL_FALLBACK != 0;
+        let mut set = cpuid::Set::new(match intel_fallback {
+            true => cpuid::VendorKind::Intel,
+            false => cpuid::VendorKind::Amd,
+        });
+
+        for entry in entries {
+            let (ident, value) = cpuid::from_raw(entry);
+            let conflict = set.insert(ident, value);
```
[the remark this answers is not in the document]
gjcolombo: Makes sense, thanks!

lib/propolis/src/cpuid.rs
gjcolombo: I think that policy makes sense, but as things stand the call to `with_vcpu_count` to set `has_smt` is technically optional, correct? If it's 100% up to the consumer then I'd expect them to have to provide a value for `has_smt` in all cases (probably by passing it as a parameter to `new`).

lib/propolis/src/cpuid.rs
gjcolombo: Section 6.4.13 of the RFD gets into this a bit, though embarrassingly for me it has some unresolved comments that I really need to get back to... Basically my view is that we can either

1. reflect the CCX topology we'd like to have now, so that if/when we have scheduling enlightenments that actually schedule vCPUs according to their vCCX topology, the enlightenments "light up" underneath running VMs without requiring them to restart so we can change guest CPUID
2. put everyone in the same NUMA domain into a single virtual unified cache, in the hopes that this will save guest schedulers from doing useless work to account for L3 locality (but at the cost of requiring a VM to stop and start to pick up new CPUID settings later)

In the RFD I favored option 1, but I don't feel very strongly about it at the moment, since we have relatively few production VMs right now and aren't currently updating them without stopping them first. And we still have significant runway to make policy changes here. So I don't think we have to change anything in this PR; mostly I think I need to revisit 314 and make sure we get to ground on whether we're eventually going to do this.

(With all that said, I don't think we have to do a full-blown core scheduler to have some level of L3 locality--if we end up pinning vCPUs to host LPs then it would be at least somewhat straightforward to try to be CCX-aware when we assign those mappings.)

lib/propolis/src/cpuid.rs
gjcolombo: I'm probably missing something, but I'm not sure I follow. IIUC the builder only applies these options when someone calls `execute`, which consumes the whole builder. Before that point I don't think it matters whether you call `with_cpu_topo` or `with_vcpu_count` first, since those are just setting options in the builder, and after that point there's no way to call `with_vcpu_count` anymore because the builder was moved. If all that's right, when would the builder end up getting a CPU count after applying the topo fixups?

gjcolombo (commented):
Looking good, just a couple of things I want to make absolutely sure I'm grokking.

lib/propolis/src/cpuid.rs
pfmooney: I was thinking about tossing the error in the `with_cpu_topo` call, but I guess the `execute` call could have a `thiserror` enum covering that case.  I'll take a look at what we might emit there.

lib/propolis/src/cpuid.rs
pfmooney: > I don't think we have to do a full-blown core scheduler to have some level of L3 locality--if we end up pinning vCPUs to host LPs then it would be at least somewhat straightforward to try to be CCX-aware when we assign those mappings.

But which further complicates the bin-packing problem.  Since we're likely to revise the CPUID data when we add other HVM enlightenments, I'm inclined to go with a flat NUMA domain for now, knowing we'll exposing a more sophisticated model later.

lib/propolis/src/cpuid.rs
pfmooney: With the cache topology fixing as a small exception, `has_smt` is more or less worthless without `num_vcpu` too, so I think it's fine to group them that way for now?

Honestly I expect the Specializer stuff to evolve as we build more things with it, so I'm not to worried about getting it perfect out of the gate.

gjcolombo (approved):
Let's roll!

pfmooney:
Merged in c996261b36e208dac84c1023512ef93c9c0ece92

PR #527, opened 2023-09-21
Revert abort-on-panic in 'dev' cargo profile

pfmooney:
Since proper unwinding is required for handling `should_panic` tests cases, the existing default of abort-on-panic was rather inconvenient. This returns the 'dev' profile to using unwind-on-panic, and cleans up the 'phd' profile, which can use 'dev' as well for its unwinding.


gjcolombo (approved):

pfmooney:
Merged in 6d815bbd3ef1d2ff98a40770f9b673f611e1884e

Issue #528, opened 2023-09-23
PHD: want support for migration between distinct Propolis binaries


gjcolombo:
Today PHD requires its invoker to pass a path to a Propolis server binary on the command line. This binary is used to run all the Propolis servers under test. PHD should allow users to supply multiple Propolis server binaries, e.g. as artifacts, and then should allow individual tests to specify which binary to use (or binaries, if testing live migration).

hawkw:
I think #609 fixes this!

PR #529, opened 2023-09-23
PHD: improve artifact store

gjcolombo:
Rework the PHD artifact store to reduce complexity and add some features:

- Instead of specifying a single remote URI per downloadable artifact, have the
  artifact manifest specify a list of remote URI bases from which any
  downloadable artifact with the "remote server" source type can be obtained.
- Instead of having separate tables for guest OS images and bootroms, create one
  table of artifacts whose kinds are distinguished by an enum.
- Make the store remember whether it has previously verified the hash of an
  extant artifact. This substantially reduces runtime for runs that work with
  large disk images.
- Add support for a "Propolis server" artifact type. The runner doesn't use this
  yet but will begin making use of this in a future PR. See #528.
- Add support for Buildomat as an artifact source.

Update the PHD README and default artifact file to reflect these changes.

Also convert from PathBuf to camino::Utf8PathBuf in a few places to eliminate
some awkward type conversions.


phd-tests/framework/src/artifacts/store.rs
```diff
@@ -0,0 +1,288 @@
...
+use super::{manifest::Manifest, ArtifactSource};
+use anyhow::bail;
+use camino::{Utf8Path, Utf8PathBuf};
+use ring::digest::{Digest, SHA256};
+use std::collections::BTreeMap;
+use std::fs::File;
+use std::io::{BufReader, Read, Seek, SeekFrom, Write};
+use std::sync::Mutex;
+use std::time::Duration;
+use tracing::info;
```
pfmooney, line 16: small style nit: group the crate-local (`crate::` and `super::`), `std`, and cargo imports together in their own sections to stay consistent the rest of the existing style?

pfmooney (approved):

PR #534, opened 2023-09-27
Update tungstenite crates to 0.20

pfmooney:
This upgrades tungstenite and tokio-tungstenite to 0.20 in order to address CVE-2023-43669/GHSA-9mcr-873m-xcxp.


pfmooney:
I was a bit unsure of the updated buffer size tuning (resulting from the field deprecation).  The serial console appear to work fine on a test instance, so I'm assuming those numbers are acceptable for now.

PR #539, opened 2023-10-03
Update and lock GHA dependencies

pfmooney:
This locks the dependencies used by the GitHub Actions checks to specific git hashes (with the corresponding version in a comment) to bring it in line with the corresponding portion of RFD434.


pfmooney:
Merged in b916540036fabeec444e1e0addf5f3b831794ca0

PR #548, opened 2023-10-11
Begin making Accessor interface more robust

pfmooney:
This fixes leaky (and broken) behavior for when nodes are dropped an should be removed from the hierarchy.  A small base of tests have been added to cover a few of those cases.

The ability to disable access for swaths of the hierarchy is still missing, but this should lay the groundwork for more easily adding that in the future.

FYI: The coming overhaul of the block IO interfaces depends on this work.

### DIFF
--- a/lib/propolis/src/accessors.rs
+++ b/lib/propolis/src/accessors.rs
@@ -2,6 +2,15 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+//! Hierarchical access control for emulated resources
+//!
+//! Device emulation logic requires access to resources which may be
+//! subsequently moderated by intervening parts of the emulation.
+//!
+//! For example: A PCI device performs DMA to guest memory.  If bus-mastering is
+//! disabled on the device, or any parent bridge in its bus hierarchy, then its
+//! contained emulation should fail any DMA accesses.
+
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use std::marker::PhantomData;
 use std::sync::{Arc, Mutex, MutexGuard, Weak};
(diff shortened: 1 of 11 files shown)


lib/propolis/src/hw/pci/topology.rs
```diff
@@ -267,30 +267,29 @@ impl Builder {
 
         let pio_bus = &machine.bus_pio;
         let mmio_bus = &machine.bus_mmio;
-        let acc_mem = machine.acc_mem.child();
-        let acc_msi = machine.acc_msi.child();
```
pfmooney, line 271: Reviewer(s) please note:  Creating these children this way was a bug!  The bus(es) created in the topology spawned their own children from the local `acc_mem` and `acc_msi`.  Those locals are dropped when the function returns, effectively orphaning those buses.

gjcolombo (approved):
Code changes LGTM with a couple of small suggestions and for-my-understanding questions, and then one bigger question about the future semantics of obtaining a guard and revoking access to a device while guards are active.

lib/propolis/src/hw/pci/topology.rs
pfmooney: Yes, precisely.  (Ask me how I know)

lib/propolis/src/accessors.rs
```diff
@@ -271,3 +580,184 @@ impl std::fmt::Debug for MsiAccessor {
...
+            .unwrap()
+            .adopt(children[3].as_ref().unwrap(), None);
+
+        // The adopted nodes should have access (and see the updated val)
+        for child in children[3..].iter().map(|c| c.as_ref().unwrap()) {
+            let guard = child.access().expect("resource is accessible");
+            assert_eq!(guard.load(Ordering::Relaxed), tval);
+        }
+    }
+
```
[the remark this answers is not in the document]
gjcolombo: LGTM, thanks!

lib/propolis/src/accessors.rs
```diff
@@ -10,189 +19,458 @@ use crate::vmm::{MemCtx, VmmHdl};
...
 
+const ID_NULL: AccId = 0;
 const ID_ROOT: AccId = 1;
 
-type TreeBackref<T> = Arc<Mutex<Tree<T>>>;
-
+struct TreeNode<T> {
+    parent_id: AccId,
+    node_ref: Weak<Node<T>>,
+    children: Option<BTreeSet<AccId>>,
```
luqmana, line 28: Is the `Option` here just to be able to `take` the children below? `std::mem::take` would also work without the Option.

pfmooney: More to just avoid allocating the Set at all if there are no children.

pfmooney: I can change it if you think that's needless optimization.

luqmana: ah, `BTreeSet::new` doesn't allocate so should be fine without the Option.

pfmooney: Done

pfmooney:
Merged as 980867102a3aaf19c90f8e210bfd707aaf195400

PR #551, opened 2023-10-13
Update crucible to get fix 1000

leftwo:
Use regions_dataset path for apply_smf (#1000)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -690,7 +690,7 @@ dependencies = [
 [[package]]
 name = "crucible"
 version = "0.0.1"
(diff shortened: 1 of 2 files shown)


pfmooney:
I think we can skip this, since the crucible change does not impact the upstairs (or anything which propolis is consuming here).

pfmooney:
Alan's investigation found that omicron requires matching crucible versions in propolis and other components, despite the fact that it has no measurable impact on behavior for this change.  It's due to the fact that propolis-server is a dependency of sled-agent in omicron, via the simulated sled -agent component.

We should probably split out the mock server at some point so that we don't drag the entirety of propolis into omicron as a dependency, and can skip crucible updates which do not impact Upstairs (and thus propolis itself).

pfmooney (approved):

PR #552, opened 2023-10-13
Make qemu debug port output configurable

pfmooney:
The default of dumping `debug.out` files everywhere that propolis-standalone is run is a bit unnecessary if that output isn't desired.

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -267,6 +267,7 @@ impl<'a> MachineInitializer<'a> {
     }
 
     pub fn initialize_qemu_debug_port(&self) -> Result<(), Error> {
+        // TODO: Make the output file configurable
         let dbg = QemuDebugPort::create(&self.machine.bus_pio);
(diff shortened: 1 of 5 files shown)


gjcolombo:
I think we might've discussed this offline, but I forget where we landed on the idea: what about using the Propolis logger as a sink for these messages (possibly sticking an extra key/value pair on them for easy filtering)? That would not only avoid having `debug.out` files, it would mean we'd have just one set of logging options to rule them all (if you don't want any logging then you can just route the entire log to /dev/null, or logging will just happen on the console if that's the default, or what-have-you).

pfmooney:
Doing that would imply adding line-buffering to the debug port output, lest it turn every byte of log output into a line of slog output.  Even with that, I'm not sure how we would appreciate having to filter those out when looking at the propolis logs.  I've been working on various bits to cut down on the background noise which we emit in those logs (unhandled MSRs/registers/etc)

gjcolombo (approved):
I did line buffering for PHD long enough ago that I forgot it's a freestanding problem we'd have to solve :sweat_smile:

This LGTM, let's go ahead and merge.

PR #558, opened 2023-10-30
Configure exitcode for reboot/halt on standalone

### LINKED ISSUE #557: standalone: option for process exit on guest reboot with stable exit status
In order to assist with the use of **propolis-standalone** for certain automated tests and jobs, it would help to have a new command line flag or configuration file property that would make the hypervisor process exit if the guest tries to reboot, like we do today when the guest requests power off, rather than allowing the reboot to proceed.  It would be good also to be able to have a particular exit status for a guest reboot (e.g., exit 10) to make it easy to determine when one has occurred, versus the guest powering itself off (presumably exit 0) or the hypervisor crashing (presumably exit 1 or abort or whatever today).

pfmooney:
Fixes #557

### DIFF
--- a/bin/propolis-standalone/src/main.rs
+++ b/bin/propolis-standalone/src/main.rs
@@ -6,6 +6,7 @@ use std::collections::VecDeque;
 use std::fs::File;
 use std::io::{Error, ErrorKind, Result};
 use std::path::Path;
+use std::process::ExitCode;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Condvar, Mutex, MutexGuard};
 use std::time::{SystemTime, UNIX_EPOCH};
@@ -193,6 +194,7 @@ struct InstState {
     instance: Option<propolis::Instance>,
     state: State,
     vcpu_tasks: Vec<propolis::tasks::TaskCtrl>,
+    exit_code: Option<u8>,
 }
 
 struct InstInner {
@@ -216,6 +218,7 @@ impl Instance {
                 instance: Some(pinst),
                 state: State::Initialize,
                 vcpu_tasks: Vec::new(),
+                exit_code: None,
             }),
             boot_gen: AtomicUsize::new(0),
             eq: EventQueue::new(),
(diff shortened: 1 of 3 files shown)


bin/propolis-standalone/src/main.rs
```diff
@@ -1058,6 +1070,6 @@ fn main() -> anyhow::Result<()> {
     );
 
     // wait for instance to be destroyed
-    inst.wait_destroyed();
-    Ok(())
+    let code = inst.wait_destroyed();
+    std::process::exit(code);
```
jclulow, line 1074: Does the move to an explicit exit here change the behaviour, because we are now not actually returning from **main()** any more?  In particular, I wonder if we previously would have called a bunch of **Drop** routines and now we're not doing that...

bin/propolis-standalone/src/main.rs
```diff
@@ -404,8 +406,17 @@ impl Instance {
                     for mut vcpu_ctrl in guard.vcpu_tasks.drain(..) {
                         vcpu_ctrl.exit();
                     }
+                    if guard.exit_code.is_none() {
+                        guard.exit_code = Some(inner.config.main.exit_on_halt);
+                    }
                 }
                 State::Reset => {
+                    if let Some(code) = inner.config.main.exit_on_reboot {
+                        guard.exit_code = Some(code);
```
jclulow, line 415: Should we do the same **is_none()** check you're doing on 409 for the **Halt** version?

jclulow (commented):
Thanks for doing this so quickly!  I have a couple of small questions.

bin/propolis-standalone/src/main.rs
pfmooney: Unless propolis components are spawning tasks which aren't themselves collected on instance destroy, we should be pretty well cleaned up by now.

That said, I've refactored and cleaned up some other bits so we can return an `ExitCode`.

bin/propolis-standalone/src/main.rs
pfmooney: I don't believe it's strictly necessary at this time, but for defense-in-depth purposes, I've added such a check. (In case we add something like exit-on-fault to bail out on a triple-fault or something)

jclulow (approved):
Huh!  TIL **ExitCode**.  Neat!

PR #560, opened 2023-10-31
Softnpu jumbo frames

rcgoodfellow:
- Add enough headroom in softnpu frame buffers to support jumbo frames.
- Remove a very noisy p9fs print statement.
- Appease clippy.


pfmooney (approved):
Please squash (at least the clippy commit) before pushing.

PR #563, opened 2023-11-08
propolis-client should go all-in on OpenAPI

pfmooney:
With Progenitor working well, we should stop depending on the handmade API in propolis-client, and instead rely solely on the generated client. This includes updating consumers to solely depend upon the generated types, rather than pulling them from an assortment of other propolis-internal (and Crucible) crates.


crates/propolis-api-types/src/instance_spec/openapi_impls.rs
gjcolombo, line 1: I think and hope this file can just go away completely. Its only purpose is (was) to convert from the handmade instance spec types to their Progenitorized equivalents so that a caller could use the spec builder and then pass the result to a Progenitor client. Assuming the builder is now building Progenitor-flavored specs to start with, none of the code in here should be needed anymore.

lib/propolis-client/src/instance_spec.rs
gjcolombo, line 1: Can/should propolis-client just re-export the identical builder from the api-types crate?

gjcolombo (approved):
Great stuff! There are a couple of files that I was hoping would be obsoleted by this move that I've left comments on.

lib/propolis-client/src/instance_spec.rs
gjcolombo: Per offline discussion, doing this makes propolis-client depend on crucible-client-types (because api-types of necessity includes the type definitions used by propolis-server's APIs, which still refer directly to the `VolumeConstructionRequest` type).

Ideally (IMO) we'd get out of this by teaching Omicron to use the instance spec-based ensure endpoint and deprecating the old ensure request, but that's way out of scope for this change.

crates/propolis-api-types/src/instance_spec/openapi_impls.rs
pfmooney: Considering it's not even included as a module, you couldn't be more right.

bin/mock-server/src/lib/api_types.rs
```diff
@@ -0,0 +1,62 @@
...
+
+progenitor::generate_api!(
+    spec = "../../openapi/propolis-server.json",
+    derives = [schemars::JsonSchema],
+    patch = {
+        InstanceProperties = {
+            derives = [
+                Clone,
+                schemars::JsonSchema,
+                Serialize,
```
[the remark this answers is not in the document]
pfmooney: I forgot to clean these up after our discussion.  I'll circle back to fix this in a subsequent change so I don't need to redo the testing I've done for landing this and its associated change in Omicron.

PR #570, opened 2023-11-20
better softnpu management command reliability

rcgoodfellow:
This PR does two things

1. Makes UART FIFO length configurable and sets softnpu's management UART FIFO length to 10 instead of 1.
2. Updates the softnpu management message reader to not attempt to deserialize messages where one of the following is true
  a. The softnpu management preamble has not been recognized.
  b. The message is not long enough to contain a preamble plus actual data.

These changes were spurred by transient issues adding routes in Falcon topologies. I traced this down to corrupted softnpu management command payloads coming through the UART. The corrupted payloads were easy to reproduce. With the changes in this PR, I'm no longer able to get corrupted payloads through the UART.


rcgoodfellow:
Something peculiar about the missing bytes. Missing bytes would happen intermittently on a message-by-message basis e.g. I could not detect a pattern there. However, when there was a missing byte in a message, it was often the 2nd byte, which felt odd.

pfmooney:
The illumos issue regarding the lost byte(s) on the UART has been written up in [16088](https://www.illumos.org/issues/16088).  The changes to the propolis UART to expand the FIFO sizing should not go in as-is.  (We may someday improve the emulation to support 16550-compatible FIFO behavior)

Do you want to keep the other changes to the softNPU logic contained in this PR?

rcgoodfellow:
> Do you want to keep the other changes to the softNPU logic contained in this PR?

Yeah, I do.

Once the illumos patch goes through and makes its way to stlouis, I'll cut a new Helios base image for Falcon. Then I'll revert the FIFO changes on this branch and we can get the remaining updates merged.

pfmooney:
With the UART fixes merged into both illumos-gate and stlouis, can we pare this down to only the softNPU fixes so it can be merged?

pfmooney (approved):

PR #572, opened 2023-11-27
implement p9fs read as direct host file to guest mem write

rcgoodfellow:
Comes with ~5x performance improvement. The following is on a 2.17G transfer.

* Before

    real     2:14.557405302
    user        5.665261225
    sys        15.781882665

* After

    real       23.913634189
    user        6.123950329
    sys        15.342021778

### DIFF
--- a/lib/propolis/src/hw/virtio/p9fs.rs
+++ b/lib/propolis/src/hw/virtio/p9fs.rs
@@ -21,7 +21,7 @@ use crate::vmm::MemCtx;
 
 use super::bits::*;
 use super::pci::{PciVirtio, PciVirtioState};
-use super::queue::{p9_write_file, write_buf, Chain, VirtQueue, VirtQueues};
+use super::queue::{write_buf, Chain, VirtQueue, VirtQueues};
 use super::VirtioDevice;
 
 use ispf::WireSize;
@@ -949,4 +949,55 @@ pub(crate) fn write_error(ecode: u32, chain: &mut Chain, mem: &MemCtx) {
     let mut out = ispf::to_bytes_le(&msg).unwrap();
     let buf = out.as_mut_slice();
     write_buf(buf, chain, mem);
+}
+
+pub(crate) fn p9_write_file(
+    file: &impl std::os::fd::AsRawFd,
+    chain: &mut Chain,
+    mem: &MemCtx,
+    count: usize,
+    offset: i64,
+) {
+    // Form the rread header. Unfortunately we can't do this with the Rread
+    // structure because the count is baked into the data field which is tied
+    // to the length of the vector and filling that vector is what we're
+    // explicitly trying to avoid here.
+    let sz = size_of::<u32>() + // size
+            size_of::<u8>()  + // typ
+            size_of::<u16>() + // tag
+            size_of::<u32>() +  // data.count
+            count; // data
+    let mut header = Vec::with_capacity(11);
+    header.extend_from_slice(&(sz as u32).to_le_bytes());
+    header.push(p9ds::proto::MessageType::Rread as u8);
+    header.extend_from_slice(&0u16.to_le_bytes());
+    header.extend_from_slice(&(count as u32).to_le_bytes());
+
+    // Send the header to the guest from the buffer constructed above. Then
+    // send the actual file data
+    let mut header_done = false;
+    let mut done = 0;
+    let _total = chain.for_remaining_type(false, |addr, len| {
+        let mut remain = len;
+        let mut copied = 0;
+        if !header_done {
+            mem.write_from(addr, &header, header.len()).unwrap();
+            header_done = true;
+            remain -= header.len();
+            copied += header.len();
+        }
+        let addr = GuestAddr(addr.0 + copied as u64);
+        let sub_mapping =
+            mem.direct_writable_region(&GuestRegion(addr, remain)).unwrap();
+
+        let len = usize::min(remain, count);
+        let off = offset + done as i64;
+        let mapped = sub_mapping.pread(file, len, off).unwrap();
+        copied += mapped;
+        done += mapped;
+
+        let need_more = done < count;
+
+        (copied, need_more)
+    });
 }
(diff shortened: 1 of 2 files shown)


lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -770,3 +770,55 @@ pub(crate) fn write_buf(buf: &[u8], chain: &mut Chain, mem: &MemCtx) {
...
 }
+
+#[cfg(feature = "falcon")]
+pub(crate) fn p9_write_file(
+    file: &impl std::os::fd::AsRawFd,
+    chain: &mut Chain,
+    mem: &MemCtx,
+    count: usize,
+    offset: i64,
+) {
```
pfmooney, line 781: Is there a reason why this can't go in `p9fs.rs`?

rcgoodfellow: Nope, I put it here as the `write_buf` function is here which is very similar. I can move it to `p9fs.rs`

pfmooney: Since it's p9-specific, I'd keep it with the rest of the p9-related things

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -974,3 +950,54 @@ pub(crate) fn write_error(ecode: u32, chain: &mut Chain, mem: &MemCtx) {
...
+    let _total = chain.for_remaining_type(false, |addr, len| {
+        let mut remain = len;
+        let mut copied = 0;
+        if !header_done {
+            mem.write_from(addr, &header, header.len()).unwrap();
+            header_done = true;
+            remain -= header.len();
+            copied += header.len();
+        }
+        let addr = GuestAddr(addr.0 + copied as u64);
```
pfmooney, line 989: Could use `GuestAddr::offset()` for this

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -974,3 +950,54 @@ pub(crate) fn write_error(ecode: u32, chain: &mut Chain, mem: &MemCtx) {
...
+    let mut done = 0;
+    let _total = chain.for_remaining_type(false, |addr, len| {
+        let mut remain = len;
+        let mut copied = 0;
+        if !header_done {
+            mem.write_from(addr, &header, header.len()).unwrap();
+            header_done = true;
+            remain -= header.len();
+            copied += header.len();
+        }
```
pfmooney, line 988: I think you could lift this into a `Chain::write` call, and the extra `header_done` logic could be skipped?

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -974,3 +950,54 @@ pub(crate) fn write_error(ecode: u32, chain: &mut Chain, mem: &MemCtx) {
...
+        let addr = GuestAddr(addr.0 + copied as u64);
+        let sub_mapping =
+            mem.direct_writable_region(&GuestRegion(addr, remain)).unwrap();
```
pfmooney, line 991: The documentation could make it more clear, but `direct_writable_region` should not be used by normal device emulation.  The `writable_region`/`readable_region` functions are what you want.

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -752,7 +752,6 @@ pub mod migrate {
     }
 }
 
-#[cfg(feature = "falcon")]
```
pfmooney, line 755: I suspect this will grumble about unused code without the guard?

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -974,3 +950,54 @@ pub(crate) fn write_error(ecode: u32, chain: &mut Chain, mem: &MemCtx) {
     let buf = out.as_mut_slice();
     write_buf(buf, chain, mem);
 }
+
+pub(crate) fn p9_write_file(
```
pfmooney, line 954: This doesn't need to be `pub`-anything

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -974,3 +950,54 @@ pub(crate) fn write_error(ecode: u32, chain: &mut Chain, mem: &MemCtx) {
...
+    file: &impl std::os::fd::AsRawFd,
```
pfmooney, line 955: Why not pass `&File` until some need compels it to be generic?

rcgoodfellow: Done

lib/propolis/src/hw/virtio/p9fs.rs
rcgoodfellow: Done

lib/propolis/src/hw/virtio/queue.rs
rcgoodfellow: Yep, fixed.

lib/propolis/src/hw/virtio/p9fs.rs
rcgoodfellow: Done

lib/propolis/src/hw/virtio/p9fs.rs
rcgoodfellow: Yeah, the code is much simpler that way. Thanks

lib/propolis/src/hw/virtio/p9fs.rs
rcgoodfellow: Done

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -974,3 +949,54 @@ pub(crate) fn write_error(ecode: u32, chain: &mut Chain, mem: &MemCtx) {
...
+        count: count as u32,
+    };
+
+    chain.write(&h, mem);
+
+    // Send the header to the guest from the buffer constructed above. Then
+    // send the actual file data
+    let mut done = 0;
+    let _total = chain.for_remaining_type(false, |addr, len| {
+        let mut copied = 0;
```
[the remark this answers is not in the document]
rcgoodfellow: Yup. That was a remnant of the original implementation that I missed in the reorg. Fixed.

PR #574, opened 2023-11-30
Modernize 16550 UART

pfmooney:
The 16550 emulation was one of the first bits of emulation sketched out in propolis, and it shows.  While there is more work to be done in improving the fidelity of emulation, the use of bitflags and the like should make this more readable.


jordanhendricks (approved):
Thanks! This looks a lot cleaner.

PR #575, opened 2023-11-30
Log when pause futures complete

### LINKED ISSUE #537: want logging of completion of pause futures
When [pausing a VM's entities](https://github.com/oxidecomputer/propolis/blob/master/bin/propolis-server/src/lib/vm/mod.rs#L953-L984), the VMM controller logs pause requests on an entity-by-entity basis, but does not log the completions of individual pause futures. This means that if pause gets stuck waiting for such a future to complete, logs won't reveal which future(s) were outstanding.

jmpesp:
Fixes #537


gjcolombo (approved):

PR #577, opened 2023-12-04
bump softnpu

rcgoodfellow:
holding in draft until softnpu changes land

### DIFF
--- a/.github/buildomat/jobs/falcon-build.sh
+++ b/.github/buildomat/jobs/falcon-build.sh
@@ -18,7 +18,7 @@ rustc --version
 
 banner check
 ptime -m cargo check --features falcon
-ptime -m clippy --features falcon --all-targets
+ptime -m cargo clippy --features falcon --all-targets
 
 banner build
 ptime -m cargo build --features falcon --release \
--- a/.github/buildomat/jobs/falcon-build.sh
+++ b/.github/buildomat/jobs/falcon-build.sh
@@ -18,6 +18,7 @@ rustc --version
 
 banner check
 ptime -m cargo check --features falcon
+ptime -m cargo xtask clippy
 
 banner build
 ptime -m cargo build --features falcon --release \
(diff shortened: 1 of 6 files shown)


pfmooney:
It looks like I should update the `clippy` xtask to specify `--no-deps`?

rcgoodfellow:
> It looks like I should update the Clippy task to specify --no-deps?

Perhaps. This is pulling in `libnet`, which will go badly on any platform which is not illumos. Looks like some of the clippy complaints would also apply to illumos, so I'll go check that out.

.github/buildomat/jobs/falcon-build.sh
```diff
@@ -18,6 +18,7 @@ rustc --version
 
 banner check
 ptime -m cargo check --features falcon
+ptime -m cargo xtask clippy
```
pfmooney, line 21: Probably just scope this to a `cargo clippy ---features falcon`?  We don't need to run the non-falcon checks.

xtask/src/task_clippy.rs
```diff
@@ -39,14 +39,17 @@ pub(crate) fn cmd_clippy(strict: bool, quiet: bool) -> Result<()> {
...
+    {
+        failed |= run_clippy(&[
+            "--features",
+            "falcon",
+            "-p",
+            "propolis-server",
+            "-p",
+            "propolis-client",
+        ])?;
+    }
```
pfmooney, line 52: I believe this part of the change can be elided now that oxidecomputer/netadm-sys@d44d9e084f39e844f8083d4d9b39a331061ebbcc has landed?  (Provided a `cargo update -p libnet` is done)

rcgoodfellow: Removed

.github/buildomat/jobs/falcon-build.sh
rcgoodfellow: done, also with `--all-targets` flag

pfmooney (approved):

PR #586, opened 2023-12-15
Make Viona more robust in the face of errors

### LINKED ISSUE #583: WS2016 guest panics Propolis after disabling/re-enabling network device from the guest
Repro steps:

- Obtain a Windows Server 2016 image with netkvm.sys from build 217 of the [Fedora virtio drivers](https://github.com/virtio-win/virtio-win-pkg-scripts/blob/master/README.md). (Note that the latest driver ISO is now `virtio-win-0.1.240.iso`, not `0.1.217`; I will retest with the newer guest driver shortly.)
- Boot the guest in a Propolis server.
- Observe that there's no network connectivity. According to Device Manager (devmgmt.msc), the network driver is loaded correctly, but DHCP doesn't work properly. Running `snoop` on the relevant VNIC shows some DHCP activity but no IP address is ever assigned.
- Run the network troubleshooter for this connection via Control Panel.

Expected: The guest network interface acquires an IP address and has connectivity with no additional work required.

Observed: Propolis panics with the following backtrace upon trying to run networking diagnostics:

```
thread 'vcpu-1' panicked at lib/propolis/src/hw/virtio/viona.rs:271:45:
not yet implemented: viona error handling
stack backtrace:
   0: rust_begin_unwind
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/panicking.rs:72:14
   2: <propolis::hw::virtio::viona::PciVirtioViona as propolis::hw::virtio::VirtioDevice>::queue_change
   3: propolis::hw::virtio::pci::PciVirtioState::legacy_write
   4: propolis::hw::virtio::pci::<impl propolis::hw::pci::device::Device for D>::bar_rw::{{closure}}
   5: propolis::util::regmap::RegMap<ID>::process
   6: propolis::hw::pci::device::<impl propolis::hw::pci::Endpoint for D>::bar_rw
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

This scenario appears to work correctly with Windows Server 2022 using the appropriate driver from the same driver ISO (i.e. the 217 build, not the new 240 build).

pfmooney:
When the Viona driver was initially drafted, it was made to work well enough on the "happy path", but lacked any sort of resilience in the face of uncommon or unexpected conditions.

Fixes #583

### DIFF
--- a/lib/propolis/src/hw/virtio/pci.rs
+++ b/lib/propolis/src/hw/virtio/pci.rs
@@ -160,7 +160,7 @@ impl<D: PciVirtio + Send + Sync + 'static> pci::Device for D {
             if result.is_err() {
                 // An error updating the VQ interrupt config should set the
                 // device in a failed state.
-                vs.set_failed(self, &mut state);
+                vs.needs_reset_locked(self, &mut state);
             }
         }
         state.intr_mode_updating = false;
@@ -320,7 +320,7 @@ impl PciVirtioState {
                         state.nego_feat = nego;
                     }
                     Err(_) => {
-                        self.set_failed(dev, &mut state);
+                        self.needs_reset_locked(dev, &mut state);
                     }
                 }
             }
@@ -334,7 +334,7 @@ impl PciVirtioState {
 
                     if qs_old.mapping.desc_addr != new_addr {
                         if dev.queue_change(queue, VqChange::Address).is_err() {
-                            self.set_failed(dev, &mut state);
+                            self.needs_reset_locked(dev, &mut state);
                         }
                     }
                 }
@@ -380,7 +380,7 @@ impl PciVirtioState {
                         // With the MSI configuration updated for the virtqueue,
                         // notify the device of the change
                         if dev.queue_change(queue, VqChange::IntrCfg).is_err() {
-                            self.set_failed(dev, &mut state);
+                            self.needs_reset_locked(dev, &mut state);
                         }
 
                         state.intr_mode_updating = false;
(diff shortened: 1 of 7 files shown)


lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -395,6 +410,22 @@ impl PciVirtioState {
...
+    /// This will require the guest OS to reset the device before it can use it
+    /// for anything further.
+    fn set_failed(
+        &self,
+        _dev: &dyn VirtioDevice,
+        state: &mut MutexGuard<VirtioState>,
+    ) {
+        if !state.status.contains(Status::FAILED) {
+            state.status.insert(Status::FAILED);
+            // XXX: interrupt needed?
```
gjcolombo, line 425: I read [section 2.1 of the Virtio 1.2 spec](http://docs.oasis-open.org/virtio/virtio/v1.0/cs04/virtio-v1.0-cs04.html#x1-100001) (versions 1.0 and 1.1 have the same language in this section) and came away thinking that

- This should set `DEVICE_NEEDS_RESET` and not `FAILED` (IIUC the latter is for guest drivers to set and the former is for devices to set, even if device failed while responding to a driver write), and
- If that's correct and the `DRIVER_OK` bit was previously set, then a configuration interrupt is required after doing this.

lib/propolis/src/hw/virtio/viona.rs
gjcolombo, line 1: Optional: As a general matter, do we want to think about setting the "needs reset" flag in cases where we transition to VRingState::Fatal? (The spec doesn't require it, so this would purely be for the benefit of guest drivers that understand what to do with the NEEDS_RESET bit.)

lib/propolis/src/hw/virtio/pci.rs
pfmooney: I'll fix this to use `NEEDS_RESET` rather than `FAILED`.  As for the configuration interrupt work, I'll open an issue to deal with that separately.

lib/propolis/src/hw/virtio/viona.rs
pfmooney: I believe (?) I've covered those bases for the `queue_update` case.  The only other applicable case (I believe) would be `queues_restart`.  I'll update that to set such a state if it happens to set a ring as failed.

pfmooney:
Merged in 5d99d9718ad01183f5115e79a254610f119bd281

PR #589, opened 2023-12-18
nvme: Wire up GetFeatures command

### LINKED ISSUE #567: NVMe controller should probably implement get features
Currently the NVMe controller does not implement the Get Features command. Most folks assume that if they can set a feature that they can get it (and it's a required command even in NVMe 1.0). Given that currently the volatile write cache and the number of queues features are implemented, being able to get those at least would probably be beneficial.

pfmooney:
Fixes #567

### DIFF
--- a/lib/propolis/src/hw/nvme/admin.rs
+++ b/lib/propolis/src/hw/nvme/admin.rs
@@ -372,8 +372,11 @@ impl NvmeCtrl {
             // Optional features
             cmds::FeatureIdent::VolatileWriteCache => {
                 // TODO: wire into actual write cache state
+                //
+                // Until that is done, indicate an enabled write cache to ensure
+                // IO flushes for the backends which require it for consistency.
                 cmds::Completion::success_val(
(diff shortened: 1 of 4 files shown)


lib/propolis/src/hw/nvme/admin.rs
```diff
@@ -278,6 +277,115 @@ impl NvmeCtrl {
...
+            cmds::FeatureIdent::AsynchronousEventConfiguration => {
+                // None of the defined events result in AEN transmission
+                cmds::Completion::success_val(0)
+            }
+
+            // Optional features
+            cmds::FeatureIdent::VolatileWriteCache => {
+                // TODO: wire into actual write cache state
+                cmds::Completion::success_val(
+                    cmds::FeatVolatileWriteCache { wce: false }.into(),
```
luqmana, line 376: Leaving the TODO is fine but just return `true` to match the current state?

luqmana (approved):
Thanks!

pfmooney:
Merged in 451f3c7b5b32d6c0a7e80a76e40171b0435c722a

PR #596, opened 2024-01-09
add QEMU pvpanic ISA device

### LINKED ISSUE #592: want pvpanic device
It would be nice to emulate the [QEMU pvpanic device](https://www.qemu.org/docs/master/specs/pvpanic.html) so that if a VM has failed or is otherwise acting wonky, we can learn whether or not the guest has reported its own ill health.

I think the work here looks something like this (N.B. this is somewhat disorganized rambling, and some of it may not hold up as we begin to code):

- Decide what device interface to present to guests: this device can function as either an ISA bus device (with the guest indicating panic via port I/O) or a PCI device (in the PCI case we need to make sure to choose a device number that doesn't conflict with the space the `instance_ensure` endpoint implicitly reserves)
- Add an [instance spec](https://github.com/oxidecomputer/propolis/blob/85dc6db36709981c1d35b0651a18e20caee0b73f/crates/propolis-api-types/src/instance_spec/mod.rs) device type that Propolis servers can use to decide whether to add one of these devices to a new VM
  - Since we're still not live-migrating VMs yet, let alone between versions, it *should* be possible to add this directly to [`DeviceSpecV0`](https://github.com/oxidecomputer/propolis/blob/master/crates/propolis-api-types/src/instance_spec/v0/mod.rs#L110) without worrying about breaking anything. However, this is a good chance to reason about what would need to happen to safely make a change like this: we probably need the device to be optional and need to make sure serde doesn't serialize this field if the device doesn't exist.
- Decide how to get panic counts out of Propolis
  - This should probably be an Oximeter metric. Today, Propolis itself only produces [one metric](https://github.com/oxidecomputer/propolis/blob/85dc6db36709981c1d35b0651a18e20caee0b73f/bin/propolis-server/src/lib/stats.rs#L36-L43), and it's from the server itself and not from any of its entities. We'll probably want to do for other entities something similar to what we do for Crucible: pass a copy of the machine initializer's [producer registry](https://github.com/oxidecomputer/propolis/blob/85dc6db36709981c1d35b0651a18e20caee0b73f/bin/propolis-server/src/lib/initializer.rs#L121) to components that accept it and allow them to define their own producer and call `register_producer` on it.
  - For test purposes it might also be useful to be able to walk up to a Propolis server and ask for its guest panic count by defining a new, possibly test-only, endpoint that finds the pvpanic entity and figures out how many panics it has seen
- Write tests
  - If we do a PCI device, we can write a PHD test that verifies the device is listed by `lspci`
  - If we have a "how many panics have you seen?" endpoint, PHD can query it

hawkw:
This branch adds support for the [pvpanic] virtual device implemented by
QEMU. This device allows guests to report kernel panics to the
hypervisor. In `propolis-server`, guest-reported kernel panics are
handled by incrementing an Oximeter metric.

The pvpanic device can be exposed to the guest either as an ISA bus I/O
port device or as a PCI bus device. This branch implements the ISA bus
device. I'd like to also add a PCI pvpanic device, but will implement
that in a subsequent pull request.

In order for the guest to detect the ISA bus pvpanic device, it's
necessary to add an entry for the panic device to the ACPI DSDT table.
This is the AML that QEMU adds to its DSDT when the ISA bus pvpanic
device is enabled:

```
//
// QEMU panic device
//
Device (PEVT)
{
    Name (_HID, "QEMU0001")  // _HID: Hardware ID
    Name (_CRS, ResourceTemplate ()  // _CRS: Current Resource Settings
    {
        IO (Decode16,
            0x0505,             // Range Minimum
            0x0505,             // Range Maximum
            0x01,               // Alignment
            0x01,               // Length
            )
    })
    OperationRegion (PEOR, SystemIO, 0x0505, One)
    Field (PEOR, ByteAcc, NoLock, Preserve)
    {
        PEPT,   8
    }

    Name (_STA, 0x0F)  // _STA: Status
    Method (RDPT, 0, NotSerialized)
    {
        Local0 = PEPT /* \_SB_.PCI0.S08_.PEVT.PEPT */
        Return (Local0)
    }

    Method (WRPT, 1, NotSerialized)
    {
        PEPT = Arg0
    }
}
```

This means that in order for guests to use this device, we need to boot
with an ACPI table that contains this entry. For testing purposes, I
modified EDK2 OVMF to add this entry to the DSDT. In the future, though,
we'll likely want Propolis to generate ACPI tables dynamically on boot
based on the instance spec.

The EDK2 changes I used for testing this are available [here][edk2].

To test this change, I ran `propolis-standalone` with an Alpine Linux
3.19 guest,, and the following device added to the VM config file:

```toml
[dev.pvpanic]
driver = "qemu-pvpanic"
enable_mmio = true
```

The guest correctly detects the panic device and loads the appropriate
kernel module. If I then trigger a panic in the guest using SysRq, like
this:

```console
$ echo 1 > /proc/sys/kernel/sysrq
$ echo c > /proc/sysrq-trigger
```

The guest crashes, and `propolis-standalone` logs:

```
dev: pvpanic
 Jan 11 18:14:13.494 DEBG guest kernel panic, guest_handled: false, host_handled: true
```

Closes #592 

[pvpanic]: https://www.qemu.org/docs/master/specs/pvpanic.html
[edk2]:
    https://github.com/oxidecomputer/edk2/commit/6ca196ffca7a91884b4bd693df7b58d833d1ad42

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -282,7 +282,7 @@ impl<'a> MachineInitializer<'a> {
         uuid: uuid::Uuid,
     ) -> Result<(), anyhow::Error> {
         if let Some(ref spec) = self.spec.devices.qemu_pvpanic {
-            if spec.enable_isa {
+            if spec.enable_mmio {
                 let pvpanic = QemuPvpanic::create(
                     self.log.new(slog::o!("dev" => "qemu-pvpanic")),
                 );
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -35,7 +35,7 @@ use crate::serial::Serial;
 use crate::server::CrucibleBackendMap;
 pub use nexus_client::Client as NexusClient;
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 
 // Arbitrary ROM limit for now
 const MAX_ROM_SIZE: usize = 0x20_0000;
@@ -292,11 +292,9 @@ impl<'a> MachineInitializer<'a> {
                 if let Some(ref registry) = self.producer_registry {
                     let producer =
                         crate::stats::PvpanicProducer::new(uuid, pvpanic);
-                    registry.register_producer(producer).map_err(|error| {
-                        anyhow::anyhow!(
-                            "failed to register PVPANIC Oximeter producer: {error}"
-                        )
-                    })?;
+                    registry.register_producer(producer).context(
+                        "failed to register PVPANIC Oximeter producer",
+                    )?;
                 }
             }
         }
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -21,6 +21,7 @@ use propolis::hw::chipset::Chipset;
 use propolis::hw::ibmpc;
 use propolis::hw::pci;
 use propolis::hw::ps2::ctrl::PS2Ctrl;
+use propolis::hw::qemu::pvpanic::{self, QemuPioPvpanic};
 use propolis::hw::qemu::{debug::QemuDebugPort, fwcfg, ramfb};
 use propolis::hw::uart::LpcUart;
 use propolis::hw::{nvme, virtio};
@@ -273,6 +274,29 @@ impl<'a> MachineInitializer<'a> {
         let poller = chardev::BlockingFileOutput::new(debug_file);
         poller.attach(Arc::clone(&dbg) as Arc<dyn BlockingSource>);
         self.inv.register(&dbg)?;
+        Ok(())
+    }
+
+    pub fn initialize_qemu_pvpanic(
+        &self,
+        uuid: uuid::Uuid,
+    ) -> Result<(), anyhow::Error> {
+        if self.spec.devices.qemu_pvpanic.enable_isa {
+            let counts = Arc::new(pvpanic::PanicCounts::default());
+            let piodev =
+                QemuPioPvpanic::create(&self.machine.bus_pio, counts.clone());
+            self.inv.register(&piodev)?;
+
+            if let Some(ref registry) = self.producer_registry {
+                let producer = crate::stats::PvpanicProducer::new(uuid, counts);
+                registry.register_producer(producer).map_err(|error| {
+                    anyhow::anyhow!(
+                        "failed to register PVPANIC Oximeter producer: {error}"
+                    )
+                })?;
+            }
+        }
+
         Ok(())
     }
 
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -282,7 +282,7 @@ impl<'a> MachineInitializer<'a> {
         uuid: uuid::Uuid,
     ) -> Result<(), anyhow::Error> {
         if let Some(ref spec) = self.spec.devices.qemu_pvpanic {
-            if spec.enable_mmio {
+            if spec.enable_isa {
                 let pvpanic = QemuPvpanic::create(
                     self.log.new(slog::o!("dev" => "qemu-pvpanic")),
                 );
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -21,7 +21,7 @@ use propolis::hw::chipset::Chipset;
 use propolis::hw::ibmpc;
 use propolis::hw::pci;
 use propolis::hw::ps2::ctrl::PS2Ctrl;
-use propolis::hw::qemu::pvpanic::{self, QemuPioPvpanic};
+use propolis::hw::qemu::pvpanic::QemuPvpanic;
 use propolis::hw::qemu::{debug::QemuDebugPort, fwcfg, ramfb};
 use propolis::hw::uart::LpcUart;
 use propolis::hw::{nvme, virtio};
@@ -281,19 +281,21 @@ impl<'a> MachineInitializer<'a> {
         &self,
         uuid: uuid::Uuid,
     ) -> Result<(), anyhow::Error> {
-        if self.spec.devices.qemu_pvpanic.enable_isa {
-            let counts = Arc::new(pvpanic::PanicCounts::default());
-            let piodev =
-                QemuPioPvpanic::create(&self.machine.bus_pio, counts.clone());
-            self.inv.register(&piodev)?;
+        if let Some(ref spec) = self.spec.devices.qemu_pvpanic {
+            if spec.enable_isa {
+                let pvpanic = QemuPvpanic::create();
+                pvpanic.attach_pio(&self.machine.bus_pio);
+                self.inv.register(&pvpanic)?;
 
-            if let Some(ref registry) = self.producer_registry {
-                let producer = crate::stats::PvpanicProducer::new(uuid, counts);
-                registry.register_producer(producer).map_err(|error| {
-                    anyhow::anyhow!(
-                        "failed to register PVPANIC Oximeter producer: {error}"
-                    )
-                })?;
+                if let Some(ref registry) = self.producer_registry {
+                    let producer =
+                        crate::stats::PvpanicProducer::new(uuid, pvpanic);
+                    registry.register_producer(producer).map_err(|error| {
+                        anyhow::anyhow!(
+                            "failed to register PVPANIC Oximeter producer: {error}"
+                        )
+                    })?;
+                }
             }
         }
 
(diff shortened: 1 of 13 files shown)


lib/propolis/src/hw/qemu/pvpanic.rs
```diff
@@ -0,0 +1,80 @@
...
+    const IOPORT: u16 = 0x505;
+
+    #[must_use]
+    pub fn create(pio: &PioBus, counts: Arc<PanicCounts>) -> Arc<Self> {
+        let this = Arc::new(Self { counts });
+
+        let piodev = this.clone();
+        let piofn = Arc::new(move |_port: u16, rwo: RWOp| piodev.pio_rw(rwo))
+            as Arc<PioFn>;
+        pio.register(Self::IOPORT, 1, piofn).unwrap();
```
pfmooney, line 46: The other qemu devices set a bad example, and this is not (yet) well documented, but we should try to split up the device creation from its attachment to any resources.  See something like `LpcUart` or the ps/2 controller for an example.

lib/propolis/src/hw/qemu/pvpanic.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::pio::{PioBus, PioFn};
+
+pub struct QemuPioPvpanic {
+    counts: Arc<PanicCounts>,
+}
+
+#[derive(Debug, Default)]
+pub struct PanicCounts {
+    host_handled: AtomicUsize,
+    guest_handled: AtomicUsize,
```
pfmooney, line 20: From a design standpoint, I wonder if we should just keep the counts in a Mutex, since we're not expecting high event rates or contention.


The bigger question (in my mind) being how do we want stats consumers to reach in for the data?:
1. Holding an Arc to the data bits like you have now
2. Keeping and Arc to the device itself, and proving a method on the device to emit a current snapshot of the data.

My inclination would be the latter, but it's not 100% clear.

lib/propolis/src/hw/qemu/pvpanic.rs
```diff
@@ -0,0 +1,80 @@
...
+                if c & GUEST_HANDLED != 0 {
+                    self.counts.guest_handled.fetch_add(1, Relaxed);
+                }
+            }
+        }
+    }
+}
+
+impl PanicCounts {
+    #[must_use]
```
pfmooney, line 71: Why the `must_use` when the consequences for ignoring it are basically nil?

pfmooney (commented):
Thanks for drawing this up.

lib/propolis/src/hw/qemu/pvpanic.rs
hawkw: The style I'm used to from other projects is to put `#[must_use]` on most functions that return a value without performing any side effects (e.g. accessors, constructors). There are no _consequences_ for ignoring the return value from these functions, but these functions are called only to produce the return value --- if  the return value is not used, the function may as well not have been called, so calling them without using the value is generally a mistake. 

I'm happy to remove the `#[must_use]` attributes if this style is controversial, though.

lib/propolis/src/hw/qemu/pvpanic.rs
hawkw: Ah, thanks for the advice, I'll go ahead and change that!

lib/propolis/src/hw/qemu/pvpanic.rs
pfmooney: I'm inclined to elide it, for now, in this kind of case.  Maybe we'll decide that style is a fit for propolis in the future.  That said, there are probably a bunch of existing cases where ignoring a return value will cause correctness issues, so please do not hesitate to point those out.  (I've been trying to fix some of them as time passes)

lib/propolis/src/hw/qemu/pvpanic.rs
hawkw: > From a design standpoint, I wonder if we should just keep the counts in a Mutex, since we're not expecting high event rates or contention.

I guess it's a question of whether we consider a mutex to be more or less conceptual complexity than an atomic, if performance isn't an issue here. 

My preference is normally to avoid reaching for a `Mutex` in situations where an atomic is sufficient, regardless of whether or not the data is contended enough for there to be a performance impact. The atomic counter feels conceptually simpler to me: incrementing the counter is just a one-instruction write to an integer, rather than requiring an allocation and syscalls for `pthread_mutex`...

If there's a strong preference for a mutex, though, I'm happy to change this. 



>   1. Holding an Arc to the data bits like you have now
> 
>   2. Keeping and Arc to the device itself, and proving a method on the device to emit a current snapshot of the data.
> 
> 
> My inclination would be the latter, but it's not 100% clear.

Hmm, I think you're right that it's simpler to just have a single `Arc`ed type for the device itself, and just having the Oximeter `Producer` hold a reference to the `Arc`ed device itself. The reason I didn't do that was that I thought, eventually, we would also want to support the PCI bus PVPANIC device, and we would want panics reported using the PCI device to increment the same counters as the ISA bus device, so they would need to share the counters.

Looking at it again, though, in light of your comment https://github.com/oxidecomputer/propolis/pull/596#discussion_r1446649693, I suppose we could have one `Pvpanic` struct that's potentially registered as both a PCI device and a PIO device, so we don't actually need a separate type for each device.

openapi/propolis-server.json
```diff
@@ -623,6 +623,14 @@
               "$ref": "#/components/schemas/PciPciBridge"
             }
           },
+          "qemu_pvpanic": {
```
pfmooney, line 626: So we don't forget later, would you mind updating the `propolis-server-falcon.json` version of the spec as well? 

FYI: The build with the `falcon` feature must be done on illumos for FFI/linking reasons today, unlike "vanilla" which succeeds on linux with stubs.

lib/propolis/src/hw/qemu/pvpanic.rs
pfmooney: > I guess it's a question of whether we consider a mutex to be more or less conceptual complexity than an atomic, if performance isn't an issue here.

Atomics makes me at least acknowledge that memory ordering semantics are a thing (which I remain not especially confident about), even if I'm not using them in a context where such ordering will have a meaningful impact.  For that reason (and a few more), I prefer `Mutex` over atomics unless there's a pressing reason (performance or otherwise).  Most of the time, I just want consistent access to all the fields covered by one.

> I suppose we could have one Pvpanic struct that's potentially registered as both a PCI device and a PIO device.

If we were to expose it via both interfaces, doing so with a single set of counters (be it via one device or two) definitely seems preferable.

lib/propolis/src/hw/qemu/pvpanic.rs
```diff
@@ -0,0 +1,107 @@
...
+        pio.register(Self::IOPORT, 1, piofn).unwrap();
+    }
+
+    pub fn host_handled_count(&self) -> usize {
+        self.counts.lock().unwrap().host_handled
+    }
+
+    pub fn guest_handled_count(&self) -> usize {
+        self.counts.lock().unwrap().guest_handled
+    }
```
pfmooney, line 70: Worth defining a struct with both values to emit all at once?  Seems like most consumers will just want one big snapshot.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -282,7 +282,7 @@ impl<'a> MachineInitializer<'a> {
         uuid: uuid::Uuid,
     ) -> Result<(), anyhow::Error> {
         if let Some(ref spec) = self.spec.devices.qemu_pvpanic {
-            if spec.enable_isa {
+            if spec.enable_mmio {
```
pfmooney, line 285: I think "ISA" was more accurate when it comes to naming, even if the linux driver calls it MMIO.

hawkw: will change it back!

gjcolombo (commented):
This is looking good! A few comments on how we want the instance spec side of this to look:

crates/propolis-api-types/src/instance_spec/v0/mod.rs
```diff
@@ -113,6 +113,19 @@ pub struct DeviceSpecV0 {
     pub network_devices: HashMap<SpecKey, NetworkDeviceV0>,
     pub serial_ports: HashMap<SpecKey, components::devices::SerialPort>,
     pub pci_pci_bridges: HashMap<SpecKey, components::devices::PciPciBridge>,
+
+    /// This field has a default value (`None`) to allow for
```
gjcolombo, line 117: turbo nit: since this appears in the OpenAPI documentation for the server, I would make these regular comments instead of doc comments, since consumers have no control over our serde inside baseball (and it appears, from the regenerated spec, that the attributes don't actually change any of the semantics of the client)

gjcolombo (approved):
LGTM with one small tweak. (I didn't leave any additional comments on it, but the device emulation also seems fine to me, modulo any other feedback @pfmooney might have.)

pfmooney (approved):

PR #604, opened 2024-01-16
phd: automatically fetch `crucible-downstairs` from Buildomat

hawkw:
Currently, in order to run the `phd` tests that involve Crucible disks,
the path to a local `crucible-downstairs` binary must be provided via a
command-line argument. When no `crucible-downstairs` binary path is
provided, the tests that use Crucible are skipped. This means that the
tests which use Crucible disks are not currently able to run on CI, as
no `crucible-downstairs` binary is present.

This branch extends the `phd-framework` crate's artifact store to
support automatically fetch a `crucible-downstairs` binary from
Buildomat when configured via the command line. This PR adds a new
`--crucible-downstairs-commit` CLI argument to `phd-runner`. If present,
this argument configures PHD to fetch a `crucible-downstairs` binary
from Buildomat.

The `--crucible-downstairs-commit` argument can either be the string
`auto`, or a 40-character Git hash. When `auto` is passed, Git revision
to fetch from Buildomat is determined automatically based on the
`propolis` workspace's Git dependency on the `crucible` crate, to ensure
that the downstairs binary matches the version of the upstairs library
that Propolis was built against. Otherwise, if a Git hash is provided,
the artifact for that commit hash is fetched from Buildomat, instead.

The behavior of the existing `--crucible-downstairs-cmd` is unchanged,
so a local `crucible-downstairs` binary can still be used rather than
downloading it from Buildomat. This way, users can still run `phd` tests
against a local Crucible build, if necessary. The new
`--crucible-downstairs-commit` argument conflicts with
`--crucible-downstairs-cmd`, so if both are provided, `phd-runner` will
print an error. If _neither_ argument is provided, tests that require
Crucible are skipped, maintaining the default behavior of not
downloading anything unless explicitly configured to.

The Buildomat job for Crucible publishes the entire Illumos Crucible
zone as a tarball, rather than separate files for each artifact.
Therefore, it was necessary to extend the `phd-framework` artifact store
to support a configuration for extracting an individual file from within
a tarball. The SHA256 used for the artifact is the SHA of the entire
tarball, since Crucible's Buildomat job publishes the tarball's SHA, but
once the tarball is downloaded, the `crucible-downstairs` binary is
extracted from the tarball into the object store.


phd-tests/framework/src/artifacts/mod.rs
```diff
@@ -13,12 +13,15 @@ pub use store::Store as ArtifactStore;
 
 pub const DEFAULT_PROPOLIS_ARTIFACT: &str = "__DEFAULT_PROPOLIS";
 
-#[derive(Clone, Debug, Serialize, Deserialize)]
+pub const CRUCIBLE_DOWNSTAIRS_ARTIFACT: &str = "crucible-downstairs";
```
gjcolombo, line 16: nit: although we're not currently supporting arbitrary Crucible artifacts, I'd consider using the two-underscores-and-screaming-snake-case convention as on line 14 for this as well. That said this is mostly stylistic preference--it's useful only to the extent that it very gently tips the scales against having a conflicting artifact in the artifact TOML, and if there *is* one it's not a hard conflict to detect

phd-tests/framework/src/artifacts/store.rs
```diff
@@ -311,6 +373,88 @@ impl Store {
             )),
         }
     }
+
+    pub fn get_crucible_downstairs(&self) -> anyhow::Result<Utf8PathBuf> {
```
gjcolombo, line 377: Thanks for refactoring all this stuff--the store looks a lot cleaner with all these bits pulled out into proper functions!

phd-tests/framework/src/artifacts/store.rs
```diff
@@ -344,3 +488,68 @@ fn hash_equals(path: &Utf8Path, expected_digest: &str) -> anyhow::Result<()> {
...
+        };
+        let path = entry.path().context("Tarball entry path was not UTF-8")?;
+        if path == bin_path {
+            let filename = bin_path
+                .file_name()
+                .expect("binary path in tarball must include a filename");
+            let out_path = dir_path.join(filename);
+            entry.unpack(&out_path).with_context(|| {
+                format!(
+                    "Failed to unpack '{bin_path} 'from tarball to {out_path}"
```
gjcolombo, line 547: nit: spacing around `bin_path`

phd-tests/framework/src/artifacts/mod.rs
hawkw: can do!

phd-tests/runner/src/config.rs
```diff
@@ -36,8 +36,19 @@ pub struct RunOptions {
...
     /// The command to use to launch Crucible downstairs servers.
+    ///
+    /// If this is not present, then a Crucible revision will be determined
+    /// based on the current Git revision of `propolis`' dependency on the
+    /// `crucible` crate.
     #[clap(long, value_parser)]
-    pub crucible_downstairs_cmd: Option<Utf8PathBuf>,
+    crucible_downstairs_cmd: Option<Utf8PathBuf>,
+
+    /// Disable Crucible.
```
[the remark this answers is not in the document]
hawkw: :+1: that makes sense! my assumption was that we wanted it to, essentially, just work out of the box, but I'm fine with changing it so that downloads are always opt-in.

gjcolombo (approved):
Looks great--thanks for working on all this!

Issue #601, opened 2024-01-11
PHD: guest OS adapter for Windows Server 2022/2019/2016


gjcolombo:
Provided that we start with images that have EMS enabled and an appropriate admin password set, it's not too hard to write an adapter that gets us to a Windows command prompt on these guest OSes. The bigger challenge is what to do from there. One of the design goals of PHD is to make tests as concise as possible, and it's hard to have concise tests if every test has to have conditional logic that decides whether to use Unix-flavored or Windows-flavored commands. An emulation layer like Cygwin might help with this, and indeed I've built a Windows Server 2022 image that includes it that seems to work fine with PHD, but I'm running into some difficulty with WS2019 that I'm still investigating (will detail this in an issue comment once I've dug in more).

gjcolombo:
Two hiccups with the Cygwin/Server 2019 configuration so far:

- When running with Cygwin 3.4.9, in some cases, trying to interact with Cygwin over the serial console causes the `bash` process it spawns and its corresponding `conhost` process to get stuck in a CPU busy loop. I took a performance trace and observed a ton of context switches where `cygwin1.dll` was calling WriteConsoleInputW and then switching out to a conhost thread, only to be switched back in immediately by that thread. The Cygwin Git repo history shows a few changes to its main input processing thread's logic (which is the logic that calls the console APIs) over the last couple of years. I'm not sure what specifically the problem is here, but it looks like downgrading to version 3.0.7 of the Cygwin packages (including the main Cygwin glue DLL) mitigates the issue.
- Server 2019's SAC channel appears to differ from Server 2022's in that the former tries very hard to emulate an 80x24 VT100 terminal while the latter does not. That is, while Server 2022 happily echoes the newlines PHD types into it and prints trailing spaces when setting up a Bash prompt and running commands, Server 2019 prefers to use VT100 cursor positioning commands to move the cursor around. PHD's terminal processing is currently extremely basic and doesn't cope with this at all. I need either to find a way to suppress this behavior in WS2019 (no luck so far) or to expand the VT logic in PHD to handle this mode of operation.

gjcolombo:
I've been thinking about what to do for WS2019 guests today and have an approach in mind. Here are my notes.

The general problem is that guests have (at least) two approaches to writing to the serial console:

1. Some guests write and echo characters and control bytes "as-is" and allow the recipient to decide how to format them for display in a terminal.
[...]

PR #607, opened 2024-01-16
PHD: add guest adapter for WS2022

gjcolombo:
The adapter assumes that

- sysprep has been run such that the guest OS is generalized and, at first boot, its hostname will be set to "PHD-WINDOWS"
- Cygwin is installed to C:\cygwin
- the local administrator account is enabled with the appropriate password

A suitable image is available on catacomb. This image passes `nproc_test`, but currently fails several other tests in ways that appear to be unrelated to the correctness of the adapter:

- The Cygwin version in the test image doesn't include a `pciutils` package in-box, so the `lspci` lifecycle test doesn't work as expected. There are Windows ports of `pciutils` available that could resolve this, but a more flexible approach (which I will probably take in a future PR) is to allow tests to ask the framework what "flavor" of user space the current guest provides so that they can tailor their commands accordingly.
- The Crucible tests unconditionally specify that the guest disk size should be 10 GiB, which is smaller than the guest image, which (unsurprisingly) leads to a very unhappy guest. This is a general problem with the way Crucible disk sizes work and will be fixed in a separate PR.

Tests: local test run of the `nproc` smoke test with a WS2022 image.

Related to #601.

### DIFF
--- a/phd-tests/framework/src/guest_os/windows_server_2022.rs
+++ b/phd-tests/framework/src/guest_os/windows_server_2022.rs
@@ -9,14 +9,24 @@
 
 use super::{CommandSequence, CommandSequenceEntry, GuestOs};
 
+/// Provides a guest OS adapter for Windows Server 2022 images. This adapter
+/// assumes the following:
+///
+/// - The image has been generalized (by running `sysprep /generalize`) and is
+///   configured so that on first boot it will skip the out-of-box experience
+///   (OOBE) and initialize the local administrator account with the appropriate
+///   password. See [MSDN's Windows Setup
+///   documentation](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/generalize?view=windows-11)
+///   for more details.
+/// - Cygwin is installed to C:\cygwin and can be launched by invoking
+///   C:\cygwin\cygwin.bat.
 pub(super) struct WindowsServer2022;
 
 impl GuestOs for WindowsServer2022 {
     fn get_login_sequence(&self) -> CommandSequence {
         CommandSequence(vec![
-            // The image is generalized and needs to be specialized. Let it boot
-            // once, then wait for it to boot again and only then start waiting
-            // for the command prompt to become available.
+            // Assume the image will need to reboot one last time after being
+            // specialized.
             CommandSequenceEntry::WaitFor(
                 "Computer is booting, SAC started and initialized.",
             ),
--- a/phd-tests/framework/src/guest_os/windows_server_2022.rs
+++ b/phd-tests/framework/src/guest_os/windows_server_2022.rs
@@ -1,0 +1,67 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Guest OS adaptations for Windows Server 2022 images. This adapter assumes
+//! the guest OS image has been generalized, that Cygwin has been installed, and
+//! that the local administrator user was configured at image generation time to
+//! have the appropriate password.
+
+use super::{CommandSequence, CommandSequenceEntry, GuestOs};
+
+pub(super) struct WindowsServer2022;
+
+impl GuestOs for WindowsServer2022 {
+    fn get_login_sequence(&self) -> CommandSequence {
+        CommandSequence(vec![
+            // The image is generalized and needs to be specialized. Let it boot
+            // once, then wait for it to boot again and only then start waiting
+            // for the command prompt to become available.
+            CommandSequenceEntry::WaitFor(
+                "Computer is booting, SAC started and initialized.",
+            ),
+            CommandSequenceEntry::WaitFor(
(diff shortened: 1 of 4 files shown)


phd-tests/framework/src/guest_os/windows_server_2022.rs
```diff
@@ -0,0 +1,67 @@
...
+            CommandSequenceEntry::WaitFor(
+                "Use any other key to view this channel.",
+            ),
+            CommandSequenceEntry::WriteStr(""),
+            CommandSequenceEntry::WaitFor("Username:"),
+            CommandSequenceEntry::WriteStr("Administrator"),
+            CommandSequenceEntry::WaitFor("Domain  :"),
+            CommandSequenceEntry::WriteStr(""),
+            CommandSequenceEntry::WaitFor("Password:"),
+            CommandSequenceEntry::WriteStr("0xide#1Fan"),
```
pfmooney, line 43: If we're going to be encoding more expectations about image login credentials, it might make sense to have some top-level markdown doc listing them so that folks don't need to go fishing in the source for them?

phd-tests/framework/src/guest_os/windows_server_2022.rs
```diff
@@ -0,0 +1,67 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Guest OS adaptations for Windows Server 2022 images. This adapter assumes
+//! the guest OS image has been generalized, that Cygwin has been installed, and
+//! that the local administrator user was configured at image generation time to
+//! have the appropriate password.
```
hawkw, line 8: not at all important, nitpicky: would it make sense for this to be a doc comment on `WindowsServer2022` rather than at the module level?

phd-tests/framework/src/guest_os/windows_server_2022.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use super::{CommandSequence, CommandSequenceEntry, GuestOs};
+
+pub(super) struct WindowsServer2022;
+
+impl GuestOs for WindowsServer2022 {
+    fn get_login_sequence(&self) -> CommandSequence {
+        CommandSequence(vec![
+            // The image is generalized and needs to be specialized. Let it boot
```
hawkw, line 17: out of curiosity: what does it mean for the image to be "generalized"/"specialized", exactly? and why does booting the image twice do what we need?

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -612,8 +612,8 @@ impl TestVm {
     async fn send_serial_str_async(&self, string: &str) -> Result<()> {
         let mut bytes = Vec::new();
         bytes.extend_from_slice(string.as_bytes());
-        self.send_serial_bytes_async(bytes).await?;
-        self.send_serial_bytes_async(vec![b'\n']).await
+        bytes.extend_from_slice(&[b'\n']);
```
hawkw, line 615: nitpicky, not a big deal: since this is a single character, why do we need to call `extend_from_slice` with a single-character-long slice? can't we just write:
```suggestion
        bytes.push(b'\n');
```

hawkw (approved):
I had a couple questions and some _very_ minor nitpicks, but this looks good to me overall!

phd-tests/framework/src/guest_os/windows_server_2022.rs
```diff
@@ -9,14 +9,24 @@
...
 
+/// Provides a guest OS adapter for Windows Server 2022 images. This adapter
+/// assumes the following:
+///
+/// - The image has been generalized (by running `sysprep /generalize`) and is
+///   configured so that on first boot it will skip the out-of-box experience
+///   (OOBE) and initialize the local administrator account with the appropriate
+///   password. See [MSDN's Windows Setup
+///   documentation](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/generalize?view=windows-11)
+///   for more details.
```
hawkw, line 20: this is lovely, thank you for expanding on it!

hawkw (approved):

phd-tests/framework/src/guest_os/windows_server_2022.rs
gjcolombo: Good call, done in 40aed6a.

phd-tests/framework/src/guest_os/windows_server_2022.rs
gjcolombo: My pleasure! One additional detail here (I mentioned this in DMs, but adding here as well for posterity) is that the (WIP) Windows image builder currently generates generalized images (so that we don't have to worry about any specialization conflicts if they're used as the read-only base images for multiple VMs), and I leveraged that here to create the test image I used to test this adapter. It might be possible to create an image that's technically been specialized already (so that Setup doesn't have to run and we don't have to wait for this reboot) but can still be reused across multiple VMs, but I'd have to do some research to figure out how.

PR #611, opened 2024-01-19
phd: Ensure min disk size fits read-only parents

### LINKED ISSUE #605: PHD: minimum Crucible disk size should always be at least the size of the read-only parent
Today PHD's Crucible tests unconditionally specify a 10 GiB guest disk even if the guest image being used as a read-only parent is larger than that. To avoid truncating, tests should specify a *minimum* guest disk size, and PHD should expand it to the underlying guest image size if that exceeds what the test specified.

hawkw:
Today PHD's Crucible tests unconditionally specify a 10 GiB guest disk
even if the guest image being used as a read-only parent is larger than
that. To avoid truncating, tests should specify a minimum guest disk
size, and PHD should expand it to the underlying guest image size if
that exceeds what the test specified.

This branch changes the `disk_size_gib` argument to functions that
create Crucible disks to `min_disk_size_gib`. If a disk is created with
a read-only parent, the code that constructs new Crucible disks
(`CrucibleDisk::new`) will now read the FS metadata for the parent image
to determine its size, and uses that to set the size of the created disk
to ensure that the read-only parent image fits.

Fixes #605

### DIFF
--- a/phd-tests/framework/src/disk/crucible.rs
+++ b/phd-tests/framework/src/disk/crucible.rs
@@ -11,6 +11,7 @@ use std::{
     sync::atomic::{AtomicU64, Ordering},
 };
 
+use anyhow::Context;
 use propolis_client::types::{
     CrucibleOpts, CrucibleStorageBackend, StorageBackendV0,
     VolumeConstructionRequest,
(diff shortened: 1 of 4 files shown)


gjcolombo (approved):
Thanks for pulling this together!

PR #615, opened 2024-01-23
PHD: refactor guest serial console handling

gjcolombo:
Significantly clean up PHD's serial console logic: use only one tokio
task instead of two, and abstract console operations like processing
guest bytes and waiting for strings into a trait to allow the framework
to implement different handlers for different kinds of guests.

The long-range intention is to support a second "backend" that better
handles the command sequences sent by guests that think they're driving
a VT100-compatible terminal instead of leaving that work to the
recipients of their bytes. See #601 for more details.

Log serial console output to files instead of the main PHD tracing
channel. At one point this sped up testing enough to reveal a bug in
Crucible downstairs teardown, which was killing Crucible processes but
not waiting for them to exit; fix this issue.

Finally, tweak the instance ensure path to use a nice, shiny exponential
backoff regime when waiting for Propolis servers to begin serving
instead of a hand-rolled retry loop with an arbitrary interval.

Tested with local PHD runs against Alpine and Debian 11 guests.


phd-tests/framework/src/serial/mod.rs
```diff
@@ -2,114 +2,210 @@
...
-    vt_processor: vt100::Vt100Processor,
-    wstx: tokio::sync::mpsc::Sender<Vec<u8>>,
+
+    /// Bytes sent to this channel are received by the websocket task, which
+    /// sends them to Propolis.
+    ws_tx: UnboundedSender<Vec<u8>>,
+
+    /// The buffer for this serial console, shared by the console itself and by
+    /// its websocket handler task.
+    buffer: Arc<Mutex<Box<dyn Buffer>>>,
```
hawkw, line 76: AFAICT, the buffer only needs to be shared between the `SerialConsole` struct and the websocket task so that we can mutate the `OutputWaiter` in `register_wait_for_string`/`cancel_wait_for_string`. The actual buffer is never touched by the `SerialConsole` struct, just the `OutputWaiter`. It seems like we _could_ refactor this so that we just share an `Arc<Mutex<OutputWaiter>>` between the `SerialConsole` and the websocket task, so that we don't need to also lock the mutex every time we write bytes when there's no output waiter set?

Potentially, we could even go as far as changing this code to give the websocket task exclusive ownership over the `OutputWaiter` as well, and register a new waiter/cancel the waiter by sending messages to the task over a channel. This way, we have no shared state that requires a mutex, which is potentially much more efficient.

This isn't a blocking concern, but it might be worth looking into.

phd-tests/framework/src/serial/raw_buffer.rs
```diff
@@ -0,0 +1,141 @@
...
+    wait_buffer: String,
+    waiter: Option<OutputWaiter>,
+}
+
+impl Inner {
+    /// Pushes `c` to the buffer's contents and attempts to satisfy active
+    /// waits.
+    fn push_character(&mut self, c: char) {
+        if c == '\n' {
+            writeln!(self.log, "{}", &self.line_buffer).unwrap();
```
hawkw, line 33: very tiny nit: i generally prefer to avoid going through the `fmt` machinery when there's no actual formatting going on:
```suggestion
            self.log.write_all(self.line_buffer.as_bytes()).unwrap();
```
not a blocker, of course!

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -226,18 +230,53 @@ impl TestVm {
...
+                                e
+                            )))
+                        }
+                    }
+                } else {
+                    Ok(())
+                }
+            },
+        )
+        .await?;
```
hawkw, line 279: style nit, take it or leave it: i might consider defining the closure separately to avoid the extra level of indentation. but, it's up to you...

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -226,18 +230,53 @@ impl TestVm {
...
+                    .client
+                    .instance_spec_ensure()
+                    .body(&ensure_req)
+                    .send()
+                    .await
+                {
+                    match e {
+                        propolis_client::Error::CommunicationError(e) => {
+                            info!(%e,
+                                  "retriable error from instance_spec_ensure");
```
hawkw, line 254: elsewhere i think i may have spelt this "retryable"...TIL that "retriable" is the correct one.

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -226,18 +230,53 @@ impl TestVm {
...
+                            Err(backoff::Error::Transient {
+                                err: anyhow!(
+                                    "retriable error from instance_spec_ensure \
+                                    ({:?})",
+                                    e
+                                ),
+                                retry_after: None,
+                            })
```
hawkw, line 262: style nit, take it or leave it:: there's a convenience constructor for this in `backoff`:
```suggestion
                            Err(backoff::Error::transient(anyhow!(
                                "retriable error from instance_spec_ensure ({e:?})",
                            ))
```

hawkw: also, we might prefer to use `anyhow::Context`, rather than creating a new error using `anyhow!`:
```suggestion
                            Err(backoff::Error::transient(
                                e.context("retriable error from instance_spec_ensure")
                            ))
```

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -226,18 +230,53 @@ impl TestVm {
...
+                        }
+                        _ => {
+                            error!(%e,
+                                   "permanent error from instance_spec_ensure");
+
+                            Err(backoff::Error::permanent(anyhow!(
+                                "error from instance spec ensure: {:?}",
+                                e
+                            )))
```
hawkw, line 271: nit, unimportant: we might prefer to use `anyhow::Context`:
```suggestion
                            Err(backoff::Error::permanent(
                                e.context("error from instance spec ensure")
                            ))
```

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -627,6 +670,15 @@ impl TestVm {
...
     }
+
+    /// Generates a path to a file into which the VM's serial console adapter
+    /// can log serial console output.
+    fn serial_log_file_path(&self) -> Utf8PathBuf {
+        let filename = format!("{}.serial.log", self.spec.vm_name);
+        let mut path = self.data_dir.clone();
+        path.push(filename);
+        path
+    }
```
hawkw, line 681: as a follow-up, it might be nice if the PHD framework would go back and print the serial log output when a test fails, so that the output from failing tests ends up in the main log automatically?

i was also thinking about possibly changing PHD to do this for `tracing` logs by default, a la `libtest`'s output capturing...

hawkw (commented):
This looks great! I commented on a bunch of fairly minor nits. None of them are particularly important --- feel free to ignore me. :)

phd-tests/framework/src/serial/mod.rs
```diff
@@ -2,114 +2,210 @@
...
 pub struct SerialConsole {
+    /// A handle to a tokio task that handles websocket messages to and from the
+    /// Propolis server that serves this console.
     ws_task: JoinHandle<()>,
-    vt_processor: vt100::Vt100Processor,
-    wstx: tokio::sync::mpsc::Sender<Vec<u8>>,
+
+    /// Bytes sent to this channel are received by the websocket task, which
+    /// sends them to Propolis.
+    ws_tx: UnboundedSender<Vec<u8>>,
```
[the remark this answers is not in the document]
gjcolombo: I left this alone for now--we can always reevaluate if this turns out to be a significant performance headache (right now the long pole on most of my test runs is "copying disk files around"...).

phd-tests/framework/src/serial/mod.rs
```diff
@@ -2,114 +2,210 @@
...
+
+        let ws_span = info_span!("Serial websocket task");
+        ws_span.follows_from(tracing::Span::current());
+
+        let buffer = Arc::new(Mutex::new(new_buffer(buffer_kind, log_path)?));
+        let buffer_for_task = buffer.clone();
+        let ws_task = tokio::spawn(
+            async move { websocket_handler(ws, buffer_for_task, ws_rx).await }
+                .instrument(ws_span),
+        );
```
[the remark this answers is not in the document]
gjcolombo: Nice! Fixed in e5fba95e.

phd-tests/framework/src/serial/mod.rs
gjcolombo: This is going to get a little more complicated in the future, because I intend to make it possible to switch buffering disciplines at runtime (to handle guests that e.g. use `RawBuffer` at first but will switch to `Vt80x24` midway through their startup sequences). My thinking is that `SerialConsole` would handle that case by taking the lock and swapping out the existing `Box` with a pointer to a new buffer that implements the new discipline.

I did think a bit about introducing a separate control channel so that the websocket task can just own everything (in this world it would also need to accept messages that say "change your buffering mode to X"). I can give that another whirl and see how it looks now that everything else is more or less put together.

phd-tests/framework/src/serial/raw_buffer.rs
gjcolombo: I knew there had to be a better way to do this. Thanks! Fixed in 6e75a993.

phd-tests/framework/src/test_vm/mod.rs
gjcolombo: Fixed in 84f94e9.

phd-tests/framework/src/test_vm/mod.rs
gjcolombo: I started using the convenience constructor in 84f94e93. I'm not sure about how to do the `context` bit, though--IIUC `anyhow` implements that for `Result`, not for the errors themselves. I tried throwing an extra `.context` onto the preceding call to `instance_spec_ensure` and then downcasting the error back to a `propolis_client::Error` to try to figure out whether it was transient or not, but the downcast panicked for reasons I'd need to investigate.

Because we're logging a message in either case (and we don't really care about the innards of the error in the transient case anyway), in 4ff11a67 I dropped the construction of the extra `anyhow` error and just put the Propolis client error directly into the backoff error.

phd-tests/framework/src/test_vm/mod.rs
gjcolombo: I ended up getting rid of the anyhow layer in 4ff11a67 (see above).

phd-tests/framework/src/serial/mod.rs
```diff
@@ -2,114 +2,210 @@
...
+    /// Registers with the current buffer a request to wait for `wanted` to
+    /// appear in the console buffer. When a match is found, the buffer sends
+    /// all buffered characters preceding the match to `preceding_tx`, consuming
+    /// those characters and the matched string. If the buffer already contains
+    /// one or more matches at the time this call is made, the last match is
+    /// used to satisfy the wait immediately.
+    pub fn register_wait_for_string(
+        &self,
+        wanted: String,
+        preceding_tx: UnboundedSender<String>,
```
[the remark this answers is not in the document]
gjcolombo: Fixed in c5c7d300.

phd-tests/framework/src/serial/mod.rs
gjcolombo: I like how this turned out--it also makes it a little easier to fail fast in cases where, e.g., the task exited because something went wrong with the websocket connection (previously, you could register new waits but would have to wait for them to time out; now, to register a wait, you have to send to the task, so you error out right away if the connection was lost). Retested with Alpine and Debian 11 and pushed in 119d57b9.

phd-tests/framework/src/serial/mod.rs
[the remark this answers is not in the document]
hawkw: yeah, it probably doesn't really matter! just wanted to share the `DuplexStream` API in case you hadn't seen it before :)

phd-tests/framework/src/serial/mod.rs
hawkw: :woman_shrugging: the mutex makes more sense given those future plans, i'm fine with this approach. again, the cost of any of this stuff is almost certainly not what dominates test run time. :)

phd-tests/framework/src/test_vm/mod.rs
hawkw: cool, seems fine to me --- I *thought* there was also a `context` method for errors, but maybe I hallucinated that, ChatGPT-style :)

phd-tests/framework/src/serial/mod.rs
hawkw: ah, looks like we raced on comments again. i like the approach in i like the approach you switched to in [119d57b](https://github.com/oxidecomputer/propolis/pull/615/commits/119d57b913d11c86e1b2f7a19856d4837abf68d7) a lot, this seems great! in the future, if we want to add code for switching the buffer discipline at runtime, i suppose we could do that over the command channel, too!

hawkw (approved):
looks great!

phd-tests/framework/src/test_vm/mod.rs
gjcolombo: Printing the serial console output after the fact seems tricky in PHD's current architecture (though I might just be missing something). It's the runner that notices that test cases have failed (by observing an error returned from a test case or a panic caught while executing a test), so it would need to be the one to start dumping the consoles of any active VMs, but the runner has no idea what VMs the test owns. I think we'd need more control-plane-ish `Framework` that keeps track of the VMs it creates so that it can run them down on failure to do this elegantly.

> i was also thinking about possibly changing PHD to do this for tracing logs by default, a la libtest's output capturing...

I kind of like being able to follow the progress of the run in the terminal, but if there's a `--nocapture` equivalent I can always just add that to my PHD wrapper script.

gjcolombo:
Thanks for the review! Will merge once the latest run comes out green (it IS going to come out green, right?).

PR #617, opened 2024-01-24
Update Crucible (and Omicron) dependencies

pfmooney:


Cargo.toml
```diff
@@ -78,8 +78,8 @@ oximeter-producer = { git = "https://github.com/oxidecomputer/omicron", branch =
 oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
 
 # Crucible
-crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "e71b10d2f9f1fb52818b916bae83ba15a339548d" }
-crucible-client-types = { git = "https://github.com/oxidecomputer/crucible", rev = "e71b10d2f9f1fb52818b916bae83ba15a339548d" }
+crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "eb96d243c1c711df15cd108c6c2ad9b239565e5f" }
```
mkeeter, line 81: Would you mind bumping one more time to Crucible commit [6b12017765410a52b8d13b5f2fab85842ca365e3](https://github.com/oxidecomputer/crucible/commit/6b12017765410a52b8d13b5f2fab85842ca365e3)?

lib/propolis/src/block/crucible.rs
```diff
@@ -332,28 +338,50 @@ impl From<Error> for block::Result {
...
 ) -> Result<(), Error> {
+    let block_size = info.block_size as usize;
+
+    // Calculate offset (in crucible::Block form) and length in blocksize
+    fn block_offset_count(
+        off_bytes: usize,
+        len_bytes: usize,
+        block_size: usize,
+    ) -> Result<(crucible::Block, usize), Error> {
+        if off_bytes % block_size == 0 || len_bytes % block_size == 0 {
```
mkeeter, line 355: This should be `&&` not `||`

mkeeter: Perhaps this should be a static member function, so it could be tested?

pfmooney: > This should be `&&` not `||`

Gah!  I'd flipped the check, but forgot the operator.  Thanks for catching that.

> Perhaps this should be a static member function, so it could be tested?

Sure, will do that and add a basic test or two

Cargo.toml
pfmooney: Done.

lib/propolis/src/block/crucible.rs
```diff
@@ -312,6 +315,9 @@ pub enum Error {
     #[error("backend is read-only")]
     ReadOnly,
 
+    #[error("offset or length not multiple of blocksize")]
+    BlocksizeMismatch,
```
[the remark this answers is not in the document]
pfmooney: Perhaps one day if/when we propagate more detail about the error through the block layer, but currently it gets immediately [discarded](https://github.com/oxidecomputer/propolis/blob/master/lib/propolis/src/block/crucible.rs#L325-L330) anyways

lib/propolis/src/block/crucible.rs
```diff
@@ -330,30 +336,52 @@ impl From<Error> for block::Result {
...
+    len_bytes: usize,
+    block_size: usize,
+) -> Result<(crucible::Block, usize), Error> {
+    if off_bytes % block_size == 0 && len_bytes % block_size == 0 {
+        Ok((
+            crucible::Block::new(
+                (off_bytes / block_size) as u64,
+                block_size.trailing_zeros(),
+            ),
+            len_bytes / block_size,
```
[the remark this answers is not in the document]
pfmooney: If crucible gives us a blocksize of 0, many more things are going to be grumpy

mkeeter (approved):

lib/propolis/src/block/crucible.rs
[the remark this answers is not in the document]
hawkw: okay, yup, that makes sens!

hawkw (approved):

pfmooney:
Merged in 261fe5ae06103cd12395da9c8d3c23b959d43f1d

PR #619, opened 2024-01-24
PHD: add `cargo phd` to make using PHD nicer

hawkw:
Human use of `phd-runner` has a few pain points. In particular:

- Because `phd-runner` takes the path to the `propolis-server` binary as a CLI argument, it's easy to forget to rebuild `propolis-server` after making a change, and accidentally run the tests against a stale artifact. This has bitten me a few times, and in some cases, has resulted in a non-trivial time waste, both because running the tests takes a bit and because I've spent time debugging why the tests were still failing after I fixed something, only to discover that I was testing the old binary.
- A whole bunch of command-line arguments have to be passed in every time, which is a bit annoying to remember.
- The `phd-quickstart.sh` script can be used to avoid typing out some of the CLI args, but its management of the temp dir and artifact store is not particularly sophisticated. In particular, it uses the same directory for both,[^1] making it difficult to reuse the artifact store to avoid downloading the same artifacts over and over again.

In order to make testing changes locally during development a bit nicer, I've added a new `cargo phd` Cargo subcommand for running PHD tests, which should resolve some of these issues. Because `cargo phd`'s behavior is a bit more opinionated than `phd-runner`, and it has default behaviors for a bunch of things that `phd-runner` requires be explicitly specified, I've implemented it as a separate binary that's completely opt-in. In particular, I don't think it would make sense to change the `phd-run` CI jobs to use `cargo phd`.

The primary goal of the cargo subcommand is to ensure that the `propolis-server` binary used for testing is always up-to-date with local changes. The subcommand shells out to `cargo build`, to ensure that the binary is up-to-date, and then passes the path to the freshly built `propolis-server` to `phd-runner`. This relies on cargo's change detection to avoid rebuilding the binary when the `propolis-server` codebase hasn't acutally changed.

If the user passes a `--propolis-server-cmd` argument, that value is used rather than rebuilding the `propolis-server` binary, so `cargo phd` can still be used to test a `propolis-server` binary that came from elsewhere.

In addition, `cargo phd` manages temporary dirs and the artifact store a bit more intelligently. The artifact store and temp dirs for a test run are `target/phd/artifacts` and `target/phd/tmp/{timestamp}`, respectively. This lets us create separate temp dirs for each run, to avoid clobbering logs and stuff from prior runs, but reuse the existing artifacts we may have downloaded previously, making it faster to run tests repeatedly. Because we put both directories in `target/`, they last longer than `/tmp/` would, making the persistent artifact store more useful, but they're still blown away by `cargo clean` if the developer wants to free up some space.

I'll admit that I may also have gotten a bit carried away trying to make the CLI as nice as possible... :)

[^1]: which, unfortunately, is what `phd-runner` will actually *always* do, prior to PR #618...


.cargo/config.toml
```diff
@@ -1,5 +1,6 @@
 [alias]
 xtask = "run --package xtask --quiet --"
+phd = "run --package cargo-phd --quiet --"
```
smklein, line 3: This seems like a really cool use of aliases! Not a request, just a question, but why bother using xtask when all subcommands could be implemented this way?

pfmooney: Speaking personally, I like the namespacing offered by an xtask prefix.  `cargo xtask clippy`, for example, performs the run(s) with all the necessary fanfare, and I wouldn't want it colliding with the native `cargo clippy`

hawkw: I'm happy to change this to `cargo xtask phd` if that's preferred! My initial approach was going to use Cargo artifact deps, so I didn't want it to be part of the `xtasks` crate so that compiling xtasks didn't also require compiling `propolis-server` and `phd-runner`. But, I ended up not using artifact dependencies, since they're unstable, and instead did the compilation of `propolis-server` and `phd-runner` at runtime, so that issue is no longer the case. I'm happy to move this code to the `xtasks` crate if you like!

pfmooney: I'm not opposed to it living outside of `xtask`.  I was more voicing my opposition to all xtasks being turned into aliases.

hawkw: I moved it into the `xtasks` crate in 0da065452a5283bf79259976bfcbaa6b17b97325 --- it is nice for it to show up in `cargo xtask help`.

pfmooney: Sounds good.

xtask/src/task_phd.rs
```diff
@@ -0,0 +1,349 @@
...
+            .arg(args::CRUCIBLE_COMMIT)
+            .arg(crucible_commit.unwrap_or("auto"))
+            .arg(args::PROPOLIS_BASE)
+            .arg(propolis_base_branch.unwrap_or("master"))
+            .arg(args::ARTIFACTS_TOML)
+            .arg(&artifacts_toml)
+            .arg("--artifact-directory")
+            .arg(&artifact_dir)
+            .arg("--tmp-directory")
+            .arg(&tmp_dir)
```
gjcolombo, line 151: I'd probably make at least the artifact directory overrideable. In the future when we specify more and larger artifacts in the checked-in TOML, I think it'd be valuable to be able to elect to stash those in a directory that isn't subject to being pruned by `cargo clean`.

(I suppose the tmp directory could be overridden too, but I don't think the argument for it is as strong, outside maybe of uniformity with the other switches.)

xtask/src/task_phd.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub(super) const PROPOLIS_BASE: &str = "--base-propolis-branch";
+    pub(super) const CRUCIBLE_COMMIT: &str = "--crucible-downstairs-commit";
+    pub(super) const ARTIFACTS_TOML: &str = "--artifact-toml-path";
+}
+
+fn build_bin(name: impl AsRef<str>) -> anyhow::Result<escargot::CargoRun> {
+    const PROFILE: &str = if cfg!(debug_assertions) {
+        "dev [unoptimized + debuginfo]"
+    } else {
+        "release [optimized]"
```
gjcolombo, line 170: How do you specify this with an xtask? (Do you have to manually invoke the xtask via `cargo run`?)

It's generally undesirable to build `phd-runner` with the `release` profile in this workspace, because that enables abort-on-panic, and the runner expects panics to unwind so it can catch them to report test failures. It might be nice to have an option to override the `current_release` setting here if we know we're building the runner itself.

gjcolombo (approved):
This looks nice! I'd tweak a couple of things but LGTM overall.

xtask/src/task_phd.rs
hawkw: oh, i thought i'd done this, looks like maybe i forgot to push that release!

xtask/src/task_phd.rs
hawkw: :+1:  will changee that!

hawkw: > How do you specify this with an xtask? (Do you have to manually invoke the xtask via `cargo run`?)

Yes, but we have a  cargo alias for xtasks, so you can type `cargo xtask ...` and get `cargo run` with some bonus flags:
https://github.com/oxidecomputer/propolis/blob/c596d09aa152ab8c8cf5086756ae86b16708d9b4/.cargo/config.toml#L1-L2

Currently, if you run the xtask using `cargo xtask`, you'll never be able to specify a release build, because the `cargo xtask` alias passes all arguments to the xtask binary, so if you type `cargo xtask phd --release`, the `--release` argument is passed _to the xtask binary_ rather than to `cargo`.

xtask/src/task_phd.rs
hawkw: alright, as of 6cdbbafd617d05c704f1eed1ed3101879d1a0523, you can now do:
```console
$ cargo xtask phd run --artifact-directory ~/.cache/phd/artifacts
   Compiling phd-runner
    Finished phd-runner dev [unoptimized + debuginfo] in 0.23s
  Overridden --artifact-directory "/home/eliza/.cache/phd/artifacts"
   Compiling propolis-server
    Finished propolis-server dev [unoptimized + debuginfo] in 51.95s
     Created artifact directory `/home/eliza/.cache/phd/artifacts`
     Created temp directory `target/phd/tmp/1706309041`
       Found artifacts.toml at `phd-tests/artifacts.toml`
     Running target/x86_64-unknown-linux-gnu/debug/phd-runner run \
        --propolis-server-cmd target/x86_64-unknown-linux-gnu/debug/propolis-server \
        --crucible-downstairs-commit auto \
        --base-propolis-branch master \
        --artifact-toml-path phd-tests/artifacts.toml \
        --artifact-directory /home/eliza/.cache/phd/artifacts \
        --tmp-directory target/phd/tmp/1706309041
2024-01-26T22:44:01.294551Z  WARN phd_runner: 27: Failed to enable one or more kernel options, some tests may not work error=set_vmm_globals requires illumos
```

xtask/src/task_phd.rs
hawkw: as of 06e0fa60084af15b005ca9efff16df4cbf23c10a and 3e77e05df08e75c09680d3f2d14219677784a75d, we now always build `phd-runner` in debug mode, and i've added a `--release`/`-r` flag to build `propolis-server` in release mode (if you want to do that):

```console
$ cargo xtask phd run --release
   Compiling phd-runner
    Finished phd-runner dev [unoptimized + debuginfo] in 0.25s
   Compiling propolis-server
    Finished propolis-server release [optimized] in 87.05s
       Found existing artifact directory `target/phd/artifacts`
     Created temp directory `target/phd/tmp/1706310120`
       Found artifacts.toml at `phd-tests/artifacts.toml`
     Running target/x86_64-unknown-linux-gnu/debug/phd-runner run \
        --propolis-server-cmd target/x86_64-unknown-linux-gnu/release/propolis-server \
        --crucible-downstairs-commit auto \
        --base-propolis-branch master \
        --artifact-toml-path phd-tests/artifacts.toml \
        --artifact-directory target/phd/artifacts \
        --tmp-directory target/phd/tmp/1706310120
2024-01-26T23:02:00.279763Z  WARN phd_runner: 27: Failed to enable one or more kernel options, some tests may not work error=set_vmm_globals requires illumos
```

and, if you combine `--release` with a custom `propolis-server` path, you get a warning! :)

```console
:; cargo xtask phd run --release --propolis-server-cmd target/x86_64-unknown-linux-gnu/debug/propolis-server
   Compiling phd-runner
    Finished phd-runner dev [unoptimized + debuginfo] in 0.24s
  Overridden --propolis-server-cmd "target/x86_64-unknown-linux-gnu/debug/propolis-server"
warning: setting `--release` to build propolis-server in release mode does nothing if an existing propolis binary path is provided using `--propolis-server-cmd`
       Found existing artifact directory `target/phd/artifacts`
     Created temp directory `target/phd/tmp/1706310263`
       Found artifacts.toml at `phd-tests/artifacts.toml`
     Running target/x86_64-unknown-linux-gnu/debug/phd-runner run \
        --propolis-server-cmd target/x86_64-unknown-linux-gnu/debug/propolis-server \
        --crucible-downstairs-commit auto \
        --base-propolis-branch master \
        --artifact-toml-path phd-tests/artifacts.toml \
        --artifact-directory target/phd/artifacts \
        --tmp-directory target/phd/tmp/1706310263
2024-01-26T23:04:23.328977Z  WARN phd_runner: 27: Failed to enable one or more kernel options, some tests may not work error=set_vmm_globals requires illumos
```

PR #621, opened 2024-01-25
PHD: fix run_shell_command with multiline commands

hawkw:
Currently, the `TestVm::run_shell_command` method in `phd_framework` fails incorrectly when the shell is multi-line. This is because the shell will not echo the literal command, but instead prints a character at the beginning of each continuation line. On Alpine, this is `> `. Because `run_shell_command` waits for the literal command line to be echoed, it will never see the input command again if it's more than one line, because of the addition of these continuation characters by the shell, and we time out waiting to see the command echoed.

This branch fixes this by adding continuation characters to `echo_cmd` when the input command is multi-line. I can't promise this works on all Linux images, since other shells may use a different character for multi-line continuations, but it works with the Alpine image we're currently using...

### DIFF
--- a/phd-tests/framework/src/test_vm/mod.rs
+++ b/phd-tests/framework/src/test_vm/mod.rs
@@ -623,14 +623,20 @@ impl TestVm {
     pub fn run_shell_command(&self, cmd: &str) -> Result<String> {
         self.send_serial_str(cmd)?;
 
-        let mut echo_cmd = cmd.to_string();
+        // If the command is multi-line, it won't be echoed literally.
+        // instead, it will (probably) have each line begin with an `>`. so,
+        // fix that.
+        let mut echo_cmd = cmd.trim_end().replace('\n', "\n> ");
         echo_cmd.push('\n');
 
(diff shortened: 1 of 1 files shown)


phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -623,14 +623,20 @@ impl TestVm {
     pub fn run_shell_command(&self, cmd: &str) -> Result<String> {
         self.send_serial_str(cmd)?;
 
-        let mut echo_cmd = cmd.to_string();
+        // If the command is multi-line, it won't be echoed literally.
+        // instead, it will (probably) have each line begin with an `>`. so,
```
gjcolombo, line 627: Another in my growing list of reasons to give the `GuestOs` adapters more input into how shell commands actually get issued and get echoed. No need to do that in this PR though.

gjcolombo (approved):

phd-tests/framework/src/test_vm/mod.rs
hawkw: Yup, I also thought about that. For now, I just wanted to unbreak multiline, but we should probably have a nicer way of handling this...

PR #622, opened 2024-01-25
PHD: fix missing newlines in serial.log

hawkw:
PR #615 changed PHD's serial handling code writes serial output from the guest to a file, `<name_of_test_vm>.serial.log`. Output is written to the file every time a newline is received on the serial port. However, the way this is implemented is by special-casing the receipt of a `\n` character from the serial port to write out the current line and reset it rather than appending the character to the buffer. This means that the serial output is written to the file without any newlines, resulting in a pretty illegible log file.

This commit fixes that by writing a newline character after writing the buffered line. I've also added a `flush` call after writing out a full line to ensure that lines are visible in the log file immediately.


gjcolombo (approved):
Whoops. Thanks for the fix!

Issue #387, opened 2023-05-05
Preserve modified page masks on migration sources


gjcolombo:
If a migration source calls `VmmHdl::track_dirty_pages` and migration subsequently fails, the source must try to transfer the tracked pages again if it tries to migrate again, even if the guest hasn't touched those pages again.

hawkw:
Thinking through this, here are my notes so far:

- When we call [`VmmHdl::track_dirty_pages`](https://github.com/oxidecomputer/propolis/blob/master/lib/propolis/src/vmm/hdl.rs#L237), we pass in a 4k bitmap and a start guest physical address,, and `track_dirty_pages`[^ioctl] sets a 1 in the bitmap for each page in that range that is dirty.
 - Calling `track_dirty_pages` clears the dirty bit in the virtual page table for each dirty page in the range, so it won't still look dirty after the migration request, even if it fails.
 - Right now, [we only create a single 4k bitmap](https://github.com/oxidecomputer/propolis/blob/a13ebb9a982661b1e2592a3a736b621bf2264e3d/bin/propolis-server/src/lib/migrate/source.rs#L259) and [reuse it in a loop that constructs the memory transfer requests](https://github.com/oxidecomputer/propolis/blob/a13ebb9a982661b1e2592a3a736b621bf2264e3d/bin/propolis-server/src/lib/migrate/source.rs#L278-L282)[^reuse]
 - If we want to store the dirty page information across multiple migration attempts, the naive representation of that would look sort of like a `HashMap<RangeInclusive<GuestAddr>, [u8; 4096]>`, because the page range might include more than 4096 pages.
	 - There is probably a nicer representation for this than `HashMap` but this is a sketch
	 - This means we're potentially using a lot more memory than we are right now, since the number of 4k bitmaps now scales with the size of the guest's memory, rather than only ever having one.
 - We would then need to hold onto that data structure, and if the migration attempt fails, subsequent migration attempts would bit-or their dirtiness bitmaps with a new call to `track_dirty_pages` to determine which pages need to be xferred.
	 - If a failed migration can be retried multiple times if it fails again (which I assume it can), we need to continue hanging onto the dirty-pages data, and add any pages that became dirty since the last migration attempt.
 - This naive solution seems kind of inefficient (holding onto lots of 4k bitmaps) and there are probably some annoying implementation complexities with handling ranges that start in different places, etc.
[...]

PR #623, opened 2024-01-25
PHD: add tests for migration of running processes

hawkw:
In real life, it's likely that VM migration may occur while one or more
user processes are running on the guest. We don't have a lot of testing
for what happens to userspace processes during migration. Therefore,
I've added some tests that create a process on the guest, write a bunch
of data into variables in memory, suspend the process, migrate the
guest, and attempt to read the data back out of memory on the new host
and check that it's still correct.

In addition to a simple test that performs one migration, I've also
added tests for correctness in the face of a _failed_ migration attempt.
These tests force a migration attempt to fail while a process is
running, and then check that the process hasn't been messed up after a
successful second migration attempt.

In order facilitate testing migration after a failed migration,  this PR
introduces a test-only virtual device which can be configured to fail
during migration, either while exporting the device or while importing
it. The device can be configured to fail a set number of times, to allow
tests that force a migration failure and then perform a subsequent
migration that does not fail.

Since a virtual device whose only purpose is to break Propolis seems
like something that we don't want to create in production,
`propolis-server` will only permit the migration-failure test device to
be configured in the TOML config file, and it is *not* added to the
`InstanceSpec` public API. `propolis-server` also ignores any
migration-failure devices configured in the TOML if it was built with
the "omicron-build" feature, so we won't accidentally inject failures
into release builds.

The primary intention behind these tests is to ensure that all dirty
pages of the guest's memory are migrated correctly (see #387). With the
current approach to RAM transfer, we don't run into issues with this,
even in the case of a failed migration, because the `RamPushPrePause`
phase always offers *all* guest RAM, regardless of whether it's dirty.
However, if we change the RAM transfer behavior to only offer dirty
pages both in the `RamPushPrePause` and `RamPushPostPause` migration
phases, the migration-failure tests will now fail, because the failed
migration will have cleared dirty bits on any dirty guest pages, and we
now don't transfer the guest's memory correctly on the second migration.
These tests will allow us to verify that a fix for #387, where we no
longer offer all RAM in the `RamPushPrePause` phase, is working
correctly.


lib/propolis/src/hw/testdev.rs
```diff
@@ -0,0 +1,113 @@
...
+pub struct MigrationFailures {
+    pub exports: usize,
+    pub imports: usize,
+}
+
+#[derive(Clone, Default, Deserialize, Serialize)]
+struct MigrationFailurePayloadV1 {
+    /// If set, an attempt to import this device should fail.
+    fail_import: bool,
+}
```
gjcolombo, line 38: Putting this in the payload (if I'm understanding things correctly) means that the configuration passed to the migration *source* determines whether a failure gets injected into the *destination*. The code's correct, but this feels a bit unintuitive to me--if/when we add more failure injection capabilities I reckon most of them will need to be configured directly on the process we want to fail. Is there another advantage to having the source dictate this behavior that I'm overlooking?

phd-tests/tests/testdata/dirt.sh
```diff
@@ -0,0 +1,58 @@
...
+hello"
+    i=$(( i + 1 ))
+done
+
+echo "made dirt"
+
+# suspend this process and wait for it to be resumed before checking that the
+# data still exists.
+#
+# N.B.: posix sh doesn't have a suspend builtin, but we can make our own!
```
gjcolombo, line 25: No action required here, but we've really got to start running CI on something other than Alpine.

Maybe I'll make another project board...

phd-tests/tests/src/migrate.rs
```diff
@@ -84,6 +85,147 @@ mod from_base {
...
+
+        source.launch()?;
+        source.wait_to_boot()?;
+
+        mk_dirt(&source)?;
+
+        target.migrate_from(
+            &source,
+            Uuid::new_v4(),
+            Duration::from_secs(60),
```
gjcolombo, line 110: I just landed #627, so this can use `MigrationTimeout::default` after a merge/rebase.

gjcolombo (commented):
Good stuff, thanks for working on this! I have one question about the approach to injecting import failures, but beyond that I'm excited to get this in.

lib/propolis/src/hw/testdev.rs
hawkw: Hm, yeah, that's a good point. The reason that it works this way is mostly historical: when I wrote this code, I hadn't yet decided that the failure injection device would be configured only though the `config.toml` and not through `InstanceEnsure`, so it hadn't occurred to me that there would be a _way_ to configure the device in the migration target except by configuring it in the source and having that tell the target what to do. I can change this so that the import failure configuration is set in the target!

hawkw: Addressed this in 3705717b3ae3b8944631f6809783530cd49f3e9e !

lib/propolis/src/hw/testdev.rs
```diff
@@ -0,0 +1,122 @@
...
+use crate::inventory::Entity;
+use crate::migrate::*;
+
+use serde::{Deserialize, Serialize};
+
+use slog::info;
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Arc,
+};
```
pfmooney, line 20: style turbo-nit: elsewhere in the repo, imports have been grouped as:
1. std
2. crate-internal
3. external

lib/propolis/src/hw/testdev.rs
```diff
@@ -0,0 +1,122 @@
...
+                "you have no chance to survive, make your time".to_string(),
+            ));
+        }
+        Ok(())
+    }
+}
+
+impl Schema<'_> for MigrationFailurePayloadV1 {
+    fn id() -> SchemaId {
+        (MigrationFailureDevice::NAME, 1)
```
pfmooney, line 120: Since the schema IDs effectively become part of the Committed interface in migration, I'd hesitate to use a reference to the device name here - the latter can be changed on a whim (relatively speaking) whereas changing the former will result in compatibility issues.

Since testing devices are largely an implementation detail in propolis, I'd like to preserve maximum optionality for iteration there.  It'd be nice to make it easy to add or modify devices for test cases without having to worry too much about breaking cross-version tests.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -575,6 +577,59 @@ impl<'a> MachineInitializer<'a> {
         Ok(())
     }
 
+    pub fn initialize_test_devices(
```
pfmooney, line 580: Do we want this behind a `not(omicron-build)` guard?

hawkw: Currently, when compiled with `omicron-build`, this function will skip initializing test devices and print a warning stating that they are not available in release builds:

https://github.com/oxidecomputer/propolis/blob/3705717b3ae3b8944631f6809783530cd49f3e9e/bin/propolis-server/src/lib/initializer.rs#L588-L595

We could change this function to not be compiled in at all when `omicron-build` is set, but we would no longer detect the presence of unsupported test devices and warn about them in `omicron-build` builds. If you'd prefer that, I can change it --- I just thought it was nice to warn if a test device was found in the config while running an omicron build. What do you think?

lib/propolis/src/hw/testdev.rs
```diff
@@ -0,0 +1,121 @@
...
+    fn import(
+        &self,
+        mut offer: PayloadOffer,
+        _ctx: &MigrateCtx,
+    ) -> Result<(), MigrateStateError> {
+        let import_num = self.imports.fetch_add(1, Ordering::Relaxed);
+        let fail = import_num < self.fail.imports;
+        info!(
+            self.log,
+            "exporting device";
```
gjcolombo, line 102: s/exporting/importing

bin/propolis-server/src/lib/initializer.rs
gjcolombo: I like the idea of avoiding the entire call to the init function, since it helps to avoid the sort of bug where we add another test device to `initialize_test_devices` but forget to add the `cfg` check. Another way to make this observable is to make the caller do the logging:

```rust
if cfg!(not(feature = "omicron-build")) {
    init.initialize_test_devices(...)?;
} else {
    info!("omicron-build was configured, won't enable test devices");
}
```

The latter call could be upgraded to something like `warn_if_test_devices_present`, but if I imagine being the audience of this message--I've added a test device to the Omicron package TOML, I'm testing with that package, it didn't work, so now I'm looking at the Propolis logs--I think I'd be satisfied with a log line during init that reminds me that Propolis doesn't initialize these for prod builds.

gjcolombo (approved):

bin/propolis-server/src/lib/initializer.rs
hawkw: @gjcolombo I'm fine with moving this to the caller, if you prefer that. My thinking was that it was kinda nice to only log about test devices if we actually encountered such a device in the config, rather than always logging that we won't enable them --- the information is really only relevant if there are test devices in the config, and if there aren't any, there's no reason to log it. But, I also see the value in compiling out the test device initialization entirely in prod builds. So, I'm fine with always logging.

hawkw: changed in 58ec187d81f77e1980edbda99645a49931fa2bfb

Issue #625, opened 2024-01-26
phd: try to revise Windows adapters not to assume an initial reboot is required


gjcolombo:
The guest login sequences in PHD's WS2022 adapter (and the forthcoming WS2019 adapter) assume they're going to be executing on a sysprepped image that has to be [specialized](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/how-configuration-passes-work?view=windows-11) such that they will reboot at least once during the logon sequence. One nasty consequence of this is that calls to `TestVm::wait_to_boot` work only once per disk; after that the initial reboot is no longer required and the logon sequence no longer works.

Working around this in PHD is going to be pretty ugly (it essentially requires PHD disks to know whether they've been booted from before). It would be much nicer to see if we can arrange our Windows images so that this reboot is not required.

This issue is nominally tracking PHD work, but that work is really minimal; the tricky part is figuring out how to create a test image that has the properties we want.

PR #627, opened 2024-01-26
phd: add Windows Server 2019 guest adapter

gjcolombo:
The biggest challenge with this adapter is the way Windows Server 2019's serial console-based command prompt behaves. Server 2022 acts like most Linux guests: it sends text and ASCII control characters out the serial port and allows the recipient to decide how to drive a terminal from those characters. Server 2019, on the other hand, assumes it is driving a VT100-compatible terminal: it appears to maintain an internal conception of what's on the screen, decides in response to console input and output how to update the screen, and then sends just enough command sequences out the serial port to make the recipient terminal's state match the desired state.

To make this work, add a few serial console features:

- Add a `Vt80x24` serial console backend that renders to an internal screen and defines some basic semantics for using that screen as a wait buffer.
- Add some affordances to allow a VM's serial console buffering scheme to change at runtime. This is needed because WS2019 doesn't start running in VT100- compatible mode until its command prompt actually launches.
- Allow a migration target VM to inherit the serial console task from its source VM. Use the Propolis client serial helper to transfer connections between the two.
- Allow callers to configure a serial task to insert a delay between each typed character. This appears to be needed (unfortunately) to avoid losing keystrokes to in-guest keyboard debounce logic.

Amend the `GuestOS` trait to allow guests to transform a serial console command before issuing it to the guest. This is used to allow WS2019 guests to prepend a command to clear and redraw the terminal before each command issued to the guest. This, in turn, lets the VT100 backend implement slightly simpler logic to return command outputs to tests--the backend can return everything on the screen preceding the last (and only) command prompt, and tests can assume that the only text they get back is command output (and not e.g. leftover output from prior commands).

Finally, tweak the logic for specifying migration timeouts so that they scale with guest memory size by default. The default guest memory size of 512 MiB often results in tremendously slow Windows guests; just bumping up the size without adjusting these timeouts leads to migration timeouts.

Tested with Alpine, Debian 11, Windows Server 2019, and Windows Server 2022 guests. The Windows variations fail the `lspci` lifecycle test because they don't support multiple calls to `wait_to_boot`, but otherwise pass our existing tests.

Takes another step toward #601.


gjcolombo:
This is a big PR that contains some material improvements to PHD writ large, so I wanted to go ahead and get it up for review, but there are several bits of follow-on work still to come:

- #625 will be needed before any start-stop lifecycle tests will pass with Windows guests.
- The "amend command" function in `trait GuestOs` works for now but basically pushes `TestVm::run_shell_command` to the limits of its complexity. We should refactor shell command execution so that it passes through the guest adapters, which can then decide how they want to handle prepend/append, multi-line commands, hitting the Enter key, deciding what should/shouldn't be expected on the console afterwards, etc.
- The new terminal backend doesn't do any logging yet, and its mechanism for consuming text from the buffer isn't very sophisticated. I have some vague ideas for how to do better here (e.g. detect that some action has moved the cursor to a different line and then log the line that was previously being edited) but will probably need to look at something else for a bit before I can get a toehold on them.

Will also fix clippy in my next push.

phd-tests/framework/src/guest_os/mod.rs
```diff
@@ -12,15 +12,26 @@ use serde::{Deserialize, Serialize};
...
     /// Write the specified string as a command to the guest serial console.
     WriteStr(&'static str),
+
+    /// Change the serial console buffering discipline to the supplied
+    /// discipline.
+    ChangeSerialConsoleBuffer(crate::serial::BufferKind),
+
+    /// Set a delay between writing individual bytes to the guest serial console
+    /// to avoid keyboard debouncing logic in guests.
+    SetSerialByteWriteDelay(std::time::Duration),
```
hawkw, line 34: *touches ground* something terrible happened here...i can feel it..,

phd-tests/framework/src/guest_os/mod.rs
```diff
@@ -35,6 +46,16 @@ pub(super) trait GuestOs {
...
+
+    /// Some guests need to amend incoming shell commands from tests in order to
+    /// get output to display on the serial console in a way those guests can
+    /// accept (e.g. by clearing the screen immediately before running each
+    /// command). If a guest requires this kind of amendment for a given
+    /// command, this function returns `Some(amended_command)`. Otherwise it
+    /// returns `None`.
+    fn amend_shell_command(&self, _cmd: &str) -> Option<String> {
+        None
+    }
```
hawkw, line 58: nit, take it or leave it: why not just have this be
```suggestion
    fn amend_shell_command<'a>(&self, cmd: &'a str) -> Cow<'a, str> {
        Cow::Borrowed(cmd)
    }
```

rather than having to bother with an `Option`?

phd-tests/framework/src/guest_os/windows.rs
```diff
@@ -0,0 +1,95 @@
...
+///   documentation](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/generalize?view=windows-11)
+///   for more details.
+/// - Cygwin is installed to C:\cygwin and can be launched by invoking
+///   C:\cygwin\cygwin.bat.
+/// - The local administrator account is enabled with password `0xide#1Fan`.
+pub(super) fn get_login_sequence_for(guest: GuestOsKind) -> CommandSequence {
+    assert!(matches!(
+        guest,
+        GuestOsKind::WindowsServer2019 | GuestOsKind::WindowsServer2022
+    ));
```
hawkw, line 27: nitpicky and very much not important: we _could_ avoid having to do this if we had a 
```rust
enum GuestOsKind {
    Windows(Windows), 
    // ...
}

enum Windows {
    Server2019,
    Server2022,
}
```
but i'm not sure if it's worth bothering that or not, up to you.

phd-tests/framework/src/guest_os/windows.rs
```diff
@@ -0,0 +1,95 @@
...
+    // only switch buffering modes after entering credentials.
+    if let GuestOsKind::WindowsServer2019 = guest {
+        commands.extend([
+            CommandSequenceEntry::ChangeSerialConsoleBuffer(
+                crate::serial::BufferKind::Vt80x24,
+            ),
+            // Server 2019 also likes to debounce keystrokes, so set a small
+            // delay between characters to try to avoid this. (This value was
+            // chosen by experimentation; there doesn't seem to be a guest
+            // setting that controls this interval.)
```
hawkw, line 71: thanks for writing this down!

phd-tests/framework/src/guest_os/windows_server_2019.rs
```diff
@@ -0,0 +1,45 @@
...
+        // Use Cygwin bash's `reset` instead of `clear` or `cls` to try to force
+        // Windows to clear and redraw the entire terminal before displaying any
+        // command output. Without this, Windows sometimes reprints a new
+        // command prompt to its internal screen buffer before re-rendering
+        // anything to the terminal; when this happens, it doesn't re-send the
+        // new command prompt, since it's "already there" on the output terminal
+        // (even though it may have been cleared from the match buffer).
+        let mut new_cmd = "reset; ".to_owned();
+        new_cmd.push_str(cmd);
+        Some(new_cmd)
```
hawkw, line 43: nitpick, TIOLI: could just be
```suggestion
        Some(format!("reset; {cmd}")
```

phd-tests/framework/src/serial/raw_buffer.rs
```diff
@@ -82,62 +106,37 @@ impl Inner {
...
+            match action {
+                Action::Print(c) => self.push_character(c),
+                Action::PrintString(s) => {
+                    self.push_str(&s);
+                }
+                Action::Control(ControlCode::LineFeed) => {
+                    self.push_character('\n')
+                }
+                _ => {}
+            }
```
hawkw, line 123: nit, TIOLI: per my reading of the `termwiz` docs, we can save the `Vec` allocation if we wrotee this, instead:
```suggestion
        self.parser.parse(bytes, |action| {
            match action {
                Action::Print(c) => self.push_character(c),
                Action::PrintString(s) => {
                    self.push_str(&s);
                }
                Action::Control(ControlCode::LineFeed) => {
                    self.push_character('\n')
                }
                _ => {}
            }
        });
```

phd-tests/framework/src/serial/raw_buffer.rs
```diff
@@ -82,62 +106,37 @@ impl Inner {
...
+        for action in actions {
+            match action {
+                Action::Print(c) => self.push_character(c),
+                Action::PrintString(s) => {
+                    self.push_str(&s);
+                }
+                Action::Control(ControlCode::LineFeed) => {
+                    self.push_character('\n')
+                }
+                _ => {}
```
hawkw, line 122: nit, not a big deal: do you think it's worthwhile to log other control characters, instead of eating them?  i could be convinced either way.

phd-tests/framework/src/serial/vt80x24.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        if let Some(waiter) = self.waiter.take() {
+            self.satisfy_or_set_wait(waiter);
+        }
+    }
+
+    /// Attempts to satisfy the wait described by `waiter`. If the wait is not
+    /// immediately satisfiable, stores `waiter` to try again later.
+    fn satisfy_or_set_wait(&mut self, waiter: OutputWaiter) {
+        assert!(self.waiter.is_none());
```
hawkw, line 101: maybe we should add an assertion like this:
```suggestion
        assert!(self.waiter.is_none());
        let too_long = waiter.wanted.lines().find(|line| line.len() > 80);
        assert_eq!(too_long, None, "a wait request that contains a line longer than 80 columns will never be satisfied");
```
 
i'd prefer my test to get a loud failure telling me what i did wrong rather than a wait that silently doesn't complete for unclear reasons...

phd-tests/framework/src/serial/vt80x24.rs
```diff
@@ -0,0 +1,174 @@
...
+            let (old_col, old_row) = self.surface.cursor_position();
+            self.surface.add_change(make_absolute_cursor_position(0, 0));
+            self.surface.add_change(Change::Text(" ".repeat(char_count)));
+            let seq = self
+                .surface
+                .add_change(make_absolute_cursor_position(old_col, old_row));
+            self.surface.flush_changes_older_than(seq);
+
+            // Remove the match string from the match contents and push
+            // everything else back to the listener.
```
hawkw, line 143: thanks for the comments here! i don't think i would have had any understanding of what was going on here otherwise :)

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -56,6 +59,34 @@ pub enum VmStateError {
     InstanceAlreadyEnsured,
 }
 
+/// Specifies the timeout to apply to an attempt to migrate.
+pub enum MigrationTimeout {
+    /// Time out after the specified duration.
+    Explicit(std::time::Duration),
+
+    /// Allow MIGRATION_SECS_PER_GUEST_GIB seconds per GiB of guest memory.
+    InferFromMemorySize,
```
hawkw, line 68: <3

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -396,30 +440,43 @@ impl TestVm {
         &mut self,
         source: &Self,
         migration_id: Uuid,
-        timeout_duration: Duration,
+        timeout: MigrationTimeout,
```
hawkw, line 443: nit, take it or leave it: it might be nice to add an `impl From<Duration> for MigrationTimeout` and change this to 
```suggestion
        timeout: impl Into<MigrationTimeout>,
```
so that tests that want to set explicit timeouts can just call this with a `Duration`, rather than having to import `MigrationTimeout` and construct one.

not a big deal, just a tiny API quality-of-life improvement :)

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -621,35 +690,41 @@ impl TestVm {
...
-        let mut out = self
-            .wait_for_serial_output(
-                self.guest_os.get_shell_prompt(),
-                Duration::from_secs(300),
-            )
-            .context("Failed to wait for prompt running shell command")?;
-        out.truncate(out.trim_end().len());
-        Ok(out)
+        let to_send = to_send.trim_end().replace('\n', "\n> ");
+        self.send_serial_str(&to_send)?;
```
hawkw, line 700: wait...why does this work? do Unix shells mind if you send them the `> ` character yourself in multiline strings?

hawkw (approved):
Overall, this looks great --- thanks for the detailed comments in the VT100 code, I don't think I could have followed what's going on there otherwise. 

I had some minor nitpicks to comment on, but no hard blockers from me!

phd-tests/framework/src/test_vm/mod.rs
gjcolombo: It doesn't! I butchered this and didn't notice because none of the currently checked-in tests depend on it. Should be fixed now. I checked in the test that I used to verify this (it also works on WS2022), though I could be convinced to omit it as well, on the theory that the test coverage juice may not be worth the test runtime squeeze.

gjcolombo (commented):
Thanks for the detailed feedback! "How do I Rust better"-type stuff is always especially useful for me, since my Rust skills are still, uh, getting better every day.

phd-tests/framework/src/guest_os/windows.rs
hawkw: yeah, that was what i was imagining my suggestion would grow into, eventually. i’m fine with saving it for a later PR!

phd-tests/framework/src/serial/raw_buffer.rs
hawkw: oh, ugh. i think i know a way to work around that, gimme a sec?

hawkw:
> Thanks for the detailed feedback! "How do I Rust better"-type stuff is always especially useful for me, since my Rust skills are still, uh, getting better every day.

Great, I’m very happy to leave that kind of style and API-niceness suggestions, as long as you like receiving them! And do let me know if it ever gets annoying — I’ll be the first to admit that this kind of nitpicking is not exactly the most important thing in the world!

phd-tests/framework/src/test_vm/mod.rs
hawkw: ahh, yup, i had forgotten that my hitherto-unmerged tests in #623  were the only tests that actually _use_ a multi line shell command, so i had mistakenly assumed that CI passing meant that it still worked, somehow! i’m glad we caught this!

phd-tests/tests/src/smoke.rs
hawkw, line 1: i’m very much in favor of checking in tests like this that are primarily intended to assert that the test framework itself is working correctly! but, i wonder if we ought to have some convention (maybe a module or naming scheme) to make it clear that the test is at least partially intended to make sure the framework itself is reasonable? one of the main signals that you get from a test like this failing is “hey, the test framework itself is busted, so _other_ failures might indicate a bug in the framework rather than the code under test!”


if we wanted to be _really_ fancy someday, we could maybe even mark such tests with a special `TestOutcome` (maybe via an argument to `#[phd_testcase]`?) and if they fail, we could have the framework loudly indicate to the user “HEY, I’M BUSTED, OTHER TEST RESULTS I GIVE YOU MAY BE SUSPECT!”

on the other hand, that may not really be worth the effort. i dunno…

gjcolombo: Yeah, that's a good call. 4acebd98 moves the test to a new `framework` module. Doing this in a module has the nice property that this test and all its future siblings will be nicely grouped when referenced with the `--include-filter` and `--exclude-filter` switches, and its name will hopefully suggest that a failure in that module suggests that things are in an especially bad way...

phd-tests/framework/src/serial/raw_buffer.rs
hawkw: @gjcolombo i've opened PR #631 with the way I would refactor this code to allow using `Parser::parse` rather than calling `parse_as_vec`...but then, when I came back to post this, I re-read your comment and saw you mention that you thought the current approach was nicer than having to have a separate `Inner` struct, which I hadn't remembered you saying earlier. So, we can just close that PR --- my bad.

PR #630, opened 2024-01-29
PHD: fix `cargo xtask phd` tidy not doing anything

hawkw:
`cargo xtask phd` attempts to delete test temporary directories that are older than one day. However, because the code for seeing how old a temp dir is accidentally used the dir's last accessed time, checking how old the dir is makes it look like it's always less than a day old, so stuff never gets deleted. I've changed this code to use the last modified time instead.

In addition, I've added a new `cargo xtask phd tidy` subcommand that *just* deletes old temp dirs, without building `phd-runner` or `propolis-server`. This is a bit quicker if all you want is to free up some space.

### DIFF
--- a/xtask/src/task_phd.rs
+++ b/xtask/src/task_phd.rs
@@ -79,7 +79,13 @@ pub(crate) struct RunArgs {
 
     /// Arguments to pass to `phd-runner run`.
     ///
-    /// Use `cargo xtask phd runner-help` for details on the arguments passed to `phd-runner`.
+    /// If the `--propolis-server-cmd`, `--crucible-downstairs-commit`,
+    /// `--base-propolis-branch`, `--artifact-toml-path`, or
+    /// `--artifact-directory` arguments are passed here, they will override
+    /// `cargo xtask phd`'s default values for those arguments.
+    ///
+    /// Use `cargo xtask phd runner-help` for details on the arguments passed to
+    /// `phd-runner`.
     #[clap(trailing_var_arg = true)]
     phd_args: Vec<String>,
 }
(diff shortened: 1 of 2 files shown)


xtask/src/task_phd.rs
```diff
@@ -41,9 +41,25 @@ macro_rules! cargo_warn {
...
+        .context("Failed to run cargo metadata")?;
+    let phd_dir = relativize(&meta.target_directory).join("phd");
+
+    let mut tmp_dir = phd_dir.join("tmp");
+    let now = time::SystemTime::now();
+
+    if arg_iter.peek() == Some(&"tidy") {
+        delete_old_tmps(tmp_dir, now)?;
+        cargo_log!("Tidying up", "old temporary directories...");
+        std::process::exit(0);
```
pfmooney, line 58: Shouldn't this just `return Ok(())`?

xtask/src/task_phd.rs
```diff
@@ -41,9 +41,25 @@ macro_rules! cargo_warn {
...
+    let meta = cargo_metadata::MetadataCommand::new()
+        .no_deps()
+        .exec()
+        .context("Failed to run cargo metadata")?;
+    let phd_dir = relativize(&meta.target_directory).join("phd");
+
+    let mut tmp_dir = phd_dir.join("tmp");
+    let now = time::SystemTime::now();
+
+    if arg_iter.peek() == Some(&"tidy") {
```
pfmooney, line 55: Should this command and a `--no-tidy` (rather than just the implied ENV option) option be moved into the arg parsing logic, especially so it can be visible via --help and such?

hawkw: Yeah, that's not a bad idea. We probably ought to do that for the `--release` flag that builds `propolis-server` in release mode, as well.

hawkw: Okay, 1a5e641fa4efdd5ab80a63319005cc06e40bdd53 changes the PHD subcommand to handle its own arguments using `clap`. This does have the minor downside that we no longer forward `cargo xtask phd help` directly to `phd-runner`, so I've added an additional `runner-help` subcommand for printing `phd-runner`'s help, which seems like an acceptable compromise...

gjcolombo (approved):

PR #634, opened 2024-01-31
PHD: improve ctrl-c handling

gjcolombo:
(N.B. Depends on #633.)

Install a SIGINT handler in the PHD runner and use it to broadcast to all test execution tasks that this signal was received. Add a `select!` to the tasks to preempt further execution of test cases when this happens. Dropping the test case futures also drops all the VMs and disks they create, giving the runner a chance to clean up kernel VMMs and disk backing files before exiting. (If one SIGINT is handled and a second is received, exit immediately to allow the runner to be killed even if a test case hasn't reached an `await` point.)

To ensure that Propolis processes get a chance to stop gracefully, add a pre-`exec` hook so that the framework's Propolis servers ignore SIGINT. This is needed so that hitting Ctrl-C to interrupt a foreground runner doesn't take out the Propolis servers in the same process group before `TestVm::drop` gets a chance to clean them up.

Tested with runs of Debian 11 guests, both running to completion (all tests still pass) and interrupted at various points midway (verified that kernel VMMs and disks get cleaned up as expected when they didn't before).

### DIFF
--- a/phd-tests/framework/src/test_vm/server.rs
+++ b/phd-tests/framework/src/test_vm/server.rs
@@ -4,7 +4,7 @@
 
 //! Routines and data structures for working with Propolis server processes.
 
-use std::{fmt::Debug, net::SocketAddrV4};
+use std::{fmt::Debug, net::SocketAddrV4, os::unix::process::CommandExt};
 
 use anyhow::Result;
 use camino::{Utf8Path, Utf8PathBuf};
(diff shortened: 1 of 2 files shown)


phd-tests/runner/src/execute.rs
```diff
@@ -76,32 +78,54 @@ pub async fn run_tests_with_ctx(
...
+                outcome = tc.run(test_ctx.as_ref()) => outcome
+            }
+        })
+        .await
+        {
+            Ok(outcome) => outcome,
+            Err(_) => TestOutcome::Failed(Some(
+                "test task panicked, see test logs".to_string(),
+            )),
+        };
```
hawkw, line 128: nit, take it or leave it: this could be:
```suggestion
        .await
        .unwrap_or_else(||  {
            TestOutcome::Failed(Some("test task panicked, see test logs".to_string()))
        });
```

phd-tests/runner/src/execute.rs
```diff
@@ -141,3 +165,35 @@ pub async fn run_tests_with_ctx(
...
+    let (sigint_tx, sigint_rx) = watch::channel(false);
+    tokio::spawn(async move {
+        loop {
+            sigint.recv().await;
+
+            // If a signal was previously dispatched to the channel, exit
+            // immediately with the customary SIGINT exit code (130 is 128 +
+            // SIGINT). This allows users to interrupt tests even if they aren't
+            // at an await point (at the cost of not having destructors run).
+            if *sigint_tx.borrow() {
```
hawkw, line 185: this is nice to add, good call!

phd-tests/framework/src/test_vm/server.rs
```diff
@@ -62,18 +62,52 @@ impl PropolisServer {
...
+                        vnc_addr.to_string().as_str(),
+                    ])
+                    .stdout(server_stdout)
+                    .stderr(server_stderr)
+                    .pre_exec(move || {
+                        libc::signal(libc::SIGINT, libc::SIG_IGN);
+                        Ok(())
+                    })
+                    .spawn()?
+            },
```
hawkw, line 110: style nit, definitely not a blocker: hmm, it's a little bit unfortunate that this whole thing is wrapped in `unsafe` when it's really _just_ the call to `Command::pre_exec` that's unsafe --- it's a bit unclear to the reader what the unsafe operation is here. what do you think about moving this out to something like this: 

```rust
let server = {
    let mut cmd = std::process::Command::new("pfexec");
    cmd.args([
            server_path.as_str(),
            "run",
            config_toml_path.as_str(),
            server_addr.to_string().as_str(),
            vnc_addr.to_string().as_str(),
        ])
        .stdout(server_stdout)
        .stderr(server_stderr);
    unsafe {
        cmd.pre_exec(move || {
            libc::signal(libc::SIGINT, libc::SIG_IGN);
            Ok(())
        });
    }
    cmd.spawn()?
};
let server = PropolisServer {
    server,
    address: server_addr,
};
```

hawkw (approved):
this looks good to me! i have one very small style nit to pick, but it's not as blocker.

phd-tests/framework/src/test_vm/server.rs
gjcolombo: I like it. Done in 7f547237.

PR #636, opened 2024-02-02
PHD: assume specialized Windows images

### LINKED ISSUE #625: phd: try to revise Windows adapters not to assume an initial reboot is required

gjcolombo:
Change the Windows guest adapters to assume they're operating on a fully- specialized, post-OOBE guest image. This fixes tests like the `lspci` lifecycle test that stop and start a VM and assume that it will go through the same boot sequence on its initial and subsequent attempts to start.

Also fix up the PHD README to describe the WS2019 adapter (missed this in the change that added the adapter).

Tested with WS2019 and WS2022 guests.

Fixes #625.

### DIFF
--- a/phd-tests/README.md
+++ b/phd-tests/README.md
@@ -217,6 +217,7 @@ encoded into the logon sequences for each adapter and reproduced below:
 | Alpine Linux        | `root`          |              |
 | Debian 11 (nocloud) | `root`          |              |
 | Ubuntu 20.04        | `ubuntu`        | `1!Passw0rd` |
+| Windows Server 2019 | `Administrator` | `0xide#1Fan` |
 | Windows Server 2022 | `Administrator` | `0xide#1Fan` |
 
 If you add a custom image to your artifact file, you must make sure either to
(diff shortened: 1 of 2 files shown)


hawkw (approved):
looks good to me! didn't review this one earlier because i assumed it depended on the async changes. :)

gjcolombo:
Thanks! No worries, this one stands alone (it depends on having a slightly different guest image config than the one I had been using, but I think I have one worked out now; just need to decide on the right place to check it in).

PR #638, opened 2024-02-05
standalone: Use tar for snapshot/restore

pfmooney:
While debugging unrelated bits, I wanted to be able to more easily pick apart the contents of a snapshot from propolis-standalone.  It seemed preferable to switch to a widely used format, rather than continuing to build out a bespoke inspection tool.


bin/propolis-standalone/src/snapshot.rs
```diff
@@ -55,22 +49,40 @@ struct SnapshotDevice {
...
+    let mut header = tar::Header::new_gnu();
+    header.set_mode(0o444);
+    header.set_mtime(now.as_secs());
+
+    info!(log, "Serializing VM config");
+    {
+        let config_bytes = toml::to_string(config)?.into_bytes();
+        header.set_size(config_bytes.len() as u64);
+        header.set_cksum();
+        builder.append_data(&mut header, CONFIG_NAME, &config_bytes[..])?;
```
luqmana, line 83: You can skip the `header.set_cksum()` call since [`Builder::append_data`](https://docs.rs/tar/latest/tar/struct.Builder.html#method.append_data) will do it for you:
>The checksum for the header will be automatically updated via the `set_cksum` method after setting the path.

bin/propolis-standalone/src/snapshot.rs
```diff
@@ -81,59 +93,78 @@ pub(crate) async fn save(
...
+    {
+        let mut dir_header = tar::Header::new_gnu();
+        dir_header.set_entry_type(tar::EntryType::Directory);
+        dir_header.set_mode(0o555);
+        dir_header.set_mtime(now.as_secs());
+        dir_header.set_size(0);
+        dir_header.set_cksum();
+
+        builder.append_dir(&mut dir_header, format!("{DEVICE_DIR}/"))?;
+        builder.append_dir(&mut dir_header, format!("{MEMORY_DIR}/"))?;
```
luqmana, line 115: Can skip creating format strings
```suggestion
        builder.append_dir(&mut dir_header, DEVICE_DIR)?;
        builder.append_dir(&mut dir_header, MEMORY_DIR)?;
```

bin/propolis-standalone/src/snapshot.rs
```diff
@@ -475,3 +481,140 @@ fn import_global(hdl: &VmmHdl, state: &VmGlobalState) -> std::io::Result<()> {
...
+        self.0.append_data(header, path, io::empty())
+    }
+
+    fn rawfd(&mut self) -> RawFd {
+        self.0.get_mut().as_raw_fd()
+    }
+    fn inner_builder(&mut self) -> &mut tar::Builder<File> {
+        &mut self.0
+    }
+    fn info_file(self) -> io::Result<File> {
```
luqmana, line 534: Typo? `into_file`

luqmana (approved):
A couple small nits but looks good otherwise. Using standard formats/tooling definitely is nice here! I can totally see natural extensions to this like adding compression support.

PR #639, opened 2024-02-06
Remove Inventory mechanism

pfmooney:
The Inventory abstraction in propolis-lib has outlived its necessity. Created initially to keep devices in scope after VM initialization, it was somewhat cumbersome to deal with, and was providing more friction that value when it came to accessing the contained Entity bits.  With its removal, the Instance abstraction can go as well, since it was simply holding the Machine and the Inventory together.  In this new world, propolis-lib consumers are left to hold Lifecycle (formerly Entity) trait objects in whatever way seems appropriate to them.

With the Inventory gone, the parent-child relationships possible through the former Entity trait are also on the chopping block.  The 440FX/PIIX4 chipset devices have been rearranged to more accurately reflect what is present in "real" devices.


hawkw:
Hmm, it looks like the "migration from base" PHD tests [have actually caught an incompatibility](https://buildomat.eng.oxide.computer/wg/0/artefact/01HP048PY1C3RV9D9SPQBJ7T9T/eNALVhkg3IpsEXK5ZowKAONFey061ceZFCwTaWoD61rQ4EgO/01HP049NBNKK9PGZH6GC0Z8QQT/01HP08SPW8H5ER9Q51KX7WK0Z3/phd-runner.log?format=x-bunyan#L1454)!

pfmooney:
> Hmm, it looks like the "migration from base" PHD tests [have actually caught an incompatibility](https://buildomat.eng.oxide.computer/wg/0/artefact/01HP048PY1C3RV9D9SPQBJ7T9T/eNALVhkg3IpsEXK5ZowKAONFey061ceZFCwTaWoD61rQ4EgO/01HP049NBNKK9PGZH6GC0Z8QQT/01HP08SPW8H5ER9Q51KX7WK0Z3/phd-runner.log?format=x-bunyan#L1454)!

Indeed they have!  That failure is expected, given the device shuffling.

hawkw:
> > Hmm, it looks like the "migration from base" PHD tests [have actually caught an incompatibility](https://buildomat.eng.oxide.computer/wg/0/artefact/01HP048PY1C3RV9D9SPQBJ7T9T/eNALVhkg3IpsEXK5ZowKAONFey061ceZFCwTaWoD61rQ4EgO/01HP049NBNKK9PGZH6GC0Z8QQT/01HP08SPW8H5ER9Q51KX7WK0Z3/phd-runner.log?format=x-bunyan#L1454)!
> 
> Indeed they have! That failure is expected, given the device shuffling.

Ah, okay. It's good to know the tests are working, in any case. :)

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -186,7 +192,10 @@ impl<'a> MachineInitializer<'a> {
             );
             pci_builder.add_bridge(desc)?;
         }
-        let pci_topology = pci_builder.finish(self.inv, self.machine)?;
+        let pci::topology::FinishedTopology {
+            topology: pci_topology,
+            bridges: _bridges,
```
gjcolombo, line 197: The old `pci_builder` used to register the created bridges into the inventory. To match its behavior I think the bridges should be added to `devices` here instead of being dropped (they already have a `Lifecycle` impl).

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -98,46 +76,65 @@ pub fn build_instance(
...
 }
 
 pub struct MachineInitializer<'a> {
-    log: slog::Logger,
-    machine: &'a Machine,
-    inv: &'a Inventory,
-    spec: &'a InstanceSpecV0,
-    producer_registry: Option<ProducerRegistry>,
+    pub(crate) log: slog::Logger,
+    pub(crate) machine: &'a Machine,
```
gjcolombo, line 105: Not necessary for this PR, just a forward-looking question: how challenging do you think it'd be to have a fake Machine? I'm thinking about what it'd take to write unit tests of the form "I added this initializer function and expect that when I call it with spec X the appropriate things will show up in maps A, B, and C."

gjcolombo (approved):
I think this direction makes sense and like the new arrangement of the chipset devices. Couple of little things but otherwise LGTM.

bin/propolis-server/src/lib/initializer.rs
pfmooney: I think that as the core abstractions become more polished, such a thing will become easier.  `Machine` is obviously doing much less (as of this change) than `Instance` -- mostly just holding references to top-level items.  It seems likely that with a bit more clean-up (particularly around memory maps, and mediating access to the VmmHdl) that `Machine` could be discarded as well.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -498,15 +543,15 @@ impl<'a> MachineInitializer<'a> {
                             slog::o!("component" => format!("nvme-{}", name)),
                         ),
                     );
-                    let id =
-                        self.inv.register_instance(&nvme, bdf.to_string())?;
-                    let _ = self.inv.register_child(child, id).unwrap();
+                    self.devices
+                        .insert(format!("pci-nvme-{}", bdf), nvme.clone());
```
hawkw, line 547: tiny, unimportant nit: this could be
```suggestion
                        .insert(format!("pci-nvme-{bdf}"), nvme.clone());
```

hawkw (approved):

lib/propolis/src/hw/pci/mod.rs
```diff
@@ -150,11 +171,16 @@ impl Bdf {
...
-        match (bnum, loc) {
-            (Some(b), Some(l)) => Some(Self { bus: b, location: l }),
-            _ => None,
+        if let Some(location) = BusLocation::new(dev, func) {
+            Some(Self { bus: BusNum::new(bus), location })
+        } else {
+            None
+        }
+    }
+    pub const fn new_unchecked(bus: u8, dev: u8, func: u8) -> Self {
```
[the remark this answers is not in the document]
pfmooney: If it's alright, I'm instead going to document the panic behavior.

lib/propolis/src/lifecycle.rs
```diff
@@ -0,0 +1,89 @@
...
+/// - [Lifecycle::resume]
+/// - [Lifecycle::paused]
+/// - [Lifecycle::reset]
+/// - [Lifecycle::halt]
+/// - [Lifecycle::migrate]
+pub trait Lifecycle: Send + Sync + 'static {
+    /// Unique name for devices for a given type
+    ///
+    /// Intended to be `const` once stabilized for trait functions.
+    fn type_name(&self) -> &'static str;
```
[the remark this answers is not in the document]
pfmooney: This is an aspect of the `Entity`/`Lifecycle` system I'd like to revisit in the somewhat near future -- possibly for outright removal.

We're working toward a future state where devices can be instantiated directly from migration payloads (rather than created and then hydrated w/ the migration state).  Depending on the details of that, we might nix the `type_name` stuff entirely.

With all that in mind, I'm incline to leave it for now.

bin/propolis-server/src/lib/initializer.rs
pfmooney: Done.

lib/propolis/src/hw/chipset/i440fx.rs
```diff
@@ -366,17 +171,100 @@ impl Piix4HostBridge {
...
+            pci_topology,
+            pci_cfg,
+            pcie_cfg,
+
+            pin_power,
+            pin_reset,
+        })
+    }
+
+    pub fn attach(self: &Arc<Self>, machine: &Machine) {
```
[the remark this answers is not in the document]
pfmooney: Definitely a "Doc it hurts when I do this.  Then don't do that" situation for now.  I'd like to see a better abstraction around MMIO/PIO attachments at some point, which would make it easier to avoid such mistakes.

lib/propolis/src/hw/pci/mod.rs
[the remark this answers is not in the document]
hawkw: That's also fine with me!

PR #640, opened 2024-02-08
PHD: Use ZFS clones for file-backed disks

gjcolombo:
(N.B. Stacked on #634.) 

When creating a file-backed disk from an artifact, create a ZFS snapshot and of the dataset containing the artifact, then create and mount a clone of the snapshot and back the disk from the clone. This is much, much less expensive than copying the underlying file and markedly improves performance for test runs with disks of nontrivial size. (On my local test machine, a full test run with a 2 GiB Debian nocloud image took 688 seconds before this change and 610 seconds after, about a 10% speedup.)

This is unfortunately not likely to move the needle much on CI times because the Alpine test image CI uses is very small. However, this should make it much easier to switch to non-Alpine CI guest(s) without needlessly bloating CI times.

Tested with full runs against Debian 11 and Alpine guests.

### DIFF
--- a/phd-tests/framework/src/disk/file.rs
+++ b/phd-tests/framework/src/disk/file.rs
@@ -30,12 +30,14 @@ impl BackingFile {
         artifact_path: &Utf8Path,
         data_dir: &Utf8Path,
     ) -> anyhow::Result<Self> {
-        if let Ok(file) = ZfsClonedFile::create_from_path(artifact_path) {
-            return Ok(Self::Zfs(file));
+        match ZfsClonedFile::create_from_path(artifact_path) {
+            Ok(file) => return Ok(Self::Zfs(file)),
+            Err(error) => warn!(
+                %artifact_path,
+                %error,
+                "failed to make ZFS clone of backing artifact, will copy it"
+            ),
         }
-
-        warn!(%artifact_path,
-              "failed to make ZFS clone of backing artifact, will copy it");
 
         let mut disk_path = data_dir.to_path_buf();
         disk_path.push(format!("{}.phd_disk", Uuid::new_v4()));
--- a/phd-tests/framework/src/disk/file.rs
+++ b/phd-tests/framework/src/disk/file.rs
@@ -4,13 +4,10 @@
 
 //! Abstractions for disks with a raw file backend.
 
-use std::path::{Path, PathBuf};
-
+use camino::Utf8Path;
 use propolis_client::types::{FileStorageBackend, StorageBackendV0};
-use tracing::{error, info};
-use uuid::Uuid;
 
-use crate::guest_os::GuestOsKind;
+use crate::{guest_os::GuestOsKind, zfs::ClonedFile as ZfsClonedFile};
 
 /// An RAII wrapper for a disk wrapped by a file.
 #[derive(Debug)]
@@ -18,8 +15,8 @@ pub struct FileBackedDisk {
     /// The name to use for instance spec backends that refer to this disk.
     backend_name: String,
 
-    /// The path at which the disk is stored.
-    disk_path: PathBuf,
+    /// The backing file for this disk.
+    file: ZfsClonedFile,
 
     /// The kind of guest OS image this guest contains, or `None` if the disk
     /// was not initialized from a guest OS artifact.
@@ -31,23 +28,15 @@ impl FileBackedDisk {
     /// the specified artifact on the host file system.
     pub(crate) fn new_from_artifact(
         backend_name: String,
-        artifact_path: &impl AsRef<Path>,
-        data_dir: &impl AsRef<Path>,
+        artifact_path: &impl AsRef<Utf8Path>,
         guest_os: Option<GuestOsKind>,
     ) -> Result<Self, super::DiskError> {
-        let mut disk_path = data_dir.as_ref().to_path_buf();
-        disk_path.push(format!("{}.phd_disk", Uuid::new_v4()));
-        info!(
-            source = %artifact_path.as_ref().display(),
-            disk_path = %disk_path.display(),
-            "Copying source image to create temporary disk",
-        );
-
-        std::fs::copy(artifact_path, &disk_path)?;
+        let artifact = ZfsClonedFile::create_from_path(artifact_path.as_ref())?;
+        let disk_path = artifact.path().into_std_path_buf();
 
         // Make sure the disk is writable (the artifact may have been
         // read-only).
-        let disk_file = std::fs::File::open(&disk_path)?;
+        let disk_file = std::fs::File::open(disk_path)?;
         let mut permissions = disk_file.metadata()?.permissions();
 
         // TODO: Clippy is upset that `set_readonly(false)` results in
@@ -57,7 +46,7 @@ impl FileBackedDisk {
         permissions.set_readonly(false);
         disk_file.set_permissions(permissions)?;
 
-        Ok(Self { backend_name, disk_path, guest_os })
+        Ok(Self { backend_name, file: artifact, guest_os })
     }
 }
 
@@ -66,7 +55,7 @@ impl super::DiskConfig for FileBackedDisk {
         (
             self.backend_name.clone(),
             StorageBackendV0::File(FileStorageBackend {
-                path: self.disk_path.to_string_lossy().to_string(),
+                path: self.file.path().to_string(),
                 readonly: false,
             }),
         )
@@ -74,16 +63,5 @@ impl super::DiskConfig for FileBackedDisk {
 
     fn guest_os(&self) -> Option<GuestOsKind> {
         self.guest_os
-    }
-}
-
-impl Drop for FileBackedDisk {
-    fn drop(&mut self) {
-        info!(path = %self.disk_path.display(), "Deleting guest disk image");
-        if let Err(e) = std::fs::remove_file(&self.disk_path) {
-            error!(?e,
-                   path = %self.disk_path.display(),
-                   "Failed to delete guest disk image");
-        }
     }
 }
--- a/phd-tests/framework/src/disk/file.rs
+++ b/phd-tests/framework/src/disk/file.rs
@@ -4,10 +4,75 @@
 
 //! Abstractions for disks with a raw file backend.
 
-use camino::Utf8Path;
+use camino::{Utf8Path, Utf8PathBuf};
 use propolis_client::types::{FileStorageBackend, StorageBackendV0};
+use tracing::{debug, error, warn};
+use uuid::Uuid;
 
 use crate::{guest_os::GuestOsKind, zfs::ClonedFile as ZfsClonedFile};
+
+/// Describes the method used to create the backing file for a file-backed disk.
+#[derive(Debug)]
+enum BackingFile {
+    /// The disk is a ZFS clone of the original artifact.
+    Zfs(ZfsClonedFile),
+
+    /// The disk is a hard copy of the original artifact.
+    HardCopy(Utf8PathBuf),
+}
+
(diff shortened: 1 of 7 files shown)


gjcolombo:
Oops. The phd-run job stores artifacts on a tmpfs mount, which makes the runner upset when it tries to figure out which ZFS dataset they're on. Two threads to pull on here: (1) make PHD fall back to file copies if the artifact can't be cloned, (2) see if there's a better directory to use for artifacts so that CI runs can take advantage of this feature.

gjcolombo:
OK, this seems to be working now: the logs from the most recent CI run don't have the warnings/errors about ZFS clones failing to work as intended, and the run before that shows that the fallback behavior works as intended.

jclulow:
In the lab* target environments, if you want a ZFS pool (reasonable!) you'll need to create one on one of the available SSDs.

jclulow:
I believe that's what's going on here, for example: https://github.com/oxidecomputer/omicron/blob/a5430a6465f457175ca1cf0c0053ffc47f25bf62/.github/buildomat/jobs/deploy.sh#L223

gjcolombo:
Oh nice, beat me to it. Thanks for the pointer! (I see now from a perusal of the lab image post-boot script that I'm definitely getting away with a crime here, abetted by the small size of the Alpine image, which just so happens to fit into the `/work` ramdisk...)

I'll move this back to draft status while I work on this a bit more. I think I'll also want to change the framework not to presume anything about the name of the zpool under which it should try to place the clones.

jclulow:
Yeah I think two things to note:

- if you've got clones in pool B you need the original in B as well, as you can't clone a snapshot across two pools 
- it's always good to be able to just take a ZFS dataset name (which might be a child of the pool, rather than at the top level) and work exclusively under that; that's what the various Helios image building things do, so that they can coexist on multiuser systems, or work in different CI shapes, etc

gjcolombo:
I have (hopefully) tightened this up a bit in the newest changes. The phd-run script now creates a pool spanning the runner's SSDs (as in the omicron-deploy job) and uses this pool's default dataset's root directory to store downloaded artifacts. I also tweaked the framework so that, when cloning artifact A, the clone is made a descendant of A's original dataset (the framework was already looking this up to figure out what to snapshot) instead of sticking it under whateer happens to be named `rpool` and hoping for the best.

I ran this locally (where I have a bespoke artifact TOML that specifies some handmade artisan locally-blended OS images, not all of which live in the same directory as the artifact dir I pass to PHD) in addition to the CI run on 2b025bd and things looked OK (tests pass, nothing obviously leaked).

phd-tests/framework/src/disk/file.rs
```diff
@@ -4,22 +4,84 @@
...
+
+    /// The disk is a hard copy of the original artifact.
+    HardCopy(Utf8PathBuf),
+}
+
+impl BackingFile {
+    /// Creates a new backing file from the artifact with the supplied
+    /// `artifact_path`. If possible, this routine will create a ZFS clone of
+    /// the dataset containing the file; otherwise it will fall back to creating
+    /// a hard copy of the original artifact.
```
hawkw, line 28: the first thing i was looking for while reviewing was whether we fall back to not making ZFS clones if it doesn't work --- i'm glad to see that we do that!

phd-tests/framework/src/disk/file.rs
```diff
@@ -4,22 +4,84 @@
...
+    fn create_from_source(
+        artifact_path: &Utf8Path,
+        data_dir: &Utf8Path,
+    ) -> anyhow::Result<Self> {
+        if let Ok(file) = ZfsClonedFile::create_from_path(artifact_path) {
+            return Ok(Self::Zfs(file));
+        }
+
+        warn!(%artifact_path,
+              "failed to make ZFS clone of backing artifact, will copy it");
```
hawkw, line 38: nit, not a big deal: it might be nice to include the error in the warning?
```suggestion
        match ZfsClonedFile::create_from_path(artifact_path) {
            Ok(file) => return Ok(Self::Zfs(file)),
            Err(error) => warn!(
                %artifact_path,
                %error,
               "failed to make ZFS clone of backing artifact, will copy it",
            ),
        }

```

phd-tests/framework/src/disk/file.rs
```diff
@@ -31,23 +93,19 @@ impl FileBackedDisk {
...
+        let artifact = BackingFile::create_from_source(
+            artifact_path.as_ref(),
+            data_dir.as_ref(),
+        )?;
+        let disk_path = artifact.path().into_std_path_buf();
 
         // Make sure the disk is writable (the artifact may have been
         // read-only).
-        let disk_file = std::fs::File::open(&disk_path)?;
+        let disk_file = std::fs::File::open(disk_path)?;
```
hawkw, line 108: nit: `File::open` takes `impl AsRef<Path>`, so I don't _think_ the explicit conversion to a `std::path::PathBuf` is actually necessary?

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -905,3 +866,54 @@ impl Drop for TestVm {
...
+
+/// Attempts to ensure that the Propolis server referred to by `client` is in
+/// the `Destroyed` state by stopping any VM that happens to be running in that
+/// server.
+///
+/// This function is best-effort.
+async fn try_ensure_vm_destroyed(client: &Client) {
+    match client.instance_get().send().await.map(|r| r.instance.state) {
+        Ok(InstanceState::Destroyed) => return,
+        Err(e) => warn!(?e, "error getting instance state from dropped VM"),
```
hawkw, line 878: tiny nit: the name `e` will show up in the logs, so i might change this to 
```suggestion
        Err(error) => warn!(?error, "error getting instance state from dropped VM"),
```

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -905,3 +866,54 @@ impl Drop for TestVm {
...
+    }
+
+    debug!("trying to ensure Propolis server VM is destroyed");
+    if let Err(e) = client
+        .instance_state_put()
+        .body(InstanceStateRequested::Stop)
+        .send()
+        .await
+    {
+        error!(%e, "error stopping VM to move it to Destroyed");
```
hawkw, line 889: similarly:
```suggestion
    if let Err(error) = client
        .instance_state_put()
        .body(InstanceStateRequested::Stop)
        .send()
        .await
    {
        error!(%error, "error stopping VM to move it to Destroyed");
```

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -905,3 +866,54 @@ impl Drop for TestVm {
...
+    let check_destroyed = || async {
+        match client.instance_get().send().await.map(|r| r.instance.state) {
+            Ok(InstanceState::Destroyed) => Ok(()),
+            Ok(state) => Err(backoff::Error::transient(anyhow::anyhow!(
+                "instance not destroyed yet (state: {:?})",
+                state
+            ))),
+            Err(e) => {
+                error!(%e, "failed to get state of VM being destroyed");
+                Err(backoff::Error::permanent(e.into()))
```
hawkw, line 902: and also:
```suggestion
            Err(error) => {
                error!(%error, "failed to get state of VM being destroyed");
                Err(backoff::Error::permanent(error.into()))
```

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -905,3 +866,54 @@ impl Drop for TestVm {
...
+        backoff::ExponentialBackoff {
+            max_elapsed_time: Some(std::time::Duration::from_secs(5)),
+            ..Default::default()
+        },
+        check_destroyed,
+    )
+    .await;
+
+    if let Err(e) = destroyed {
+        error!(%e, "VM not destroyed after 5 seconds");
```
hawkw, line 917:
```suggestion
    if let Err(error) = destroyed {
        error!(%error, "VM not destroyed after 5 seconds");
```

phd-tests/framework/src/zfs.rs
```diff
@@ -0,0 +1,321 @@
...
+    args: &[&str],
+) -> anyhow::Result<std::process::Output> {
+    debug!(verb, ?args, "executing ZFS command");
+
+    let output = std::process::Command::new("pfexec")
+        .arg("zfs")
+        .arg(verb)
+        .args(args)
+        .output()
+        .with_context(|| format!("running `zfs {verb}` with args {args:?}"))?;
```
hawkw, line 301: lovely, thanks for adding this <3

phd-tests/framework/src/zfs.rs
```diff
@@ -0,0 +1,321 @@
...
+
+    // An unmounted object's mount point displays as "-", so this token should
+    // always be present, even for unmounted objects.
+    let mount_point = words.nth(3).ok_or_else(|| {
+        anyhow::anyhow!("`zfs list` didn't produce a mount point")
+    })?;
+
+    let mount_point = mount_point
+        .starts_with('/')
+        .then_some(mount_point)
```
hawkw, line 281: cute :)

phd-tests/framework/src/zfs.rs
```diff
@@ -0,0 +1,321 @@
...
+            .strip_prefix(&containing_dataset.mount_point)
+            .context("getting relative path to file to clone")?;
+
+        let snapshot = Snapshot::create_from_dataset(&containing_dataset.name)?;
+        Self::create_from_paths_and_snapshot(
+            containing_dataset,
+            relative_file_path,
+            snapshot,
+        )
+        .context("creating zfs clone")
```
hawkw, line 120: tiny nit: maybe these errors should contain the file path? or do the inner errors already do that?

phd-tests/framework/src/zfs.rs
```diff
@@ -0,0 +1,321 @@
...
+        dataset: Dataset,
+        relative_file_path: &Utf8Path,
+        snapshot: Snapshot,
+    ) -> anyhow::Result<Self> {
+        let clone_name =
+            format!("{}/phd-clone-{}", dataset.name.0, Uuid::new_v4());
+
+        zfs_command("clone", &[&snapshot.name.0, &clone_name])?;
+
+        fn get_clone_mount_point(
```
hawkw, line 145: maybe worth a comment explaining that this is a function so that we can run `zfs destroy` no matter where we bail in this function?

hawkw (approved):
Overall this is very straightforward and pretty easy to understand! I commented on some small nits.

phd-tests/framework/src/zfs.rs
gjcolombo: Done in dece78c.

phd-tests/framework/src/disk/file.rs
gjcolombo: Done in 409c7bd.

phd-tests/framework/src/test_vm/mod.rs
gjcolombo: Done here (and below) in 93b17b55.

phd-tests/framework/src/zfs.rs
gjcolombo: They... do not! Added in fd6ac0b.

phd-tests/framework/src/disk/file.rs
gjcolombo: Good call. Done in c2133469b.

phd-tests/framework/src/zfs.rs
gjcolombo: Added a note in dece78ce.

PR #645, opened 2024-02-15
PHD: use `clap` for more `cargo xtask phd` args

hawkw:
Currently, `cargo xtask phd` does a bunch of "bash-style" arg parsing, looping over any unrecognized arguments. This was intended as a way of passing through arguments to `phd-runner`. However, because `clap` will stick any unrecognized arguments in `trailing_var_args`, we can just use normal `clap` arg parsing to handle the arguments that we default before passing through to `phd-runner`.

Now, the `cargo xtask phd` command has nicer help text for its overridden arguments, and it can use `clap`'s built-in validation of conflicting arguments. This seems nicer.

### DIFF
--- a/xtask/src/task_phd.rs
+++ b/xtask/src/task_phd.rs
@@ -41,6 +41,7 @@ macro_rules! cargo_warn {
 }
 
 #[derive(clap::Subcommand, Debug, Clone)]
+#[allow(clippy::large_enum_variant)]
 pub(crate) enum Cmd {
     /// Run the PHD test suite.
     Run {
@@ -68,21 +69,10 @@ pub(crate) enum Cmd {
 
 #[derive(clap::Parser, Debug, Clone)]
 pub(crate) struct RunArgs {
-    /// Build `propolis-server` in release mode.
-    #[clap(long = "release", short = 'r')]
-    propolis_release_mode: bool,
-
     /// If set, temporary directories older than one day will not be
     /// deleted.
     #[clap(long)]
     no_tidy: bool,
-
-    /// If set, skip migration-from-base tests.
-    ///
-    /// If this flag is present, `cargo xtask phd` will not pass
-    /// `--base-propolis-branch master` to the `phd-runner` command.
-    #[clap(long)]
-    no_base_propolis: bool,
 
     /// Arguments to pass to `phd-runner run`.
     ///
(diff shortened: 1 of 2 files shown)


xtask/src/task_phd.rs
```diff
@@ -95,6 +85,163 @@ pub(crate) struct RunArgs {
...
+    ///
+    /// If this argument is provided, PHD will download the latest Propolis
+    /// server artifact from Buildomat for the provided branch name, and use it
+    /// to test migration from that Propolis version to the Propolis revision
+    /// under test.
+    ///
+    /// This argument conflicts with the `--base-propolis-commit`,
+    /// `--base-propolis-cmd`, and `--no-base-propolis` arguments. If none of
+    /// these arguments are provided, `cargo xtask phd` will automatically pass
+    /// `--base-propolis-branch master` to `phd-runner`.
```
gjcolombo, line 132: If I'm reading this right the behavior of `BasePropolisArgs::configure_command` is to take the first matching option from (cmd, commit, branch, no_base) and ignore the others. Would it make sense to tell `clap` these options all conflict? (This applies to the `CrucibleArgs` as well I think.)

hawkw: This struct, as well as the `CrucibleArgs` and `PropolisArgs`, have a `clap` #[group]` attribute with [`multiple = false`](https://docs.rs/clap/latest/clap/struct.ArgGroup.html#method.multiple) on them: https://github.com/oxidecomputer/propolis/blob/66f09845d10a4c5c5b6ba1ae1cc4f23e1390376f/xtask/src/task_phd.rs#L118

This is the same as adding `conflicts_with` for each arg in the group to each individual arg. So, this validation already works:
```console
$ cargo xtask phd run --base-propolis-branch master --base-propolis-commit asdf --no-base-propolis

error: the argument '--base-propolis-branch <BASE_PROPOLIS_BRANCH>' cannot be used with:
  --base-propolis-commit <BASE_PROPOLIS_COMMIT>
  --no-base-propolis

Usage: xtask phd run --base-propolis-branch <BASE_PROPOLIS_BRANCH> [PHD_ARGS]...

For more information, try '--help'.
```

gjcolombo: Nice!

gjcolombo (approved):

PR #646, opened 2024-02-15
PHD: add Windows Server 2016 adapter & improve WS2016/2019 reliability

### LINKED ISSUE #601: PHD: guest OS adapter for Windows Server 2022/2019/2016

### LINKED ISSUE #643: phd: multiline shell commands don't work with VT100 serial console backend
PHD's `Vt80x24` terminal backend searches for strings by calling `termwiz::Surface::screen_chars_to_string()` and calling `rfind` on the result. This works fine for single-line search targets, but for multi-line targets, things are trickier, because `screen_chars_to_string` returns *entire lines* (including all their spaces) separated by `\n` characters. That is, if the virtual contents are

```
hello
world
user@host $
```

then a search for `hello\nworld` will not match, because `screen_chars_to_string` returns `hello`, then 75 spaces, then `\nworld`. PHD will need to account for this behavior when trying to watch for multiline strings.

### LINKED ISSUE #644: phd: long serial console commands time out when guest serial console debounce is in effect
The PHD framework's `run_shell_command` function takes a command to execute in a VM, sends it to the VM's serial console worker to be "typed" into the guest console, and then waits for a fixed 15-second interval for the command to be echoed back from the guest. Because this timeout begins to elapse as soon as the command is sent to the serial console task, it can easily expire before the entire command is typed if the command is long and the task has been asked to insert a delay between sending each character of an outgoing message (e.g. to avoid an in-guest keyboard debounce). 

This causes the `running_process` migration tests (which type a long shell script into the guest console) to fail for Windows Server 2019 guests.

The easiest way to fix this is to wait for the task to type the entire command before starting the clock on waiting for it to be echoed back, e.g. by sending a `oneshot` along with the string to type so that the serial console task can tell the submitter when everything has been typed.

gjcolombo:
Add a Windows Server 2016 guest adapter to PHD. It's more or less identical to the Windows Server 2019 adapter (but is still worth keeping separate since the guests support different command-line tools and Powershell cmdlets out of the box).

Tighten a few bolts to improve (but not perfect) test performance and reliability for these guests:

- Rework the way shell commands are issued:
  - `GuestOs`es now specify a serial console `CommandSequence` that the framework should execute in order to run a shell command.
  - Waiting for a string to appear in the serial console buffer no longer consumes any characters from the buffer. Instead, callers must clear the buffer manually.

With this change, it no longer matters that Windows guests don't re-echo characters that are already on-screen, because the framework no longer invalidates them until it knows they're no longer wanted. 

Additionally improve the keystroke debounce logic:

- Only apply the debounce delay if the next character to be issued is the same as the previous character. This allows the debounce timeout to be lengthened considerably without tanking test performance (which is important to minimize flakiness).

  It might be worth investigating whether we could avoid this delay by doing byte-by-byte "call and response" for characters we expect the guest to echo. This may require a more substantial refactoring of the serial console code, so I've just resorted to having a longer debounce for now.
- Change `TestVm::send_serial_bytes_async` to wait for the serial task to acknowledge that it has sent all the requested bytes to the guest before returning so that `TestVm` can time out waiting for characters to be echoed without including the time spent waiting for them to be sent.

Tested with WS2016, WS2019, and Debian 11 guests.

Fixes #601. Fixes #644. Fixes #643.

### DIFF
--- a/phd-tests/framework/src/guest_os/mod.rs
+++ b/phd-tests/framework/src/guest_os/mod.rs
@@ -13,6 +13,7 @@ mod alpine;
 mod debian11_nocloud;
 mod ubuntu22_04;
 mod windows;
+mod windows_server_2016;
 mod windows_server_2019;
 mod windows_server_2022;
 
@@ -64,6 +65,7 @@ pub enum GuestOsKind {
     Alpine,
     Debian11NoCloud,
     Ubuntu2204,
+    WindowsServer2016,
     WindowsServer2019,
     WindowsServer2022,
 }
@@ -76,6 +78,8 @@ impl FromStr for GuestOsKind {
             "alpine" => Ok(Self::Alpine),
             "debian11nocloud" => Ok(Self::Debian11NoCloud),
             "ubuntu2204" => Ok(Self::Ubuntu2204),
+            "windowsserver2016" => Ok(Self::WindowsServer2016),
+            "windowsserver2019" => Ok(Self::WindowsServer2019),
             "windowsserver2022" => Ok(Self::WindowsServer2022),
             _ => Err(std::io::Error::new(
                 std::io::ErrorKind::InvalidInput,
@@ -92,6 +96,9 @@ pub(super) fn get_guest_os_adapter(kind: GuestOsKind) -> Box<dyn GuestOs> {
             Box::new(debian11_nocloud::Debian11NoCloud)
         }
         GuestOsKind::Ubuntu2204 => Box::new(ubuntu22_04::Ubuntu2204),
+        GuestOsKind::WindowsServer2016 => {
+            Box::new(windows_server_2016::WindowsServer2016)
+        }
         GuestOsKind::WindowsServer2019 => {
             Box::new(windows_server_2019::WindowsServer2019)
         }
--- a/phd-tests/framework/src/guest_os/mod.rs
+++ b/phd-tests/framework/src/guest_os/mod.rs
@@ -27,6 +27,9 @@ pub(super) enum CommandSequenceEntry<'a> {
     /// Write the specified string as a command to the guest serial console.
     WriteStr(Cow<'a, str>),
 
+    /// Tell the serial console task to clear its buffer.
+    ClearBuffer,
+
     /// Change the serial console buffering discipline to the supplied
     /// discipline.
     ChangeSerialConsoleBuffer(crate::serial::BufferKind),
--- a/phd-tests/framework/src/guest_os/mod.rs
+++ b/phd-tests/framework/src/guest_os/mod.rs
@@ -11,6 +11,7 @@ use serde::{Deserialize, Serialize};
 
 mod alpine;
 mod debian11_nocloud;
+mod shell_commands;
 mod ubuntu22_04;
 mod windows;
 mod windows_server_2016;
@@ -19,12 +20,12 @@ mod windows_server_2022;
 
 /// An entry in a sequence of interactions with the guest's command prompt.
 #[derive(Debug)]
-pub(super) enum CommandSequenceEntry {
+pub(super) enum CommandSequenceEntry<'a> {
     /// Wait for the supplied string to appear on the guest serial console.
-    WaitFor(&'static str),
+    WaitFor(Cow<'a, str>),
 
     /// Write the specified string as a command to the guest serial console.
-    WriteStr(&'static str),
+    WriteStr(Cow<'a, str>),
 
     /// Change the serial console buffering discipline to the supplied
     /// discipline.
@@ -35,7 +36,7 @@ pub(super) enum CommandSequenceEntry {
     SetSerialByteWriteDelay(std::time::Duration),
 }
 
-pub(super) struct CommandSequence(pub Vec<CommandSequenceEntry>);
+pub(super) struct CommandSequence<'a>(pub Vec<CommandSequenceEntry<'a>>);
 
 pub(super) trait GuestOs: Send + Sync {
     /// Retrieves the command sequence used to wait for the OS to boot and log
@@ -48,13 +49,13 @@ pub(super) trait GuestOs: Send + Sync {
     /// Indicates whether the guest has a read-only filesystem.
     fn read_only_fs(&self) -> bool;
 
-    /// Some guests need to amend incoming shell commands from tests in order to
-    /// get output to display on the serial console in a way those guests can
-    /// accept (e.g. by clearing the screen immediately before running each
-    /// command). This function amends an incoming command according to the
-    /// guest adapter's instructions.
-    fn amend_shell_command<'a>(&self, cmd: &'a str) -> Cow<'a, str> {
-        Cow::Borrowed(cmd)
+    /// Returns the sequence of serial console operations a test VM should issue
+    /// in order to execute `cmd` in the guest's shell.
+    fn shell_command_sequence<'a>(&self, cmd: &'a str) -> CommandSequence<'a> {
+        shell_commands::shell_command_sequence(
+            Cow::Borrowed(cmd),
+            crate::serial::BufferKind::Raw,
+        )
     }
 }
 
--- a/phd-tests/framework/src/guest_os/mod.rs
+++ b/phd-tests/framework/src/guest_os/mod.rs
@@ -33,7 +33,7 @@ pub(super) enum CommandSequenceEntry<'a> {
 
     /// Set a delay between writing individual bytes to the guest serial console
     /// to avoid keyboard debouncing logic in guests.
-    SetSerialByteWriteDelay(std::time::Duration),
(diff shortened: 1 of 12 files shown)


phd-tests/framework/src/guest_os/mod.rs
```diff
@@ -11,30 +11,32 @@ use serde::{Deserialize, Serialize};
...
 mod windows_server_2019;
 mod windows_server_2022;
 
 /// An entry in a sequence of interactions with the guest's command prompt.
 #[derive(Debug)]
-pub(super) enum CommandSequenceEntry {
+pub(super) enum CommandSequenceEntry<'a> {
     /// Wait for the supplied string to appear on the guest serial console.
-    WaitFor(&'static str),
+    WaitFor(Cow<'a, str>),
```
gjcolombo, line 25: This feels like it might be the wrong tool for the job, but maybe not?

Previously this was only used for the string literals that appear in guest OSes' boot sequences, but now a command *might* be an owned string, e.g. if `shell_command_sequence` had to amend it. It seemed silly to make all the literals into owned `String`s just to accommodate this case, but I'm not sure if `Cow` is the best (or even an appropriate) way for this code to have its `&'static str` cake and eat `String`s too.

hawkw: `Cow` is absolutely an appropriate way to have something that's "either a `&'static str` or a `String`" --- honestly, I think I see it used for that purpose far more than for use-cases that actually use copy-on-write...

phd-tests/framework/src/guest_os/mod.rs
```diff
@@ -11,30 +11,32 @@ use serde::{Deserialize, Serialize};
...
     /// Write the specified string as a command to the guest serial console.
-    WriteStr(&'static str),
+    WriteStr(Cow<'a, str>),
 
     /// Change the serial console buffering discipline to the supplied
     /// discipline.
     ChangeSerialConsoleBuffer(crate::serial::BufferKind),
 
     /// Set a delay between writing individual bytes to the guest serial console
     /// to avoid keyboard debouncing logic in guests.
```
hawkw, line 35: does this comment need to be updated, if the semantics of this have changed?

phd-tests/framework/src/guest_os/windows.rs
```diff
@@ -14,52 +14,60 @@ use super::{CommandSequence, CommandSequenceEntry, GuestOsKind};
...
-        CommandSequenceEntry::WriteStr(""),
-        CommandSequenceEntry::WaitFor("Password:"),
-        CommandSequenceEntry::WriteStr("0xide#1Fan"),
+        CommandSequenceEntry::WriteStr("".into()),
+        CommandSequenceEntry::WaitFor("Username:".into()),
+        CommandSequenceEntry::WriteStr("Administrator".into()),
+        CommandSequenceEntry::WaitFor("Domain  :".into()),
+        CommandSequenceEntry::WriteStr("".into()),
+        CommandSequenceEntry::WaitFor("Password:".into()),
+        CommandSequenceEntry::WriteStr("0xide#1Fan".into()),
```
hawkw, line 48: nit, take it or leave it: maybe worth adding functions like
```rust
impl<'a> CommandSequenceEntry<'a> {
    pub fn write_str(s: impl Into<Cow<'a, str>>) -> Self {
         Self::WriteStr(s.into())
    }

    pub fn wait_for(s: impl Into<Cow<'a, str>>) -> Self {
         Self::WaitFor(s.into())
    }
}
```
and then calling them instead of having to `.into()` everything? 

not a big deal, but maybe a little bit nicer?

phd-tests/framework/src/guest_os/shell_commands.rs
```diff
@@ -0,0 +1,49 @@
...
+            // line at a time and wait for each line to be echoed before
+            // starting the next. For very long commands (more than 24 lines),
+            // this avoids having to deal with lines scrolling off the buffer
+            // before they can be waited for.
+            let cmd_lines = cmd.trim_end().lines();
+            let echo_lines = echo.lines();
+            let mut seq = vec![];
+            for (cmd, echo) in cmd_lines.zip(echo_lines) {
+                seq.push(CommandSequenceEntry::WriteStr(cmd.to_owned().into()));
+                seq.push(CommandSequenceEntry::WaitFor(echo.to_owned().into()));
```
hawkw, line 42: wait, why do we need to turn these into owned strings, and then immediately into `Cow`s? can't we take a `cmd: Cow<'a, str>` and return a `CommandSequence<'a>`?

(importance: not really)

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -774,7 +782,12 @@ impl TestVm {
 
     async fn send_serial_bytes_async(&self, bytes: Vec<u8>) -> Result<()> {
         match &self.state {
-            VmState::Ensured { serial } => serial.send_bytes(bytes),
+            VmState::Ensured { serial } => {
+                let (done_tx, done_rx) = oneshot::channel();
+                serial.send_bytes(bytes, done_tx)?;
+                done_rx.await?;
```
hawkw, line 788: nit, take it or leave it: i _might_ move the oneshot into `send_bytes`, and make it an `async fn` that creates a oneshot, sends the `tx` end to the serial worker, and then awaits the `rx` end? that way, here, we could just write `serial.send_bytes(bytes).await`...

the big reason _not_ to do that is if we want to be able to call `send_bytes()` in a non-async context,[^1] which ... i don't _think_ we do anywhere? but maybe we want to be able to do it later? 

[^1]: and either pass the `rx` end of the oneshot into something that eventually awaits it in an async context, or just throw it away because we don't care.

hawkw: actually, thinking about this some more, if `send_bytes` was a non-async function that just returned the `rx`, we could write `serial.send_bytes(bytes)?.await?;` here, but still retain the optionality to call it without immediately awaiting and pass the oneshot `rx` to someone else/chuck it in the garbage immediately/etc.

hawkw (approved):
looks good to me! i left a few suggestions, but none of them seemed like actual problems.

gjcolombo:
Thanks @hawkw! I still need to address your feedback here. I pushed bb39823 to change the serial console wait discipline in a way that should fix #643 for good:

- Waiting for characters on the serial console no longer consumes anything from the buffer
- Instead, there's an explicit "clear the buffer" command that callers have to issue
- The shell command issuance sequences clear the buffer just before they actually issue the last '\n' that executes the command

With this change Windows Server 2019 passes all the `running_process` tests that it previously failed (and existing passing tests appear to continue passing, but I'm doing one more full Debian 11 run to make sure).

phd-tests/framework/src/guest_os/shell_commands.rs
gjcolombo: For `cmd` I think it's that the `Cow` might not be borrowing anything--if it's a `Cow::Owned` then this would return a reference to data owned by the caller. (If this function took an actual `&'a str` then I think this would work, but then some of the callers would break, e.g. `WindowsServer2016::shell_command_sequence`, which creates an owned `String` that it passes down to here.) 

`echo` is owned outright at this point (this function created it by calling `replace` on line 23).

phd-tests/framework/src/guest_os/windows.rs
gjcolombo: I like this much better, thanks! The `.into()` grossness was part of made me question whether `Cow` was the right tool for this job.

phd-tests/framework/src/guest_os/mod.rs
gjcolombo: Good catch, done. (In b5b8d81 I also cleaned up a few other comments that were invalidated by bb39823.)

phd-tests/framework/src/test_vm/mod.rs
gjcolombo: Done in b5b8d81 (along with some general cleanup of this whole family of TestVm functions).

hawkw (approved):
lovely!

gjcolombo:
The phd-run job failed due to a known flake in VMM teardown that I've filed as #648. All the tests otherwise passed.

PR #647, opened 2024-02-16
PHD: add `pfexec` to xtask phd-runner invocation

hawkw:
Currently, when using `cargo xtask phd run` to run the PHD test suite, ensuring that the runner has the correct permissions requires running `pfexec cargo xtask phd`. This means that the commands run by the xtask to set up the artifact store and temp directories, and to build `phd-runner` and `propolis-server`, are run under the profile specified by `pfexec`, resulting in a bunch of files in the Cargo `target/` directory that the user doesn't actually own. This is annoying if you want to then run `cargo clean` or something.

This commit changes `cargo xtask phd run` to just execute the runner command with `pfexec` when compiled for Illumos. This saves the user from having to type it themself, and (more importantly) means that the binaries aren't built with a different owner than the user who builds them.

### DIFF
--- a/xtask/src/task_phd.rs
+++ b/xtask/src/task_phd.rs
@@ -350,7 +350,18 @@ impl Cmd {
         }
 
         let phd_runner = build_bin("phd-runner", false)?;
-        let mut cmd = phd_runner.command();
+        let mut cmd = if cfg!(target_os = "illumos") {
+            let mut cmd = Command::new("pfexec");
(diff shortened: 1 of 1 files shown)


gjcolombo (approved):

hawkw:
PHD failure looks like #648, i'm gonna just merge this instead of rerunning it to be a good buildomat citizen.

PR #651, opened 2024-02-26
PHD: improve Windows reliability

gjcolombo:
Tweak a handful of things to make Windows tests more (though not perfectly) reliable:

- Explicitly sleep between the time the first `C:\Windows\System32>` prompt is displayed and the time the next command is issued. This appears to reduce the chance that the first command's keystrokes will be eaten.
- On Windows Server 2022, use the `mode` command to set the console session's row/column count explicitly before launching Cygwin. This avoids a problem where Cygwin thought it was on a 30-column terminal and was wrapping lines unexpectedly.
- Shorten the Cygwin prompt from `Administrator@PHD-WINDOWS:$ ` to just `$ `. The test doesn't need the user/host name cue and this reduces the likelihood of a line wrapping unexpectedly on WS2016/2019.
- Handle the "cursor right" control code in the raw serial console buffer, since WS2022 uses it sometimes.
- Add a tracing affordance to the Vt80x24 buffer that prints the current screen contents line-by-line, ignoring blank lines at the end of the output. This makes the terminal contents much easier to read during debugging.
- Bump the repeated character debounce up *again* to try to cut down on 2016/2019 flakiness. In the long run I think it'll make sense to experiment with totally different approaches to this problem. For example, I wonder if things would work better if the serial task waited for individual typed characters to be echoed before sending a subsequent character, but I think doing that will require the serial task and the buffer trait to be restructured somewhat.

Tested with WS2022, WS2019, and Debian 11 guests.

### DIFF
--- a/phd-tests/framework/src/guest_os/mod.rs
+++ b/phd-tests/framework/src/guest_os/mod.rs
@@ -30,6 +30,10 @@ pub(super) enum CommandSequenceEntry<'a> {
     /// Tell the serial console task to clear its buffer.
     ClearBuffer,
 
+    /// The command executor (not the serial console task!) should sleep for
+    /// this duration.
+    Sleep(std::time::Duration),
+
     /// Change the serial console buffering discipline to the supplied
     /// discipline.
     ChangeSerialConsoleBuffer(crate::serial::BufferKind),
--- a/phd-tests/framework/src/guest_os/mod.rs
+++ b/phd-tests/framework/src/guest_os/mod.rs
@@ -27,12 +27,17 @@ pub(super) enum CommandSequenceEntry<'a> {
     /// Write the specified string as a command to the guest serial console.
     WriteStr(Cow<'a, str>),
 
+    /// Attempt to establish consistent echoing of characters to the guest
+    /// serial console by typing `send` and then waiting for `expect` to appear
+    /// within the supplied `timeout`.
+    EstablishConsistentEcho {
+        send: Cow<'a, str>,
+        expect: Cow<'a, str>,
+        timeout: std::time::Duration,
+    },
+
     /// Tell the serial console task to clear its buffer.
     ClearBuffer,
-
-    /// The command executor (not the serial console task!) should sleep for
-    /// this duration.
-    Sleep(std::time::Duration),
 
     /// Change the serial console buffering discipline to the supplied
     /// discipline.
--- a/phd-tests/framework/src/guest_os/windows.rs
+++ b/phd-tests/framework/src/guest_os/windows.rs
@@ -6,6 +6,18 @@
 
 use super::{CommandSequence, CommandSequenceEntry, GuestOsKind};
 
+const CYGWIN_CMD: &str = "C:\\cygwin\\cygwin.bat\r";
+
+/// Prepends a `reset` command to the shell command supplied in `cmd`. Windows
+/// versions that drive a VT100 terminal can use this to try to force Windows to
+/// clear and redraw the entire screen before displaying the command's output.
+/// Without this, Windows may not render the post-output command prompt if the
+/// post-command terminal state happens to place a prompt at a location that
+/// already had onen pre-command.
+pub(super) fn prepend_reset_to_shell_command(cmd: &str) -> String {
+    format!("reset && {cmd}")
+}
+
 /// Emits the login seqeunce for the given `guest`, which must be one of the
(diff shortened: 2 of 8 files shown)


phd-tests/framework/src/serial/vt80x24.rs
```diff
@@ -148,3 +152,23 @@ fn make_absolute_cursor_position(col: usize, row: usize) -> Change {
...
     }
 }
+
+fn trace_buffer_contents(contents: &str) {
+    let mut last_non_empty = None;
+    for (idx, line) in contents.lines().enumerate() {
+        if line.chars().any(|c| c != ' ') {
+            last_non_empty = Some(idx);
+        }
+    }
```
hawkw, line 162: style nit, take it or leave it: this looks like [`Iterator::position`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.position):
```suggestion
    let last_non_empty =  contents.lines().position(|line| line.chars().any(|c| c != ' '));
```

hawkw (approved):

gjcolombo (commented):
Thanks for the review, @hawkw!

phd-tests/framework/src/guest_os/windows_server_2016.rs
```diff
@@ -24,7 +24,7 @@ impl GuestOs for WindowsServer2016 {
     }
 
     fn get_shell_prompt(&self) -> &'static str {
-        "Administrator@PHD-WINDOWS:$ "
+        "$ "
```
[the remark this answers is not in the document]
hawkw: Yeah, I definitely wasn't expecting a change in this PR --- just a thought for the Eventual Future. :)

phd-tests/framework/src/guest_os/windows.rs
```diff
@@ -67,25 +69,43 @@ pub(super) fn get_login_sequence_for<'a>(
...
-        // prefer carriage returns to linefeeds. Accommodate this behavior
-        // until Cygwin is launched.
+        // There appears (from observing Windows test reliability) to be some
+        // kind of race at command prompt startup that can cause characters to
+        // be eaten if they're typed too quickly after the command prompt
+        // session launches. Sleep for a bit after the initial prompt is
+        // displayed to work around this.
         CommandSequenceEntry::wait_for("C:\\Windows\\system32>"),
-        CommandSequenceEntry::write_str("cls\r"),
+        CommandSequenceEntry::Sleep(Duration::from_secs(5)),
```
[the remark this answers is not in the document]
hawkw: Interesting, I'll take a look!

phd-tests/framework/src/guest_os/mod.rs
```diff
@@ -27,6 +27,15 @@ pub(super) enum CommandSequenceEntry<'a> {
     /// Write the specified string as a command to the guest serial console.
     WriteStr(Cow<'a, str>),
 
+    /// Attempt to establish consistent echoing of characters to the guest
+    /// serial console by typing `send` and then waiting for `expect` to appear
+    /// within the supplied `timeout`.
+    EstablishConsistentEcho {
```
hawkw, line 33: neat!

is there any particular reason we allow using arbitrary strings for this operation? could we just always use a single character, or is there a reason we might want to do this with multiple strings?

hawkw (approved):

phd-tests/framework/src/guest_os/mod.rs
gjcolombo: The string's arbitrary. It could be just a single character, though if the caller picks something that might already be in the back buffer at the time it requests this operation, it will want to issue a `ClearBuffer` command first to avoid spurious matching. (The way I look at this operation is that it gives guests a way to tell the framework to repeat a pair of `WriteStr`/`WaitFor` commands until the `WaitFor` is satisfied; the flexibility in specifying the sent/expected strings is accidental, not essential, and is mostly there to provide an analogy to the Write/Wait commands.)

hawkw: Ah, yeah, the need to ensure that the string isn't already in the buffer makes sense. :+1:

gjcolombo:
I suspect the migrate-from-base tests barfed on the last run because this branch was previously based on a commit from before #639 landed. Let's see how the force-pushed branch fares.

gjcolombo:
That's more like it.

PR #654, opened 2024-02-27
Update progenitor and omicron deps

pfmooney:
Depending a version, rather than "branch = main", meant that we had two different copies of progenitor being used.  This fixes that situation, as well as updating related dependencies (omicron, strum) for a clean build.


sunshowers (approved):

sunshowers:
Thank you!

PR #655, opened 2024-03-01
[meta] patch whoami

sunshowers:
Patch it to fix memory corruption on illumos.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -496,6 +496,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "cfg_aliases"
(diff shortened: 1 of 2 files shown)


sunshowers:
(moving to draft per internal issue)

pfmooney:
While `omicron-common` (and thus `whoami` via `tokio-postgres`) is in the dependency tree here, it is effectively unused.  Propolis gets the `omicron-common` dependency since the nexus client library crates include it for the sake of convenience.  Because propolis is not interacting directly with anything postgres-y, it should never call into that code.

We discussed one possible way forward being the refactoring of those client crates to make `omicron-common` an optional dependency (largely for `From` impls and the like for omicron-internal types), so that external consumers like propolis only end up with a `progenitor` dependency to transform the OpenAPI document into the appropriate interface.

sunshowers:
Closing this and replacing with patching omicron: https://github.com/oxidecomputer/omicron/pull/5198

PR #657, opened 2024-03-06
Misc nvme/block fixes

pfmooney:
While investigating some block perf issues, I trip over a few things in nvme (and related).

### DIFF
--- a/lib/propolis/src/block/backend.rs
+++ b/lib/propolis/src/block/backend.rs
@@ -182,6 +182,7 @@ impl Attachment {
     pub fn notify(&self) {
         // TODO: avoid thundering herd?
         self.0.req_notifier.notify_waiters();
+        let _guard = self.0.state.lock().unwrap();
         self.0.cv.notify_all();
     }
 }
--- a/lib/propolis/src/hw/nvme/mod.rs
+++ b/lib/propolis/src/hw/nvme/mod.rs
@@ -840,20 +840,23 @@ impl PciNvme {
                     // Completion Queue y Head Doorbell
                     let cq = state.get_cq(qid)?;
                     cq.notify_head(val)?;
-
-                    // We may have skipped pulling entries off some SQ due to this
-                    // CQ having no available entry slots. Since we've just freed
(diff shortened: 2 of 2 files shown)


lib/propolis/src/block/backend.rs
```diff
@@ -182,6 +182,7 @@ impl Attachment {
     pub fn notify(&self) {
         // TODO: avoid thundering herd?
         self.0.req_notifier.notify_waiters();
+        let _guard = self.0.state.lock().unwrap();
```
luqmana, line 185: Hmm, was there some subtle timing bug you ran into that this fixes? Otherwise seems like a pessimization to grab the lock here.

pfmooney: It is pessimism.  Although I've not delved into the innards of the Condvar impl, I was worried about the potential to miss a wake-up if the notifier does not hold the same lock as the waiter.  The docs definitely state that using more than one mutex with a Condvar is forbidden:

> Note that any attempt to use multiple mutexes on the same condition variable may result in a runtime panic.

While it's not explicitly stated that _not_ holding the lock during notify is unsafe, I was left uneasy about it.

I check perf (using the in-memory backend) before and after, and noticed no change with the added lock acquisition.

luqmana: >Note that any attempt to use multiple mutexes on the same condition variable may result in a runtime panic.

I think "use" here means explicitly passing one to the Condvar wait method. But given the same perf characteristics seems fine to lock first.

[`pthread_cond_broadcast(3C)`](https://illumos.org/man/3C/pthread_cond_broadcast) also calls out this gives more predictable scheduling which is always nice!

       The pthread_cond_signal() or pthread_cond_broadcast() functions may be
       called by a thread whether or not it currently owns the mutex that
       threads calling pthread_cond_wait() or pthread_cond_timedwait() have
       associated with the condition variable during their waits; however, if
       predictable scheduling behavior is required, then that mutex is locked by
       the thread calling pthread_cond_signal() or pthread_cond_broadcast().

luqmana (approved):

PR #659, opened 2024-03-07
Publish instance vCPU stats to oximeter

bnaecker:
- Add more metadata to instance and HTTP API, for tracking silo / project IDs for an instance.
- Add the `VirtualMachine` kstat target for tracking individual vCPU usage by microstate.
- Use `VirtualMachine` in place of the bare instance UUID for tracking other statistics too.
- Use `KstatSampler` to pull vCPU kernel statistics to track vCPU usage
- A few minor tweaks to existing metrics
- Update OpenAPI documents


bnaecker:
This is the companion to https://github.com/oxidecomputer/omicron/pull/4900, in which Propolis reads `kstat`s for the vCPU usage data of its managed instance. These are published to `oximeter` along with the other statistics. This also modifies the pre-existing instance-reset counter statistic, subsuming it under the new `VirtualMachine` target for identifying an instance by silo, project, and instance ID.

bnaecker:
CI is failing because the `oximeter_instruments::kstat` module is only available on illumos, while we're running a number of tests on Linux. I'm going to experiment with some `cfg` directives to work around it, hopefully it's not too nasty.

bnaecker:
Ok, this last test failure is real. One thing the PHD tests verify is that we can migrate between the PR commit and its merge base. That is failing, because the PR branch changes the instance-ensure API, by adding the project / silo ID for tracking instance vCPU metrics. I'm not sure how this is _supposed_ to work when one changes the API. @gjcolombo or perhaps @hawkw, are there any existing plans for handling breaking API changes like this?

I can imagine this case is handled pretty easily, by making this new data optional. If we're migrating an instance from before the change, then those fields can't exist and we will simply forgo producing vCPU data in any case. Are there additional test-suite changes one would like to see in such cases?

gjcolombo:
> I'm not sure how this is supposed to work when one changes the API.

All this is a unforeseen but mostly positive consequence of adding the migrate-from-base tests: we've begun to test Propolis server API versioning! The specific problem here is:

- `InstanceProperties` now contains an `InstanceMetadata`
- The Propolis instance GET endpoint returns an `Instance` that contains an `InstanceProperties`
- PHD always uses the Propolis client API defined by the OpenAPI spec that was in the commit used to build the framework
- So, when the framework (with the new `Instance` definition) asks to get properties from a downlevel Propolis that doesn't have the new definition, deserialization fails on the client (because the downlevel server didn't supply all the fields it needs)

This would be a Real Problem in production if there were ever heterogeneous Propolis client and server versions, e.g. if we were doing a live upgrade and had an upgraded Nexus trying to query state directly from a not-yet-upgraded Propolis. However, I don't think we currently have that situation, for two reasons:

- Because all our upgrades are still offline (for now), there's no way to get this kind of heterogeneity: all Propolises get upgraded all at once
- Even if that weren't true, I believe it's true that the only user of this particular API is a Propolis's overseeing sled agent, and sled agent and Propolis are supposed to be shipped in a single logical package (i.e. it shouldn't currently be possible to upgrade a sled's sled agent without installing software that will also immediately and atomically upgrade its Propolis image)

---

I think there are a couple of ways to make this sort of change in general:

- The straightforward way, which you've suggested, is to make the new field (in this case the metadata field in `InstanceProperties` an `Option` and tag it with the `#[serde(default)]` field attribute. This will get everything to deserialize properly, at the cost of making the server have to deal with cases where this field hasn't been specified, which may never occur in practice (see above).
  - There's probably a separate discussion to be had here about whether all these API types should also be labeled `deny_unknown_fields` so that a client gets an error if it specifies a field a server doesn't understand. (I tend to prefer this behavior to "the server silently drops the fields"--I specified them because they were *important*, dang it!)
- The buttoned-down way is to create a new `InstancePropertiesV2` struct and then create `v2/foo` endpoints in the server that take the new type. Then PHD can try to use v2 endpoints first and fall back to v1 endpoints if v2 isn't available. Ideally this would require a minimum of boilerplate; RFD 421 talks about how to achieve that ideal but I don't think we've yet built any of the mechanisms it talks about.

For *this specific PR*, I would be OK with the first approach, but I'd also be OK with declaring that because we're pre-live-update, breaking API changes like this are technically permissible even though they break these updateability tests. We basically did this in #639--breaking change, would have been Very Illegal if we were actually migrating anything, but because we understood what was happening we were willing to live with a reddish CI run on the PR.

bnaecker:
I talked with @gjcolombo offline after reading through his comments. I agree that it's not possible to hit this actual issue as long as we continue to stop instances during a Propolis upgrade. I'm going to confirm that's still the plan for the next release. As he pointed out, _intra-version_ migration is still working fine, so there is no "real" failure here.

Assuming @pfmooney is also OK with it, and that the plan is still to stop instances during an upgrade, we feel it's most expedient to ignore that specific CI failure for the purposes of this PR.

bnaecker:
I've confirmed that we're still planning to take instances offline during the next release cycle. This version incompatibility should not show up in that case.

bin/propolis-server/src/lib/server.rs
```diff
@@ -318,36 +320,111 @@ fn instance_spec_from_request(
...
-async fn register_oximeter(
-    server_context: &DropshotEndpointContext,
-    cfg: &MetricsEndpointConfig,
-    vm_id: uuid::Uuid,
+/// Register an Oximeter server reporting metrics from a new instance.
+///
+/// This spawns a tokio task which will indefinitely attempt to register with
+/// Nexus as an Oximeter metric producer. Once the registration succeeds, it
+/// will also arrange for the statistics produced for the managed instance to be
+/// collected.
```
pfmooney, line 328: Having this loop forever, with no means of the rest of propolis-server to signal its desired termination seems a bit off to me.  Perhaps this task can be bundled into a consolidated oximeter state in `ServiceProviders`?

bin/propolis-server/src/lib/stats.rs
```diff
@@ -4,73 +4,89 @@
 
 //! Methods for starting an Oximeter endpoint and gathering server-level stats.
 
-use dropshot::{
-    ConfigDropshot, ConfigLogging, ConfigLoggingLevel, HandlerTaskMode,
-};
+use dropshot::{ConfigDropshot, HandlerTaskMode};
 use omicron_common::api::internal::nexus::ProducerEndpoint;
 use omicron_common::api::internal::nexus::ProducerKind;
+use oximeter::types::ProducerRegistry;
```
pfmooney, line 10: nit: consolidate w/ `types::` below

bin/propolis-server/src/lib/stats.rs
```diff
@@ -4,73 +4,89 @@
...
 use oximeter::{
     types::{Cumulative, Sample},
-    Metric, MetricsError, Producer, Target,
+    Metric, MetricsError, Producer,
 };
+use oximeter_producer::Error;
```
pfmooney, line 15: nit: combine w/ `use` below

bin/propolis-server/src/lib/stats.rs
```diff
@@ -135,57 +164,77 @@ pub async fn start_oximeter_server(
...
-    id: Uuid,
-    server: &Server,
+///
+/// This attempts to initialize kstat-based metrics for vCPU usage data. This
+/// may fail, in which case those metrics will be unavailable.
+pub async fn register_server_metrics(
+    registry: &ProducerRegistry,
+    virtual_machine: VirtualMachine,
+    #[cfg(all(not(test), target_os = "illumos"))] log: &Logger,
+    #[cfg(not(all(not(test), target_os = "illumos")))] _log: &Logger,
```
pfmooney, line 191: nit: Please let's not have `cfg()` soup inside the function arguments

bin/propolis-server/src/lib/stats/pvpanic.rs
```diff
@@ -26,11 +26,19 @@ pub struct PvpanicProducer {
 
 /// An Oximeter `Metric` that specifies the number of times an instance's guest
 /// reported a guest-handled kernel panic using the QEMU `pvpanic` device.
+//
+// NOTE: We may want to collapse these into a single metric with a field
+// indicating whether the guest or host handle bit was set, but it's not clear
+// that these are truly mutually exclusive. It could also be done with two
+// boolean fields, which would allow either to be set.
```
pfmooney, line 33: File a ticket about this?

bin/propolis-server/src/lib/stats/virtual_machine.rs
```diff
@@ -0,0 +1,530 @@
...
+use chrono::DateTime;
+use chrono::Utc;
+use oximeter::types::Cumulative;
+use oximeter::FieldType;
+use oximeter::FieldValue;
+use oximeter::Metric;
+use oximeter::Sample;
+use oximeter::Target;
+use std::collections::BTreeMap;
+use uuid::Uuid;
```
pfmooney, line 25: style nit: please collapse down these includes

bin/propolis-server/src/lib/stats/virtual_machine.rs
```diff
@@ -0,0 +1,530 @@
...
+// Copyright 2024 Oxide Computer Company
+
+//! Types for tracking statistics about virtual machine instances.
+
+// Propolis is built in a variety of configurations, including checks and tests
+// run on non-illumos machines where kstats are meaningless. This is a big
+// hammer, but a large number of the values in this module are not referenced in
+// those configurations, and so this is more straightfoward than littering the
+// code with cfg directives.
+#![cfg_attr(any(test, not(target_os = "illumos")), allow(dead_code))]
```
pfmooney, line 14: As a follow-up item, I'd like to update kstat-rs to be able to stub out its function bindings, so it can be included (albeit in a non-functional matter) on platforms which lack the library.  Not at all blocking for this, though.

crates/propolis-api-types/src/lib.rs
```diff
@@ -186,6 +194,13 @@ pub struct InstanceProperties {
     pub vcpus: u8,
 }
 
+impl InstanceProperties {
+    /// Return the name of the VM as managed by the VMM.
+    pub fn vm_name(&self) -> String {
+        self.id.to_string()
```
pfmooney, line 200: return the `&str`, and leave it up to the caller if they want to turn it into an owned string?

crates/propolis-api-types/src/lib.rs
```diff
@@ -186,6 +194,13 @@ pub struct InstanceProperties {
     pub vcpus: u8,
 }
 
+impl InstanceProperties {
+    /// Return the name of the VM as managed by the VMM.
```
pfmooney, line 198: Perhaps clearer as "Name of the VMM resource backing this VM"?

bin/propolis-server/src/lib/server.rs
```diff
@@ -36,13 +36,15 @@ use propolis_api_types::instance_spec::{
 
 use propolis_server_config::Config as VmTomlConfig;
 use rfb::server::VncServer;
+use slog::info;
 use slog::{error, o, warn, Logger};
```
pfmooney, line 40: nit: consolidate with include below

bin/propolis-server/src/lib/server.rs
```diff
@@ -105,7 +107,7 @@ pub enum VmControllerState {
     Destroyed {
         /// A copy of the instance properties recorded at the time the instance
         /// was destroyed, used to serve subsequent `instance_get` requests.
-        last_instance: api::Instance,
+        last_instance: Box<api::Instance>,
```
pfmooney, line 110: Why did this get Boxed?  Did the compiler complain about size?

bin/propolis-server/src/lib/server.rs
```diff
@@ -318,36 +320,111 @@ fn instance_spec_from_request(
...
     log: Logger,
-) -> anyhow::Result<ProducerRegistry> {
-    assert!(server_context.services.oximeter_stats.lock().await.is_none());
-    assert!(server_context
-        .services
-        .oximeter_server_task
-        .lock()
+) {
+    assert!(services.oximeter_stats.lock().await.is_none());
+    assert!(services.oximeter_server_task.lock().await.is_none());
```
pfmooney, line 337: Maybe use `replace` when doing the assignments, and assert then that the old values were none?  (Or both, for the ol' belt-n-suspenders)

bin/propolis-server/src/lib/spec.rs
```diff
@@ -608,20 +608,29 @@ impl ServerSpecBuilder {
 #[cfg(test)]
 mod test {
     use crucible_client_types::VolumeConstructionRequest;
+    use propolis_api_types::InstanceMetadata;
     use propolis_api_types::Slot;
```
pfmooney, line 612: nit: combine includes

pfmooney:
> Assuming @pfmooney is also OK with it, and that the plan is still to stop instances during an upgrade, we feel it's most expedient to ignore that specific CI failure for the purposes of this PR.

Yeah, that's fine by me.

bin/propolis-server/src/lib/server.rs
bnaecker: Yeah, there was a clippy lint about the size of the enum in CI. I'm not sure why I hit it now.

bnaecker: For reference: https://github.com/oxidecomputer/propolis/actions/runs/8197097085/job/22418514347#step:5:1753

bin/propolis-server/src/lib/stats/virtual_machine.rs
bnaecker: Yeah, good idea. I filed https://github.com/oxidecomputer/kstat-rs/issues/1 to track this.

crates/propolis-api-types/src/lib.rs
bnaecker: `InstanceProperties::id` is a `Uuid`, so we have to return either the ID itself or an owned string, I think?

bin/propolis-server/src/lib/stats/pvpanic.rs
bnaecker: https://github.com/oxidecomputer/propolis/issues/660

bin/propolis-server/src/lib/stats.rs
bnaecker: Yeah, this was my annoyance showing through. I'll host it up somewhere.

bin/propolis-server/src/lib/server.rs
bnaecker: Done in d61fd69

bin/propolis-server/src/lib/stats.rs
bnaecker: Done in d61fd69

bin/propolis-server/src/lib/stats.rs
bnaecker: Done in d61fd69

bin/propolis-server/src/lib/stats.rs
bnaecker: Move to the function itself in d61fd69

bin/propolis-server/src/lib/stats/virtual_machine.rs
bnaecker: Done in d61fd69

crates/propolis-api-types/src/lib.rs
bnaecker: Reworded in d61fd69

bin/propolis-server/src/lib/server.rs
bnaecker: Done in d61fd69

bin/propolis-server/src/lib/server.rs
bnaecker: Kept these asserts, and added them on the value returned from `Option::replace` as well, in d61fd69.

bin/propolis-server/src/lib/spec.rs
bnaecker: Done in d61fd69

bnaecker:
Thanks for the review @pfmooney. Let me know if you have new comments.

bin/propolis-server/src/lib/server.rs
```diff
@@ -173,6 +174,21 @@ impl VmControllerState {
...
+    /// production.
+    registration_task: Mutex<Option<tokio::task::JoinHandle<()>>>,
+
+    /// The host task for this Propolis server's Oximeter server.
+    server_task: Mutex<Option<tokio::task::JoinHandle<()>>>,
+
+    /// The metrics wrapper for "server-level" metrics, i.e., metrics that are
+    /// tracked by the server itself (as opposed to being tracked by a component
+    /// within an instance).
+    stats: Mutex<Option<crate::stats::ServerStatsOuter>>,
```
pfmooney, line 189: Do these need to be accessed independently, or would it make sense to instead wrap the whole `OximeterState` in a mutex in `ServiceProviders`?

bin/propolis-server/src/lib/server.rs
```diff
@@ -318,36 +342,127 @@ fn instance_spec_from_request(
...
+                    retryable,
+                    msg,
+                } if retryable => {
+                    omicron_common::backoff::BackoffError::transient(msg)
+                }
+                _ => omicron_common::backoff::BackoffError::permanent(
+                    e.to_string(),
+                ),
+            })
+        };
```
pfmooney, line 390: Should `OximeterState` also hold some sort of notification channel to allow `register_with_nexus` or the backoff loop to bail out promptly if the server is being shutdown?

crates/propolis-api-types/src/lib.rs
pfmooney: Ah, oversight on my part.

bin/propolis-server/src/lib/server.rs
bnaecker: Yeah, I was not sure either. The stats might need to be accessed independently, but they're the only thing I expect to be accessed with any frequency. I'll try out hosting the mutex up a layer.

bin/propolis-server/src/lib/server.rs
bnaecker: Sure, I can do that, and signal it where we shutdown all the provider tasks.

bnaecker: Done in e295e14

bin/propolis-server/src/lib/server.rs
bnaecker: Done in e295e14

pfmooney (approved):

bin/propolis-server/src/lib/stats/virtual_machine.rs
```diff
@@ -0,0 +1,526 @@
...
+    fn test_usage() -> VcpuUsage {
+        VcpuUsage {
+            state: "run".to_string(),
+            vcpu_id: 0,
+            datum: Cumulative::new(100),
+        }
+    }
+
+    #[test]
+    fn test_no_schema_changes() {
```
gjcolombo, line 431: Curiosity about metrics in general, not a problem with this change: is the schema ever allowed to change (e.g. if we were to break things down into a different set of microstates, or report some additional statistics, or something else)? Are metrics versioned? Again, not a problem with this PR, just curious about what the ground rules are here.

bin/propolis-server/src/lib/stats/virtual_machine.rs
```diff
@@ -0,0 +1,526 @@
...
+
+        // Now convert the aggregated occupancy times into samples.
+        for (state, datum) in occupancy_by_state.into_iter() {
+            let metric = VcpuUsage { vcpu_id, state, datum };
+            let sample =
+                Sample::new_with_timestamp(snapshot_time, vm, &metric)?;
+            out.push(sample);
+        }
+    }
+    Ok(out)
```
gjcolombo, line 382: Would it be worthwhile to assert here that `out.len() <= original_capacity`? (I'm trying to think of good ways to produce a warning if we start collecting more statistics but don't bump the value of `N_VCPU_MICROSTATES`/whatever constant ultimately governs the size of the sample buffer.)

gjcolombo (approved):
LGTM, just a couple of little questions/comments.

bin/propolis-server/src/lib/stats/virtual_machine.rs
bnaecker: The short answer is "no", unfortunately.

The longer answer is, "it depends on the changes". I've put a bit of thought into this over the years, but it's getting close to time to handle this robustly and correctly. Here's the current state of affairs, broken down by the kind of change.

It is always fine to report _new_ metrics, such as tracking memory usage or guest VNIC usage. That is a completely new timeseries, and can be done whenever we want. This is done by adding a new type which implements `oximeter::Metric`, and generating samples from it. That could still use the `VirtualMachine` target I've added this PR, just with a new metric.

Breaking things down into a different set of microstates would be allowed, but sort of scary. The onus is currently on the person running to query to inspect when those changes occur and handle them. If you did something like sum across all microstates, then the query is still valid; if you looked at a particular microstate, which was further broken down as part of a change, you would see discontinuities. Nothing would _break_ per se, but you've now got a different set of timeseries.

The most difficult (read: disallowed) change would be adding or removing a field of an existing timeseries. That's what this test is for. Right now, `oximeter` will retrieve your (now changed) samples; notice that its schema is different; report a warning; and helpfully chuck your sample into the bin.

This is really Not Great. I think it's certainly possible to handle the addition of _new_ fields gracefully, by updating the schema; using default or NULL values for old samples; and / or handling the schema change on the query side. The _removal_ of fields or a change in their type is pretty difficult, and I'm not sure how we'd handle it in general. At this still-early stage of development, we can always use the nuclear option of completely expunging the timeseries database or an individual timeseries, during an upgrade. That only works though if you've got one version of the schema deployed.

One thing that could help identify changes is to compute an ID or hash for the schema itself, and then effectively consider two versions of the schema to be two separate timeseries. The query side would then handle running queries against both, or returning an error if that can't be done.

Another way to manage this is to put some API in place in Nexus and / or `oximeter` itself to explicitly manage the update of a schema. `oximeter` could then either expunge old data, or add new (default) values for fields, etc. This is sort of complicated because we store the schema in ClickHouse itself, which has extremely weak consistency guarantees. I've floated the idea of moving schema to Cockroach, but that's a big change that I wouldn't take on without good reason.

Anyway, that's the long and unsatisfying story at this point :)

bin/propolis-server/src/lib/stats/virtual_machine.rs
gjcolombo: > Hmm, I'm not sure what we'd be guarding against in that case. I had thought of this only as a value to pre-allocate with, but it seems like you're concerned about a correctness issue. Could you explain that more?

I don't think there's really a correctness issue as such, and in fact the question I meant to ask seems pretty fringy now that I think about it some more.

My real question here is about `KSTAT_LIMIT_PER_VCPU`: that constant reads to me as very lightly load-bearing in the sense that (per its doc comment) it's chosen so that the `KstatSampler` will buffer at least 64 samples even if the producer isn't being polled regularly. If we were to start querying for more states/collecting additional `Sample`s here that buffer would shrink. This is definitely not the end of the world, but it seems like the sort of thing that would be nice to catch if it happens, and this seemed like it might be a place where we could do that.

This is all definitely well into "optional/nice to have" territory, though--the code should work fine as-is.

bnaecker: Ah I see. So one place we'd catch that is the self-statistics that the `KstatSampler` generates. It notes any dropped samples, and produces counters for them which themselves go into the timeseries database. The tradeoff there is really just space. We should have _some_ limit, but 64 is really arbitrary.

gjcolombo: Nice--if we already have some other drop detection then I think this is even less of a concern.

bin/propolis-server/src/lib/stats/virtual_machine.rs
pfmooney: Considering the relative brittleness in the face of changes to the emitted metrics, do you think it would make sense to device them in a module or crate, such that it's clear that any subsequent changes would require the care you mention?

We've started something similar to this with the migration payload data (living in child modules of the various devices, perhaps to be moved to a crate eventually).  The intent there was to make it easier to spot if/when one of those data structures is changed.

bnaecker: @pfmooney Could you explain more what that looks like? I'm not entirely sure I follow. Does that mean something like naming the target/metric struct with a version ID, such as `VirtualMachineV1`?

The test here will catch if the schema changes (though not the microstates), since the expected schema file is checked in as part of this PR. It's not really sufficient, but it will at least get more eyes on any change that modifies those schema.

pfmooney: > Does that mean something like naming the target/metric struct with a version ID, such as `VirtualMachineV1`?

Not specifically that, no.

> The test here will catch if the schema changes (though not the microstates)

But a change in the reported microstates (or rather what we distill them down into) would be a problem, right?

Having the scheme checked in for the test is good.  I guess we can revisit later if/when we have a good mechanism to constrain the microstates too?

bin/propolis-cli/src/main.rs
```diff
@@ -187,11 +196,14 @@ async fn new_instance(
     memory: u64,
     disks: Vec<DiskRequest>,
     cloud_init_bytes: Option<String>,
+    silo_id: Uuid,
+    project_id: Uuid,
```
hawkw, line 200: I don't love seeing two arguments with the same type and different meanings, it seems quite easy to accidentally swap them and not notice --- would it make sense to use @sunshowers' `newtype_uuid` crate for this?

bin/propolis-server/src/lib/server.rs
```diff
@@ -173,6 +174,21 @@ impl VmControllerState {
     }
 }
 
+/// Objects related to Propolis's Oximeter metric production.
+pub struct OximeterState {
+    /// A task spawned at initial startup to register with Nexus for metric
+    /// production.
```
hawkw, line 180: it would be nice if this doc comment explained the circumstances under which the `Notify` is woken?

bin/propolis-server/src/lib/server.rs
```diff
@@ -218,10 +230,18 @@ impl ServiceProviders {
             // Wait for the serial task to exit
             let _ = serial_task.task.await;
         }
-        if let Some(server) = self.oximeter_server_task.lock().await.take() {
+
+        // Clean up oximeter tasks and statistic state.
+        let mut oximeter_state = self.oximeter_state.lock().await;
+        if let Some((shutdown, task)) = oximeter_state.registration_task.take()
+        {
+            shutdown.notify_one();
```
hawkw, line 238: nit, take it or leave it: since this is used to signal shutdown only a single time, it might make more sense to use a `tokio::sync::oneshot` of `()`s? `Notify` is a bit more complex as it can be woken multiple times. not a big deal.

bin/propolis-server/src/lib/stats/virtual_machine.rs
bnaecker: A change in the microstates would be confusing, at the very least. The data would continue to be reported and stored in ClickHouse just fine, but it'd be hard to follow, I think. Suppose we decided to break the `"waiting"` microstate into `"waiting_for_this"` and `"waiting_for_that"`. When that change happened, in the database you'd see the end of the timeseries containing the `"waiting"` state (no more samples from it), and the genesis of the two with the other state names (no samples before the change, only after).

I think it's possible to get a consistent data representation in this case, by aggregating the two later states. Since we've explicitly set this up so that their sum is the same as the `"waiting"` state before, that would be OK. But you're certainly correct that doing something that is not a simple, linear mapping between the states before and after the change is much harder to grok.

> I guess we can revisit later if/when we have a good mechanism to constrain the microstates too?

So here's one possible way to check this now:

- Create an array of some number of states, spanning the set of lower-level `kstat` vCPU microstates we expect to produce. Make the number of each state different.
- Iterate through those, and convert to the state we report to `oximeter`.
- Sum the count of each of those
- Assert that is the same as we expect from the input we've created.

This is pretty contrived, but I think it would catch a change in the mapping from `kstat`-layer states into `oximeter`-layer. That seems worth doing, no?

bin/propolis-server/src/lib/stats/virtual_machine.rs
```diff
@@ -0,0 +1,526 @@
...
+        return Some(kstat.ks_instance);
+    }
+    None
+}
+
+// Produce `Sample`s for the `VcpuUsage` metric from the relevant kstats.
+fn produce_vcpu_usage<'a>(
+    vm: &'a VirtualMachine,
+    vcpu_stats: impl Iterator<Item = &'a (DateTime<Utc>, Kstat<'a>, Data<'a>)> + 'a,
+) -> Result<Vec<Sample>, Error> {
```
hawkw, line 334: IIUC, we'll bail here if we encounter any invalid values in the kstats. what's the rationale for that versus returning a `Vec` that only contains the good samples?

bin/propolis-server/src/lib/stats/virtual_machine.rs
```diff
@@ -0,0 +1,526 @@
...
+            "Should have samples for 'run' and 'idle' states"
+        );
+        for ((sample, state), x) in
+            samples.iter().zip(["idle", "run"]).zip([2, 2])
+        {
+            let st = sample
+                .fields()
+                .iter()
+                .find_map(|f| {
+                    if f.name == "state" {
```
hawkw, line 509: these state names are also used above, would it make sense for them to be constants?

bin/propolis-server/src/lib/stats/virtual_machine.rs
bnaecker: Ok, I've added a test, which I'll include in the next commit. It looks tautological, but I think it's useful. It catches the following:

- Removing one of the two kstat-states that map to the oximeter `"idle"` state
- Mapping that state to a new oximeter state
- Mapping both of those states to a single oximeter state, but with a different name.
- Mapping a new kstat state into an existing oximeter state

The one thing it does not catch is if the `kstat` layer grows a new state, or removes one. This just checks the mapping between the two. We could certainly add such a test, provided we can spawn VMs so that the `kstat`s show up in userland.

bin/propolis-server/src/lib/stats/virtual_machine.rs
bnaecker: It's a good question. I'm sort of operating under the assumption that if fetching one fails, they all will. I believe that's reasonable, based on my experience with the `kstat` library itself, but I could be missing something. I have a similar issue in omicron (https://github.com/oxidecomputer/omicron/issues/4863), where we could actually opt to produce explicitly-missing samples in this case.

I would be happy to make a follow-up issue to do that, if it seems like a good idea.

bin/propolis-server/src/lib/stats/virtual_machine.rs
bnaecker: This test was added in 7c4cbe7

bin/propolis-server/src/lib/server.rs
bnaecker: Done in 7c4cbe7

bin/propolis-server/src/lib/server.rs
bnaecker: Changed to a oneshot in 7c4cbe7

bin/propolis-server/src/lib/stats/virtual_machine.rs
bnaecker: Made the state names all consts, and did a few other bits in 7c4cbe7

bin/propolis-cli/src/main.rs
bnaecker: Sure, that seems like an easy win!

bnaecker: Done in 60e6c07

bnaecker:
Thanks @pfmooney, @gjcolombo, and @hawkw for all the input on this one. Let me know if y'all have any more thoughts.

bin/propolis-server/src/lib/stats/virtual_machine.rs
hawkw: Thanks, I think that's entirely reasonable --- I was just curious about the motivation here. It might be nice to add a comment about this, but it's certainly not a blocker.

hawkw (approved):
looks good to me, thanks for addressing my suggestions!

PR #661, opened 2024-03-13
server: separate statuses of inbound and outbound migrations

### LINKED ISSUE #508: want migration status endpoints to report source and target migration IDs separately
Propolis's instance status endpoint only reports the ID of the most recent migration that it participated in, irrespective of whether this was a migration in or a migration out. This causes sled agent to have to assume that when Propolis reports a migration ID that sled agent and Nexus don't know about, it's because the Propolis was previously initialized via migration in.

If instead Propolis separately reported incoming and outgoing migration IDs, sled agent could raise warnings of a misbehaving Propolis if it ever saw one report an in-progress migration out whose ID wasn't known to sled agent. Such cases are bugs because Nexus and the sled agents involved in a migration are supposed to agree on the new migration's ID before ever communicating it to a Propolis.

This is a nice-to-have to simplify some sled agent logic. It's not a showstopper for control plane-directed live migration.

gjcolombo:
**N.B.** This is a breaking change to the Propolis API that will need a corresponding Omicron change before it can be merged there.

Change the way the server reports migration statuses:

- Instead of asking clients to identify a specific migration to query, create a `migration-status` endpoint that reports on all migrations currently known to the server. (The server wasn't keeping any history of prior migration attempts anyway, so specifying an ID wasn't buying very much; if the client cares about a specific ID they can compare it to the IDs the server now returns.)
- When reporting migration status (either through the migration status endpoint or the instance monitor endpoint), report the ID and status of both the inbound migration (if there was one) and the most recent attempted outbound migration.

This should allow us to simplify a few things in sled agent. Today:

- Sled agent has to handle mismatched migration IDs carefully to cover cases such as this one:
  - Propolis was initialized via incoming migration with ID X
  - Sled agent knows about outgoing migration with ID Y
  - Propolis reports an instance state change that still bears ID X This sort of case is easier to reason about (or at least to make fewer assumptions about) when the outbound and inbound migrations are properly distinguished.
- If Propolis reports a status of "destroyed, migration finished," it's ambiguous whether the VM was migrated into and then stopped or was migrated out of. With these changes it's obvious that a successful migration-out occurred.

Additional fixes/improvements:

- The `pending_migration_id` field in the VM controller wasn't being used for anything; now it is.
- Add a log line to the code path that publishes a new externally-visible instance state.

Tests:

- cargo test
- local PHD run (migration-from-base tests are broken because of the breaking API change)
- ran an ad hoc live migration with propolis-cli to test both that the CLI works properly and that the right messages appear in the Propolis logs

Fixes #508.


gjcolombo:
> This should be OK to review as-is

> falcon Failing after 4m — Failure! 

Oops. I think this is just an out-of-date OpenAPI doc--will work on the fix.

crates/propolis-api-types/src/lib.rs
```diff
@@ -77,15 +77,16 @@ pub struct InstanceMigrateStartRequest {
...
+    pub state: MigrationState,
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq)]
 pub struct InstanceMigrateStatusResponse {
-    pub migration_id: Uuid,
-    pub state: MigrationState,
+    pub migration_in: Option<InstanceMigrationStatus>,
+    pub migration_out: Option<InstanceMigrationStatus>,
 }
```
gjcolombo, line 90: NTS: These structs need doc comments (in particular it should be made clearer that it's possible for a server to report both a prior migration in and a current migration out, and that migration statuses don't get cleared by the conclusion of a migration).

bin/propolis-cli/src/main.rs
```diff
@@ -533,12 +533,40 @@ async fn migrate_instance(
...
+                };
+
+                // The destination should start reporting migration status as
+                // soon as the ensure request completes. The source may not
+                // have a migration status yet because the request from the
+                // destination needs to arrive first.
+                let Some(migration) = migration else {
+                    if role == "dst" {
+                        anyhow::bail!(
+                            "{role} instance's migration ID wasn't set"
```
hawkw, line 552: considered nitpicky: since this is inside an `if role == "dst"`, do we really need to format it here?

crates/propolis-api-types/src/lib.rs
```diff
@@ -77,15 +77,32 @@ pub struct InstanceMigrateStartRequest {
...
+    /// The current phase the migration is in.
+    pub state: MigrationState,
 }
 
+/// The statuses of the most recent live migrations a Propolis VM has
+/// participated in. Note that Propolis will continue reporting the status of a
+/// completed migration unless it is supplanted by a new migration in the same
+/// direction. This means it's possible for both the inbound and outbound
+/// migration statuses to be set, e.g. if a Propolis was initialized via
+/// migration in and is now the source of a migration out.
```
hawkw, line 96: It might be nice if this comment discussed the motivation behind reporting both migrations here, rather than e.g. just an `enum MigrationRole` or similar?

hawkw (approved):
Excited to get this merged! I just had a couple fairly minor nits.

crates/propolis-api-types/src/lib.rs
```diff
@@ -83,15 +83,37 @@ pub struct InstanceMigrateStartRequest {
...
+///
+/// This structure only reports the status of the most recent migration in a
+/// single direction. That is, if a migration in or out fails, and a new
+/// migration attempt begins, the new migration's status replaces the old's.
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq)]
 pub struct InstanceMigrateStatusResponse {
-    pub migration_id: Uuid,
-    pub state: MigrationState,
+    /// The status of the most recent attempt to initialize the current instance
+    /// via migration in, or None if the instance has never been a migration
```
hawkw, line 111: extremely nitpicky (I don't actually care about this): 
```suggestion
    /// via migration in, or `None` if the instance has never been a migration
```

crates/propolis-api-types/src/lib.rs
```diff
@@ -83,15 +83,37 @@ pub struct InstanceMigrateStartRequest {
...
+    /// target.
+    pub migration_in: Option<InstanceMigrationStatus>,
+    /// The status of the most recent attempt to migrate out of the current
+    /// instance, or None if the instance has never been a migration source.
```
hawkw, line 115: similarly:
```suggestion
    /// instance, or `None` if the instance has never been a migration source.
```

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -108,43 +139,98 @@ where
...
+    fn get_migration_status(&self) -> ApiMigrateStatusResponse {
+        self.api_state_tx.borrow().migration.clone()
+    }
+
+    /// Sets the published instance and/or migration state and increases the
+    /// state generation number.
+    ///
+    /// # Panics
+    ///
+    /// Panics if both `instance_state` and `migration_state` are `None`.
```
hawkw, line 156: this is a bit of a bummer, I kind of wonder if we should instead have an 
```rust
enum PublishedState {
    Instance(ApiInstanceState),
    Migration(PublishedMigrationState),
    Both(ApiInstanceState, PublishedMigrationState),
}
```
or similar? On the other hand, if that's too annoying to work with, I don't think it's a huge deal...

hawkw (approved):
Excited to get this one merged! I had some very unimportant nits, but feel free to ignore them --- I'd like to get this wired through to sled-agent as soon as possible. :)

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: This was bothering me too on self-review. I'll have a fix in my next push.

hawkw:
Looks like the OpenAPI specs need to be regenerated: https://github.com/oxidecomputer/propolis/actions/runs/9551503512/job/26325948032?pr=661

gjcolombo:
Once more with feeling!

PR #662, opened 2024-03-15
nvme: Expose MDTS value with basic logic

pfmooney:
While the interaction between the block backends and frontends like NVMe could be improved to include negotiation of parameters such as MDTS, it would be nice to define a reasonable value in the mean time.

### DIFF
--- a/lib/propolis/src/hw/nvme/mod.rs
+++ b/lib/propolis/src/hw/nvme/mod.rs
@@ -411,6 +411,20 @@ impl NvmeCtrl {
         b << (self.ns_ident.lbaf[(self.ns_ident.flbas & 0xF) as usize]).lbads
     }
 
+    /// Check a given request transfer size against any MDTS configured on the
+    /// controller.
+    ///
+    /// Returns `true` if the size is acceptable for the MDTS.
+    fn valid_for_mdts(&self, sz: u64) -> bool {
+        match self.ctrl_ident.mdts {
+            0 => true,
+            mdts => {
+                let limit = (self.ctrl.cap.mpsmin_sz() as u64) << mdts;
+                sz <= limit
+            }
+        }
+    }
+
     fn update_block_info(&mut self, info: block::DeviceInfo) {
         let nsze = info.total_size;
         self.ns_ident = bits::IdentifyNamespace {
@@ -496,7 +510,11 @@ pub struct PciNvme {
 
 impl PciNvme {
     /// Create a new pci-nvme device with the given values
-    pub fn create(serial_number: String, log: slog::Logger) -> Arc<Self> {
+    pub fn create(
+        serial_number: String,
+        mdts: Option<u8>,
+        log: slog::Logger,
+    ) -> Arc<Self> {
         let builder = pci::Builder::new(pci::Ident {
             vendor_id: VENDOR_OXIDE,
             device_id: PROPOLIS_NVME_DEV_ID,
@@ -526,6 +544,7 @@ impl PciNvme {
             ssvid: VENDOR_OXIDE,
             sn,
             ieee: OXIDE_OUI,
+            mdts: mdts.unwrap_or(0),
             // We use standard Completion/Submission Queue Entry structures with no extra
             // data, so required (minimum) == maximum
             sqes: NvmQueueEntrySize(0).with_maximum(sqes).with_required(sqes),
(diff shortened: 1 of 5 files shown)


lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -496,7 +496,11 @@ pub struct PciNvme {
 
 impl PciNvme {
     /// Create a new pci-nvme device with the given values
-    pub fn create(serial_number: String, log: slog::Logger) -> Arc<Self> {
+    pub fn create(
+        serial_number: String,
+        mdts: Option<u8>,
```
luqmana, line 501: Since we're hardcoding this arg in all call sites right now can we just move it into the body for now? As-is I think it can be a bit hard to get correct and ensure a valid value is given. Case in point, `MDTS` of 20 means we're saying a max size of `2^20 * CAP.MPSMIN (4K) = 4GiB` and not the intended `1MiB`.

lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -519,13 +523,21 @@ impl PciNvme {
         let mut sn: [u8; 20] = [0u8; 20];
         sn[..sz].clone_from_slice(&serial_number.as_bytes()[..sz]);
 
+        // If MDTS is specified, ensure that it is either 0 (meaning no limit)
+        // or is at least as large as a single page.
+        let mdts = match mdts {
+            None | Some(0) => 0,
+            Some(v) => v.max(PAGE_SHIFT as u8),
+        };
```
luqmana, line 531: Since we're hardcoding the arg in all call sites right now can we just move it into the body for now? As-is I think it can be a bit hard to get correct and ensure a valid value is given. Case in point, 20 means we're saying a max size of `2^20 * CAP.MPSMIN (4K) = 4GiB` and not the intended `1MiB` 😅️

(Also, should that `max` have been a `min`?)

luqmana (changes requested):

lib/propolis/src/hw/nvme/mod.rs
pfmooney: Yeah, this is wrong.  OVMF in particular freaks out, since it uses a u32 for calculating the max transfer size.  I'll be posting an updated wad shortly.

pfmooney: The obvious math bug (deviating from the spec) should be fixed now.

pfmooney:
Test notes: I was able to force an illumos guest into issuing a too-large IO in order to observe its rejection from the device.

lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -411,6 +411,21 @@ impl NvmeCtrl {
...
 
+    /// Check a given request transfer size against any MDTS configured on the
+    /// controller.
+    ///
+    /// Returns `true` if the size is acceptable for the MDTS.
+    fn valid_for_mdts(&self, sz: u64) -> bool {
+        match self.ctrl_ident.mdts {
+            0 => true,
+            mdts => {
+                // MPSMIN is currently locked to PAGE_SIZE
```
luqmana, line 422: nit: I know we assume `PAGE_SIZE` various places but can just grab `MPSMIN` off of `self` here so there's one less place to remember.

lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -496,7 +511,11 @@ pub struct PciNvme {
 
 impl PciNvme {
     /// Create a new pci-nvme device with the given values
-    pub fn create(serial_number: String, log: slog::Logger) -> Arc<Self> {
+    pub fn create(
+        serial_number: String,
+        mdts: Option<u8>,
```
luqmana, line 516: I know we eventually want this coordinated with the block layer but given both call sites just hardcode the same value now I'm inclined to move it into the body here. We already have a whole bunch of other controller config/capabilities hardcoded there.

pfmooney: I'd like to leave it as-is for now.  It's good incentive to make sure a tunable like this is addressed when some of the other migration/block changes go in.

lib/propolis/src/hw/nvme/mod.rs
pfmooney: Done.

lib/propolis/src/hw/nvme/mod.rs
luqmana: sounds good

luqmana (approved):

pfmooney:
Merged in 6dceb9ef69c217cb78a2018bbedafbc19f6ec1af

PR #663, opened 2024-03-15
Bump crucible and update APIs

mkeeter:
This bumps Crucible to the latest `main`.  https://github.com/oxidecomputer/crucible/pull/1206 makes some changes to the `BlockIO::write` API, so it requires minor updates to Propolis as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -741,7 +741,7 @@ dependencies = [
 [[package]]
 name = "crucible"
 version = "0.0.1"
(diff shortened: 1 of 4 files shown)


leftwo (approved):
So, I've not much experience with using the `MaybeUninit` stuff, but, from my reading
of their documentation, this seems good.

There is a buch of similar looking kinds of shenanigans going on in this file, so It's
probably best if we have another set of eyes to make sure this is all good.

I say shenanigans in a positive light here :)

pfmooney (approved):

PR #669, opened 2024-03-21
Add `IntrPin::import_state` and migrate LPC UART pin states

hawkw:
Currently, interrupt pin state for userspace-emulated devices, like the
LPC UART, is accounted for both by the kernel in Bhyve and userspace in
Propolis. When userspace wishes to assert an IRQ for such a device, it
records the assertion of the IRQ line in userspace and then calls the
`ioctl` that asserts the IRQ in kernelspace.  However, when such devices
are _migrated_, only the kernelspace Bhyve state is migrated, and any
userspace-tracked pin assertion counts are not. This means that these
states may become out of sync during a migration, potentially resulting
in a lost interrupt.

This branch introduces a mechanism for migrating userspace's
understanding of pin counts, in addition to migrating kernelspace's.
Because incrementing the userspace-tracked pin counts through the
`IntrPin::assert` or `IntrPin::set_state` methods may generate an
`ioctl` that changes the kernelspace state of that interrupt pin, we
cannot just call those methods while importing userspace pin assertion
counts. Instead, I've introduced a new `import_state` method to the
`IntrPin` trait, which is intended to be called only during imports.
This method sets the userspace-tracked pin state, the same as
`set_state`, but will *not* perform an `ioctl` call. This way, we can
import the userspace state independently of the kernelspace state,
without triggering additional interrupts that don't actually exist. I've
modified the LPC UART to use this new mechanism.

I've manually tested this change by running a migration between two lab
machines while mashing keys into the serial console, and the console
does not appear to get stuck.

### DIFF
--- a/lib/propolis/src/intr_pins.rs
+++ b/lib/propolis/src/intr_pins.rs
@@ -27,6 +27,8 @@ pub trait IntrPin: Send + Sync + 'static {
             self.deassert();
         }
     }
+
+    fn import_state(&self, is_asserted: bool);
 }
 
 /// Describes the operation to take with an interrupt pin.
@@ -92,6 +94,17 @@ impl LegacyPIC {
         Some(LegacyPin::new(irq, Arc::downgrade(self)))
     }
 
+    fn import_irq(&self, op: PinOp, irq: u8) {
+        assert!(irq < PIN_COUNT);
+
+        let mut inner = self.inner.lock().unwrap();
+
+        // Update our tracked pin level count, but *don't* actually perform the
+        // ioctl to assert the bhyve interrupt, since the kernelspace pin states
+        // are imported separately.
+        inner.pins[irq as usize].process_op(&op);
+    }
+
     fn do_irq(&self, op: PinOp, irq: u8) {
         assert!(irq < PIN_COUNT);
 
@@ -153,6 +166,17 @@ impl IntrPin for LegacyPin {
         let asserted = self.asserted.lock().unwrap();
         *asserted
     }
+    fn import_state(&self, is_asserted: bool) {
+        let mut asserted = self.asserted.lock().unwrap();
+        if *asserted != is_asserted {
+            *asserted = is_asserted;
+            if let Some(pic) = Weak::upgrade(&self.pic) {
+                let op =
+                    if is_asserted { PinOp::Assert } else { PinOp::Deassert };
+                pic.import_irq(op, self.irq);
+            }
+        }
+    }
 }
 
 /// Interrupt pin which calls a provided function on rising and falling edges.
@@ -188,6 +212,13 @@ impl IntrPin for FuncPin {
         let inner = self.0.lock().unwrap();
         inner.level
     }
+    fn import_state(&self, is_asserted: bool) {
+        let mut inner = self.0.lock().unwrap();
+        if inner.level != is_asserted {
+            inner.level = is_asserted;
+            (inner.func)(inner.level);
+        }
+    }
 }
 struct FPInner {
     level: bool,
@@ -203,4 +234,5 @@ impl IntrPin for NoOpPin {
     fn is_asserted(&self) -> bool {
         false
     }
+    fn import_state(&self, _: bool) {}
 }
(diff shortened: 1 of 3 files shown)


hawkw:
Clippy is [now complaining about some code that didn't change in this branch](https://github.com/oxidecomputer/propolis/actions/runs/8381496400/job/22953147966), must be lints added in the latest Rust version.

lib/propolis/src/intr_pins.rs
```diff
@@ -27,6 +27,8 @@ pub trait IntrPin: Send + Sync + 'static {
             self.deassert();
         }
     }
+
+    fn import_state(&self, is_asserted: bool);
```
pfmooney, line 31: I know the rest of this trait is missing doc comments, but for something like this which is less obvious in its use, a good explanation would be useful for both consumers and implementers.

lib/propolis/src/intr_pins.rs
```diff
@@ -188,6 +212,13 @@ impl IntrPin for FuncPin {
         let inner = self.0.lock().unwrap();
         inner.level
     }
+    fn import_state(&self, is_asserted: bool) {
+        let mut inner = self.0.lock().unwrap();
+        if inner.level != is_asserted {
+            inner.level = is_asserted;
+            (inner.func)(inner.level);
+        }
+    }
```
pfmooney, line 221: I'm not certain we want to call the underlying function here?  Do we want the edge to be visible while importing state?

pfmooney (commented):
This looks good from a high level.

lib/propolis/src/intr_pins.rs
hawkw: I wasn't sure what to do for `FuncPin`, since it's not actually exercised in this change...I'll look into the uses of it and see what their functions actually do.

lib/propolis/src/intr_pins.rs
hawkw: How do you feel about the comment from 1412f9cddb1c52ef8d0ecc610e1def32cb0e08a7?  I'd love any suggestions on how to describe this more clearly.

pfmooney: I suppose I would focus more on the edge not being visible to the _consumer_.  In the case of `FuncPin`, we're not emitting guest-sourced events to the greater hypervisor (propolis-server).  Including a specific example of the kernel-vmm-managed state being left unchanged as well would probably also be appropriate.

> This method is intended for use when importing a guest.

I'd probably move this up in the description, since it's key info about the motivations here.

hawkw: Okay, I gave it another pass. Hopefully this one isn't too wordy.

lib/propolis/src/intr_pins.rs
```diff
@@ -28,6 +28,13 @@ pub trait IntrPin: Send + Sync + 'static {
         }
     }
 
+    /// Set the state of this interrupt pin *without* treating the state change
+    /// as an edge in the interrupt line.
```
pfmooney: I would have kept this as the top-line bit of the comment.  It's a nice summary to be backed up by the additional detail.

jordanhendricks (approved):

lib/propolis/src/intr_pins.rs
hawkw: Oh, okay, I think I misread your comment in https://github.com/oxidecomputer/propolis/pull/669#discussion_r1536254841 as suggesting that we replace that line with one about imports. I'll put it back!

pfmooney: My fault for not making it clear in my comment.

pfmooney (approved):

lib/propolis/src/intr_pins.rs
hawkw: Put it back in https://github.com/oxidecomputer/propolis/pull/669/commits/2243baa115db2aa8152f1f047e9e0e56f9be35c4 --- what do you think?

pfmooney: turbo nit: could probably do without the emphasis on the second `*without*`

otherwise... ship it

hawkw: good call, removed it. going to go ahead and merge now, since all the stuff
that's changed since the last CI build is docs.

PR #672, opened 2024-03-27
Make block attachment mechanism more generic

### LINKED ISSUE #671: want generic device attachment interface
As part of the long-since-merged block layer rework, propolis gained the concept of backend providers (file, crucible) attaching to frontend device emulation (nvme, virtio).  This was useful for decoupling the initialization of those respective backend and device instances from each other and their operation.  This is a similar pattern we could extend to other resources such as chardevs (UARTs), NICs (viona), and potentially HID (ramfb, keyboard) use cases.  It would be nice if that attachment logic was generic enough to avoid too much duplicated logic between such applications.

pfmooney:
The frontend/backend attachment pattern will be more useful for other device types (chardev, viona, etc).  While those new consumers are not yet built, we can abstract out the bits most likely to be common.

Fixes #671

### DIFF
--- a/bin/propolis-standalone/src/main.rs
+++ b/bin/propolis-standalone/src/main.rs
@@ -316,6 +316,7 @@ impl Instance {
         state: State,
         guard: &MutexGuard<InstState>,
         first_boot: bool,
+        log: &slog::Logger,
     ) {
         for (name, device) in guard.inventory.devs.iter() {
             match state {
@@ -359,8 +360,16 @@ impl Instance {
                 }
             }
             State::Halt => {
-                for (_name, be) in guard.inventory.block.iter() {
-                    be.halt();
+                for (name, be) in guard.inventory.block.iter() {
+                    be.stop();
+                    if let Err(err) = be.detach() {
+                        slog::error!(
+                            log,
+                            "Error during detach of block backend {}: {:?}",
+                            name,
+                            err
+                        );
+                    }
                 }
             }
             _ => {}
@@ -425,6 +434,7 @@ impl Instance {
                         State::Run,
                         &guard,
                         inner.boot_gen.load(Ordering::Acquire) == 0,
+                        &log,
                     );
                     if needs_resume {
                         let machine = guard.machine.as_ref().unwrap();
@@ -449,6 +459,7 @@ impl Instance {
                         State::Quiesce,
                         &guard,
                         false,
+                        &log,
                     );
                     let machine = guard.machine.as_ref().unwrap();
                     machine.hdl.pause().expect("pause should complete");
@@ -461,7 +472,12 @@ impl Instance {
                     }
                 }
                 State::Halt => {
-                    Self::device_state_transition(State::Halt, &guard, false);
+                    Self::device_state_transition(
+                        State::Halt,
+                        &guard,
+                        false,
+                        &log,
+                    );
                     for mut vcpu_ctrl in guard.vcpu_tasks.drain(..) {
                         vcpu_ctrl.exit();
                     }
@@ -481,7 +497,12 @@ impl Instance {
                         cur_ev = Some(InstEvent::ReqHalt);
                         continue;
                     }
-                    Self::device_state_transition(State::Reset, &guard, false);
+                    Self::device_state_transition(
+                        State::Reset,
+                        &guard,
+                        false,
+                        &log,
+                    );
                     let machine = guard.machine.as_ref().unwrap();
                     machine.reinitialize().unwrap();
                     machine.vcpu_x86_setup().unwrap();
@@ -1104,7 +1125,7 @@ fn setup_instance(
                         .register_instance(&vioblk, &bdf.to_string());
                     guard.inventory.register_block(&backend, name);
 
-                    block::attach(backend, vioblk.clone());
+                    block::attach(vioblk.clone(), backend).unwrap();
                     chipset_pci_attach(bdf, vioblk);
                 }
                 "pci-virtio-viona" => {
@@ -1139,7 +1160,7 @@ fn setup_instance(
                     guard.inventory.register_instance(&nvme, &bdf.to_string());
                     guard.inventory.register_block(&backend, name);
 
-                    block::attach(backend, nvme.clone());
+                    block::attach(nvme.clone(), backend).unwrap();
                     chipset_pci_attach(bdf, nvme);
                 }
                 qemu::pvpanic::DEVICE_NAME => {
--- a/lib/propolis/src/attachment.rs
+++ b/lib/propolis/src/attachment.rs
@@ -1,0 +1,279 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
+
+/// Frontend (emulated device) attachment point
+pub struct FrontAttachment<T>(Mutex<AttachPoint<T>>);
+impl<T> FrontAttachment<T> {
+    pub fn new() -> Self {
+        Self(Mutex::new(AttachPoint::default()))
+    }
+    pub fn access<R>(
+        &self,
+        when_attached: impl FnOnce(&T, &AttachState) -> R,
+    ) -> Option<R> {
+        self.0.lock().unwrap().access_fallback(when_attached, |_state| ())
+    }
+    pub fn access_fallback<R>(
+        &self,
+        when_attached: impl FnOnce(&T, &AttachState) -> R,
+        fallback: impl FnOnce(&AttachState) -> (),
+    ) -> Option<R> {
+        self.0.lock().unwrap().access_fallback(when_attached, fallback)
+    }
+}
+
+/// Backend (resource provider) attachment point
+pub struct BackAttachment<T>(Mutex<AttachPoint<T>>);
+impl<T> BackAttachment<T> {
+    pub fn new() -> Self {
+        Self(Mutex::new(AttachPoint::default()))
+    }
+    pub fn access<R>(
+        &self,
+        when_attached: impl FnOnce(&T, &AttachState) -> R,
+    ) -> Option<R> {
+        self.0.lock().unwrap().access_fallback(when_attached, |_state| ())
+    }
+    pub fn access_fallback<R>(
+        &self,
+        when_attached: impl FnOnce(&T, &AttachState) -> R,
+        fallback: impl FnOnce(&AttachState) -> (),
+    ) -> Option<R> {
+        self.0.lock().unwrap().access_fallback(when_attached, fallback)
+    }
+}
+
+enum AttachPoint<T> {
+    Attached(Arc<(AttachState, T)>),
+    Detached(AttachState),
+}
+impl<T> Default for AttachPoint<T> {
+    fn default() -> Self {
+        Self::Detached(AttachState::new())
+    }
+}
+impl<T> AttachPoint<T> {
+    fn access_fallback<R>(
+        &self,
+        when_attached: impl FnOnce(&T, &AttachState) -> R,
+        fallback: impl FnOnce(&AttachState) -> (),
+    ) -> Option<R> {
+        match self {
+            AttachPoint::Attached(data) => {
+                Some(when_attached(&data.1, &data.0))
+            }
+            AttachPoint::Detached(state) => {
+                fallback(state);
+                None
+            }
+        }
+    }
+}
+
+/// State of an Attachment ([FrontAttachment] or [BackAttachment])
+#[derive(Clone)]
+pub struct AttachState(Arc<AttachStateInner>);
+impl AttachState {
+    /// Is this Attachment attached to a corresponding peer?
+    ///
+    /// Because the [AttachState] can be cloned to outlive its initial access
+    /// via [FrontAttachment::access] or [BackAttachment::access], the ability
+    /// to check attachedness remains relevant.
+    pub fn is_attached(&self) -> bool {
+        self.0.is_attached.load(Ordering::Acquire)
+    }
+    /// Is the Attachment frontend paused?
+    pub fn is_paused(&self) -> bool {
+        self.0.is_paused.load(Ordering::Acquire)
+    }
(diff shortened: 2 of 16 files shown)


lib/propolis/src/block/attachment.rs
```diff
@@ -0,0 +1,362 @@
...
+            // into BackAttachment::access()
+            data.next_req(state).map_err(|_e| (data.clone(), state.clone()))
+        })? {
+            Ok(req) => Some(req),
+            Err((be, state)) => be.block_for_req(&state),
+        }
+    }
+
+    /// Get a [Waiter], through which one can asynchronously poll for new
+    /// requests via [`Waiter::for_req()`].  Will return [`None`] if no device
```
hawkw, line 221: Very silly nit: inconsistent use of backticks in the links here:
```suggestion
    /// Get a [`Waiter`], through which one can asynchronously poll for new
    /// requests via [`Waiter::for_req()`].  Will return [`None`] if no device
```

lib/propolis/src/block/attachment.rs
```diff
@@ -0,0 +1,362 @@
...
+    pub fn for_req(&self) -> WaitForReq<'_> {
+        WaitForReq { be: &self.0, att_state: &self.1, wait: self.1.notified() }
+    }
+}
+
+pin_project! {
+    /// Future returned from [`Waiter::wait_for_req()`]
+    ///
+    /// This future is not fused, so it can be repeatedly polled for additional
+    /// [`Request`]s as they become available.
```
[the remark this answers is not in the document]
pfmooney: No existing consumer depends on this style of behavior (subsequent calls to `WaitForReq::poll()` after receiving a `Ready`), so I'll just remove the note for now.  Turning this into a `Stream` would complicate existing consumers, since they would then need to take care of pinning the future themselves (since the pin projection makes `WaitForReq` `!Unpin`)

If a future consumer wanted `Stream` behavior, it wouldn't be hard to impl the trait as a call to this unfused `poll()`

lib/propolis/src/block/attachment.rs
pfmooney: I'm leaving them around the `for_req` references, because otherwise the nvim-lsp renderer complains (mildly) about the markdown formatting and the underscore.  Otherwise I'll clean up the rest.

lib/propolis/src/attachment.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    fn new() -> Self {
+        Self(Arc::new(AttachStateInner::default()))
+    }
+    fn merge(front: &Self, back: &Self) -> Self {
+        Self(Arc::new(AttachStateInner {
+            is_attached: AtomicBool::new(false),
+            is_paused: AtomicBool::new(front.is_paused()),
+            is_stopped: AtomicBool::new(back.is_stopped()),
```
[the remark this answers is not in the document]
pfmooney: Added a comment here to explain that.

bin/propolis-standalone/src/main.rs
```diff
@@ -357,8 +357,13 @@ impl Instance {
                 }
             }
             State::Halt => {
-                for (_name, be) in guard.inventory.block.iter() {
-                    be.halt();
+                for (name, be) in guard.inventory.block.iter() {
+                    be.stop();
+                    if let Err(err) = be.detach() {
+                        eprintln!(
```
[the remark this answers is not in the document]
pfmooney: Done

lib/propolis/src/attachment.rs
```diff
@@ -0,0 +1,287 @@
...
+        if !self.0.is_stopped.swap(true, Ordering::Release) {
+            self.0.notify.notify_waiters();
+        }
+    }
+    /// Mark Attachment backend as started (no longer stopped)
+    pub fn start(&self) {
+        if self.0.is_stopped.swap(false, Ordering::Release) {
+            self.0.notify.notify_waiters();
+        }
+    }
```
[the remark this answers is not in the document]
pfmooney: I would have assumed the `watch` API was more hostile to synchronous callers, but it's mostly reasonable, so that's good to know.

All the same, it'd still require pinging the Condvar for the synchronous waiters, so I think I'm going to leave this as-is for now.  Maybe conditions will dictate otherwise in the future.

lib/propolis/src/attachment.rs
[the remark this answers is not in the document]
pfmooney: Yeah, I agree.  I wanted to get the more widely used `block::attachment` stuff documented first, and didn't want to burn up a bunch of effort on the general `attachment` stuff until it was further through CR (suggesting some measure of usefulness)

lib/propolis/src/attachment.rs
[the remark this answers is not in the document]
hawkw: Synchronous waiters _could_ potentially use a runtime's [`Handle::block_on`](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on) and do `handle.block_on(watch.changed())`...but i think the condvar is probably much more efficient, as the `block_on` would spawn a task. So, I agree, let's carry on with this.

lib/propolis/src/tasks.rs
```diff
@@ -412,3 +416,94 @@ impl Drop for CtrlHeld<'_> {
...
+            match result {
+                Err(e) => first_err = Some(e),
+                Ok(hdl) => guard.push(hdl),
+            }
+        }
+        if let Some(e) = first_err {
+            Err(e)
+        } else {
+            Ok(())
+        }
```
hawkw, line 494: style nit, take it or leave it: we could maybe simplify this to
```suggestion
        let mut first_err = Ok(());
        for result in threads {
            match result {
                Err(e) => first_err = Err(e),
                Ok(hdl) => guard.push(hdl),
            }
        }
        first_err
```

hawkw: also, is the name `first_err` really correct? isn't this actually `last_err`, since any subsequent error we see will overwrite it?

lib/propolis/src/tasks.rs
```diff
@@ -412,3 +416,94 @@ impl Drop for CtrlHeld<'_> {
...
+pub struct ThreadGroup(Mutex<Vec<thread::JoinHandle<()>>>);
+impl ThreadGroup {
+    pub fn new() -> Self {
+        Self(Mutex::new(Vec::new()))
+    }
+
+    /// Add to group of contained threads
+    ///
+    /// The first (if any) [std::io::Error] encounted while among the `threads`
+    /// items, will determine the return from this function.  Despite any
```
hawkw, line 475: tiny grammar nit, unimportant:
```suggestion
    /// items will determine the return from this function.  Despite any
```

lib/propolis/src/tasks.rs
```diff
@@ -412,3 +416,94 @@ impl Drop for CtrlHeld<'_> {
         }
     }
 }
+
+/// Holds a group of tokio task [task::JoinHandle]s to be later joined as a
+/// group when they have all concluded.
+pub struct TaskGroup(Mutex<Vec<task::JoinHandle<()>>>);
```
hawkw, line 422: This feels like it could, potentially, be a good use case for [`tokio::task::JoinSet`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html) rather than `FuturesUnordered`, which is a more efficient way to join a group of spawned tasks. However, it offers a somewhat different API surface; you can't add `JoinHandle`s to a `JoinSet` and must, instead, spawn the tasks on the `JoinSet` using its own `JoinSet::spawn` family of methods.[^1]

Definitely not something we have to do in this PR, just thought I'd mention it.

[^1]: This is because `JoinSet`'s optimizations rely on it being integrated with the runtime and only able to operate on spawned tasks, rather than arbitrary futures which may not be tasks.

lib/propolis/src/attachment.rs
```diff
@@ -0,0 +1,292 @@
...
+enum AttachPoint<T> {
+    Attached(Arc<(AttachState, T)>),
+    Detached(AttachState),
+}
+impl<T> Default for AttachPoint<T> {
+    fn default() -> Self {
+        Self::Detached(AttachState::new())
+    }
+}
+
```
hawkw, line 89: nit, take it or leave it: it occurs to me that the `access` and `access_fallback` methods on `FrontAttachment` and `BackAttachment` have the same shape, and could be defined in terms of a single shared implementation on `AttachPoint`:
```suggestion
impl<T> FrontAttachment<T> {
    pub fn new() -> Self {
        Self(Mutex::new(AttachPoint::default()))
    }
    pub fn access<R>(
        &self,
        when_attached: impl FnOnce(&T, &AttachState) -> R,
    ) -> Option<R> {
        self.0.lock().unwrap().access(when_attached)
    }
    pub fn access_fallback<R>(
        &self,
        when_attached: impl FnOnce(&T, &AttachState) -> R,
        fallback: impl FnOnce(&AttachState) -> (),
    ) -> Option<R> {
        self.0.lock().unwrap().access_fallback(when_attached, fallback)
    }
}

/// Backend (resource provider) attachment point
pub struct BackAttachment<T>(Mutex<AttachPoint<T>>);
impl<T> BackAttachment<T> {
    pub fn new() -> Self {
        Self(Mutex::new(AttachPoint::default()))
    }
    pub fn access<R>(
        &self,
        when_attached: impl FnOnce(&T, &AttachState) -> R,
    ) -> Option<R> {
        self.0.lock().unwrap().access(when_attached)
    }
    pub fn access_fallback<R>(
        &self,
        when_attached: impl FnOnce(&T, &AttachState) -> R,
        fallback: impl FnOnce(&AttachState) -> (),
    ) -> Option<R> {
        self.0.lock().unwrap().access_fallback(when_attached, fallback)
    }
}

enum AttachPoint<T> {
    Attached(Arc<(AttachState, T)>),
    Detached(AttachState),
}
impl<T> Default for AttachPoint<T> {
    fn default() -> Self {
        Self::Detached(AttachState::new())
    }
}
impl<T> AttachPoint<T> {
    fn access<R>(
        &self,
        when_attached: impl FnOnce(&T, &AttachState) -> R,
    ) -> Option<R> {
        match self {
            AttachPoint::Attached(data) => {
                Some(when_attached(&data.1, &data.0))
            }
            AttachPoint::Detached(..) => None,
        }
    }
    fn access_fallback<R>(
        &self,
        when_attached: impl FnOnce(&T, &AttachState) -> R,
        fallback: impl FnOnce(&AttachState) -> (),
    ) -> Option<R> {
        match self {
            AttachPoint::Attached(data) => {
                Some(when_attached(&data.1, &data.0))
            }
            AttachPoint::Detached(state) => {
                fallback(state);
                None
            }
        }
    }
}
```

It's up to you whether you think the de-duplication is worth it, though.

hawkw (changes requested):
This looks good to me! I left a few more comments, but they're mostly just little style suggestions that you're welcome to take or leave.

hawkw (approved):
whoops, i meant to "approve" in my last review rather than  "request changes"

lib/propolis/src/tasks.rs
hawkw: Yeah, I don't think it's worth the effort to change everything that uses this type in this PR, I just thought I'd mention it for future reference.

lib/propolis/src/tasks.rs
pfmooney: I've reworked this to resemble your suggestion, with the logic fixed so it does emit the first error.

lib/propolis/src/tasks.rs
pfmooney: Fixed several things here, including the comma

lib/propolis/src/attachment.rs
pfmooney: I did some combining here.  With the consolidation into `AttachPoint`, it does seem a bit silly to have `FrontAttachment` and `BackAttachment` structs which are identical save for their names.  That being said, it's probably not worth the simplicity of a single attachment type if we lose the compiler protection from attaching two frontends or two backends together.

lib/propolis/src/tasks.rs
```diff
@@ -412,3 +416,95 @@ impl Drop for CtrlHeld<'_> {
...
+        let mut guard = self.0.lock().unwrap();
+        let mut status = Ok(());
+        for result in threads {
+            match result {
+                Err(e) => {
+                    // record the first error
+                    if status.is_ok() {
+                        status = Err(e);
+                    }
+                }
```
hawkw, line 491: thanks for addressing my feedback here, this looks correct now.

take it or leave it: if you like, you _could_ fold the `if` into a match guard, like this:
```suggestion
                // record the first error
                Err(e) if status.is_ok() => {
                    status = Err(e);
                },
                Err(_) => {},
```

lib/propolis/src/tasks.rs
```diff
@@ -412,3 +416,95 @@ impl Drop for CtrlHeld<'_> {
...
+                Ok(hdl) => guard.push(hdl),
+            }
+        }
+        status
+    }
+
+    /// Block until all contained [thread::JoinHandle]s have been joined,
+    /// returning any resulting [ThreadErr]s after doing so.
+    pub fn block_until_joined(&self) -> Option<Vec<ThreadErr>> {
```
hawkw, line 500: nitpick, take it or leave it: i'm a bit unconvinced that `Option<Vec<_>>` is really worth it here...why not just represent the "no errors" case with an empty `Vec`? this type introduces the possibility of multiple "empty" cases (`None` and `Some(vec![])`), which calling code might need to handle – although we will never currently return a `Some` with an empty vec, the caller still may need to deal with it. 

this isn't a huge deal, though, and if returning an option is useful when calling this, i'm fine with it.

hawkw (approved):
looks good to me, thanks for addressing the last round of feedback!

lib/propolis/src/tasks.rs
pfmooney: I considered that, but thought it was clearer as-is.

lib/propolis/src/tasks.rs
pfmooney: I kinda rather `Some(errors)`, and leaving it up to the callee to do the right thing (not return `Some([])`).  Considering how minimal usage of this facility is so far, the cost of changing it later is pretty low.

pfmooney:
FYI: No substantive update in eae45bfdce1900fd88f4e6307b66c3a666be21f0, just rebasing & squashing.

PR #677, opened 2024-04-01
Bump Crucible for backpressure tuning

mkeeter:
Bump backpressure delay for byte-based backpressure (#1240)
`crutest` cleaning; adding rand-read/write tests (#1233)


pfmooney (approved):

mkeeter:
(Holding off on this in case we want https://github.com/oxidecomputer/crucible/pull/1243 as well)

leftwo:
I think yeah, we do want 1243 as well.

leftwo (approved):

PR #678, opened 2024-04-04
Rework storage for Accessors

### LINKED ISSUE #675: accessor Node::poison deadlocks with racing drop
I had a propolis-standalone instance deadlock while destroying a VM during halt.   The offending stack is as follows:
```
stack pointer for thread 39 [state loop]: fffffc7fe39ff8e0
[ fffffc7fe39ff8e0 libc.so.1`__lwp_park+0x17() ]
  fffffc7fe39ff930 libc.so.1`mutex_lock_impl+0x113(1bcd790, 0)
  fffffc7fe39ff950 libc.so.1`mutex_lock+0x13(1bcd790)
  fffffc7fe39ffb80 <propolis::accessors::Node<T> as core::ops::drop::Drop>::drop::h4f759e432fa68c42+0x90()
  fffffc7fe39ffba0 alloc::sync::Arc<T,A>::drop_slow::h7e635989f14d1904+0x12()
  fffffc7fe39ffca0 propolis::accessors::Accessor<T>::poison::h4c27da6f9d986cc6+0x563()
  fffffc7fe39ffcc0 propolis::vmm::machine::Machine::do_destroy::h4bf1bb9dda81db12+0x24()
  fffffc7fe39ffd00 <propolis::vmm::machine::Machine as core::ops::drop::Drop>::drop::hf798d7b8b32cf6f4+0x14()
  fffffc7fe39ffe70 propolis_standalone::Instance::state_loop::h9bb359840508caaa+0xecf()
  fffffc7fe39ffed0 std::sys_common::backtrace::__rust_begin_short_backtrace::h55883990d859e2f8 +0x3e()
  fffffc7fe39fff60 core::ops::function::FnOnce::call_once{{vtable.shim}}::h1f25fcf57c71bbd5+0x85()
  fffffc7fe39fffb0 std::sys::unix::thread::Thread::new::thread_start::ha7be144980000e90+0x29()
  fffffc7fe39fffe0 libc.so.1`_thrp_setup+0x77(fffffc7feecb2a40)
  fffffc7fe39ffff0 libc.so.1`_lwp_start()
```

This is in poison:
```rust
    fn poison(&self) -> Option<Arc<T>> {
        self.lock_tree(|mut tguard, ent| {
            let id = ent.id;
            drop(ent);
            if id != ID_ROOT {
                drop(tguard);
                panic!("tree poisoning only allowed at root");
            }

            // Remove the resource from the tree...
            let top_res = tguard.resource_root.take();
            // ... and poison all nodes too
            if top_res.is_some() {
                tguard.for_each(ID_ROOT, |_id, tnode| {
                    if let Some(node) = tnode.node_ref.upgrade() {
                        node.0.lock().unwrap().resource.take();
                    }
                });
            }

            top_res
        })
```

When `take()`-ing the resource from nodes in the tree during the `for_each()`, if the node in question is dropped by its holder, then when the local copy we got from the `upgrade()` call is dropped, we'll find ourself in `Node::drop` which will attempt to lock the tree (which is already held) and deadlock.

pfmooney:
The initial storage (trees/nodes) and lifecycle model for Accessors left open a race between node self-cleanup (removing the associated entry from a tree on drop) and racing node re-parenting (as part of node adoption or orphaning).

This changes how nodes are cleaned up from their containing tree, and simplifies indexing of nodes in said tree to use (opaque) pointers, rather than arbitrary numeric IDs.

Fixes #675

### DIFF
--- a/lib/propolis/src/accessors.rs
+++ b/lib/propolis/src/accessors.rs
@@ -12,54 +12,98 @@
 //! contained emulation should fail any DMA accesses.
 
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
+use std::ffi::c_void;
 use std::marker::PhantomData;
+use std::ptr::NonNull;
 use std::sync::{Arc, Mutex, MutexGuard, Weak};
 
 use crate::vmm::{MemCtx, VmmHdl};
 
-type AccId = usize;
-
-const ID_NULL: AccId = 0;
-const ID_ROOT: AccId = 1;
+/// Key type for identifying nodes referenced by `Tree`.
+#[derive(Ord, PartialOrd, Eq, PartialEq, Debug, Copy, Clone)]
+pub struct NodeKey(NonNull<Node<c_void>>);
+impl<T> From<&Arc<Node<T>>> for NodeKey {
+    fn from(value: &Arc<Node<T>>) -> Self {
+        let raw = Arc::as_ptr(value) as *const Node<c_void>;
+        let inner = unsafe { NonNull::new_unchecked(raw as *mut Node<c_void>) };
+        NodeKey(inner)
+    }
+}
+impl std::fmt::Display for NodeKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:p}", self.0.as_ptr())
+    }
+}
+// Safety: While the key uses a pointer (!Send) type internally, it is for
+// unique identification purposes only, and is never meant to be dereferenced,
+// copied from, or transformed into a reference of any kind.
+unsafe impl Send for NodeKey {}
 
 struct TreeNode<T> {
-    parent_id: AccId,
+    /// [NodeKey] of the parent to this node
+    ///
+    /// Holds [None] if the node is the root of the [Tree]
+    parent_key: Option<NodeKey>,
+    /// [Weak] reference back to the node.  This access is needed if the node
+    /// undergoes adoption (being moved to a different [Tree])
     node_ref: Weak<Node<T>>,
-    children: BTreeSet<AccId>,
+    /// List of keys to child nodes (if any)
+    children: BTreeSet<NodeKey>,
+    /// Display name for [Tree::print()]-ing
     name: Option<String>,
 }
 impl<T> TreeNode<T> {
     fn new(
-        parent_id: AccId,
+        parent_key: NodeKey,
         node_ref: Weak<Node<T>>,
         name: Option<String>,
     ) -> Self {
-        Self { parent_id, node_ref, children: BTreeSet::new(), name }
+        Self {
+            parent_key: Some(parent_key),
+            node_ref,
+            children: BTreeSet::new(),
+            name,
+        }
+    }
+    fn new_root(node_ref: Weak<Node<T>>) -> Self {
+        Self {
+            parent_key: None,
+            node_ref,
+            children: BTreeSet::new(),
+            name: None,
+        }
     }
 }
+
 struct Tree<T> {
+    /// Underlying resource (if any) that this hierarchy is granting access to
     resource_root: Option<Arc<T>>,
-    nodes: BTreeMap<AccId, TreeNode<T>>,
-    next_id: AccId,
+    /// Key of the root node of this hierarchy
+    ///
+    /// Only when the tree is being initialized, should `root_key` be [None]
+    root_key: Option<NodeKey>,
+    /// Nodes within this hierarchy
+    nodes: BTreeMap<NodeKey, TreeNode<T>>,
+    /// Weak self-reference, used when building [TreeNode] entries as nodes are
+    /// added to the tree.  Held as a convenience, instead of requiring it to be
+    /// passed in by the caller.
+    self_weak: Weak<Mutex<Tree<T>>>,
 }
 impl<T> Tree<T> {
-    /// Get the next [AccId] for a node to be added into this tree.
-    fn next_id(&mut self) -> AccId {
-        let res = self.next_id;
-        self.next_id += 1;
-        res
-    }
-
-    /// Record a node in the tree, given its ID and the ID of its parent.
+    /// Record a node in the tree
     fn add_child(
         &mut self,
-        parent: AccId,
-        child: AccId,
-        child_node: &Arc<Node<T>>,
+        parent: NodeKey,
         name: Option<String>,
-    ) {
+    ) -> Arc<Node<T>> {
+        let child_node = Arc::new(Node(Mutex::new(NodeEntry {
+            tree: Weak::upgrade(&self.self_weak).expect("tree ref still live"),
+            resource: self.resource_root.clone(),
+        })));
+
+        let child_key = NodeKey::from(&child_node);
         let conflict = self.nodes.insert(
-            child,
+            child_key,
             TreeNode::new(parent, Arc::downgrade(&child_node), name),
         );
         assert!(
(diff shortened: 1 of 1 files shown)


lib/propolis/src/accessors.rs
```diff
@@ -12,54 +12,84 @@
...
+    /// [NodeKey] of the parent to this node
+    ///
+    /// Holds [NodeKey::NULL] if the node is the root of the [Tree]
+    parent_key: NodeKey,
+    /// [Weak] reference back to the node.  This access is needed if the node
+    /// undergoes adoption (being moved to a different [Tree])
     node_ref: Weak<Node<T>>,
-    children: BTreeSet<AccId>,
+    /// List of keys to child nodes (if any)
+    children: BTreeSet<NodeKey>,
```
hawkw, line 52: nit, not important: is the ordering provided by the `BTreeSet` load-bearing here? it might be nice if the comment also explained why.

lib/propolis/src/accessors.rs
```diff
@@ -71,93 +101,81 @@ impl<T> Tree<T> {
...
-                    // already hold the lock on it.
-                    debug_assert_eq!(adopt_root.id, child_id);
-                    adopt_root.id = new_child_id;
-                    adopt_root.tree = Arc::clone(tree_ref);
-                    adopt_root.resource = self.resource_root.clone();
-                }
+                let mut ent = node.0.lock().unwrap();
+                ent.tree = Arc::clone(&tree_ref);
+                ent.resource = self.resource_root.clone();
+                drop(ent);
```
hawkw, line 137: silly style nit, feel free to disagree: in general, I prefer to use scopes with curly-braces for stuff like this, rather than explicit `drop`s, it makes the portion of code for which the lock is held a little more visually obvious to readers.

also, i notice that we do it that way later in the file, so the consistency might be nice?

lib/propolis/src/accessors.rs
```diff
@@ -12,54 +12,84 @@
...
-
-const ID_NULL: AccId = 0;
-const ID_ROOT: AccId = 1;
+/// Key type for identifying nodes referenced by `Tree`.
+#[derive(Ord, PartialOrd, Eq, PartialEq, Debug, Copy, Clone)]
+#[repr(transparent)]
+pub struct NodeKey(*const Node<c_void>);
+impl NodeKey {
+    const NULL: Self = Self(std::ptr::null());
+}
```
hawkw, line 27: IMO, it might be nicer to represent this as `NonNull<Node<c_void>>`, allowing us to use `Option<NodeKey>` rather than treating `NodeKey` as nullable (and using `NonNull` means that `Option<NodeKey>` should be represented the same as `NodeKey(ptr::null())` due to niche optimization). 

This way, code that touches a nullable `NodeKey` (the parent ID) is forced to handle the null case, while the compiler ensures that other `NodeKey`s that are not the parent key are never null.

lib/propolis/src/accessors.rs
```diff
@@ -525,11 +518,33 @@ impl<T> Accessor<T> {
...
+    /// but poses a challenge: When the `Tree` is performing node adoption or
+    /// orphaning, it must reach back out into the Nodes it owns via
+    /// `TreeNode::node_ref`.  This poses a race for when the last
+    /// `Arc<Node<T>>` is dropped, either by the `Accessor`, or the logic in the
+    /// `Tree`.  When the latter wins, it still holds the tree lock, posing a
+    /// deadlock situation which is otherwise impossible to prevent.
+    ///
+    /// As such it is expected that the Accessor will perform the tree
+    /// deregistration of the node when it is being dropped.  No other
+    /// structures should hold the `Arc<Node<T>>`.
```
hawkw, line 543: Thanks for the comment, this is very helpful!

hawkw (commented):
I'm not very familiar with the context around this tree structure and how it's used, so I've tried to focus on the implementation of this change. Overall, this seems good, and most of my comments were style nits. I think it _might_ be worth using `NonNull` rather than `*const` for the key type, so that the null case can be represented using an `Option`. 

Regarding the poisoning operation on trees, what does this actually represent and when do we do it? A comment on the method might be nice, IMO, although I understand that there wasn't one previously, so it doesn't necessarily have to be added in this PR...I'll look at the rest of the codebase and try to give a more thorough review

lib/propolis/src/accessors.rs
pfmooney: The ordering is not load-bearing at all.  I picked the btree version over hash due to this comment in the latter:

> In the current implementation, iterating over set takes O(capacity) time instead of O(len) because it internally visits empty buckets too.

I don't plan to comment on the choice, since it's not an important detail of the implementation. (If that seems alright)

pfmooney:
Thanks for taking a look.

> Regarding the poisoning operation on trees, what does this actually represent and when do we do it?

It's there mainly for the top of the accessor hierarchy to quickly withdraw access to the underlying resource in as prompt a manner as possible.  Consumers may be in the middle of an access (with a guard held) which would delay the process somewhat, but absent that, it should allow the resource to be freed quickly.  I've added some detail to the `Accessor::poision()` method.

lib/propolis/src/accessors.rs
pfmooney: This should be cleaned up, as well as the other instances I found.  (I didn't bother going through the tests, though)

lib/propolis/src/accessors.rs
pfmooney: I was (and remain) a bit uncomfortable moving that load-bearing `drop` logic into the `Accessor`, so I wanted to make the reasoning clear to future readers (including myself)

lib/propolis/src/accessors.rs
pfmooney: I waffled a bit on this while writing it.  It makes some accesses a bit more annoying, and further suggests to one examining `NodeKey` that the pointer it holds might actually be useful for something (beyond a unique ID).  The help from the compiler is probably worth it, though.

lib/propolis/src/accessors.rs
hawkw: fine with me!

lib/propolis/src/accessors.rs
```diff
@@ -12,54 +12,98 @@
...
 struct Tree<T> {
+    /// Underlying resource (if any) that this heirarchy is granting access to
     resource_root: Option<Arc<T>>,
-    nodes: BTreeMap<AccId, TreeNode<T>>,
-    next_id: AccId,
+    /// Key of the root node of this hierarchy
+    ///
+    /// Only when the tree is being initialized, should `root_key` be [None]
+    root_key: Option<NodeKey>,
+    /// Nodes within this heirarchy
```
hawkw, line 85: spelling nit:
```suggestion
    /// Nodes within this hierarchy
```

lib/propolis/src/accessors.rs
```diff
@@ -502,11 +497,32 @@ impl<T> Accessor<T> {
...
+                child_guard.rename_node(child_key, name);
+                parent_guard.adopt(parent_key, &mut child_guard);
+            });
+        });
     }
 
     /// Poison (remove the underlying resource) from the root node of a
-    /// hierarchy.
+    /// hierarchy.  This is meant to provide the root holder of the resource the
+    /// means to promptly remove access to it during events such as tear-down.
```
hawkw, line 525: Thanks, this is helpful!

hawkw (approved):
Looks good to me, thanks for adding the comment on `poison`!

lib/propolis/src/accessors.rs
pfmooney: *sigh* I don't know why I haven't enabled `:set spell` all the time.

PR #680, opened 2024-04-15
Add NPT ops API definitions from illumos#15639

pfmooney:


pfmooney:
The `header-check` test for bhyve-api passes when run against up-to-date illumos-gate sources:
```
     Running test/main.rs (target/debug/deps/main-17e3b0e49faaeb6c)
RUNNING ALL TESTS
PASSED 1175 tests
```

hawkw (approved):

PR #687, opened 2024-04-25
Propagate ReplaceResult up; return disk status

jmpesp:
Bump crucible rev to 5a78ef8e, picking up:

- Remove `ReadResponse` in favor of `RawReadResponse`
- Fix typo in DTrace upstairs_info
- replace needing no work should not be an error
- Add some DTrace scripts to the package.
- More Pantry updates for Region replacement
- Send the correct task count for reconciliations
- Raw extent cleanup

Crucible's `target_replace` now returns a `ReplaceResult` instead of `()`, so propagate that back to the caller.

Also adds a status endpoint to propolis-server to query whether or not a disk's volume is active.


openapi/propolis-server-falcon.json
```diff
@@ -112,6 +112,41 @@
         }
       }
     },
+    "/instance/disk/{id}/status": {
```
leftwo, line 115: The only question I have is if this propolis-server-falcon needs any plumbing
to pass this to crucible if it needs to.  Given the other endpoints must already
be doing the same thing, it's probably okay.

leftwo (approved):

bin/propolis-server/src/lib/server.rs
```diff
@@ -1144,9 +1170,17 @@ async fn instance_issue_crucible_vcr_request(
...
+        match backend.vcr_replace(old_vcr_json, &new_vcr_json).await {
+            Ok(replace_result) => replace_result,
+
+            Err(e) => {
+                return Err(HttpError::for_bad_request(
+                    Some(e.to_string()),
+                    e.to_string(),
+                ));
+            }
+        };
```
pfmooney, line 1183: Why the explicit return vs the `?` operator (like above)?

lib/propolis/src/block/crucible.rs
```diff
@@ -278,6 +280,10 @@ impl CrucibleBackend {
             );
         }
     }
+
+    pub async fn volume_is_active(&self) -> io::Result<bool> {
```
pfmooney, line 284: Why make the trip through std::io::Error (and Result) here, vs letting the CrucibleError through?

crates/propolis-api-types/src/lib.rs
```diff
@@ -396,6 +396,16 @@ pub struct VCRRequestPathParams {
...
 }
 
+#[derive(Deserialize, JsonSchema)]
+pub struct VolumeStatusPathParams {
+    pub id: Uuid,
+}
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
+pub struct VolumeStatus {
+    pub active: bool,
```
pfmooney, line 406: Do we foresee any other pieces of status information being useful in the future?

jmpesp: None immediately come to mind. I could see cases where Nexus may want to query what Volume Construction Request is currently in the instance spec (to compare with the latest that is stored in CRDB), but it could just use `GET /instance/spec` in that case.

bin/propolis-server/src/lib/server.rs
jmpesp: Sorry, do you mean why doesn't it look like

```rust
    let replace_result =
        backend.vcr_replace(old_vcr_json, &new_vcr_json).await.map_err(
            |e| HttpError::for_bad_request(Some(e.to_string()), e.to_string()),
        )?;
```

? I could change it to that, yeah.

lib/propolis/src/block/crucible.rs
jmpesp: good point, done in 473bdfe

jmpesp:
Note 231d6f4b8ff8a0ea7a1fc972ed8f71a48282c6e7 enables Crucible's `notify-nexus` feature when `omicron-build` is set.

bin/propolis-server/src/lib/server.rs
jmpesp: cool, done in eb21550

lib/propolis/Cargo.toml
```diff
@@ -54,4 +54,4 @@ rand.workspace = true
 default = []
 crucible-full = ["crucible", "crucible-client-types", "oximeter", "nexus-client"]
 falcon = ["libloading", "p9ds", "dlpi", "ispf", "rand", "softnpu-lib", "num_enum"]
-omicron-build = []
+omicron-build = ["crucible/notify-nexus"]
```
pfmooney, line 57: We should be requiring this feature directly via the propolis-server deps, not through `omicron-build` here in propolis-lib.

jmpesp: I first tried

```patch
diff --git a/bin/propolis-server/Cargo.toml b/bin/propolis-server/Cargo.toml
index 6e30f39..4b71754 100644
--- a/bin/propolis-server/Cargo.toml
+++ b/bin/propolis-server/Cargo.toml
@@ -77,7 +77,7 @@ default = []
 
 # When building to be packaged for inclusion in the production ramdisk
 # (nominally an Omicron package), certain code is compiled in or out.
-omicron-build = ["propolis/omicron-build"]
+omicron-build = ["propolis/omicron-build", "crucible/notify-nexus"]
 
 # Falcon builds require corresponding bits turned on in the dependency libs
 falcon = ["propolis/falcon", "propolis_api_types/falcon"]
 ```
 
 but saw
 
 ```
 cargo build
error: failed to load manifest for workspace member `/home/james/src/oxidecomputer/propolis/bin/propolis-server`

Caused by:
  failed to parse manifest at `/home/james/src/oxidecomputer/propolis/bin/propolis-server/Cargo.toml`

Caused by:
  feature `omicron-build` includes `crucible/notify-nexus`, but `crucible` is not a dependency
 ```

Comment (pfmooney) [reply lib/propolis/Cargo.toml:57]: I think this would be more appropriate:
```diff
 default = []
-crucible-full = ["crucible", "crucible-client-types", "oximeter", "nexus-client"]
+crucible-full = ["crucible", "crucible/notify-nexus", "crucible-client-types", "oximeter", "nexus-client"]
 falcon = ["libloading", "p9ds", "dlpi", "ispf", "rand", "softnpu-lib", "num_enum"]
-omicron-build = ["crucible/notify-nexus"]
+omicron-build = []
```

We're expecting to have the nexus-involved behavior whenever `crucible-full` is set.  Why would we only build it in when `omicron-build` is set?

jmpesp: I was only trying to avoid cases where an Upstairs would try to notify Nexus when in CI or when building propolis-server to run on its own (not in a full setup). Is that the case with `crucible-full` being set? Sorry, I get the intention of the `omicron-build` feature but not the `crucible-full` one

jmpesp: Following up with stuff from the chat: I pushed eaa60e4 which adds a note about oxidecomputer/crucible#1280, which tracks the work to configure Nexus notifications at runtime instead of compile time.

pfmooney: Thanks!

pfmooney (approved):

PR #694, opened 2024-05-01
smbios: add enums/bitflags types for some SMBIOS fields

hawkw:
Currently, most of the SMBIOS fields that we populate are magic numbers, with comments describing what that magic number indicates. For SMBIOS enum fields, this isn't too bad, but for bitfields like the various "characteristics" fields, it's a bit of a shame, since the hex constant doesn't always make it obvious which bit corresponds to which characteristic.

To make this a bit nicer, this commit adds enums for most of the SMBIOS enum fields we currently populate, and bitflags for most of the bitflags fields. This way, we can refer to them by name rather than as magic numbers. Also, I've made the `Default` implementation for the various enum fields return the "unknown" (0x2) variant. Eventually, we may want to default-initialize more SMBIOS tables --- but, we should probably hold off on that until *all* the enum fields in the tables we populate are Rust types with `Default` impls, so that we don't have to manually initialize some of them to 0x2.

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -853,7 +853,7 @@ impl<'a> MachineInitializer<'a> {
 
     fn generate_smbios(&self) -> smbios::TableBytes {
         use propolis::cpuid;
-        use smbios::table::{type0, type4};
+        use smbios::table::{type0, type1, type4};
 
         let rom_size =
             self.state.rom_size_bytes.expect("ROM is already populated");
@@ -885,7 +885,7 @@ impl<'a> MachineInitializer<'a> {
             uuid: self.properties.id.to_bytes_le(),
 
             // power switch
-            wake_up_type: 0x06,
+            wake_up_type: type1::WakeUpType::PowerSwitch,
             ..Default::default()
         };
 
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -865,7 +865,7 @@ impl<'a> MachineInitializer<'a> {
                 .unwrap(),
             bios_rom_size: ((rom_size / (64 * 1024)) - 1) as u8,
             // Characteristics-not-supported
-            bios_characteristics: type0::BiosCharacteristics::UNKNOWN,
+            bios_characteristics: type0::BiosCharacteristics::UNSUPPORTED,
             bios_ext_characteristics: type0::BiosExtCharacteristics::ACPI
                 | type0::BiosExtCharacteristics::UEFI
                 | type0::BiosExtCharacteristics::IS_VM,
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -853,7 +853,7 @@ impl<'a> MachineInitializer<'a> {
 
     fn generate_smbios(&self) -> smbios::TableBytes {
         use propolis::cpuid;
-        use smbios::table::{type0, type1, type4};
+        use smbios::table::{type0, type1, type16, type4};
 
         let rom_size =
             self.state.rom_size_bytes.expect("ROM is already populated");
@@ -953,11 +953,9 @@ impl<'a> MachineInitializer<'a> {
         let memsize_bytes = (self.properties.memory as usize) * MB;
         let mut smb_type16 = smbios::table::Type16 {
             // system board
-            location: 0x3,
+            location: type16::Location::SystemBoard,
             // system memory
-            array_use: 0x3,
-            // unknown
-            error_correction: 0x2,
+            array_use: type16::ArrayUse::System,
             num_mem_devices: 1,
             ..Default::default()
         };
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -883,7 +883,6 @@ impl<'a> MachineInitializer<'a> {
                 .unwrap_or_default(),
             uuid: self.properties.id.to_bytes_le(),
 
-            // power switch
             wake_up_type: type1::WakeUpType::PowerSwitch,
             ..Default::default()
         };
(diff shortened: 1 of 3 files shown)


bin/propolis-server/src/lib/initializer.rs
```diff
@@ -864,9 +865,10 @@ impl<'a> MachineInitializer<'a> {
                 .unwrap(),
             bios_rom_size: ((rom_size / (64 * 1024)) - 1) as u8,
             // Characteristics-not-supported
```
pfmooney, line 867: style nit: Since you're turning these into nice named values, the accompanying comments can probably be tossed out now (here and below)

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -864,9 +865,10 @@ impl<'a> MachineInitializer<'a> {
...
-            bios_characteristics: 0x8,
-            // ACPI + UEFI + IsVM
-            bios_ext_characteristics: 0x1801,
+            bios_characteristics: type0::BiosCharacteristics::UNKNOWN,
```
pfmooney: Here and in -server, this should be not-supported, rather than unknown.

lib/propolis/src/firmware/smbios/table.rs
```diff
@@ -138,19 +324,121 @@ impl Table for Type4 {
...
+    #[derive(Debug, Default, Copy, Clone, PartialEq, Eq)]
+    #[repr(u8)]
+    pub enum ProcType {
+        /// Other
+        Other = 0x01,
+        /// Unknown
+        #[default]
+        Unknown = 0x02,
+        /// Central Processor
+        Cpu = 0x03,
```
pfmooney, line 352: nit: Why Cpu instead of Central, to stay consistent with the other entries?

bin/propolis-standalone/src/main.rs
```diff
@@ -876,31 +879,32 @@ fn generate_smbios(params: SmbiosParams) -> anyhow::Result<smbios::TableBytes> {
...
     let mut smb_type16 = smbios::table::Type16 {
         // system board
-        location: 0x3,
+        location: type16::Location::SystemBoard,
         // system memory
-        array_use: 0x3,
+        array_use: type16::ArrayUse::System,
         // unknown
-        error_correction: 0x2,
+        error_correction: type16::ErrorCorrection::Unknown,
```
pfmooney, line 907: Omitted for standalone, but not here?

pfmooney (commented):
For testing notes, it'd be nice to see that `dmidecode` (or `smbios` on illumos) output remained unchanged before and after.


Thanks for polishing these bits up

bin/propolis-standalone/src/main.rs
hawkw: ah, the default impl populates this with `Unknown` now, and I went back and
forth on whether I wanted to use the default impl for `Unknown` fields when we
don't default to unknown for stuff that isn't an enum/bitflags yet. I'll go back
and add it in standalone.

hawkw:
@pfmooney: I've addressed your review feedback (thanks for the quick review & suggestions!) and I've added `serde::{Serialize, Deserialize}` impls for all the new types that should serialize/deserialize them as raw numbers and validate when deserializing --- let me know what you think!

I'd also like add a PHD test that runs `dmidecode` in a guest and checks its output, but I'm not sure if it's installed out of the box in our test Alpine image...

lib/propolis/src/firmware/smbios/table.rs
```diff
@@ -184,6 +711,213 @@ impl Table for Type16 {
...
+    }
+
+    enum_try_from! {
+        /// Memory array error correction field.
+        ///
+        /// See Table 74 in section 7.17.3 of [the SMBIOS Reference
+        /// Specification][DSP0136] for details.
+        ///
+        /// [DSP0136]:
+        ///     https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.7.0.pdf
```
pfmooney, line 794: Could probably put this reference at the top somewhere, rather than repeating it at each bitfield.

lib/propolis/src/firmware/smbios/table.rs
```diff
@@ -46,14 +57,284 @@ impl Table for Type0 {
...
+                        $val => Ok($Enum::$Variant),
+                    )+
+                    _ => Err(concat!(
+                        "invalid ", stringify!($enum), " value, expected one of: ",
+                        stringify!($Variant1)," (", stringify!($val1), ")",
+                        $(", ", stringify!($Variant), " (", stringify!($val), ")" ),*
+                    )),
+                }
+            }
+        }
```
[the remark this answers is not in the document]
hawkw: Ah, I didn't realize we already had a dep on `strum`, that'll make this much neater!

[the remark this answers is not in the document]
pfmooney: A convenient crate, for sure

lib/propolis/src/firmware/smbios/table.rs
```diff
@@ -46,14 +68,303 @@ impl Table for Type0 {
...
+                        dbg!(serde_json::to_string(dbg!(variant))).unwrap();
+                    let deserialized =
+                        dbg!(serde_json::from_str::<'_, $Enum>(&serialized))
+                            .unwrap();
+                    assert_eq!(*variant, deserialized);
+                }
+            }
+        )+
+    }
+}
```
pfmooney, line 209: Nice to cut down on all that boilerplate with these.

lib/propolis/src/firmware/smbios/table.rs
```diff
@@ -46,14 +68,303 @@ impl Table for Type0 {
...
+                                };
+
+                                let comma = if has_remaining {
+                                    ", "
+                                } else {
+                                    ""
+                                };
+                                write!(err, "{or}{variant:?} ({:#04x}){comma}", variant as $repr).unwrap();
+                                first = false;
+                            }
```
pfmooney, line 102: Might be simpler to impl this in terms of `slice::join`, switching on `VARIANTS.len()` if you want to preserve the fancy behavior for the 1 and 2 item cases.

Up to you.

pfmooney (approved):
Looks good.  I assume you've done the `dmidecode` diff testing to ensure that nothing has changed on the guest-visible output?

hawkw:
Looks like we're good: 
```console
$ colordiff dmidecode-master.txt dmidecode-branch.txt
colordiff 1.0.21 (http://www.colordiff.org/)
(C)2002-2022 Dave Ewart, davee@sungate.co.uk

```

pfmooney (approved):

PR #700, opened 2024-05-03
Standardize on `mod test`

pfmooney:
Keep a consistent style of `mod test`, rather than a mix of that and `mod tests`.  Enforce this new rule with a rudimentary check, added as `xtask style`.

### DIFF
--- a/xtask/src/task_style.rs
+++ b/xtask/src/task_style.rs
@@ -13,9 +13,17 @@ use crate::util::*;
 fn check_test_names() -> Result<()> {
     let wroot = workspace_root()?;
 
+    // Get listing of all tests (excluding doctests)
     let mut cmd = Command::new("cargo");
     let child = cmd
-        .args(["test", "--workspace", "--", "--list", "--format=terse"])
+        .args([
+            "test",
+            "--workspace",
+            "--all-targets",
+            "--",
+            "--list",
+            "--format=terse",
+        ])
         .stdin(Stdio::null())
         .stdout(Stdio::piped())
         .stderr(Stdio::null())
@@ -32,15 +40,11 @@ fn check_test_names() -> Result<()> {
                 _ => return None,
             };
 
-            // skip doctests which follow "<path> - <testname> (line <num>)"
-            if test_name.contains(" - ") {
-                return None;
-            }
-
             // Check for `mod tests` instead of `mod test` as the last component of
-            // the test name;
-            match test_name.split("::").collect::<Vec<_>>().iter().nth_back(1) {
-                Some(&"tests") => {
+            // the test name (prior to the test function name itself);
+            let mut name_parts = test_name.rsplit("::");
+            match (name_parts.next(), name_parts.next()) {
+                (_fn_name, Some("tests")) => {
                     Some(test_name.rsplit_once("::").unwrap().0.to_owned())
                 }
                 _ => None,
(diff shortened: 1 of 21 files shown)


xtask/src/task_style.rs
```diff
@@ -0,0 +1,64 @@
...
+            // Check for `mod tests` instead of `mod test` as the last component of
+            // the test name;
+            match test_name.split("::").collect::<Vec<_>>().iter().nth_back(1) {
+                Some(&"tests") => {
+                    Some(test_name.rsplit_once("::").unwrap().0.to_owned())
+                }
+                _ => None,
+            }
+        })
+        .collect::<BTreeSet<_>>();
```
hawkw, line 49: nice that this is alphabetized for a consistent ordering in the output!

xtask/src/task_style.rs
```diff
@@ -0,0 +1,64 @@
...
+            // Look for "<test name>: test"
+            let test_name = match line.rsplit_once(": ") {
+                Some((p, "test")) => p,
+                _ => return None,
+            };
+
+            // skip doctests which follow "<path> - <testname> (line <num>)"
+            if test_name.contains(" - ") {
+                return None;
+            }
```
hawkw, line 38: nit, take it or leave it: we could, alternatively, filter out doctests by changing the command from `cargo test --workspace -- ...` to `cargo test --tests --lib --workspace -- ...`, to enable only lib tests and integration tests. that way, `cargo test` will never print the names of doctests (and, it might be a bit faster, as we also don't compile the doctests at all).

xtask/src/task_style.rs
```diff
@@ -0,0 +1,64 @@
...
+
+            // Check for `mod tests` instead of `mod test` as the last component of
+            // the test name;
+            match test_name.split("::").collect::<Vec<_>>().iter().nth_back(1) {
```
hawkw, line 42: nit, take it or leave it: [the `Split` iterator implements `DoubleEndedIterator`](https://doc.rust-lang.org/stable/std/iter/trait.DoubleEndedIterator.html#tymethod.next_back), so this could just be:
```suggestion
            match test_name.split("::").next_back() {
```

xtask/src/task_style.rs
```diff
@@ -0,0 +1,64 @@
...
+                Some(&"tests") => {
+                    Some(test_name.rsplit_once("::").unwrap().0.to_owned())
+                }
+                _ => None,
+            }
```
hawkw, line 47: actually, can't this just be:
```suggestion
            match test_name.rsplit_once("::") {
                Some((path, &"tests")) => Some(path),
                _ => None,
            }
```

hawkw (approved):
This looks good to me overall! I had some smallish suggestions for the actual style check implementation, but no blockers on the change.

xtask/src/task_style.rs
pfmooney: I'm picking the second-to-last item (`nth_back` is 0-indexed), not the last.

I've reworked this to skip the `collect()` allocation.

xtask/src/task_style.rs
pfmooney: See above

xtask/src/task_style.rs
pfmooney: Looks like `--all-targets` achieves this, so I'll rip out that complication in the parsing logic.

xtask/src/task_style.rs
hawkw: Ah, yup, I can't read. Carry on.

hawkw (approved):
looks good to me!

pfmooney:
Merged in 6d4fe85574b71fef17e99af56f07b866c77cf6d7

Issue #701, opened 2024-05-06
Accurate versions in SMBIOS tables


luqmana:
The version information in the BIOS (type 0) and System (type 1) tables are currently hardcoded. We could instead set them to the OVMF and propolis versions (/commit hash?), respectively.

hawkw:
Since Propolis itself is not published to crates.io, we don't really increment the crate version number very often --- it will almost always be 0.1.0. I think the Git metadata is much likelier to be useful.

We may also want to include the bhyve API version in a system version string?

pfmooney:
One point worth noting: If we put version data about propolis and/or bhyve in the SMBIOS tables, it will potentially become out of date when the instance is live-migrated.

hawkw:
Oh, hmm, that's a good point @pfmooney. It would sure be a shame to break guests just because we migrated to a VMM built from a different Git commit. 

I do have a branch implementing the Propolis version in the type 1 table (https://github.com/oxidecomputer/propolis/pull/702), but I'm happy to drop it if we think this is actually a bad idea.

hawkw:
Regarding the OVMF version, since Propolis just gets the bootrom from a path to an arbitrary file in the TOML config, it doesn't have an easy way to know what the OVMF version is. Probably, we want to make the reported BIOS version configurable in the config file, and ensure that it's set to whatever OVMF version Propolis gets?

pfmooney:
> I'm happy to drop it if we think this is actually a bad idea.

I don't think it's a _bad_ idea, but one of limited utility, at least for now.  I'd rather not clutter up the build until we have a clear reason.

> Probably, we want to make the reported BIOS version configurable in the config file, and ensure that it's set to whatever OVMF version Propolis gets?

At some point, I imagine we'll have some sort of selectable bootrom profiles (complete with version metadata), but until then, setting it via the config is probably the right path.  (And I wouldn't even bother with wiring it up in standalone)

pfmooney:
> It would sure be a shame to break guests just because we migrated to a VMM built from a different Git commit.

I don't believe this would actually be breaking.  Once `fw_cfg` is fixed (a WIP) in propolis, we'll simply migrate the generated contents, so they remain fixed after the instance initialization code sets them the first time.

rmustacc:
I missed the discussion here when I wrote up [this comment](https://github.com/oxidecomputer/propolis/pull/702#issuecomment-2099669009) in #702.



[...]

PR #702, opened 2024-05-07
teach Propolis a -v flag with Git metadata

hawkw:
ixi note: since I'm basically taking the vergen bits Eliza had set up, I'm editing this PR in place after having clipped out the SMBIOS stuff. also it uses vergen 10.0 now rather than 6 or whatever it was. the below is my reworking Eliza's commit a bit given the different integration with the  binaries, so it's .. pretty similar to the old one but also different.

Currently, Propolis, both -server and -standalone, opaquely *run* with no
hint available about the provenance of the binary. For binaries that
come from product software releases, this is basically fine, as those
packages include a sha256 of the packaged Propolis zone and one can work
from a binary to a source commit.

For any development-time experience, random `propolis-standalone`s we've
left around, or VMM binaries one finds on a particularly scrumptious USB
stick in a parking lot, a `--version flag telling you where the binary
probably came from would go a long way.

This branch adds a `build.rs` script that uses [the `vergen` crate][1]
to emit information about the Git revision that Propolis was build from.
Now, we can generate a version string that describes the git branch,
commit hash, and commit depth. This is generated in a
`propolis::version()` function, which also includes the detected bhyve
and viona API versions at runtime. This results in version strings like:

```
propolis-standalone v0.1.0-930 (DIRTY b428c40ef) eliza/somebios-version, bhyve API v18, viona API v6
```

In addition to the `--version` flag described above, this commit also
has propolis-server and propolis-standalone print version information
when run in a way that they will later run a VM. In release software
this can provide evidence of a propolis-server which generated a log
file. For the rest of us, hopefully the log line isn't lost to
backscroll if an issue has occurred..

The start of both -server and -standalone logs will now have something like:

```
Aug 12 02:06:49.389 INFO Running v0.1.0-932 (c948d02) eliza/somebios-version, bhyve API v18, viona API v6
```

[1]: https://docs.rs/vergen


luqmana:
Version in SMBIOS tables aside for now, I like having this available via just `--version`. Great for double checking one-off binaries laying about!

lib/propolis/src/lib.rs
```diff
@@ -35,3 +35,46 @@ pub mod vmm;
...
+pub fn version() -> &'static str {
+    lazy_static::lazy_static! {
+        static ref VERSION: String = {
+            use std::fmt::Write;
+
+            let git = option_env!("VERGEN_GIT_BRANCH")
+                .and_then(|branch| Some((branch, option_env!("VERGEN_GIT_SHA")?)))
+                .and_then(|(branch, sha)| Some((branch, sha, option_env!("VERGEN_GIT_COMMIT_COUNT")?)));
+
+            let mut version = format!("Propolis v{}", env!("CARGO_PKG_VERSION"));
```
luqmana, line 48: Drop the Propolis prefix here? Using this with `#[clap(version)]` means you already get the prefix.

hawkw: hmm, yeah, we can do that --- my thinking was that the prefix made it obvious which was the propolis version in the SMBIOS version string, but we could also just stick the prefix in front of that specifically.

hawkw:
Since this shouldn't actually result in differing versions after live migration (per @pfmooney's comment in https://github.com/oxidecomputer/propolis/issues/701#issuecomment-2098894092), I'm marking this as ready for review.

lib/propolis/src/lib.rs
```diff
@@ -35,3 +35,47 @@ pub mod vmm;
...
+            let mut version = format!("v{}", env!("CARGO_PKG_VERSION"));
+            if let Some((branch, sha, commit)) = git {
+                write!(version, "-{commit} ({sha}) {branch}, ")
+                    .expect("writing to a string never fails");
+            } else {
+                version.push_str(" <unknown git commit>, ");
+            }
+            match bhyve_api::api_version() {
+                Ok(v) => {
+                    write!(version, "Bhyve API v{v}")
```
pfmooney, line 57: With everything else in the version being fixed at compile time, do we want to include runtime data (the bhyve api version) in the string?

pfmooney, line 57: nit: `bhyve` should not be capitalized

lib/propolis/src/firmware/smbios/table.rs
```diff
@@ -381,6 +381,22 @@ impl Table for Type1 {
     }
 }
 
+impl Default for Type1 {
+    fn default() -> Self {
+        Type1 {
+            manufacturer: SmbString::default(),
+            product_name: SmbString::default(),
+            version: SmbString::try_from(crate::version())
+                .expect("version string should not contain NULs"),
```
pfmooney, line 390: If we remove the bhyve API version from the string, we could include this condition as part of the `version()` unit tests

rmustacc:
So I'm not sure if we want this going in SMBIOS per se. I get the benefit as a developer, but I'm less sold from a customer perspective. What are the semantics of the version that they should use? Are we committing to the format? This leaks information about the host version in a much more concrete way and I'll admit I expected this to be something about the virtual machine itself or something about the overall system as opposed to the virtualization part of it.

For example:
* qemu: `Version: pc-i440fx-jammy`
* AWS Nitro doesn't use it at all
* On both Gigabyte and Dell systems (AMD and Intel respectively) I see it as being the string `001`/`01`.

I think if we want to say something about propolis itself, then this would be where we want the actual hardware platform revision to go in as this seems much more about the virtual platform.

hawkw:
@rmustacc I'm also not totally sure what the intended use-case of this is --- perhaps @luqmana (who opened issue #701 for adding this) has thoughts about why we would want to expose Propolis versions in SMBIOS?

luqmana:
> @rmustacc I'm also not totally sure what the intended use-case of this is --- perhaps @luqmana (who opened issue #701 for adding this) has thoughts about why we would want to expose Propolis versions in SMBIOS?

No particular use case beyond just tracking whether we wanted to hardcode it at all. I do think @rmustacc's point about identifying the "Virtual Platform" revision would make more sense. Either way, there's clearly a couple ways to go here.

Thanks for putting this together so quickly @hawkw and I think if we drop the SMBIOS bits it's still worth merging just to have the `--version` output. In which case, let's drop the bhyve api version from it as well. @pfmooney made the good point that that's not some fixed version tied to the binary itself. If anything, [`ApiVersion::current()`](https://github.com/oxidecomputer/propolis/blob/4591a60e74a20df9ab7f62a0e45fabd1ec667747/crates/bhyve-api/src/lib.rs#L590) would be more useful as quick way to figure out what a random propolis binary was built against without having to find the corresponding commit first.

hawkw:
I'll go ahead and back out the SMBIOS changes, I agree that it's nice to get commit information in the CLI version output regardless of what we do about SMBIOS. And I think including the bhyve version we were built against, rather than the current version detected at runtime, seems like a good call.

lib/propolis/src/lib.rs
iximeow: fwiw I think the detected bhyve API version we see at runtime is helpful here _in addition to_ `ApiVersion::current()`. something like "`propolis-standalone --version`  and `uname -v`" would get you a decent bit about the environment for an issue, then. there's not (that I know of) a `bhyve -opt`-like to discover the system's bhyve API version, so if we didn't include it in the version you'd have to go from the OS commit to the bhyve version, seems kinda.. annoying.

iximeow (commented):
just to note: I think it'd be nice to have this at some point even though we've since gotten programmable type 1 table information. I agree that the SMBIOS version makes more sense as the virtual platform version, but that doesn't preclude making `--version` a useful CLI argument

iximeow:
having talked with Eliza i'm cordycepsing this PR into putting version information into non-SMBIOS places

lib/propolis/src/firmware/smbios/table.rs
```diff
@@ -381,6 +381,22 @@ impl Table for Type1 {
     }
 }
 
+impl Default for Type1 {
```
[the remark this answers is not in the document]
iximeow: since i was reworking things, i ended up squashing the two commits here and removing this: https://github.com/oxidecomputer/propolis/pull/702/commits/90a2bc40404b29954aed780086d3cbe73b7bb8b9

lib/propolis/src/lib.rs
iximeow: i've made this even worse! in that the version string includes the discovered viona version as well now :)

iximeow: it is now not (and the new mention of viona is appropriately lowercase as well)

iximeow:
@leftwo :gift: finally the random propolis-servers that we run on bench sleds will tell us _what they are_ :)

papertigers:
The test flake here should be fixed in #1198, when that lands I will rerun the tests.

Cargo.toml
```diff
@@ -190,9 +190,9 @@ tracing-bunyan-formatter = "0.3.3"
 tracing-subscriber = "0.3.14"
 usdt = { version = "0.6", default-features = false }
 uuid = "1.3.2"
+vergen-git2 = { version = "10.0.0" }
```
papertigers, line 193: There's also a version of this crate that uses gix (pure rust git) vs libgit2. https://docs.rs/vergen-gix/latest/vergen_gix/

iximeow: (to mention here after we'd talked about it elsewhere: I'd picked `git2` mostly just because we're already going to be using libgit2 to interact with the Propolis repo via .. `git` ... but I'm not deeply attached to any particular choice)

leftwo (approved):
Hmm, I make take some ideas here and update what we do in crucible

iximeow:
I'm going to merge this with the failing test-gimlet job mostly because I don't want to further perturb this branch, but it would be useful to get the change in. the failing job here was due to the flake Mike mentioned above, and on trying to rerun it I've found what looks like a bug maybe relating to jobs with dependencies jobs which ran long ago? anyway, that's written up at https://github.com/oxidecomputer/buildomat/issues/119.

I think I'd have to push another commit here to rerun everything, and _that_ will fail because of the contemporary VIONA_CURRENT_INTERFACE_VERSION issue, so one way or the other _something_ in CI will not be happy today :(

PR #704, opened 2024-05-16
Update Omicron deps

bnaecker:
- Update `oximeter`, Omicron, and Crucible dependencies
- Remove defunct `base_route` field from producer info


pfmooney:
If this is updating the crucible dep, shouldn't it include a change to Cargo.toml?

bnaecker:
Updated in c9e2d38

pfmooney (approved):

leftwo (approved):
@pfmooney's approval is good enough, but I'll add mine as well :)

Issue #667, opened 2024-03-20
fw_cfg device should migrate contents


pfmooney:
The entries made available via the qemu `fw_cfg` device are all constructed (via a builder pattern) prior to starting the instance, being stored in an otherwise immutable data structure.  The only state from the device which is migrated is that which describes any transfers in progress (active entry selector, DMA address, etc).  As more things are added to fw_cfg, (such as generated ACPI tables)  it would probably make sense to migrate the contents too, rather than relying on the destination instance generating them exactly the same.

pfmooney:
Addressed in #705

PR #705, opened 2024-05-16
Modernize fw_cfg emulation

pfmooney:
Being one of the early parts of the prototypical Propolis, the fw_cfg emulation was rather unpolished, and lacked the ability for entries to be added or removed after instance initialization.

This change should improve it on several fronts:
- Entries can be added and/or removed during runtime
- Association of any RamFB device is more flexible and can also be modified during runtime
- fw_cfg entries are properly migrated, instead of requiring identical population on the destination instance.
- A modicum of test coverage is present

### DIFF
--- a/lib/propolis/src/hw/qemu/fwcfg.rs
+++ b/lib/propolis/src/hw/qemu/fwcfg.rs
@@ -2,22 +2,27 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use std::collections::BTreeMap;
+use std::collections::{btree_map, BTreeMap};
+use std::io::Write;
+use std::mem::size_of;
+use std::ops::{Deref, DerefMut};
 use std::sync::{Arc, Mutex, MutexGuard};
 
 use crate::accessors::MemAccessor;
 use crate::common::*;
+use crate::hw::qemu::ramfb::RamFb;
 use crate::migrate::*;
 use crate::pio::{PioBus, PioFn};
 use crate::vmm::MemCtx;
 use bits::*;
 
-use byteorder::{ByteOrder, BE, LE};
+use zerocopy::AsBytes;
 
-pub type Result = std::result::Result<(), &'static str>;
+const SIGNATURE_VALUE: &[u8; 4] = b"QEMU";
 
 #[allow(unused)]
 #[derive(Copy, Clone)]
+#[repr(u16)]
 pub enum LegacyId {
     Signature = 0x0000,
     Id = 0x0001,
@@ -47,7 +52,7 @@ pub enum LegacyId {
     FileDir = 0x0019,
 }
 impl LegacyId {
-    fn name(&self) -> &'static str {
+    const fn name(&self) -> &'static str {
         match self {
             LegacyId::Signature => "signature",
             LegacyId::Id => "id",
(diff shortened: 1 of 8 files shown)


pfmooney:
Commit message should be updated to reflect that it fixes #667 prior to integration

hawkw:
The [PHD failures](https://buildomat.eng.oxide.computer/wg/0/artefact/01HY1B2G1MBMNS2Y6RH8T63Z20/qtyE43kKLA8xzLElWCD5i3Fo09EHpgiUiGhpKqQ3lQCUGZvX/01HY1B31RXBGEN3Q2GNQSEZK48/01HY1FB7BFBER6KPMCQWKJBRG0/phd-runner.log?format=x-bunyan#L1993) are indicating that this revision is no longer migration-compatible with the current `master` (868cc546b53c862ec0180f2f0896041df51f953b), e.g. https://buildomat.eng.oxide.computer/wg/0/artefact/01HY1B2G1MBMNS2Y6RH8T63Z20/qtyE43kKLA8xzLElWCD5i3Fo09EHpgiUiGhpKqQ3lQCUGZvX/01HY1B31RXBGEN3Q2GNQSEZK48/01HY1FB7BFBER6KPMCQWKJBRG0/phd-runner.log?format=x-bunyan#L378. Is that expected, or should it work?

pfmooney:
> The [PHD failures](https://buildomat.eng.oxide.computer/wg/0/artefact/01HY1B2G1MBMNS2Y6RH8T63Z20/qtyE43kKLA8xzLElWCD5i3Fo09EHpgiUiGhpKqQ3lQCUGZvX/01HY1B31RXBGEN3Q2GNQSEZK48/01HY1FB7BFBER6KPMCQWKJBRG0/phd-runner.log?format=x-bunyan#L1993) are indicating that this revision is no longer migration-compatible with the current `master` ([868cc54](https://github.com/oxidecomputer/propolis/commit/868cc546b53c862ec0180f2f0896041df51f953b)), e.g. https://buildomat.eng.oxide.computer/wg/0/artefact/01HY1B2G1MBMNS2Y6RH8T63Z20/qtyE43kKLA8xzLElWCD5i3Fo09EHpgiUiGhpKqQ3lQCUGZvX/01HY1B31RXBGEN3Q2GNQSEZK48/01HY1FB7BFBER6KPMCQWKJBRG0/phd-runner.log?format=x-bunyan#L378. Is that expected, or should it work?

It is expected.  The migration payload is [substantially different](https://github.com/oxidecomputer/propolis/pull/705/commits/67195b05db3fd768dd71c23ecca967d8ef9a7bdf#diff-157a2fe73b817b6bc752a8072808061e2f8f83ff15fcee71bec5699c135539f5R729-R775)

hawkw:
Thanks, that's what I had figured but I thought it was worth double-checking.

I kind of wonder if we want to run just the migration-compatibility tests in a separate buildomat job from the other PHD tests, so that we can more clearly distinguish between test failures that indicate a commit broke something, and commits which introduce incompatible migration payload changes...

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -396,259 +304,353 @@ impl FwCfg {
...
+    pub fn remove(&self, selector: u16) -> Option<Entry> {
+        let mut state = self.state.lock().unwrap();
+        let entry = state.directory.remove(selector)?;
+        Self::validate_selected(&mut state);
+        Some(entry)
+    }
+    pub fn remove_named(&self, name: &str) -> Option<Entry> {
+        let mut state = self.state.lock().unwrap();
+        let selector = state.directory.named_selector(name)?;
+        let entry = state.directory.remove(selector).unwrap();
```
hawkw, line 356: nit, take it or leave it: maybe this `unwrap()` could be an `expect` with a message noting that we already would have returned `None` if there was no selector for that name? mostly for the benefit of a reader rather than because i expect this to panic.

not a big deal either way though.

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -396,259 +304,353 @@ impl FwCfg {
...
+                        .map_err(|_| FwCfgErr::OpUnsuccessful)?;
+                    Ok(wo.bytes_read())
+                } else {
+                    Ok(0)
+                }
             }
-        }
-        state.offset = end_offset;
-        Ok(())
+        }?;
```
hawkw, line 560: this is, admittedly, a very minor nitpick, but i find the `Ok`-wrapping here a little bit unnecessarily convoluted --- perhaps this would be a bit clearer:
```rust
let len = match entry {
    Entry::FileDir | Entry::Bytes(_) => return Err(FwCfgErr::ReadOnly),
    Entry::RamFb => {
       if let Some(ramfb) = state.ramfb.as_ref() {
            ramfb
                .fwcfg_rw(RWOp::Write(&mut wo))
                .map_err(|_| FwCfgErr::OpUnsuccessful)?;
            wo.bytes_read()
        } else {
            0
        }
    }
};
```

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -396,259 +304,353 @@ impl FwCfg {
...
+                    GuestAddr(req.addr.get()),
+                    buf_len,
                 ))
-                .ok_or("bad GPA")?;
-            let mut wo = WriteOp::from_mapping(offset as usize, mapping);
-            self.xfer(selector, RWOp::Write(&mut wo))?;
+                .ok_or(FwCfgErr::BadAddr)?;
+            let mut wo = WriteOp::from_mapping(0, map);
+            self.write(state, &mut wo)?;
+        } else if opts.contains(FwCfgDmaCtrl::SKIP) {
```
hawkw, line 648: The fact that we will do a read operation even if the `WRITE` or `SKIP` bit is set, and do a write regardless of whether the `SKIP` bit is set, felt a bit sketchy to me --- until I looked at [the QEMU documentation](https://www.qemu.org/docs/master/specs/fw_cfg.html#guest-side-dma-interface) and saw that it describes this exact behavior. Perhaps we ought to have a comment explaining that?

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -738,4 +808,190 @@ mod bits {
...
+    pub struct FwCfgFile {
+        size: BE32,
+        select: BE16,
+        reserved: u16,
+        name: [u8; FWCFG_FILENAME_LEN],
+    }
+    impl FwCfgFile {
+        pub fn new(size: u32, select: u16, name: &str) -> Self {
+            let name_len = name.as_bytes().len();
+            assert!(name_len < FWCFG_FILENAME_LEN);
```
hawkw, line 823: probably not an issue at present since all of our names are hard-coded string literals, but [the QEMU docs explicitly state that this is an ASCII string](https://www.qemu.org/docs/master/specs/fw_cfg.html#file-directory-key-0x0019-fw-cfg-file-dir), and this code would happily accept non-ASCII UTF-8 characters. perhaps we ought to add an `assert!(name.is_ascii())` or something here? if we were to eventually start accepting configurable user-defined fw_cfg file entries, we would want to remember to validate that their names are ASCII-only...

hawkw (approved):
Overall, this looks good to me --- I wasn't very familiar with the previous code, but I've tried to review the new code to the best of my ability. I had a few small nitpicks and questions, but no major issues; the question about enforcing ASCII-only strings is the only thing I'm actually concerned about and it's not a problem with the present code.

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -396,259 +304,353 @@ impl FwCfg {
...
+        Self::validate_selected(&mut state);
+        Some(entry)
+    }
+
+    /// Ensure that the selected entry (if any) is still valid after a change to
+    /// the directory, clearing the selection if there was trouble.
+    ///
+    /// This should not happen unless the VMM chooses to remove an item which
+    /// the running guest had selected.  Doing so is rather unsporting.
+    fn validate_selected(state: &mut MutexGuard<State>) {
```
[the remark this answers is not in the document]
pfmooney: Done.

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -86,305 +91,208 @@ pub enum LegacyX86Id {
...
-            let item = Entry { content };
-            self.entries.insert(sel, item);
-            self.name_to_sel.insert(name, sel);
-            Ok(())
-        } else {
-            Err("entry already exists")
-        }
+    fn next(&mut self) -> Option<Self::Item> {
+        let (name, selector) = self.iter.next()?;
+        let (entry, _name) = self.entries.get(selector).unwrap();
```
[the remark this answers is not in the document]
pfmooney: Done.

lib/propolis/src/hw/qemu/fwcfg.rs
pfmooney: Done.

lib/propolis/src/hw/qemu/fwcfg.rs
pfmooney: I thought it was clearer with the single bail-out point, so I'm inclined to leave it.

lib/propolis/src/hw/qemu/fwcfg.rs
pfmooney: Yeah a comment is a good suggestion.  Hopefully the addition is suitable?

lib/propolis/src/hw/qemu/fwcfg.rs
pfmooney: The consuming code in Linux doesn't seem to care.  I added a comment to the `insert` method suggestion that ASCII be used, but eschewed enforcement for now.

lib/propolis/src/hw/qemu/fwcfg.rs
hawkw: Looks good to me!

lib/propolis/src/hw/qemu/fwcfg.rs
hawkw: Mm, if we're reasonably confident that Linux guests will survive a stray Unicode character, the comment is probably fine. Thanks for looking into it!

pfmooney: I think Linux is much more likely to be grumpy about the entry keys implying a nonsensical directory structure rather than some non-ASCII Unicode.

PR #708, opened 2024-05-22
Update and prune dependencies

pfmooney:


pfmooney:
The `termwiz` dep cannot be updated until repnop/mac_address#47 is merged, and a release is cut, since one of the inner wezterm crates uses it for uuid generation.

hawkw (approved):

PR #711, opened 2024-06-14
Don't return None from BlockData::block_for_req when attachment is paused

### LINKED ISSUE #709: Guest reboot got stuck at unmounting /boot/efi
I ran into this issue with ubuntu instance reboot both while rebooting with API and in-guest reboot command.

From the serial console, I could see that the guest failed to make any progress at this point when I did `sudo reboot`:
```
[  OK  ] Stopped Update UTMP about System Boot/Shutdown.
[  OK  ] Stopped Network Time Synchronization.
[  OK  ] Stopped Create Volatile Files and Directories.
[  OK  ] Stopped target Local File Systems.
[  OK  ] Stopped target Mounted snaps.
         Unmounting /boot/efi...
         Unmounting /run/snapd/ns/lxd.mnt...
```
With instance reboot API, it got stuck at start up instead:
```
[  OK  ] Started Session 1 of User ubuntu.
[  301.807462] cloud-init[1378]: loading tables: 2 for warehouse: 10
[  302.097101] cloud-init[1378]: loading tables: 2 for warehouse: 9
         Starting Download data for…led at package install time...
[  OK  ] Finished Download data for…ailed at package install time.
         Starting Time & Date Service...
```

### LINKED ISSUE #710: PHD: raw buffer backend doesn't properly consume characters after matching
The raw buffer's `satisfy_or_set_wait` function doesn't do what its doc comment says: https://github.com/oxidecomputer/propolis/blob/3c0f998399ff13d605f3cf0bb724190f96eff4a2/phd-tests/framework/src/serial/raw_buffer.rs#L74-L110

This copies the buffer contents that preceded the match string so they can be returned to the caller, but doesn't actually clear the wait buffer.

This turns out not to be a problem today because our existing test cases rely on `run_shell_command` to send commands to the guest, and the default shell command sequence explicitly clears the back buffer of its own volition (specifically to keep the newlines it sends from appearing in command output). But new test cases that can't use `run_shell_command` for some reason (e.g. a guest reboot test that's not guaranteed to get another shell prompt after inputting the reboot command) are likely to run afoul of this.

gjcolombo:
Change `BlockData::block_for_req` so that it doesn't return early when the device/backend pair is paused. (The async block backend's request future doesn't have this issue.)

Add PHD tests for guest reboot (both in-guest and API-driven) to verify the fix; the non-Crucible variants of these tests fail without this fix and pass with them. The Crucible variants (which use the async driver) pass without issue, but it's useful to get coverage there, too.

Fix PHD's raw-buffering serial adapter so that, if a caller asks to wait for a string to appear in the buffer, the buffer properly trims its contents when the wait is satisfied. This is needed to keep the new tests from passing trivially; without the fix, the first login sequence to appear on the serial console satisfies all of the "wait for guest to boot" calls in the rest of the test.

Tests: PHD runs of Alpine and Debian 11 w/new test cases.

Fixes #709. Fixes #710.

### DIFF
--- a/lib/propolis/src/block/attachment.rs
+++ b/lib/propolis/src/block/attachment.rs
@@ -70,16 +70,12 @@ impl BlockData {
                         // conditions under protection of the lock before
                         // finally blocking.
                         let guard = self.lock.lock().unwrap();
-                        match check_state(att_state) {
-                            Err(ReqError::Stopped | ReqError::Detached) => {
-                                return None;
-                            }
-                            Ok(())
-                            | Err(ReqError::Paused | ReqError::NonePending) => {
-                                let _guard = self.cv.wait(guard).unwrap();
-                                continue;
-                            }
+                        if !att_state.is_attached() || att_state.is_stopped() {
+                            return None;
                         }
+
+                        let _guard = self.cv.wait(guard).unwrap();
+                        continue;
                     }
                     _ => {
                         return None;
--- a/lib/propolis/src/block/attachment.rs
+++ b/lib/propolis/src/block/attachment.rs
@@ -66,15 +66,20 @@ impl BlockData {
                 }
                 Err(err) => match err {
                     ReqError::NonePending | ReqError::Paused => {
-                        let guard = self.lock.lock().unwrap();
                         // Double-check for attachment-related "error"
                         // conditions under protection of the lock before
                         // finally blocking.
-                        if check_state(att_state).is_err() {
-                            return None;
+                        let guard = self.lock.lock().unwrap();
+                        match check_state(att_state) {
+                            Err(ReqError::Stopped | ReqError::Detached) => {
+                                return None;
+                            }
+                            Ok(())
+                            | Err(ReqError::Paused | ReqError::NonePending) => {
+                                let _guard = self.cv.wait(guard).unwrap();
+                                continue;
+                            }
                         }
-                        let _guard = self.cv.wait(guard).unwrap();
-                        continue;
                     }
                     _ => {
                         return None;
--- a/phd-tests/framework/src/serial/raw_buffer.rs
+++ b/phd-tests/framework/src/serial/raw_buffer.rs
@@ -95,7 +95,8 @@ impl RawBuffer {
             "checking wait on raw serial buffer"
         );
         if let Some(idx) = self.wait_buffer.rfind(&waiter.wanted) {
-            let out = self.wait_buffer[..idx].to_owned();
+            let out = self.wait_buffer.drain(..idx).collect();
+            self.wait_buffer = self.wait_buffer.split_off(waiter.wanted.len());
 
             // Because incoming bytes from Propolis may be processed on a
             // separate task than the task that registered the wait, this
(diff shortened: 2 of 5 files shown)


gjcolombo:
The new Clippy failures seem to be Rust 1.79-related--I only started seeing them locally after a `rustup update`. I'll investigate these and deal with them in a separate PR.

The failed PHD run seems to be due to an error downloading an artifact from Buildomat:

```
phd-runner: test phd_tests::migrate::from_base::migration_from_base_and_back ... FAILED: constructing test VM
    
    Caused by:
        0: building environment for new VM
        1: setting up VM execution environment
        2: request or response body error: error reading a body from connection: stream error received: unexpected internal error encountered
        3: error reading a body from connection: stream error received: unexpected internal error encountered
        4: stream error received: unexpected internal error encountered

    Stack backtrace:
       0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
                 at /home/build/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.83/src/backtrace.rs:27:14
       1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
                 at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/result.rs:1964:27
       2: phd_framework::artifacts::buildomat::<impl phd_framework::artifacts::DownloadConfig>::download_buildomat_uri::{{closure}}
                 at ./oxidecomputer/propolis/phd-tests/framework/src/artifacts/buildomat.rs:258:21
       3: phd_framework::artifacts::store::<impl phd_framework::artifacts::DownloadConfig>::download_buildomat_artifact::{{closure}}
                 at ./oxidecomputer/propolis/phd-tests/framework/src/artifacts/store.rs:589:72
       4: phd_framework::artifacts::store::StoredArtifact::ensure::{{closure}}
                 at ./oxidecomputer/propolis/phd-tests/framework/src/artifacts/store.rs:116:22
       5: phd_framework::artifacts::store::Store::get_propolis_server::{{closure}}
                 at ./oxidecomputer/propolis/phd-tests/framework/src/artifacts/store.rs:392:65
```

I've submitted this for a rerun.

lib/propolis/src/block/attachment.rs
```diff
@@ -66,15 +66,20 @@ impl BlockData {
...
+                        match check_state(att_state) {
+                            Err(ReqError::Stopped | ReqError::Detached) => {
+                                return None;
+                            }
+                            Ok(())
+                            | Err(ReqError::Paused | ReqError::NonePending) => {
+                                let _guard = self.cv.wait(guard).unwrap();
+                                continue;
+                            }
                         }
```
pfmooney, line 82: I would just make the calls to `is_attached()` and `is_stopped()` here, instead of using `check_state`.  Handling the `NonePending` error case is misleading, since `check_state` never emits it.  It suggest that more is being checked than really is.

pfmooney (commented):
Thanks for fixing this.

lib/propolis/src/block/attachment.rs
gjcolombo: Good call, I'll fix this up.

gjcolombo: Should be fixed in 099e2be.

pfmooney (approved):

PR #713, opened 2024-06-25
Allow boot order config in propolis-standalone

pfmooney:
While the propolis-server bits of this aren't entirely hammered out, I figured I'd get the initial pieces in.

### DIFF
--- a/bin/propolis-standalone/README.md
+++ b/bin/propolis-standalone/README.md
@@ -23,6 +23,9 @@ memory = 1024
 # Exit propolis-standalone process with <code> if instance reboots (default: unset)
 # exit_on_reboot = <code>
 
+# Override boot order (via communication to OVMF bootrom)
+# boot_order = ["net0", "block0"]
+
 [block_dev.alpine_iso]
 type = "file"
 path = "/path/to/alpine-extended-3.12.0-x86_64.iso"
--- a/bin/propolis-standalone/src/main.rs
+++ b/bin/propolis-standalone/src/main.rs
@@ -981,7 +981,9 @@ fn generate_bootorder(config: &config::Config) -> anyhow::Result<fwcfg::Entry> {
                 order.add_pci(get_pci_path().location, "device");
             }
             dev => {
-                anyhow::bail!("Unsupported boot device type: {dev}");
+                anyhow::bail!(
+                    "Boot device '{name}' is unsupported device type: {dev}"
+                );
             }
         }
     }
@@ -1298,7 +1300,13 @@ fn setup_instance(
     // It is "safe" to generate bootorder (if requested) now, given that PCI
     // device configuration has been validated by preceding logic
     if config.main.boot_order.is_some() {
-        fwcfg.insert_named("bootorder", generate_bootorder(&config)?).unwrap();
+        fwcfg
+            .insert_named(
+                "bootorder",
+                generate_bootorder(&config)
+                    .context("Failed to generate boot order")?,
+            )
+            .unwrap();
     }
 
     fwcfg.attach(pio, &machine.acc_mem);
--- a/bin/propolis-standalone/src/main.rs
+++ b/bin/propolis-standalone/src/main.rs
@@ -952,6 +952,42 @@ fn generate_smbios(params: SmbiosParams) -> anyhow::Result<smbios::TableBytes> {
     Ok(smb_tables.commit())
 }
 
+fn generate_bootorder(config: &config::Config) -> anyhow::Result<fwcfg::Entry> {
+    let names = config.main.boot_order.as_ref().unwrap();
+
+    let mut order = fwcfg::formats::BootOrder::new();
+    for name in names.iter() {
+        let dev = config
+            .devices
+            .get(name)
+            .ok_or(anyhow::anyhow!("Could not find device: {name}"))?;
+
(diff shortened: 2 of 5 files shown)


bin/propolis-standalone/README.md
```diff
@@ -23,6 +23,9 @@ memory = 1024
 # Exit propolis-standalone process with <code> if instance reboots (default: unset)
 # exit_on_reboot = <code>
 
+# Override boot order (via communication to OVMF bootrom)
```
hawkw, line 26: nitpicky, not a big deal: maybe worth explicitly noting here that this is an ordered list of device names (and not `block_dev` names)?

bin/propolis-standalone/src/main.rs
```diff
@@ -952,6 +952,42 @@ fn generate_smbios(params: SmbiosParams) -> anyhow::Result<smbios::TableBytes> {
...
 
+fn generate_bootorder(config: &config::Config) -> anyhow::Result<fwcfg::Entry> {
+    let names = config.main.boot_order.as_ref().unwrap();
+
+    let mut order = fwcfg::formats::BootOrder::new();
+    for name in names.iter() {
+        let dev = config
+            .devices
+            .get(name)
+            .ok_or(anyhow::anyhow!("Could not find device: {name}"))?;
```
hawkw, line 963: nitpicky: this error message doesn't really indicate that they occurred while trying to generate the boot order. maybe we should say something like 
```suggestion
            .ok_or(anyhow::anyhow!("Could not find boot device: {name}"))?;
```
or, change the caller to add `.context("Failed to generate boot order from config file")` or something to the call to `generate_bootorder`, so that the user sees an error indicating which part of the config file was invalid, before indicating _why_ it was invalid.

bin/propolis-standalone/src/main.rs
```diff
@@ -952,6 +952,42 @@ fn generate_smbios(params: SmbiosParams) -> anyhow::Result<smbios::TableBytes> {
...
+                order.add_disk(get_pci_path().location);
+            }
+            "pci-nvme" => {
+                order.add_nvme(get_pci_path().location, 0);
+            }
+            driver if driver.starts_with("pci-") => {
+                order.add_pci(get_pci_path().location, "device");
+            }
+            dev => {
+                anyhow::bail!("Unsupported boot device type: {dev}");
```
hawkw, line 984: it might be worth including the name of the device that was a type we didn't support? maybe something like:
```suggestion
                anyhow::bail!("Boot device '{name}' is an unsupported device type: {dev}");
```

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -1044,3 +1045,98 @@ mod test {
...
+        /// ```
+        pub fn add_pci(&mut self, loc: pci::BusLocation, kind: &str) {
+            self.0.push(Self::format_pci(loc, kind));
+        }
+
+        /// Add an NVMe disk.  This assumes namespace 1, as our NVMe emulation
+        /// does not currently support multiple namespaces in a device.
+        pub fn add_nvme(&mut self, loc: pci::BusLocation, eui: u64) {
+            // The decoding in OVMF demands that the bootorder entry identify
+            // nvme devices as vendor=0x8086 and device=5845.  While that
```
hawkw, line 1086: well, that's cool :| 

(it _might_ be worth including a link to the relevant code in OVMF, but it's certainly not a big deal to me)

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -1044,3 +1045,98 @@ mod test {
...
+        pub fn finish(self) -> Entry {
+            let Self(mut entries) = self;
+            entries.push("HALT\0".to_owned());
+
+            Entry::Bytes(entries.join("\n").to_string().into())
+        }
+
+        fn format_pci(loc: pci::BusLocation, name: &str) -> String {
+            let (slot, func): (u8, u8) = (loc.dev.into(), loc.func.into());
+            format!("/pci@i0cf8/{name}@{slot:x},{func:x}")
```
hawkw, line 1104: for my own education: what is the meaning of `i0cf8`? i did some googling and encountered a bunch of examples of the Open Firmware bootorder format, but didn't find anything explaining what this string means...

hawkw (approved):
This looks good to me! I had a couple minor suggestions about the error messages, but no blocking concerns.

lib/propolis/src/hw/qemu/fwcfg.rs
pfmooney: The IO ports for PCI config space access are `0xcf8 `(address) and `0xcfc` (data)

bin/propolis-standalone/README.md
pfmooney: I thought that this was clear contextually, both with the inclusion of `net0`, and that `block0`'s backend is not named the same.

hawkw: Yeah, that's fair!

hawkw:
Latest changes look good, thanks for the error message improvements!

pfmooney:
Thanks for takin' a look.

pfmooney:
Merged in 59868677c70f3cd03f03e12584ad1056da8b5459

Issue #706, opened 2024-05-16
Instance (re)creation from migration payload


pfmooney:
Up until now, when a a Propolis instance is migrated, the rough "shape" of that instance (devices, etc) is created on the target VMM using the instance specification, after which it is "re-hydrated" using the resulting migration payload.  Some of the details of that target instance depend on recreating identical data during initialization, such as `fw_cfg` contents (see #667) or the VPD info exposed via PCI devices.

It would be nice if the migration payloads were suitably enhanced to make it possible to reliably recreate an instance from only said migration payload (perhaps excluding basic details like vCPU count, etc).

Items which still require attention:
- [ ] `fw_cfg` data - #667 
- [ ] PCI VPD bits (vendor, model, etc)

PR #682, opened 2024-04-17
migration: refrain from offering all pages when possible

hawkw:
Presently, when transferring guest RAM as part of a migration, Propolis will
always offer *all* guest pages in the initial pre-pause RAM transfer phase.
After the guest is paused, Propolis will then offer any pages which were marked
as dirty since the pre-pause RAM transfer, ensuring the most recent guest memory
is transferred. In the initial pre-pause RAM push, however, we cannot offer only
dirty RAM, because tracking which pages are dirty using the
`VM_TRACK_DIRTY_PAGES` ioctl clears the dirty bit, and we may need to perform a
subsequent migration should this one fail. That subsequent migration will have a
bad time if it only offers dirty pages and we have cleared the dirty bits on any
pages that were touched only prior to the previous failed migration. This is a
shame, because we may end up transferring a bunch of pages that the guest has
never actually touched and doesn't care about.

Now, thanks to @pfmooney's upstream work in [illumos-gate#3266], we can
be more efficient about what pages are transferred in the initial push, without
dropping a bunch of pages on the ground if we retry a failed migration. Using
the `VNO_OP_SET_DIRTY` operation performed by the `VM_NPT_OPERATION` ioctl that
Patrick has added to bhyve, we can put back any dirty bits that were cleared
during a failed migration, allowing a subsequent migration to also offer only
dirty pages. Since this ioctl is only available with the latest bhyve (v17), we
will still fall back to the previous behavior of offering all guest RAM if
 `VM_NPT_OPERATION` is unavailable.

Initially, I had planned to just change this code to use the `VNO_OP_GET_DIRTY`
operation, which accesses but does not clear the dirty bits. This way, I
figured, we could just always transfer all dirty pages. However, we *do*
actually want to be clearing dirty bits initially, because we want the
post-pause RAM push to only have to transfer pages that have been touched
_since_ the pre-pause RAM push phase. If we used `VNO_OP_GET_DIRTY` to read the
dirty bits without clearing them for the initial pre-pause RAM push, the
post-pause RAM push would offer all the same pages, which defeats the purpose of
splitting the RAM push into two pages in the first place. Thus, instead, we do
clear the dirty bits, but we store which pages were dirty and put them back
using `VNO_OP_SET_DIRTY` should the migration fail.

The PHD tests I added way back in #623 test retrying a failed migration with
running processes, which should exercise this change. I've run the tests on
`atrium`, which has an Illumos kernel that includes Patrick's bhyve changes:

<details open>

[...]

PR #715, opened 2024-07-01
server: refactor instance initialization

### LINKED ISSUE #432: server: `assertion failed: server_context.services.oximeter_stats.lock().await.is_none()` due to race in teardown
Repro steps: seen in Omicron instance stress.

The basic problem here is

- `register_oximeter` asserts that the server context's `oximeter_stats` is `None`
- `register_oximeter` gets called from `instance_ensure_common`
  - the call is tied to instance ensure because the Oximeter registration uses the VM ID as a series identifier, and that's not known until instance ensure time
- meanwhile, during instance shutdown, `ServiceProviders::stop` calls `take_controller` before `oximeter_stats.take()`

The net result of this is that during shutdown, there's a small window in which a new instance can be ensured (because the controller was taken) but the Oximeter registration hasn't been taken yet, which causes the assertion to fail.

There are several possible ways to fix this:

- use a bigger lock to protect the `ServiceProviders` instead of letting them be locked individually
- reorder the teardown steps so that the VM controller is taken last (but I don't like this very much; anything that requires this sort of strict ordering forces us to remember what that ordering is, and we also have to be sure it's safe to drop `oximeter_stats` first)
- push `oximeter_stats` into the VM controller itself (might be OK, the only place we use it is to count instance reboot requests, and that could easily be pushed down to the controller)

---

**Triage:** marking as MVP. The reasons this comes up in Omicron stress in the first place appear to be that (a) the control plane reuses sled IDs and Propolis IDs when stopping and starting VMs, and (b) sled agent will reuse Propolis zones if possible (i.e. there seems to be a race that will allow sled agent to reuse a zone with a Propolis that previously hosted a stopped VM). These behaviors appear to cause lots of other problems (e.g. sled agent crashes due to zones being in unexpected states), so we probably want to address this behavior in the control plane, which will mitigate this issue.

gjcolombo:
Refactor propolis-server to prepare for more work on instance specs and the live migration protocol. This change does a few things; in rough order of importance, they are:

### Initialize migration targets' VM components after starting the migration task

Before this change, a request to create a VM that included a migration source would initialize all the VM's components first (using the parameters passed to the ensure API), then run the live migration protocol to import state into those components. Now the instance init code starts the migration protocol first and creates VM components at the LM task's behest. This enables a couple of things:

- The path to initializing devices from their migration payloads is now somewhat straighter; see #706.
- It's also much easier to have migration sources tell migration targets what components to create (instead of requiring the control plane to specify identical component manifests) if we want to do so.

### Clarify who is able change a VM's configuration

Before this change, calls to change a Crucible volume's configuration were not synchronized with other VM state changes (e.g. ongoing live migrations). They also took a lock that protected a running VM's instance spec, but not any of its component collections; this is fine when configuring a backend in-place, but dangerous for other (future) operations like device hotplug. To help stabilize this area:

- A server's Propolis components and its instance spec are now encapsulated in a single struct (`VmObjects`).
- Dropshot API handlers can lock the active `VmObjects` for shared access if they just need to read or query state.
- APIs that need to mutate state need to submit that request to the VM's state driver task (the `ActiveVm` and `VmObjects` structs' impls and visibility markers prevent APIs from acquiring an exclusive lock directly). This allows the driver task to apply disposition rules to new requests (e.g. "no new requests to mutate if they follow a request to migrate out").

### No state machine in the Dropshot server layer

Before this change, the propolis-server Dropshot server context contained a `VmControllerState` type, which tracked whether the server thought there was an extant VM, and a `ServiceProviders` type that owned references to the "external" services offered from the VM's components (serial console access, VNC access, and Oximeter metrics). Move all this logic down into the `vm` module so that the API layer is responsible just for dispatching incoming calls and translating responses into appropriate HTTP errors.

### General cleanup

The old `vm` module was a huge mess of types and associated functions. Break this up into several sub-modules for readability.

## Testing notes

Tested with `cargo test` and a full PHD run with a Debian 11 guest and Crucible and migrate-from-base tests enabled.

Fixes #432.


gjcolombo:
Despite the draft status, this is more or less ready to review, and comments are very welcome. I noticed at least one more bolt I want to tighten (in external request queue cleanup) and am going to fix it before making the PR official.

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -349,11 +359,41 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
...
+                .await
+                .unwrap();
+
+            let response = self.response_rx.recv().await.unwrap();
+            match response {
+                MigrateSourceResponse::RedirtyingFailed(has_failed) => {
+                    if has_failed {
+                        self.dirt = None;
+                    }
+                }
```
hawkw, line 389: I'm not sure if I love the idea of doing a somewhat complex message-passing handshake thingy to send a bit of state that's only relevant to the code within this module out of this module, into some other task, and then ask it to please give us that bit of state that only we care about back. I don't see any obvious _bugs_ with this,[^1] it just feels quite convoluted and it makes me wonder if there's a simpler solution.

(my [original approach](https://github.com/oxidecomputer/propolis/pull/682/files/721a0e4313a5975065d107040f958b5dfb4864dc#r1569141450) was, for the record, just to stick a `static AtomicBool` in this module and move on with our lives...)

[^1]: It _does_ seem a lot less efficient to stick this in a queue, wait until every other request in the queue ahead of it has been processed, and then wait to get our response back, but the performance effects probably aren't actually meaningful in a protocol where we have to _actually send all the guest memory_... :)

hawkw (commented):
I'm still trying to wrap my head around this change --- there's a lot going on here. Here's a cursory review of the first couple files in the diff, as I continue reading the rest.

bin/propolis-server/src/lib/migrate/source.rs
gjcolombo: It seems very doable just to execute migration protocols on the driver task itself and just omit the message passing entirely. I'm going to try to put that together tomorrow and put up a second PR stacked on this one (assuming it pans out).

pfmooney (commented):
I need to spend more time going through the server bits, but here are some round-1 comments.

lib/propolis/src/tasks.rs
```diff
@@ -419,37 +419,35 @@ impl Drop for CtrlHeld<'_> {
...
+        let mut guard = self.0.lock().await;
         guard.extend(tasks);
     }
 
     /// Block until all held tasks have been joined, returning any resulting
     /// [task::JoinError]s after doing so.
-    pub fn block_until_joined(&self) -> Option<Vec<task::JoinError>> {
-        let mut guard = self.0.lock().unwrap();
+    pub async fn block_until_joined(&self) -> Option<Vec<task::JoinError>> {
+        let mut guard = self.0.lock().await;
```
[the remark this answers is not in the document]
gjcolombo: Done in d6d5700

lib/propolis/src/block/mod.rs
```diff
@@ -244,7 +245,7 @@ pub trait Backend: Send + Sync + 'static {
     ///
     /// Spawning of any tasks required to do such request processing can be done
     /// as part of this start-up.
-    fn start(&self) -> anyhow::Result<()>;
+    fn start(&self) -> BoxFuture<'_, anyhow::Result<()>>;
```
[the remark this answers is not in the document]
gjcolombo: (We discussed this one a bit offline.)

I think most of the ergonomic problems in the implementations of this trait come from having callers who dynamically dispatch to it, which prevents the use of standard-issue `async fn`s in the trait. I revised this in aacaff8 to use the `async_trait` crate to at least get rid of some of the extra callee-side paperwork this created and filed #716 to track removing the Box allocation for non-async impls.

The other option is to convert the state driver back into an OS thread (as it was before this change), but as more async-heavy work moves into that context (e.g. driving live migration) so that it can call routines that call `block_on`, but I'm reluctant to go that way, because I anticipate more async-heavy work (like driving live migration) to move onto the state driver task over time.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -110,7 +110,7 @@ pub struct MachineInitializerState {
 pub struct MachineInitializer<'a> {
     pub(crate) log: slog::Logger,
     pub(crate) machine: &'a Machine,
-    pub(crate) devices: DeviceMap,
+    pub(crate) devices: LifecycleMap,
```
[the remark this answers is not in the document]
gjcolombo: Done in d712bf6.

bin/propolis-server/src/lib/vm/objects.rs
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    /// Yields an exclusive lock guard referring to the underlying object
+    /// collection.
+    ///
+    /// This function is only visible within the `vm` module so that only the
+    /// state driver can obtain a mutable reference to the underlying objects.
+    pub(super) async fn write(&self) -> RwLockWriteGuard<VmObjectsLocked> {
+        self.inner.write().await
+    }
```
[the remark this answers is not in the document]
gjcolombo: Good suggestions both; done in 9f23e19.

bin/propolis-server/src/lib/vm/objects.rs
```diff
@@ -0,0 +1,455 @@
...
+
+    /// Yields a mutable reference to the VM's current instance spec.
+    pub(crate) fn instance_spec_mut(&mut self) -> &mut InstanceSpecV0 {
+        &mut self.instance_spec
+    }
+
+    /// Yields the VM's current kernel VMM handle.
+    pub(crate) fn machine(&self) -> &Machine {
+        &self.machine
+    }
```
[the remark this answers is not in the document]
gjcolombo: Done in 9f23e19 (along with a similar change for access to the memory context, since that's also quite common and quite verbose).

bin/propolis-server/src/lib/vm/objects.rs
```diff
@@ -0,0 +1,497 @@
...
+    inner: RwLock<VmObjectsLocked>,
+}
+
+/// A collection of objects that should eventually be wrapped in a lock and
+/// stored in a `VmObjects` structure. See [`VmObjectsLocked`].
+pub(super) struct InputVmObjects {
+    pub instance_spec: InstanceSpecV0,
+    pub vcpu_tasks: Box<dyn VcpuTaskController>,
+    pub machine: Machine,
+    pub lifecycle_components: DeviceMap,
```
pfmooney, line 53: Since we're keeping the `DeviceMap` name for now, would we want something to reflect that here as well, rather than `lifecycle_components`?  Up to you.

lib/propolis/src/block/mod.rs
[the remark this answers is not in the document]
pfmooney: Yeah, moving back to an OS thread for pushing the state driver forward seems less than ideal.  With #716 filed, this works for me.

bin/propolis-server/src/lib/migrate/source.rs
gjcolombo: I've put up a stacked PR, #720, that runs live migration work on the state driver (and makes things a bit more generic so that ~~if~~ when we have multiple protocol versions in the future, the driver doesn't have to care about which one was negotiated). That change gets rid of all the message passing, and should also take care of the other comments above (much less type/unwrap gore).

pfmooney (commented):
Sorry, I had a lingering unpublished comment

bin/propolis-server/src/lib/server.rs
```diff
@@ -751,8 +347,16 @@ async fn instance_spec_get(
...
 ) -> Result<HttpResponseOk<api::InstanceGetResponse>, HttpError> {
-    let (instance, _) = instance_get_common(&rqctx).await?;
-    Ok(HttpResponseOk(api::InstanceGetResponse { instance }))
+    instance_get_common(&rqctx).await.map(|full| {
+        HttpResponseOk(api::InstanceGetResponse {
+            instance: api::Instance {
+                properties: full.properties,
+                state: full.state,
+                disks: vec![],
+                nics: vec![],
```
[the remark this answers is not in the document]
gjcolombo: Filed #723 for this.

bin/propolis-server/src/lib/vm/objects.rs
gjcolombo: Done in 8606927.

gjcolombo:
> This is going to need some extra work to rebase on top of https://github.com/oxidecomputer/propolis/commit/99fa5647fa1a0baa747344ff7abab9f6e9155a8c; trying to sort that out now.

This is done in e0ed049. I tested it by running `pstack` on a few Propolis servers and verifying that `pstack` reports the expected number of `tokio-rt-vmm` threads.

bin/propolis-server/src/lib/vm/mod.rs
```diff
@@ -2,1181 +2,579 @@
...
-    /// server).
-    External(ExternalRequest),
-}
-
-/// Commands issued by the state driver back to its VM controller. These are
-/// abstracted into a trait to allow them to be mocked out for testing without
-/// having to supply mock implementations of the rest of the VM controller's
-/// functionality.
-#[cfg_attr(test, mockall::automock)]
-trait StateDriverVmController {
```
gjcolombo, line 1005: The functions in this trait moved to `VmObjects`, but the trait itself is gone--it existed to provide a way to mock these functions out for state driver testing. I haven't lost a lot of sleep over this because (a) PHD gives us a lot of signal as to whether these state transitions are working properly, and (b) mocks as test doubles in Rust have never sparked joy for me (compared to, say, the C++ codebase I worked in at $OLDJOB). I'm open to suggestions as to how to replace this (maybe some kind of `FakeVmObjects` is warranted, though I'm not sure exactly how it would look).

gjcolombo (commented):
@hawkw I've taken the liberty of leaving a few comments on the LHS of the bigger files that got broken up (server.rs, vm/mod.rs) to try to help point out where code has moved to (since there are a lot of things that just moved without really changing). As always, let me know if I can answer any questions!

lib/propolis/src/tasks.rs
```diff
@@ -436,20 +436,21 @@ impl TaskGroup {
 
     /// Block until all held tasks have been joined, returning any resulting
     /// [task::JoinError]s after doing so.
-    pub fn block_until_joined(&self) -> Option<Vec<task::JoinError>> {
-        let mut guard = self.0.lock().unwrap();
-        let workers = std::mem::replace(&mut *guard, Vec::new());
+    pub async fn block_until_joined(&self) -> Option<Vec<task::JoinError>> {
```
hawkw, line 439: nit: since this is an `async fn` now, it probably should no longer be called `block_until_joined`, since it, you know...doesn't block. maybe `wait_until_joined`? (and perhaps update the doc comment as well)

hawkw (commented):
Overall, this looks good --- most of my suggestions were on fairly minor style nits. The one thing I actually had some concerns about was the implementation of the `InputQueue` in `state_driver.rs`, which appears to be an attempt to bridge between synchronous producers and an asynchronous consumer --- IIUC, there is probably a much easier way to do this that avoids the footgunny `tokio::task::block_in_place`. If there isn't, it would be nice to know more about why we have to do it this way.

Besides that, this looks good! Much nicer overall.

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -2,1383 +2,696 @@
...
+    /// Waits for an event to arrive on the input queue and returns it for
+    /// processing.
+    ///
+    /// External requests and guest events are stored in separate queues. If
+    /// both queues have events when this routine is called, the guest event
+    /// queue takes precedence.
+    fn wait_for_next_event(&self) -> InputQueueEvent {
+        // `block_in_place` is required to avoid blocking the executor while
+        // waiting on the condvar.
+        tokio::task::block_in_place(|| {
```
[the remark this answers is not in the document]
gjcolombo: Fixed in f168f30. (We discussed this offline, but for the room: using `send_blocking` isn't quite enough here--the reason we can't use a regular mpsc is that the `GuestEvent` subchannel's `enqueue` routine looks at the contents of the queue before deciding whether to push anything. So instead I've kept the queues the same and used a `Notify` as a signaling mechanism--hopefully all the synchronization is still correct.)

bin/propolis-server/src/lib/vm/objects.rs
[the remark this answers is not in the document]
gjcolombo: No worries--these were all really good suggestions! I took care of them en masse in c9b1bf6.

lib/propolis/src/tasks.rs
gjcolombo: Fixed in c4c80f2.

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -2,1383 +2,708 @@
...
-    fn set_published_state(&mut self, state: PublishedState) {
-        let (instance_state, migration_state) = match state {
-            PublishedState::Instance(i) => (Some(i), None),
-            PublishedState::Migration(m) => (None, Some(m)),
-            PublishedState::Complete(i, m) => (Some(i), Some(m)),
-        };
+            // Notifiers in this module must use `notify_one` so that their
+            // notifications will not be lost if they arrive after this routine
+            // checks the queues but before it actually polls the notify.
+            self.notify.notified().await;
```
hawkw, line 134: I think we should be using [`Notified::enable`](https://docs.rs/tokio/latest/tokio/sync/futures/struct.Notified.html#method.enable) here, to subscribe to a potential wakeup _before_ checking the event queues. This way, if there's a race condition where new elements are pushed to the queue but we have not yet started to wait for a notification, we won't miss the wakeup.

I think the correct code would look something like this:

```rust
        let notified = self.notify.notified();
        tokio::pin!(notified);
        loop {
            // Pre-subscribe to a wakeup from `self.notify` so that we don't
            // miss any notification that is sent while checking the event queues.
            notify.as_mut().enable();

            // Check whether there are any messages in either queue.
            {
                let mut guard = self.inner.lock().unwrap();
                if let Some(guest_event) = guard.guest_events.pop_front() {
                    return InputQueueEvent::GuestEvent(guest_event);
                } else if let Some(req) = guard.external_requests.pop_front() {
                    return InputQueueEvent::ExternalRequest(req);
                }
            }
            
            // Await a notification if the queues were empty.
            notified.as_mut().await;
            // Reset the `notified` future in case we must await an additional
            // notification.
            notified.set(self.notify.notified()
        }
```

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -2,1383 +2,708 @@
...
-              "migration" => ?migration);
+impl guest_event::VcpuEventHandler for InputQueue {
+    fn suspend_halt_event(&self, when: Duration) {
+        let mut guard = self.inner.lock().unwrap();
+        if guard
+            .guest_events
+            .enqueue(guest_event::GuestEvent::VcpuSuspendHalt(when))
+        {
+            self.notify.notify_one();
+        }
```
hawkw, line 171: nit, take it or leave it: this gets repeated enough times that _maybe_ we should have a 
```rust
impl InputQueue {
    // ...
    fn enqueue_guest_event(&self, event: guest_event::GuestEvent) {
        if self.inner.lock().unwrap().guest_events.enqueue(event) {
             self.notify.notify_one()
        }
    }
    // ...
}
```

not a big deal either way, though.

lib/propolis/src/tasks.rs
```diff
@@ -434,22 +434,27 @@ impl TaskGroup {
...
-    pub fn block_until_joined(&self) -> Option<Vec<task::JoinError>> {
-        let mut guard = self.0.lock().unwrap();
-        let workers = std::mem::replace(&mut *guard, Vec::new());
+    /// Waits until all the workers in this task group have completed.
+    ///
+    /// # Return value
+    ///
+    /// `None` if all the tasks completed successfully. `Some` if at least one
+    /// task failed; the wrapped value is a `Vec` of all of the returned errors.
+    pub async fn join_all(&self) -> Option<Vec<task::JoinError>> {
```
hawkw, line 443: Now that this is async, I think we could could conceivably replace it with a new implementation using [`tokio::task::JoinSet`](https://docs.rs/tokio/latest/tokio/task/join_set/struct.JoinSet.html) instead of a `Vec<JoinHandle<()>>`. This would be more efficient, as it wouldn't have to poll every `JoinHandle` in a loop. I'm not sure if this is hot enough that it matters if we do that or not, though, and it would be fine to do it in a follow-up.

hawkw (changes requested):
Sorry to hold up merging this a second time, but I think there is a potential race in the new implementation of `InputQueue::wait_for_next_event`. I left a comment describing how we can fix it. 

Besides that, everything else looks good --- thanks for addressing my remaining feedback! I'll happily approve this once the race has been addressed.

hawkw (approved):
Nevermind, I was wrong about that --- the race I was concerned about only exists in a multi-consumer use case, which isn't what this is. Sorry for the false alarm!

PR #719, opened 2024-07-03
Issue vlan-enabling ioctl when run in falcon

Nieuwejaar:


crates/viona-api/src/lib.rs
```diff
@@ -19,6 +19,19 @@ impl VionaFd {
...
+                ioctls::VNA_IOC_SET_PROMISC,
+                vioc_promisc::VIONA_PROMISC_ALL_VLAN as usize,
+            )
+            .is_err()
+        {
+            // Until/unless this support is integrated into stlouis/illumos,
+            // this is an expected failure.   This is needed to use vlans,
+            // but shouldn't affect any other use case.
+            println!("failed to enable promisc mode on viona");
+        }
```
pfmooney, line 34: Doing this as a feature-gated compile-time decision seems inappropriate to me.  Why not make it a parameter to `PciVirtioViona::new()` and optionally perform the ioctl there?

pfmooney: There are plausible reasons why a non-falcon consumer might want to set such a promisc mode, even though there are none presently.  I'd prefer having that painted out through the interface rather than the viona-api crate changing its implicit behavior due to feature presence/absence.

Nieuwejaar: It isn't even available in `stlouis` yet.

For context: This new ioctl allows the `viona` driver to receive all vlan-tagged packets, expanding on the existing `VIONA_PROMISC_ALL` ioctl.  I imagine this could be integrated to upstream `illumos` without objection.  The problem is that there is no corresponding mechanism to allow `propolis` to indicate that the driver should transmit all vlan-tagged packets.  It's hard-coded into the driver.  Until I come up with an integrate-able method for handling the `tx` path, `falcon` has to use a non-standard `viona` driver, which it will install as part of its setup.

pfmooney: Hmm, that's unfortunate.  While `VIONA_PROMISC_ALL_VLAN` remains in "private" forks, it should probably live behind a `falcon` cfg guard.  Without that, the header checks would otherwise fail on vanilla bits.

With that said, I would still keep the ioctl call itself in propolis-lib, rather than as part of the `VionaFd` constructor.  The viona-api crate is mostly meant as a thin convenience shim over the FFI bits.

Nieuwejaar: I've moved the call upstack to `PciVirtioViona::new()` and rebased.

Nieuwejaar: @pfmooney, would you mind taking another look at this when you have time?  Thanks.

crates/viona-api/sys/src/lib.rs
```diff
@@ -67,6 +68,14 @@ mod structs {
         pub vrs_qsize: u16,
         pub vrs_qaddr: u64,
     }
+
+    #[repr(C)]
+    pub enum vioc_promisc {
```
pfmooney, line 73: This should be renamed to `viona_promisc_t` to match the enum in `viona_io.h`?

crates/viona-api/sys/src/lib.rs
```diff
@@ -67,6 +68,14 @@ mod structs {
...
+        VIONA_PROMISC_NONE = 0,
+        VIONA_PROMISC_MULTI,
+        VIONA_PROMISC_ALL,
+        VIONA_PROMISC_ALL_VLAN,
```
pfmooney, line 77: The `falcon` feature should be painted through into `viona_api_sys`, and the `ALL_VLAN` variant hidden with `cfg()` until such time as it's available in at least stlouis.

crates/viona-api/sys/src/lib.rs
Nieuwejaar: I'm happy to do that, but all of the other definitions in this file omit the `_t`.  It's not a direct comparison, as the others are `struct`s rather than `enum`s, but leaving off the `_t` feels like a better fit.

crates/viona-api/sys/src/lib.rs
Nieuwejaar: Done.  I also rebased.  I don't usually do that mid-review, since it makes it harder to see incremental changes, but the total diff is small enough that it's not hard to keep track of.

crates/viona-api/sys/src/lib.rs
pfmooney: It's `viona_promisc_t`, not `vioc_promisc_t`.

Nieuwejaar: Sheesh.  That's embarrassing.  I had the two files open in side-by-side windows and couldn't see it.  Sorry for the clumsy oversight.

pfmooney: No worries.  One could argue that the name for that enum should be made consistent with the other structs in that header, considering the other names.

pfmooney (approved):

PR #720, opened 2024-07-03
Run live migration directly on the state driver task

gjcolombo:
Stacked on #715.

#715 changes propolis-server's VM state driver from an OS thread to a tokio task. This allows live migration protocols to run directly on the state driver without the use of `block_on`, so do just that. Specifically:

* Define `SourceProtocol` and `DestinationProtocol` traits that describe the routines the state driver uses to run a generic migration irrespective of its protocol version.
* Move the `migrate::source_start` and `migrate::dest_initiate` routines into factory functions that connect to the peer Propolis, negotiate protocol versions, and return an appropriate protocol impl.
* Use the protocol impls to run migration on the state driver task. Remove all the types and constructs used to pass messages between this task and migration tasks.

Also, try to improve the interface between the `vm` module and the `migrate` module for inbound migrations. Inbound live migration is tricky to manage because (with #715) the various steps needed to create Propolis VM objects are interleaved with the steps in the live migration protocol; if migration fails, the precise set of steps needed to drive the VM state machine to its final state depends on exactly how many VM initialization steps the migration protocol has taken so far. To try to hide this complexity, define some `VmEnsure` objects that migrations can use either to fully initialize a VM or to unwind correctly if a failure occurs.

Tested with a full PHD run using a Debian 11 guest.


bin/propolis-server/src/lib/vm/ensure.rs
gjcolombo, line 1: The `VmEnsure` types in this file are net new code. The "initialize VM components" free function in here moved in from `startup.rs` unchanged.

I arranged the initialization phases this way in part to leave room to allow future versions of the LM protocol to have more of a say in exactly what instance spec is used to initialize VM objects. Right now it's just the spec from the ensure request, but I have my eyes on using (at least part of) the instance spec received from the source in the sync phase instead.

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -114,77 +117,106 @@ enum RamOfferDiscipline {
...
+        vm_objects: &VmObjects,
+        publisher: &mut StatePublisher,
+        persistent_state: &mut PersistentState,
+    ) -> Result<(), MigrateError>;
+}
 
+/// Negotiates a live migration protocol version with a target who has connected
+/// over `conn`. If this is successful, returns a `SourceProtocol`
+/// implementation that can be used to run the requested migration.
+pub(crate) async fn initiate<T: MigrateConn>(
```
gjcolombo, line 144: Most of the logic in this function was taken from the (removed) `source_start` function in `migrate/mod.rs`.

phd-tests/tests/src/migrate.rs
```diff
@@ -161,6 +161,15 @@ mod running_process {
             .unwrap_err();
         info!(%error, "first migration failed as expected");
 
+        // Also verify tha the target reports that it failed.
```
gjcolombo, line 164: typo here and below, will fix

bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -24,105 +27,321 @@ use crate::migrate::probes;
...
-        if let Ok(e) = codec::Message::Error(err.clone()).try_into() {
-            let _ = proto.conn.send(e).await;
+    Ok(match selected {
+        Protocol::RonV0 => RonV0::new(log, migration_id, conn, local_addr),
+    })
+}
+
+/// A helper type for abstracting away type-level differences between the
+/// different phases of instance ensure.
+enum EnsureState<'a> {
```
gjcolombo, line 138: This wrapper enum and its usage below are the other major source of net-new code in this PR.

bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -24,105 +27,321 @@ use crate::migrate::probes;
...
+}
 
-/// Launches an attempt to migrate into a supplied instance using the supplied
-/// source connection.
-pub async fn migrate<T: AsyncRead + AsyncWrite + Unpin + Send>(
+/// Connects to a live migration source using the migration request information
+/// in `migrate_info`, then negotiates a protocol version with that source.
+/// Returns a [`DestinationProtocol`] implementation for the negotiated version
+/// that the caller can use to run the migration.
+pub(crate) async fn initiate<'e>(
```
gjcolombo, line 59: This function is more or less identical to the old `dest_initiate` function in the root migration module.

gjcolombo (commented):
This PR is just expansive enough that it's hard to tell what code moved and what code is new, so here are a few comments pointing out what's what (and one noting a typo that I need to go fix).

bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -24,105 +27,321 @@ use crate::migrate::probes;
...
 use super::protocol::Protocol;
+use super::MigrateConn;
+
+/// The interface to an arbitrary version of the target half of the live
+/// migration protocol.
+//
+// Use `async_trait` here to help generate a `Send` bound on the futures
+// returned by the functions in this trait.
+#[async_trait::async_trait]
+pub(crate) trait DestinationProtocol<'e> {
```
hawkw, line 43: not 100% sure if I get why this lifetime parameter is necessary? but AFAICT it doesn't seem to be causing any downstream complexity, so :woman_shrugging:

bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -24,105 +27,321 @@ use crate::migrate::probes;
...
+    /// created objects.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the state machine is not in the `NotStarted` state.
+    async fn activate(self) -> anyhow::Result<Self> {
+        let not_started = match self {
+            EnsureState::NotStarted(vm) => vm,
+            _ => unreachable!("can only activate a VM once"),
+        };
```
hawkw, line 155: style nit, unimportant: also write this as:
```suggestion
        let EnsureState::NotStarted(not_started) = self else {
            unreachable!("can only activate a VM once");
        };
```
which is one line shorter...but the match is also fine.

bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -24,105 +27,321 @@ use crate::migrate::probes;
...
-    Ok(())
+    /// If the state machine is in the `Active` state, locks the VM objects
+    /// shared and returns the corresponding guard. Returns `None` otherwise.
+    async fn vm_objects(&self) -> Option<VmObjectsShared> {
+        let objects = match self {
+            EnsureState::NotStarted(_) => return None,
+            EnsureState::Active(vm) => vm.vm_objects(),
+        };
+
+        Some(objects.lock_shared().await)
```
hawkw, line 201: unimportant style nit, take it or leave it: this could be just
```suggestion
        match self {
            EnsureState::NotStarted(_) => None,
            EnsureState::Active(vm) => Some(vm.vm_objects().lock_shared().await),
        }
```
although if you think that's less clear, the current code is fine too...

bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -24,105 +27,321 @@ use crate::migrate::probes;
...
 }
 
-impl<T: AsyncRead + AsyncWrite + Unpin + Send> DestinationProtocol<T> {
+#[async_trait::async_trait]
+impl<'e, T: MigrateConn + Sync> DestinationProtocol<'e> for RonV0<'e, T> {
+    async fn run<'ensure>(
+        mut self,
+        ensure: VmEnsureNotStarted<'ensure>,
+    ) -> Result<VmEnsureActive<'ensure>, MigrateError>
+    where
```
hawkw, line 233: I'm not totally sure if this actually makes things all that much better or not (haven't actually tried it yet)  but I kind of wonder whether it would be a bit nicer if `ensure_state` were a local variable in this function, rather than a field on `RonV0`, and we passed it into the methods on `RonV0`. That way, I _think_ we could avoid `take`ing and `unwrap`ping it as often.

I might try sketching that out in a PR against this branch, if you're interested in seeing what I mean?

bin/propolis-server/src/lib/migrate/destination.rs
gjcolombo: Embarrassingly, I got here mostly by throwing things at rustc until it compiled :grimacing: But, if I understand things right (and I very well may not), the issues here are:

- If there's no extra lifetime bound on the trait itself, then when we `impl DestinationProtocol for RonV0<'e>`, the compiler is unsure that the `'e` references in `self` won't be moved/copied into the `'ensure` return value. This produces a "lifetime may not live long enough" error with a hint that an extra `'e: 'ensure` bound would fix the problem.
- But adding `where 'e: 'ensure` just to this impl produces a stronger bound on the lifetimes in the impl than were expressed in the original trait, which is also a compiler error. The easiest way out of that is to name the extra lifetime in the trait and add the appropriate `where` bound to the trait definition, as I've done here.

With all that said: all this goes away if `RonV0` doesn't hold onto an `EnsureState` (as you suggested below), and I like the way that change is looking in my early hacking. If it holds up it should eliminate the need for this extra lifetime.

bin/propolis-server/src/lib/migrate/destination.rs
gjcolombo: Done in 6943610

bin/propolis-server/src/lib/migrate/destination.rs
gjcolombo: Done in 6943610

bin/propolis-server/src/lib/migrate/destination.rs
gjcolombo: Done in cbd51fa and 3c17194: I was worried this would end up being too verbose, but I think it's actually much simpler (and, taken to its conclusion, lets us get rid of the `EnsureState` enum entirely). Thanks for the suggestion!

bin/propolis-server/src/lib/migrate/destination.rs
gjcolombo: The latest push did indeed resolve this by getting rid of the relevant `RonV0` member (and in fact by getting rid of `EnsureState` entirely).

bin/propolis-server/src/lib/vm/ensure.rs
```diff
@@ -0,0 +1,359 @@
...
+            },
+        ));
+
+        match initialize_vm_objects_from_spec(
+            self.log,
+            &input_queue,
+            &self.ensure_request.properties,
+            &self.ensure_request.instance_spec,
+            self.ensure_options,
+        )
```
hawkw, line 112: not a big deal, but: any reason this can't be a method on `VmEnsureNotStarted`, since it seems like most of its parameters are just fields on `self` that we could avoid passing explicitly?

bin/propolis-server/src/lib/vm/ensure.rs
```diff
@@ -0,0 +1,359 @@
...
+        self.vm
+            .make_active(
+                self.log,
+                self.input_queue.clone(),
+                &self.vm_objects,
+                vm_services,
+            )
+            .await;
+
+        let _ = self.ensure_response_tx.send(Ok(InstanceEnsureResponse {
```
hawkw, line 206: what does it actually mean if the receiver has gone away, here? should we give up on the ensure?

bin/propolis-server/src/lib/vm/ensure.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    #[cfg(not(feature = "omicron-build"))]
+    init.initialize_test_devices(&options.toml_config.devices)?;
+    #[cfg(feature = "omicron-build")]
+    info!(log, "`omicron-build` feature enabled, ignoring any test devices");
+
+    #[cfg(feature = "falcon")]
+    init.initialize_softnpu_ports(&chipset)?;
+    #[cfg(feature = "falcon")]
+    init.initialize_9pfs(&chipset)?;
```
hawkw, line 331: unimportant style nit, take it or leave it: why not
```suggestion
    #[cfg(feature = "falcon")] {
        init.initialize_softnpu_ports(&chipset)?;
        init.initialize_9pfs(&chipset)?;
    }
```

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -253,87 +253,58 @@ pub(super) async fn run_state_driver(
...
+                    };
+                }
+            };
 
-                vm.vm_init_failed(objects.is_some()).await;
-                let _ = ensure_result_tx
-                    .send(Err(VmError::InitializationFailed(e)));
+            match migration.run(ensure).await {
+                Ok(activated) => activated,
+                Err(_) => {
```
hawkw, line 286: what happens to the error returned by the migration attempt here? is it already recorded/logged someplace, or is it just getting thrown away here?

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -253,87 +253,58 @@ pub(super) async fn run_state_driver(
...
+                    return StateDriverOutput {
+                        state_publisher,
+                        final_state: InstanceState::Failed,
+                    }
+                }
+            }
```
hawkw, line 292: these matches feel a bit...visually unfortunate...but i couldn't immediately come up with a nicer way to write this that still does all the requisite error handling, so i dunno. probably fine as is.

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -114,77 +117,106 @@ enum RamOfferDiscipline {
...
+            .await?;
+            error!(
+                log,
+                "destination side did not begin migration version handshake: \
+                 {:?}",
+                x
+            );
+            return Err(MigrateError::Initiate);
+        }
+    };
```
hawkw, line 208: style nit, take it or leave it: there are two `match`es here, one of which handles whether or not we receive a text message on the websocket stream, and another which handles whether the text message can be parsed successfully. currently, the second match is nested inside of the first one. IMHO, this code would be a lot easier to read if the two matches were in sequence, rather than nested inside of each other, like this: 
```rust

    let dst_protocols = match conn.next().await {
        Some(Ok(tungstenite::Message::Text(dst_protocols))) => dst_protocols,
        x => {
            conn.send(tungstenite::Message::Close(Some(CloseFrame {
                code: CloseCode::Protocol,
                reason: "did not begin with version handshake.".into(),
            })))
            .await?;
            error!(
                log,
                "destination side did not begin migration version handshake: \
                 {:?}",
                x
            );
            return Err(MigrateError::Initiate);
        },
    };

    info!(log, "destination offered protocols: {}", dst_protocols);
    let selected = match protocol::select_protocol_from_offer(&dst_protocols) {
        Ok(Some(selected)) => {
            info!(log, "selected protocol {:?}", selected);
            conn.send(tungstenite::Message::Text(
                selected.offer_string(),
            ))
            .await?;
            selected
        }
        Ok(None) => {
            let src_protocols = protocol::make_protocol_offer();
            error!(
                log,
                "no compatible destination protocols";
                "dst_protocols" => &dst_protocols,
                "src_protocols" => &src_protocols,
            );
            return Err(MigrateError::NoMatchingProtocol(
                src_protocols,
                dst_protocols,
            ));
        }
        Err(e) => {
            error!(log, "failed to parse destination protocol offer";
                    "dst_protocols" => &dst_protocols,
                    "error" => %e);
            return Err(MigrateError::ProtocolParse(
                dst_protocols,
                e.to_string(),
            ));
        }
    };
```
(I couldn't leave this as a "suggestion" because it contains deleted lines apparently :/)

This way, we are saying "first, try to receive a text message, or bail out if we get something weird. then, try to parse that message". I think that's a bit easier to follow for the reader.

bin/propolis-server/src/lib/migrate/source.rs
```diff
@@ -310,22 +391,72 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> SourceProtocol<T> {
...
-        self.run_phase(MigratePhase::RamPushPostPause).await?;
-        self.run_phase(MigratePhase::TimeData).await?;
-        self.run_phase(MigratePhase::DeviceState).await?;
-        self.run_phase(MigratePhase::RamPull).await?;
-        self.run_phase(MigratePhase::ServerState).await?;
-        self.run_phase(MigratePhase::Finish).await?;
-
-        info!(self.log(), "Source Migration Successful");
-        Ok(())
+        let result: Result<(), MigrateError> = async {
```
hawkw, line 394: why is the type annotation necessary here? Don't all the calls to `run_phase` return the same type?

bin/propolis-server/src/lib/migrate/destination.rs
```diff
@@ -24,176 +27,298 @@ use crate::migrate::probes;
...
+                reason: "did not respond to version handshake.".into(),
+            })))
+            .await?;
+            error!(
+                log,
+                "source instance failed to negotiate protocol version: {:?}", x
+            );
+            return Err(MigrateError::Initiate);
+        }
+    };
```
hawkw, line 128: style nit, take it or leave it: there are two `match`es here, one of which handles whether or not we receive a text message on the websocket stream, and another which handles whether the text message can be parsed successfully. currently, the second match is nested inside of the first one. IMHO, this code would be a lot easier to read if the two matches were in sequence, rather than nested inside of each other, like this: 
```rust
    let selected_protocol = match conn.next().await {
        Some(Ok(tungstenite::Message::Text(selected_protocol))) => selected_protocol,
        x => {
            conn.send(tungstenite::Message::Close(Some(CloseFrame {
                code: CloseCode::Protocol,
                reason: "did not respond to version handshake.".into(),
            })))
            .await?;
            error!(
                log,
                "source instance failed to negotiate protocol version: {:?}", x
            );
            return Err(MigrateError::Initiate);
        }
    };

    info!(log, "source negotiated protocol {}", selected_protocol);
    let selected = match super::protocol::select_protocol_from_offer(
        &selected_protocol,
    ) {
        Ok(Some(selected)) => selected,
        Ok(None) => {
            let offered = super::protocol::make_protocol_offer();
            error!(log, "source selected protocol not on offer";
                "offered" => &offered,
                "selected" => &selected_protocol);

            return Err(MigrateError::NoMatchingProtocol(
                selected_protocol,
                offered,
            ));
        }
        Err(e) => {
            error!(log, "source selected protocol failed to parse";
                "selected" => &selected_protocol);

            return Err(MigrateError::ProtocolParse(
                selected_protocol,
                e.to_string(),
            ));
        }
    };
```
(I couldn't leave this as a "suggestion" because it contains deleted lines apparently :/)

This way, we are saying "first, try to receive a text message, or bail out if we get something weird. then, try to parse that message". I think that's a bit easier to follow for the reader.

hawkw (approved):
This looks good to me, I like the changes to remove the `EnsureState` enum! I left some more style suggestions --- let me know what you think!

bin/propolis-server/src/lib/migrate/source.rs
gjcolombo: They do, but for some reason the compiler needs a hint as to the error type. (It's fine with `_` as the success type.)

hawkw: weird. that's fine, i guess.

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: d6177e0 should help with this.

gjcolombo:
Thanks as always, @hawkw! Except where I commented otherwise the style feedback is incorporated into 6feb3c8.

hawkw:
Looks great to me!

PR #722, opened 2024-07-13
Import RFB and overhaul VNC

pfmooney:
This imports the RFB crate from [its standalone repo](https://github.com/oxidecomputer/rfb) and then overhauls its interfaces (as well as the consumer code in propolis-server) in order to prepare for subsequent improvements to the VNC experience in the future.


pfmooney:
FYI: I left the commits after the import largely un-squashed for now in the hopes that it would make review a bit easier, but plan to squash them down prior to integration.

gjcolombo (approved):
The overall direction of this makes sense to me, particularly given the details that it removed from services.rs. Thanks for putting this together!

I looked through the RFB and VNC module changes in moderate detail and didn't see anything amiss, though I wouldn't say I understand the RFB protocol terribly deeply. Still, the overall thrust of the refactor looks good to me.

PR #724, opened 2024-07-22
Move instance timeseries definitions to TOML

bnaecker:
- Use central TOML definitions for the vCPU usage and PVPANIC timeseries. Note that this change is _breaking_ because we pull in the backwards-incompatible definitions that include the sled identifiers.
- Add the sled-identifiers into the `InstanceMetadata` passed to Proplis via its instance-ensure API request. The sled-agent can provide those from its populated baseboard when provisioning instances.
- Support letting the oximeter producer server use internal DNS to look up Nexus for metric registration. This helps avoid failures to register if the Nexus we use originally disappears during reconfiguration of those services. This doesn't happen yet, but the work is ongoing in Omicron.
- Add more ergonomic argument to control metric registration via `propolis-cli`, ensuring backwards compatibility. Also add arguments for the sled-identifiers.


bnaecker:
I expect this to fail the PHD migration tests in the same way that https://github.com/oxidecomputer/propolis/pull/659 did, because I've added some new metadata to the instance-ensure request that the code running on `main` will neither provide nor understand. We're still expecting to stop all instances during an upgrade, so this should be acceptable, but I'd love a gut check on that too.

Once we're happy with this, I'll include a PR in Omicron that updates the pinned Propolis here, and _also_ deletes the past incompatible timeseries definitions.

bin/propolis-server/src/lib/server.rs
```diff
@@ -46,16 +47,22 @@ use crate::vnc::PropolisVncServer;
...
 impl MetricsEndpointConfig {
-    pub fn new(propolis_addr: SocketAddr, metric_addr: SocketAddr) -> Self {
-        Self { propolis_addr, metric_addr }
+    pub fn new(
+        listen_addr: IpAddr,
+        registration_addr: Option<SocketAddr>,
+    ) -> Self {
+        Self { listen_addr, registration_addr }
     }
 }
```
pfmooney, line 67: Since all the members are `pub`, why not do direct struct initialization where this is used?

bin/propolis-server/src/lib/stats/pvpanic.rs
```diff
@@ -4,13 +4,14 @@
...
 use chrono::Utc;
-use oximeter::{
-    types::{Cumulative, Sample},
-    Metric, MetricsError, Producer,
-};
+use oximeter::{types::Sample, Metric, MetricsError, Producer};
 use propolis::hw::qemu::pvpanic;
 use std::sync::Arc;
 
+oximeter::use_timeseries!("virtual-machine.toml");
```
pfmooney, line 11: Should we be expecting a `virtual-machine.toml` be included in the repo as part of this change?

bin/propolis-server/src/lib/stats/pvpanic.rs
```diff
@@ -4,13 +4,14 @@
...
+use self::virtual_machine::PvPanicGuestHandled;
+use self::virtual_machine::PvPanicHostHandled;
```
pfmooney, line 13: style nit: combine into one import

bin/propolis-server/src/lib/stats/virtual_machine.rs
```diff
@@ -14,11 +14,13 @@
...
-    types::Cumulative, FieldType, FieldValue, Metric, Sample, Target,
-};
+use oximeter::{types::Cumulative, FieldType, FieldValue, Sample, Target};
+use std::borrow::Cow;
 use std::collections::BTreeMap;
-use uuid::Uuid;
+
+oximeter::use_timeseries!("virtual-machine.toml");
+use self::virtual_machine::VcpuUsage;
+use self::virtual_machine::VirtualMachine as VirtualMachineTarget;
```
pfmooney, line 23: style nit: combine into one import

openapi/propolis-server.json
```diff
@@ -913,11 +913,30 @@
           "silo_id": {
             "type": "string",
             "format": "uuid"
+          },
```
pfmooney, line 916: Need to generate an updated doc for falcon as well

bin/propolis-cli/src/main.rs
```diff
@@ -219,14 +248,22 @@ async fn new_instance(
     cloud_init_bytes: Option<String>,
     silo_id: TypedUuid<SiloKind>,
     project_id: TypedUuid<ProjectKind>,
+    sled_id: TypedUuid<SledKind>,
+    sled_model: String,
+    sled_revision: u32,
+    sled_serial: String,
```
pfmooney, line 254: It's time to start wrapping some of these args in structs, now that we're at a baker's dozen

crates/propolis-api-types/src/lib.rs
```diff
@@ -36,6 +36,10 @@ pub struct InstancePathParams {
 pub struct InstanceMetadata {
     pub silo_id: Uuid,
     pub project_id: Uuid,
+    pub sled_id: Uuid,
+    pub sled_serial: String,
+    pub sled_revision: u32,
+    pub sled_model: String,
```
pfmooney, line 42: Should we add a (phd?) test for checking that these values are updated when an instance is migrated?

bin/propolis-server/src/lib/stats/pvpanic.rs
bnaecker: No, that lives [here](https://github.com/oxidecomputer/omicron/blob/main/oximeter/oximeter/schema/virtual-machine.toml) in Omicron. Part of the work to support updates to timeseries schema includes centralizing them in Omicron. There are obviously trade-offs here, and this is where we've landed for now.

openapi/propolis-server.json
bnaecker: Yeah, thank you. That was done in ec187cd.

crates/propolis-api-types/src/lib.rs
bnaecker: Sure that seems good. I've not written any PHD tests, so might take me a bit.

bin/propolis-server/src/lib/server.rs
bnaecker: Done in 8a71e3d

bin/propolis-server/src/lib/stats/pvpanic.rs
bnaecker: Done in 8a71e3d

bin/propolis-server/src/lib/stats/virtual_machine.rs
bnaecker: Done in 8a71e3d

bin/propolis-cli/src/main.rs
bnaecker: Done in 8a71e3d

bin/propolis-server/src/lib/stats/pvpanic.rs
hawkw: I think it might be worth including a comment on this line explaining that, so that someone reading this file knows where to look to see the schema?

crates/propolis-api-types/src/lib.rs
hawkw: You'll probably want to start by looking at the [existing PHD tests for live migration](https://github.com/oxidecomputer/propolis/blob/master/phd-tests/tests/src/migrate.rs). Hopefully, the test-support library should be *pretty* straightforward, but do let me know if you have questions!

bin/propolis-server/src/lib/stats/pvpanic.rs
bnaecker: Commented in 4d5453a

crates/propolis-api-types/src/lib.rs
bnaecker: I added a test in 4d5453a. It feels a bit contrived, since we're mostly testing the internal `TestVm`'s handling of instance metadata, so I'm on the fence about it. Happy to keep or remove, depending on what others think.

gjcolombo: With the caveat that I haven't looked at the non-PHD parts of the PR very carefully yet: I think the test would feel less contrived if the metadata were part of `struct VmSpec` instead of being randomly generated in `start_local_vm`. I'd arrange that like this:

- add an `InstanceMetadata` to `struct VmSpec`
- have `VmConfig::vm_spec` generate random IDs (right now this is in `start_local_vm`)
- add a `VmSpec` function that generates new sled IDs/revisions (in your change this is happening in `instance_ensure_internal`); call this function from `Framework::spawn_successor_vm` (just like the call to `refresh_crucible_backends`)

Then most of the changes to `struct TestVm` can go away, and your new test would look something like this:

```rust
async fn migration_ensures_instance_metadata(ctx: &Framework) {
    // Create a source instance, and fetch the instance metadata its metrics are
    // generated with.
    let mut source = ctx
        .spawn_default_vm("migration_ensures_instance_metadata_source")
        .await?;
    let mut target = ctx
        .spawn_successor_vm(
            "migration_ensures_instance_metadata_target",
            &source,
            None,
        )
        .await?;

    source.launch().await?;
    source.wait_to_boot().await?;
    let source_expected = source.vm_spec().metadata;
    let source_metadata = source.get_spec().await?.properties.metadata;
    assert_eq!(source_metadata, source_expected);

    // Migrate the instance to a new server, and refetch the metadata.
    target
        .migrate_from(&source, Uuid::new_v4(), MigrationTimeout::default())
        .await?;
    let target_expected = target.vm_spec().metadata;
    let target_metadata = target.get_spec().await?.properties.metadata;
    assert_eq!(target_metadata, target_expected);

    // gjc: I might not keep the "project and silo IDs shouldn't change"
    // requirement; the control plane has to uphold that, not Propolis. You
    // could have a unit test for `VmSpec::refresh_sled_info` that makes sure
    // this invariant is upheld, though.
}
```

WDYT? (Does this suggestion make any sense in view of what's going on in the rest of the PR? It very well may not, in which case you should feel free to ignore it :))

bnaecker: Thanks for the suggestion @gjcolombo. That does seem slightly less contrived, in that at least the manipulation of the source and target metadata is contained and mostly transparent to the tests. I've made those updates in 7981b96, with a few minor visibility modifications to glue it all together. Thanks!

bin/propolis-server/src/lib/stats.rs
```diff
@@ -194,6 +184,7 @@ pub async fn register_server_metrics(
     virtual_machine: VirtualMachine,
     log: &Logger,
 ) -> anyhow::Result<ServerStatsOuter> {
+    // TODO(ben): Fetch sled-agent identifiers here, or before and pass them in.
```
gjcolombo, line 187: More for my edification than anything: the sled agent IDs end up in the `VirtualMachine`, correct? What else is there to plumb here? (The caller of this function has an `&InstanceProperties` already, so it should be able to furnish whatever IDs are missing here.)

phd-tests/tests/src/migrate.rs
```diff
@@ -396,3 +396,54 @@ async fn run_serial_history_test(
...
+        "Target instance was not populated with the correct instance metadata"
+    );
+
+    // Check that the source / target sled identifiers are different.
+    //
+    // In general, we expect that the sled serial and ID _should_ be different;
+    // the model and revision _may_ be different (if migrating to a different
+    // class of compute sled entirely); and that the project / silo IDs should
+    // be the same. That's the job of the control plane, though, not Propolis,
+    // so here we just check that the sled serial and UUIDs are different.
```
gjcolombo, line 440: I might reword this a bit: the sled ID and serial are expected to be different here because `spawn_successor_vm`'s behavior is to change them, and the target is expected to report its own identifiers and not the source's. (`spawn_successor_vm` behaves that way because that roughly emulates what the control plane would do, but it's technically fine for them to differ in this respect, so I might move the bulk of this comment to `refresh_sled_identifiers` and just describe things here in terms of `spawn_successor_vm`.)

gjcolombo (approved):
LGTM with a couple of little comments about the comments.

bin/propolis-server/src/lib/stats.rs
bnaecker: Oh this was an outdated comment, thanks for catching that. I was originally going to have Propolis call back to the sled-agent for these, but it's much simpler to just pass them in the API request as I ended up with here. Will remove.

phd-tests/tests/src/migrate.rs
bnaecker: Done in 4e6ebde

bin/propolis-server/src/lib/stats.rs
bnaecker: Removed in 4e6ebde

bin/propolis-server/src/main.rs
```diff
@@ -21,6 +23,49 @@ use propolis_server::{
...
+    }
+}
+
+impl FromStr for MetricRegistration {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "disable" => Ok(Self::Disable),
+            "dns" => Ok(Self::Dns),
```
hawkw, line 50: nitpicky, take it or leave it: should we, perhaps, use [`str::eq_ignore_ascii_case`](https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case) here, in the interest of "be conservative in what you emit, liberal in what you accept"?

bin/propolis-server/src/main.rs
```diff
@@ -35,16 +80,30 @@ enum Args {
...
         #[clap(
             name = "VNC_IP:PORT",
             default_value_t = SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 5900),
             action
         )]
         vnc_addr: SocketAddr,
+
+        /// Logging level for the server
+        #[clap(long, default_value_t = slog::Level::Info, value_parser = parse_log_level)]
+        log_level: slog::Level,
```
hawkw, line 106: Adding this feels like a good idea, but it also seems pretty much completely unrelated to this change --- what's the motivation for doing it as part of this PR?

bin/propolis-server/src/main.rs
```diff
@@ -161,6 +213,48 @@ fn build_logger() -> slog::Logger {
...
+///
+/// This will return None if metrics are explicitly disabled.
+fn build_metric_configuration(
+    log: &Logger,
+    metrics_addr: MetricRegistration,
+    listen_addr: IpAddr,
+) -> anyhow::Result<Option<MetricsEndpointConfig>> {
+    let cfg = match metrics_addr {
+        MetricRegistration::Disable => {
+            debug!(
```
hawkw, line 234: i feel like i would probably log this at `Info` (at least) --- this seems like the kind of information that i would always want to see when the process starts up at the default log level, and it only gets logged once...

hawkw (approved):
Overall, this looks good to me! I commented on a couple pretty minor nitpicks, but they're not terribly important.

bin/propolis-cli/src/main.rs
```diff
@@ -647,6 +676,22 @@ async fn main() -> anyhow::Result<()> {
...
+                    .unwrap_or_else(TypedUuid::new_v4)
+                    .into_untyped_uuid(),
+                sled_id: sled_id
+                    .unwrap_or_else(TypedUuid::new_v4)
+                    .into_untyped_uuid(),
+                sled_model: sled_model
+                    .unwrap_or_else(|| String::from("fake-gimlet")),
+                sled_revision: sled_revision.unwrap_or(1),
+                sled_serial: sled_serial
+                    .unwrap_or_else(|| String::from("fake-serial")),
```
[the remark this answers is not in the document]
bnaecker: Good idea, will do. I didn't do the UUIDs for the reason you mention, but should have done the strings. Thanks!

bin/propolis-server/src/main.rs
bnaecker: I wanted to see the `oximeter_producer::Server` log messages to confirm that it was appropriately using internal DNS. That's all logged at `debug` or `trace`, so I needed a way to increase verbosity at the top level.

bin/propolis-server/src/main.rs
bnaecker: Done in 56792ba

bin/propolis-cli/src/main.rs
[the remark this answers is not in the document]
bnaecker: Done in 56792ba

bin/propolis-server/src/main.rs
bnaecker: Done in 56792ba

bnaecker:
Thanks for the input everyone! The only issue here is the expected failure to migrate between this branch and the merge base, since I changed the API. Merging shortly!

Issue #716, opened 2024-07-02
Device and block backend lifecycle ops could return immediately if no async work is required


gjcolombo:
`trait Lifecycle`'s `paused` function requires all its implementations to return a `BoxFuture` even if pausing requires no async work (e.g. because there's nothing to pause!). This function could instead return an enum with "Done" and "Pending" variants where only the latter contains a `BoxFuture`, allowing trivial impls to return immediately without a heap allocation.

#715 proposes to introduce a similar issue in the `block::Backend` trait (though masks it a bit with the use of `async_trait`). Several storage backends don't have any `async` callees in their start and stop methods and would benefit from a similar enum.

PR #725, opened 2024-07-22
Use block_in_place for device pause/halt

pfmooney:
This should address cases where propolis-server was found to be wedged when the viona interrupt poller task happened to be assigned to the same tokio thread as the synchronous pause/halt action.

### DIFF
--- a/bin/propolis-server/src/lib/vm/objects.rs
+++ b/bin/propolis-server/src/lib/vm/objects.rs
@@ -344,9 +344,11 @@ impl VmObjectsLocked {
 
     /// Pauses all of a VM's devices.
     async fn pause_devices(&self) {
-        self.for_each_device(|name, dev| {
-            info!(self.log, "sending pause request to {}", name);
-            dev.pause();
+        tokio::task::block_in_place(|| {
+            self.for_each_device(|name, dev| {
+                info!(self.log, "sending pause request to {}", name);
+                dev.pause();
+            });
         });
 
         struct NamedFuture {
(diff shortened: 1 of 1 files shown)


jmpesp:
Took this branch for a spin - it fixed the wedge for me as well, thanks!

hawkw (approved):

bin/propolis-server/src/lib/vm/objects.rs
```diff
@@ -344,9 +344,11 @@ impl VmObjectsLocked {
 
     /// Pauses all of a VM's devices.
     async fn pause_devices(&self) {
-        self.for_each_device(|name, dev| {
-            info!(self.log, "sending pause request to {}", name);
-            dev.pause();
+        tokio::task::block_in_place(|| {
```
hawkw, line 347: maybe worth a comment noting why this is considered a blocking op? up to you, though.

gjcolombo (approved):
Not for this PR, but does it make sense to try to extend the work #716 to cover pause? (It seems like it might be more idiomatic to make `pause` return either "ok I'm paused" or the relevant future, rather than having callers call back into `paused`.)

pfmooney:
> It seems like it might be more idiomatic to make pause return either "ok I'm paused" or the relevant future

Yeah, that seems like the way to go, IMO.

bin/propolis-server/src/lib/vm/objects.rs
pfmooney: Will do.

pfmooney:
Merged in 0a3a26baf66b97e1666a1f566aa3e93a9a5a0c1d

PR #727, opened 2024-07-26
PciVirtio should properly import lintr state

pfmooney:
Guests should rarely (read: never) be using legacy pin interrupts on virtio (or really any PCI) devices, but if they do, we should properly import that state during a migration.

### DIFF
--- a/lib/propolis/src/hw/virtio/pci.rs
+++ b/lib/propolis/src/hw/virtio/pci.rs
@@ -123,10 +123,7 @@ impl<D: PciVirtio + Send + Sync + 'static> pci::Device for D {
     }
     fn interrupt_mode_change(&self, mode: pci::IntrMode) {
         let vs = self.virtio_state();
-        vs.set_intr_mode(self.pci_state(), mode.into());
-
-        // Make sure the correct legacy register map is used
-        vs.map_which.store(mode == pci::IntrMode::Msix, Ordering::SeqCst);
+        vs.set_intr_mode(self.pci_state(), mode.into(), false);
     }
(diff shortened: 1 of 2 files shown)


hawkw:
Clippy [looks like it's recently decided to care about indentation in doc comments](https://github.com/oxidecomputer/propolis/actions/runs/10118491882/job/27985389319?pr=727) which is great and all, but not related to this change.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -667,19 +667,25 @@ impl IsrState {
...
     }
-    /// Write ISR value
-    fn write(&self, intr_queue: bool, intr_cfg: bool) {
-        self.sync_pin(|inner| {
-            inner.intr_queue = intr_queue;
-            inner.intr_cfg = intr_cfg;
-        });
+    /// Import ISR value
+    ///
+    /// Sets the internal ISR value without propogating the state to the
```
gjcolombo, line 675: nit: s/propogating/propagating

gjcolombo (approved):
LGTM, thanks for putting this together.

I migrated a VM with a virtio-net device between Propolises with this change, and confirmed that connectivity was preserved across migrations (#731) and that the virtio device register layout stayed in the MSI-X configuration (#730).

lib/propolis/src/hw/virtio/pci.rs
pfmooney: *shakes fist at `set spell`*

pfmooney: Fixed, along with another nit I found while I was in the neighborhood

PR #728, opened 2024-07-26
Fix typo in server metric CLI argument

bnaecker:
A previous commit changed the name of the metric registration CLI command, but did not also change the name of flag in the SMF manifest or method script. This fixes the original introduced typo, keeping the SMF parts unchanged.


bnaecker:
I goofed in #724 by renaming the server CLI flag from `--metric-addr` to `--metrics-addr` without realizing it. This puts it pack.

hawkw (approved):

PR #733, opened 2024-07-31
split migrate-from-base tests into a separate PHD run

### LINKED ISSUE #732: want separate buildomat job for PHD migrate-from-base tests
PHD's `migrate::from_base` tests break whenever there's a breaking change to the Propolis API or migration protocol. Such breaking changes are allowed under our current offline-only, all-instances-stopped upgrade discipline. But when someone intentionally makes such a change, the sole `phd-run` job in CI goes red, and the change author needs to look through the results by hand to make sure that only the expected tests have failed.

Using the `--include-filter` and `--exclude-filter` PHD options, it should be possible to run these tests in a separate Buildomat job to make it easier to parse the test results. This extra job can be removed when these sorts of changes are no longer allowed.

gjcolombo:
Move the existing phd-run script into a callable phd-run-with-args.sh that passes its arguments through to its phd-runner invocation. Then create separate "run PHD" jobs that call this script, respectively including and excluding the migrate-from-base tests from their runs.

Fixes #732.

### DIFF
--- a/.github/buildomat/jobs/phd-run.sh
+++ b/.github/buildomat/jobs/phd-run.sh
@@ -13,6 +13,12 @@
 #: job = "phd-build"
 #:
 
+# This job runs all the PHD test cases that don't involve upgrading from an
+# earlier version of Propolis.
+#
+# These tests should always pass even in the presence of breaking changes to the
+# Propolis API or live migration protocol.
+
 cp /input/phd-build/out/phd-run-with-args.sh /tmp/phd-run-with-args.sh
 chmod a+x /tmp/phd-run-with-args.sh
 exec /tmp/phd-run-with-args.sh --exclude-filter "phd_tests::migrate::from_base"
--- a/.github/buildomat/phd-run-with-args.sh
+++ b/.github/buildomat/phd-run-with-args.sh
@@ -47,15 +47,15 @@ ls $runner
 ls $artifacts
 ls $propolis
 
-args=()
-args+=($runner '--emit-bunyan' 'run')
-args+=('--propolis-server-cmd' $propolis)
-args+=('--base-propolis-branch' 'master')
-args+=('--crucible-downstairs-commit' 'auto')
-args+=('--artifact-toml-path' $artifacts)
-args+=('--tmp-directory' $tmpdir)
-args+=('--artifact-directory' $artifactdir)
-args+=($@)
+args=(
+    $runner '--emit-bunyan' 'run'
+    '--propolis-server-cmd' $propolis
+    '--crucible-downstairs-commit' 'auto'
+    '--artifact-toml-path' $artifacts
+    '--tmp-directory' $tmpdir
+    '--artifact-directory' $artifactdir
+    $@
+)
 
 # Disable errexit so that we still upload logs on failure
 set +e
--- a/.github/buildomat/phd-run-with-args.sh
+++ b/.github/buildomat/phd-run-with-args.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+# Unpacks and executes the PHD runner in the Buildomat environment, passing
+# through to the runner any arguments that were passed to this script.
+
+set -o errexit
+set -o pipefail
+set -o xtrace
+
+indir="/input"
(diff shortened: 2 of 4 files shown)


.github/buildomat/phd-run-with-args.sh
```diff
@@ -0,0 +1,89 @@
...
+
+args=()
+args+=($runner '--emit-bunyan' 'run')
+args+=('--propolis-server-cmd' $propolis)
+args+=('--base-propolis-branch' 'master')
+args+=('--crucible-downstairs-commit' 'auto')
+args+=('--artifact-toml-path' $artifacts)
+args+=('--tmp-directory' $tmpdir)
+args+=('--artifact-directory' $artifactdir)
+args+=($@)
```
hawkw, line 58: nitpicky, not a big deal: IIRC bash ignores whitespace in array literals, so i *believe* this can be a single assignment and still be multi-line:
```suggestion
args=(
    $runner '--emit-bunyan' 'run'
    '--propolis-server-cmd' $propolis
    '--base-propolis-branch' 'master'
    '--crucible-downstairs-commit' 'auto'
    '--artifact-toml-path' $artifacts
    '--tmp-directory' $tmpdir
    '--artifact-directory' $artifactdir
    $@
)
```

hawkw: (i may have made the tabs the wrong width in my suggestion...)

.github/buildomat/jobs/phd-run.sh
```diff
@@ -13,86 +13,6 @@
...
-if find /dev/vmm -mindepth 1 -maxdepth 1 | read ; then
-	echo "VMM handles leaked:"
-	find /dev/vmm -type c -exec basename {} \;
-	exitcode=2
-fi
-
-exit $exitcode
+cp /input/phd-build/out/phd-run-with-args.sh /tmp/phd-run-with-args.sh
+chmod a+x /tmp/phd-run-with-args.sh
+exec /tmp/phd-run-with-args.sh --exclude-filter "phd_tests::migrate::from_base"
```
hawkw, line 18: FWIW, if the `--base-propolis-branch`, `--base-propolis-commit`, and `--base-propolis-cmd` flags are all not provided, `phd-runner` will automagically disable all tests that require a base Propolis (even if they're not in that module). 

We may want to change this so that the `--base-propolis-branch master` argument is only passed by the migration from base tests run, rather than passing for both runs it here and then filtering out those tests. That way, we can avoid downloading the base Propolis binary in the non-migration-from-base test run, which might make the tests a *little* faster, and means that they won't spuriously fail if the Buildomat job hasn't finished running for the latest commit to `master`.

We could still keep the `--exclude-filter` as well, so that we don't even try to run those tests, rather than starting them and then realizing they've been disabled.

.github/buildomat/phd-run-with-args.sh
```diff
@@ -0,0 +1,89 @@
...
+propolis="$phddir/propolis-server"
+
+ls $runner
+ls $artifacts
+ls $propolis
+
+args=()
+args+=($runner '--emit-bunyan' 'run')
+args+=('--propolis-server-cmd' $propolis)
+args+=('--base-propolis-branch' 'master')
```
hawkw, line 53: As I mentioned above, I think it might make more sense to have only the migration-from-base job pass this argument, so that we don't have to download a base Propolis artifact when we're not running the tests that need it.

hawkw (commented):
Thanks for taking the time to do this! I left a few small suggestions.

hawkw (approved):
Looks good to me, feel free to merge as soon as we know it works!

It *might* be good to include either a comment in the scripts or in a README or something explaining the motivation for having separate jobs, so that a future reader knows why we split them? But, it's not a big deal.

gjcolombo:
Thanks for the review, @hawkw! 

> It might be good to include either a comment in the scripts or in a README or something explaining the motivation for having separate jobs, so that a future reader knows why we split them? But, it's not a big deal.

Can't hurt. I'll amend the scripts to include this before merging.

Issue #301, opened 2023-01-17
In-memory disk backends need better migration support


gjcolombo:
Migrating a disk with an in-memory backend replaces the disk's data with whatever data is specified in the migration target's backend spec, which might not be the same as the data on the source. Instead, the backend should migrate the source's data to the target. (We may also want some provision or other in the sync phase to make sure either that the destination has no data or, for read-only disks, that the data in the destination backend matches the data in the source backend.)

Our only use of these disks right now is read-only cloud-init disks, so for MVP it suffices just to check that the backends present the same data (and could even outright reject migration of instances with writable in-memory disks). This lets us defer figuring out what backend configuration a migration target should have (for compatibility purposes) if the source is just going to overwrite all that data anyway.

PR #740, opened 2024-08-07
PHD: add support for in-memory disk backends

gjcolombo:
Add an affordance for creating in-memory disk backends in PHD and populating them with data in a FAT filesystem (similar to what propolis-standalone and omicron do to create cloud-init data volumes). Add a small test that such volumes can be mounted and read as expected.

A future PR will use this for integration testing of #301, but this stands alone so thought I would put it up as its own PR.

### DIFF
--- a/phd-tests/framework/src/disk/fat.rs
+++ b/phd-tests/framework/src/disk/fat.rs
@@ -118,9 +118,10 @@ struct File {
 #[derive(Debug, Error)]
 pub enum Error {
     #[error(
-        "insufficient space for new file: {0} sectors required, {1} available"
+        "insufficient space for new file: {required} sectors required, \
+        {available} available"
     )]
-    NoSpace(usize, usize),
+    NoSpace { required: usize, available: usize },
 }
 
 /// Builds a collection of files that can be extruded as an array of bytes
@@ -150,7 +151,10 @@ impl FatFilesystem {
         let bytes = contents.as_bytes();
         let sectors_needed = Sectors::needed_for_bytes(bytes.len());
         if sectors_needed > self.sectors_remaining {
-            Err(Error::NoSpace(sectors_needed.0, self.sectors_remaining.0))
+            Err(Error::NoSpace {
+                required: sectors_needed.0,
+                available: self.sectors_remaining.0,
+            })
         } else {
             self.files
                 .push(File { path: path.to_owned(), contents: bytes.to_vec() });
--- a/phd-tests/framework/src/disk/fat.rs
+++ b/phd-tests/framework/src/disk/fat.rs
@@ -9,6 +9,7 @@ use std::io::{Cursor, Write};
 
 use anyhow::Context;
 use fatfs::{FileSystem, FormatVolumeOptions, FsOptions};
+use newtype_derive::*;
 use thiserror::Error;
 
 /// The maximum size of disk this module can create. This is fixed to put an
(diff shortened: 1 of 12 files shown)


phd-tests/framework/src/disk/fat.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub fn add_file_from_str(
+        &mut self,
+        path: &str,
+        contents: &str,
+    ) -> Result<(), Error> {
+        let bytes = contents.as_bytes();
+        let sectors_needed = Sectors::needed_for_bytes(bytes.len());
+        let sectors_available = usable_sectors().0 - self.sectors_used.0;
+        if sectors_available < sectors_needed.0 {
+            Err(Error::NoSpace(sectors_needed.0, sectors_available))
```
gjcolombo, line 145: This module could use some unit tests for cases like this.

phd-tests/framework/src/disk/fat.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// N.B. `path` is interpreted relative to the root of the file system and
+    ///      should not have a leading '/'.
+    pub fn add_file_from_str(
+        &mut self,
+        path: &str,
+        contents: &str,
+    ) -> Result<(), Error> {
+        let bytes = contents.as_bytes();
+        let sectors_needed = Sectors::needed_for_bytes(bytes.len());
+        let sectors_available = usable_sectors().0 - self.sectors_used.0;
```
hawkw, line 143: is it possible for `self.sectors_used` to have exceeded `usable_sectors()` here? I couldn't find the check for that...

phd-tests/framework/src/disk/mod.rs
```diff
@@ -25,20 +26,30 @@ use crate::{
...
 /// Errors that can arise while working with disks.
 #[derive(Debug, Error)]
 pub enum DiskError {
     #[error("Disk factory has no Crucible downstairs artifact")]
     NoCrucibleDownstairs,
 
+    #[error(
+        "can't create a disk of the requested type with the requested source"
+    )]
+    SourceNotSupported,
```
hawkw, line 42: nitpicky, not super important: it might be nice if this error included the requested type and the requested source?

phd-tests/framework/src/disk/mod.rs
```diff
@@ -212,11 +254,29 @@ impl DiskFactory {
...
+    pub(crate) async fn create_in_memory_disk<'d>(
+        &self,
+        name: String,
+        source: &DiskSource<'d>,
+        readonly: bool,
+    ) -> Result<Arc<InMemoryDisk>, DiskError> {
+        let contents = match source {
+            DiskSource::Artifact(name) => {
+                let (path, _) = self.get_guest_artifact_info(name).await?;
+                std::fs::read(path)?
```
hawkw, line 274: nitpicky: perhaps worthwhile to include the path in the error here?

phd-tests/framework/src/disk/fat.rs
gjcolombo: It's not supposed to be (any addition that would cause that post-condition should return `NoSpace` instead), but I agree that this wasn't very obvious, so I flipped the polarity of the tracker in bccf54b, which I think improves things.

phd-tests/framework/src/disk/fat.rs
gjcolombo: Fixed in 113adc1.

phd-tests/framework/src/disk/mod.rs
gjcolombo: Done in 85c0b95.

phd-tests/framework/src/disk/mod.rs
gjcolombo: Done in 85c0b95.

hawkw (approved):
some last obnoxious nitpicks, but no blockers from  me!

phd-tests/framework/src/disk/fat.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /// Adds a file with the supplied `contents` that will appear in the
+    /// resulting file system in the supplied `path`.
+    ///
+    /// N.B. `path` is interpreted relative to the root of the file system and
+    ///      should not have a leading '/'.
+    pub fn add_file_from_str(
+        &mut self,
+        path: &str,
```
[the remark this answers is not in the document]
gjcolombo: The `fatfs` crate speaks VFAT, so it can handle long filenames containing UTF-16 characters. (There are still some restrictions--e.g. it will reject an attempt to create a file if it can't find an unused MS-DOS short name for it--but those errors will get bubbled up when a caller tries to create a disk that runs afoul of these rules.)

There is one class of error we can catch up front, though (and it's one that makes me disinclined to make this a `Path`): if the supplied path contains a separator, `fatfs` will (correctly) try to traverse the filesystem's current directory tree and put the file in the requested directory. That will never work in PHD because the `FatFilesystem` struct doesn't (currently) let users create directories. So I've tweaked this in 3e77600 to make it clearer that this needs to be a bare filename (and to bail on adding a file if the name contains a '/' character).

Issue #735, opened 2024-08-01
instance specs could be improved


gjcolombo:
# The problem(s)

The instance spec types in the `propolis_api_types` crate are an important conceptual part of the live migration protocol, but there are several things about the way I implemented them that leave things to be desired:

- **Well-formed specs are not necessarily valid**

It's easy to construct an instance spec that type-checks but can't be used to start an instance. For example, it's trivial to have a spec with a storage device with no valid backend.
- **Specs have a complex hierarchy of component types**

Instance specs are composed of device specs and backend specs, each of which are further divided into their own fields and sub-collections. This makes adding new components while respecting the spec versioning rules something of a hassle (and the extra structure doesn't even guarantee the specs are valid!).
- **Not all components have names**

Live migration assumes that all (migratory) VM components have names. This allows Propolis to send a source component's saved state to the correct counterpart on the target. While many spec components are stored in a `HashMap<String, Component>`, some are not, so Propolis has to generate names for them, which then have to be kept consistent between Propolis versions.

- **There are three different spec builders**

One lives in `propolis_api_types`. `propolis-server` then wraps that in its own builder, which can also convert config TOML entries and `InstanceEnsureRequest` components to spec components. Yet a third builder lives in `propolis-client`; it's basically the same as the builder in `propolis_api_types`, except that it operates on Progenitor-generated types instead of the raw types.

- **`propolis_api_types` contains a lot more than just types**

For a crate that's ostensibly just a bag of types, this crate has a *lot* of code, most of which is only ever used by Propolis servers. Other Propolis components that want Just The Types, Please shouldn't need to pull in all the server-only methods.

- **Some component types are conditionally compiled**

The SoftNPU component definitions in `propolis_api_types` are behind Falcon feature guards. This forces us to maintain two separate OpenAPI documents, one with Falcon enabled and one without, which can lead to annoying rake-steps in CI if one forgets to check the build for one of the two variants. (In fact, as I write this, PHD builds are broken when Falcon is enabled because of a missing `patch` directive in the Falcon client generator...)

# Proposed solutions

I propose the following changes to fix these problems:

- **Create an internal `Spec` type for propolis-server**

This type is the VM component manifest type that Propolis server uses internally (in particular during VM setup). This spec isn't `Serialize`, isn't part of the API, and isn't used in the live migration protocol. Because this spec can be structured however we like without concern for versioning it, it can (and will) be structured so that illegal states are unrepresentable.

[...]

PR #741, opened 2024-08-12
instance spec refactor: move some impls out of api_types

gjcolombo:
(3/? of the instance spec refactoring series; see #735.)

One of the goals of this series of PRs is to get utility functions and impls out of the propolis_api_types crate and into propolis-server. Start that process by moving the instance spec builder and the `pci_path` associated functions into propolis-server.

Also patch up a couple of error variants that weren't using debug format specifiers for strings.

Tests: cargo test, PHD.

### DIFF
--- a/bin/propolis-server/src/lib/spec/builder.rs
+++ b/bin/propolis-server/src/lib/spec/builder.rs
@@ -2,14 +2,31 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-//! A builder for V0 instance specs.
+//! A builder for instance specs.
 
 use std::collections::BTreeSet;
 
-use crate::instance_spec::{components, v0::*, PciPath};
+use propolis_api_types::instance_spec::{
+    components::{
+        board::Board,
+        devices::{PciPciBridge, QemuPvpanic, SerialPort, SerialPortNumber},
+    },
+    v0::{
+        DeviceSpecV0, InstanceSpecV0, NetworkBackendV0, NetworkDeviceV0,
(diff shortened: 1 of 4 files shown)


bin/propolis-server/src/lib/spec/builder.rs
```diff
@@ -22,35 +39,45 @@ pub enum SpecBuilderError {
     PciPathInUse(PciPath),
 
     #[error("Serial port {0:?} is already specified")]
-    SerialPortInUse(components::devices::SerialPortNumber),
+    SerialPortInUse(SerialPortNumber),
 
+    #[cfg(feature = "falcon")]
```
hawkw, line 44: Hmm, this forces anyone who wants to match on a `SpecBuilderError` to have a "falcon" feature of their own --- I wonder if it would be better to not make the error variant conditional on "falcon", for that reason?  This would be fine to do, since the error doesn't include any falcon-only type... It could just be documented that it's only returned when falcon is enabled, so that consumers can match without having their own falcon feature flag.

On the other hand, I don't think there are presently any consumers who match on this, so it probably doesn't really matter. But, having a structured error enum makes it _possible_ for other code to do that.

hawkw (approved):
looks good to me!

bin/propolis-server/src/lib/spec/builder.rs
hawkw: upon further consideration, i realize that this type is not actually a public API, even though it's `pub`...it's only exposed to the `propolis-server` binary, which has a "falcon" feature anyway. Never mind!

gjcolombo: I demoted it to `pub(crate)` anyway in 6b5ad3e to make the intent a little clearer here.

PR #743, opened 2024-08-19
instance specs: move InstanceEnsureRequest processing into a module

gjcolombo:
(4/X in the PR series for #735.)

Move more code around:

- Move `InstanceEnsureRequest` parsing to its own module.
- Move the rest of the config TOML parsing logic (specifically surrounding device names) into the `config_toml` module.
- Shift some types around to allow parsed storage/network device descriptions to be shared between these modules.

There are still relatively few functional changes here, but getting all these modules onto the common "parsed device" types lays most of the remaining foundation for having an "internal" spec type that can take the place of the server's use of `InstanceSpecV0`.

Tests: cargo test, PHD (still no breaking changes to the API... yet!).


hawkw (approved):
No concerns from me --- I assume you intend to address the things you've already left comments on?

gjcolombo:
Yup! Just need to get my nose out of the RFDs I'm editing :)

PR #744, opened 2024-08-20
instance specs: remove extra layer of builder

gjcolombo:
(5/X in the instance spec rework; see #735.)

Now that the instance spec builder from `propolis_api_types` has moved into `propolis-server`, the server crate actually has two layers of builder: the "inner" `SpecBuilder` originally lived in the API types crate, and the "outer" `ServerSpecBuilder` was meant to transform server-only sources of spec elements (config TOML elements and API requests) into components to pass to the inner builder. Now that everything is in one crate, this layering is redundant, so collapse everything into a single `SpecBuilder` type and make the server use that.

Once again, this is mostly a matter of shuffling existing code around. The only substantial not-just-a-copy-and-paste change is that the server now has to enable a pvpanic device explicitly instead of it being included for free with a new spec builder. (The next PR in this series will do more than simply rearrange deck chairs!)

Tests: cargo test, PHD, booted a propolis-server ad hoc and made sure all the expected devices were there.

### DIFF
--- a/bin/propolis-server/src/lib/server.rs
+++ b/bin/propolis-server/src/lib/server.rs
@@ -35,7 +35,6 @@ use propolis_api_types::instance_spec::{self, VersionedInstanceSpec};
 pub use propolis_server_config::Config as VmTomlConfig;
 use rfb::tungstenite::BinaryWs;
 use slog::{error, warn, Logger};
-use thiserror::Error;
 use tokio::sync::MutexGuard;
 use tokio_tungstenite::tungstenite::protocol::{Role, WebSocketConfig};
 use tokio_tungstenite::WebSocketStream;
@@ -103,18 +102,12 @@ impl DropshotEndpointContext {
     }
 }
 
-#[derive(Debug, Error)]
-enum SpecCreationError {
-    #[error(transparent)]
-    SpecBuilderError(#[from] SpecBuilderError),
-}
-
 /// Creates an instance spec from an ensure request. (Both types are foreign to
 /// this crate, so implementing TryFrom for them is not allowed.)
 fn instance_spec_from_request(
     request: &api::InstanceEnsureRequest,
     toml_config: &VmTomlConfig,
-) -> Result<VersionedInstanceSpec, SpecCreationError> {
+) -> Result<VersionedInstanceSpec, SpecBuilderError> {
     let mut spec_builder = SpecBuilder::new(&request.properties);
 
     spec_builder.add_devices_from_config(toml_config)?;
@@ -288,7 +281,7 @@ async fn instance_ensure(
                 HttpError::for_bad_request(
                     None,
                     format!(
-                        "failed to generate instance spec from request: {}",
+                        "failed to generate instance spec from request: {:#?}",
                         e
                     ),
                 )
--- a/bin/propolis-server/src/lib/server.rs
+++ b/bin/propolis-server/src/lib/server.rs
@@ -29,6 +29,7 @@ use internal_dns::ServiceName;
 pub use nexus_client::Client as NexusClient;
 use oximeter::types::ProducerRegistry;
 use propolis_api_types as api;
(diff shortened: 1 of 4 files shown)


bin/propolis-server/src/lib/server.rs
```diff
@@ -105,7 +106,7 @@ impl DropshotEndpointContext {
 #[derive(Debug, Error)]
 enum SpecCreationError {
     #[error(transparent)]
-    SpecBuilderError(#[from] ServerSpecBuilderError),
+    SpecBuilderError(#[from] SpecBuilderError),
 }
```
hawkw, line 110: uh...do we actually need this enum?

bin/propolis-server/src/lib/spec/builder.rs
```diff
@@ -25,12 +30,27 @@ use propolis_api_types::instance_spec::{
...
+use super::ParsedSoftNpu;
 
 /// Errors that can arise while building an instance spec from component parts.
 #[allow(clippy::enum_variant_names)]
 #[derive(Debug, Error)]
 pub(crate) enum SpecBuilderError {
+    #[error("error parsing config TOML")]
+    ConfigToml(#[from] ConfigTomlError),
+
+    #[error("error parsing device in ensure request")]
```
hawkw, line 51: should these include the underlying error?
```suggestion
    #[error("error parsing config TOML: {0}")]
    ConfigToml(#[from] ConfigTomlError),

    #[error("error parsing device in ensure request: {0}")]
```

or do they always get formatted using the verbose format that will display the whole source chain?

bin/propolis-server/src/lib/spec/config_toml.rs
```diff
@@ -94,7 +89,24 @@ impl TryFrom<&config::Config> for ParsedConfig {
...
+                .map_or_else(
+                    || Ok(false),
+                    |v| {
+                        v.as_bool().ok_or_else(|| {
+                            ConfigTomlError::EnablePcieParseFailed(
+                                v.to_string(),
+                            )
+                        })
+                    },
+                )?,
```
hawkw, line 106: _whew_

hawkw: (personally, i find `map_or_else` _way_ harder to interpret than `.map(...).or_else(...)`, but...i know Clippy will probably yell at you for that...)

hawkw (approved):
Overall, I have no complaints; this looks great --- the number of different builder types was something that always bothered me a bit. I left a couple small suggestions.

bin/propolis-server/src/lib/server.rs
gjcolombo: We do not!

bin/propolis-server/src/lib/spec/builder.rs
gjcolombo: There's only one caller who actually consumes this error, so I switched it to use the `#?` specifier.

bin/propolis-server/src/lib/spec/config_toml.rs
gjcolombo: I played this a little differently (do the `ok_or_else` inside the first `map` and then use `transpose` to extract that result) but have thus removed the `map_or_else` and trimmed the indentation here.

bin/propolis-server/src/lib/spec/config_toml.rs
```diff
@@ -94,16 +94,13 @@ impl TryFrom<&config::Config> for ParsedConfig {
...
-                        })
-                    },
-                )?,
+                .map(|v| {
+                    v.as_bool().ok_or_else(|| {
+                        ConfigTomlError::EnablePcieParseFailed(v.to_string())
+                    })
+                })
+                .transpose()?
+                .unwrap_or(false),
```
hawkw, line 103: this is way less unpleasant IMO, thanks!

hawkw (approved):
let's go!

PR #745, opened 2024-08-21
update to hyper v1

ahl:


ahl:
@pfmooney I'm going to update this with the progenitor and dropshot dependencies that I'm working on landing, but I was hoping you could get eyes on this

bin/propolis-server/src/lib/serial/mod.rs
```diff
@@ -14,10 +14,10 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use crate::serial::history_buffer::{HistoryBuffer, SerialHistoryOffset};
+use dropshot::WebsocketConnectionRaw;
```
pfmooney, line 17: Why the `use` here, when the later references are made to `dropshot::WebsocketConnectionRaw`, rather than the bare name?

pfmooney (commented):
One small style nit

bin/propolis-server/src/lib/serial/mod.rs
ahl: um... bad programming? I'll fix it

PR #746, opened 2024-08-22
Publish richer virtual disk statistics to oximeter

bnaecker:
- Use the `virtual_disk:*` timeseries defined in Omicron to publish statistics about the block operations performed on virtual disks.
- Add concept of a completion callback to the existing block tracking object, and expose a way to set the callback on a `block::Device`.
- Create object for managing all the statistics for one virtual disk backed by Crucible. This is shared between the existing oximeter producer, and the block devices, via the above completion callback mechanism. As I/Os complete, the stats are updated, and oximeter collects them periodically.
- Straighten out a bunch of the initialiazation logic for the pre-existing metrics, producer server, and registry. This grew pretty organically, and was getting unwieldy and confusing.
- Use `virtual_machine:reset` definition from TOML
- Move kstat sampler creation to machine initialization process, rather than buried in the producer server startup. This lets us use the sampler in-line when initializing the vCPUs, where it should be much easier to re-use when we want to add more kstat-based metrics, such as Viona link stats.
- Simplify and cleanup the `ServerStats` types, and remove the kstat sampler from it, which never should have been there at all.
- Bump kstat-rs dep, which includes the new version that is cross-platform. This gets rid of a bunch of cfg-soup in the stats modules.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,18 +71,18 @@ p9ds = { git = "https://github.com/oxidecomputer/p9fs" }
 softnpu = { git = "https://github.com/oxidecomputer/softnpu" }
 
 # Omicron-related
-internal-dns = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
-nexus-client = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
-omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
+internal-dns = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+nexus-client = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
 omicron-zone-package = "0.9.0"
-oximeter-instruments = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition", default-features = false, features = ["kstat"] }
-oximeter-producer = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
-oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
-sled-agent-client = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
+oximeter-instruments = { git = "https://github.com/oxidecomputer/omicron", branch = "main", default-features = false, features = ["kstat"] }
+oximeter-producer = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+sled-agent-client = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
 
 # Crucible
-crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "db11baa532bf6fad73f22db0d6f2c914ab67b2fa" }
-crucible-client-types = { git = "https://github.com/oxidecomputer/crucible", rev = "db11baa532bf6fad73f22db0d6f2c914ab67b2fa" }
+crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "d037eeb9a56a8a62ff17266f340c011224d15146" }
+crucible-client-types = { git = "https://github.com/oxidecomputer/crucible", rev = "d037eeb9a56a8a62ff17266f340c011224d15146" }
 
 # External dependencies
 anyhow = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,18 +71,20 @@ p9ds = { git = "https://github.com/oxidecomputer/p9fs" }
 softnpu = { git = "https://github.com/oxidecomputer/softnpu" }
 
 # Omicron-related
-internal-dns = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-nexus-client = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+internal-dns = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
+nexus-client = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
+omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
 omicron-zone-package = "0.9.0"
-oximeter-instruments = { git = "https://github.com/oxidecomputer/omicron", branch = "main", default-features = false, features = ["kstat"] }
-oximeter-producer = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-sled-agent-client = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+oximeter-instruments = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition", default-features = false, features = ["kstat"] }
+oximeter-producer = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
+oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
+sled-agent-client = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
 
 # Crucible
-crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "e58ca3693cb9ce0438947beba10e97ee38a0966b" }
-crucible-client-types = { git = "https://github.com/oxidecomputer/crucible", rev = "e58ca3693cb9ce0438947beba10e97ee38a0966b" }
+#crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "e58ca3693cb9ce0438947beba10e97ee38a0966b" }
+#crucible-client-types = { git = "https://github.com/oxidecomputer/crucible", rev = "e58ca3693cb9ce0438947beba10e97ee38a0966b" }
+crucible = { path = "/home/bnaecker/crucible/upstairs" }
+crucible-client-types = { path = "/home/bnaecker/crucible/crucible-client-types" }
 
 # External dependencies
 anyhow = "1.0"
@@ -120,7 +122,7 @@ http = "0.2.9"
 hyper = "0.14"
 indicatif = "0.17.3"
 inventory = "0.3.0"
-kstat-rs = "0.2.3"
+kstat-rs = "0.2.4"
 lazy_static = "1.4"
 libc = "0.2"
 mockall = "0.12"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,10 +81,8 @@ oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "new-vir
 sled-agent-client = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
 
 # Crucible
-#crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "e58ca3693cb9ce0438947beba10e97ee38a0966b" }
-#crucible-client-types = { git = "https://github.com/oxidecomputer/crucible", rev = "e58ca3693cb9ce0438947beba10e97ee38a0966b" }
-crucible = { path = "/home/bnaecker/crucible/upstairs" }
-crucible-client-types = { path = "/home/bnaecker/crucible/crucible-client-types" }
+crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "db11baa532bf6fad73f22db0d6f2c914ab67b2fa" }
+crucible-client-types = { git = "https://github.com/oxidecomputer/crucible", rev = "db11baa532bf6fad73f22db0d6f2c914ab67b2fa" }
 
 # External dependencies
 anyhow = "1.0"
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -11,13 +11,16 @@ use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::serial::Serial;
+use crate::stats::virtual_disk::VirtualDiskProducer;
 use crate::stats::virtual_machine::VirtualMachine;
+use crate::stats::{create_kstat_sampler, track_vcpu_kstats};
 use crate::vm::{BlockBackendMap, CrucibleBackendMap, DeviceMap};
 use anyhow::{Context, Result};
 use crucible_client_types::VolumeConstructionRequest;
 pub use nexus_client::Client as NexusClient;
 use oximeter::types::ProducerRegistry;
 
+use oximeter_instruments::kstat::KstatSampler;
 use propolis::block;
 use propolis::chardev::{self, BlockingSource, Source};
 use propolis::common::{Lifecycle, GB, MB, PAGE_SIZE};
@@ -542,14 +545,15 @@ impl<'a> MachineInitializer<'a> {
                 .await?;
 
             self.block_backends.insert(backend_name.clone(), backend.clone());
-            match device_interface {
+            let block_dev: Arc<dyn block::Device> = match device_interface {
                 DeviceInterface::Virtio => {
                     let vioblk = virtio::PciVirtioBlock::new(0x100);
 
                     self.devices
                         .insert(format!("pci-virtio-{}", bdf), vioblk.clone());
                     block::attach(vioblk.clone(), backend).unwrap();
-                    chipset.pci_attach(bdf, vioblk);
+                    chipset.pci_attach(bdf, vioblk.clone());
+                    vioblk
                 }
                 DeviceInterface::Nvme => {
                     // Limit data transfers to 1MiB (2^8 * 4k) in size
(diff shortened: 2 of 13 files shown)


bnaecker:
This still needs some testing, but it's a larger chunk than I intended so I'd like to get early eyes on it.  I took this opportunity to rework a lot of the metrics initialization, which had become pretty Byzantine.

The main point here is to publish richer virtual disk metrics than what we currently get from Crucible. This adds the same counters we had before, and also failure counters and histograms tracking I/O latency and I/O sizes. These use the same block device tracking machinery that was added to handle USDT probes, adding a completion callback to that struct. We now set the completion callback to update the statistics for each virtual disk. Those stats are shared with the oximeter producer registry, so they're published along with everything else.

The extras that came along for the ride here are a mishmash of cleanup tasks, mostly listed in the commit message. This PR also relies on a forthcoming PR in Omicron that adds the schema for the new statistics. I'll link that once I post it, and follow up with testing notes as well.

bin/propolis-server/src/lib/stats/virtual_disk.rs
```diff
@@ -0,0 +1,381 @@
...
+        disk_id: Uuid,
+        metadata: &InstanceMetadata,
+    ) -> Self {
+        let disk = VirtualDisk {
+            attached_instance_id: instance_id,
+            block_size,
+            disk_id,
+            project_id: metadata.project_id,
+            silo_id: metadata.silo_id,
+            sled_id: metadata.sled_id,
```
leftwo, line 213: The `sled_id` here is going to be the sled the disk was attached to when it was in use, right?
We don't want to confuse that with the location(s) of all the downstairs that could be what underpins the virtual disk that Propolis sees.

lib/propolis/src/block/crucible.rs
```diff
@@ -154,6 +154,16 @@ impl CrucibleBackend {
         }))
     }
 
+    /// Return the block size of this Crucible backend, if it can be determined.
+    pub async fn block_size(&self) -> Option<u32> {
+        self.state
```
leftwo, line 159: I'm not sure what happens if we call this before the upstairs is active.
When you run this, do you get a block size back from the volume?
If it works, then you are calling it after activation or I don't know what I'm talking about.  Possibly both.

bin/propolis-server/src/lib/stats/virtual_disk.rs
bnaecker: That's right, it's the ID of the sled hosting the instance. I think the description in the `virtual-disk.toml` schema should make this clear, but let me know if that's lacking in some way.

lib/propolis/src/block/crucible.rs
bnaecker: Welp, I'm about to test it, so I'll let you know! We also get the block size in the API request from the sled-agent, so it's possible we can pass it down from there to where we need it in another way, rather than asking Crucible.

bnaecker:
Talked offline with @ahl. He questions the value of putting the sled identifiers on this kind of timeseries at all, which I can understand. I might have been overzealous in my attempt to address https://github.com/oxidecomputer/omicron/issues/5267.

bnaecker:
Ok, I've done an initial round of testing. The Omicron side is still blocked on https://github.com/oxidecomputer/omicron/pull/6409 and this WIP branch: https://github.com/oxidecomputer/omicron/tree/new-virtual-disk-timeseries-definition. But incorporating both of those locally and creating a VM does result in the expected timeseries in ClickHouse.

Here are all of them:

```
0x〉\l
oximeter_collector:collections
virtual_disk:bytes_read
virtual_disk:bytes_written
virtual_disk:failed_flushes
virtual_disk:failed_reads
virtual_disk:failed_writes
virtual_disk:flushes
virtual_disk:io_latency
virtual_disk:io_size
virtual_disk:reads
virtual_disk:writes
virtual_machine:pv_panic_guest_handled
virtual_machine:pv_panic_host_handled
virtual_machine:reset
virtual_machine:vcpu_usage
```

Digging into individual things looks like I expect:

```
0x〉get virtual_disk:reads | last 1

virtual_disk:reads

 attached_instance_id: 6d39bc42-74d3-4f34-a035-6a068a4e4da3
 block_size: 4096
 disk_id: d82074ff-7098-483f-90f3-d7fca35baba6
 project_id: d50c3c4d-0008-45a3-a285-3b9c8ef84d38
 silo_id: 7fec5b38-3c43-4ed1-ab05-5152401c46a8
 sled_id: faf169b9-bf53-4a53-8f52-68c94dc377bd
 sled_model: fake-gimlet
 sled_revision: 1
 sled_serial: fake-serial
   [2024-08-22 21:44:10.723454479, 2024-08-22 21:54:40.861305986]: [109]
0x〉get virtual_disk:io_size | filter io_kind == "read" | last 1

virtual_disk:io_size

 attached_instance_id: 6d39bc42-74d3-4f34-a035-6a068a4e4da3
 block_size: 4096
 disk_id: d82074ff-7098-483f-90f3-d7fca35baba6
 io_kind: read
 project_id: d50c3c4d-0008-45a3-a285-3b9c8ef84d38
 silo_id: 7fec5b38-3c43-4ed1-ab05-5152401c46a8
 sled_id: faf169b9-bf53-4a53-8f52-68c94dc377bd
 sled_model: fake-gimlet
 sled_revision: 1
 sled_serial: fake-serial
   [2024-08-22 21:44:10.723879738, 2024-08-22 21:55:10.862958317]: [0: 105, 4096: 1, 8192: 3, 16384: 0, 32768: 0, 65536: 0, 131072: 0, 262144: 0, 524288: 0, 1048576: 0, 2097152: 0, 4194304: 0, 8388608: 0, 16777216: 0, 33554432: 0, 67108864: 0, 134217728: 0, 268435456: 0, 536870912: 0, 1073741824: 0, min: 512, max: 8192, mean: 1211.8899082568807, std_dev: 1407.1530457991503, p50: 748.4546500654893, p90: 2197.500566706399, p99: 6740.062811662923]
```

So that shows we had 109 total reads, and the I/O size histogram shows that those were broken out into 105 in the range [0, 4KiB), 1 in [4KiB, 8KiB), and 3 in [8KiB, 16KiB). So there are the same number of items in both, meaning we're consistent at least.

Cargo.toml
```diff
@@ -71,18 +71,18 @@ p9ds = { git = "https://github.com/oxidecomputer/p9fs" }
 softnpu = { git = "https://github.com/oxidecomputer/softnpu" }
 
 # Omicron-related
-internal-dns = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-nexus-client = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+internal-dns = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
```
bnaecker, line 74: Note to self: Point back to Omicron main when https://github.com/oxidecomputer/omicron/pull/6420 merges

Cargo.toml
```diff
@@ -71,18 +71,18 @@ p9ds = { git = "https://github.com/oxidecomputer/p9fs" }
...
-sled-agent-client = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+oximeter-instruments = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition", default-features = false, features = ["kstat"] }
+oximeter-producer = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
+oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
+sled-agent-client = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
 
 # Crucible
-crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "e58ca3693cb9ce0438947beba10e97ee38a0966b" }
-crucible-client-types = { git = "https://github.com/oxidecomputer/crucible", rev = "e58ca3693cb9ce0438947beba10e97ee38a0966b" }
+crucible = { git = "https://github.com/oxidecomputer/crucible", rev = "db11baa532bf6fad73f22db0d6f2c914ab67b2fa" }
```
bnaecker, line 84: Note to self: Point back to rev on Crucible main when https://github.com/oxidecomputer/crucible/pull/1429 merges

bnaecker:
There's one kind of structural question my testing revealed. These virtual disk stats are updated all the time, but only sampled when `oximeter` asks for it. That means we get samples every 30s or 10s or whatever we pick. We could decouple these, by sampling the stats locally more frequently, and publishing more than one sample's worth of data to `oximeter`. That does require a different architecture though, since we'd need a small polling loop pushing data to a queue somewhere.

I think it largely depends on how densely we want to sample the data. I'd defer to @leftwo and @ahl on that probably. I reduced the collection interval to 10s, which still feels borderline, but I admit that's a WAG.

bin/propolis-server/src/lib/vm/ensure.rs
```diff
@@ -180,6 +180,12 @@ impl<'a> VmEnsureNotStarted<'a> {
             state: MachineInitializerState::default(),
         };
 
+        // If we are publishing metrics to oximeter, then create a kstat-sampler
+        // now. This is used to track the vCPUs today, but will soon be used to
+        // track instance datalink stats as well. It can be provided to
+        // `initialize_network_devices()` at that time.
+        let kstat_sampler = init.create_kstat_sampler();
```
gjcolombo, line 187: Is it worthwhile to create the sampler if the `producer_registry` in the `options` is `None`? (In particular, does the `kstat_sampler`'s memory usage grow over time if it never gets collected from? We'll have a producer in production Propolises, of course, but it's sometimes useful to be able to run long-running tests with a freestanding propolis-server binary.)

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -564,18 +568,59 @@ impl<'a> MachineInitializer<'a> {
...
+                        &self.properties.metadata,
+                    );
+                    if let Err(e) = registry.register_producer(stats.clone()) {
+                        slog::error!(
+                            self.log,
+                            "Could not register virtual disk producer, \
+                            metrics will not be produced";
+                            "disk_id" => %disk_id,
+                            "error" => ?e,
+                        );
```
leftwo, line 613: If we can't register stats, will this bail out of the loop early?

It seems like we are in the loop:
`for (name, device_spec) in &self.spec.devices.storage_devices {`

And, if we `retrun Ok`, will the caller even know that not all of their devices were not handled.

The same is true about if we can't get a block size above, we could bail early and not attach other disks.

bin/propolis-server/src/lib/stats.rs
```diff
@@ -46,24 +51,12 @@ const VCPU_KSTAT_INTERVAL: std::time::Duration =
...
-    pub count: Cumulative<u64>,
-}
-
-/// The full set of server-level metrics, collated by
-/// [`ServerStatsOuter::produce`] into the types needed to relay these
-/// statistics to Oximeter.
+/// Shared type for tracking metrics about the Propolis API server itself.
 #[derive(Clone, Debug)]
-struct ServerStats {
+struct ServerStatsInner {
```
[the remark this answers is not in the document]
bnaecker: I was expecting it to go away as well. But the `Reset` metric _is_ currently published to `oximeter`, so I think some form of this needs to stick around. At that point, it's similar to the current block stats: there's some state shared between the Propolis server, which updates the reset count, and the `oximeter_producer::Server`, which publishes the value when `oximeter` polls for it.

It's possible it could be shared differently, so I'll take another look at it.

leftwo (commented):
Does this leave in place the statistics that the crucible upstairs logs as well?

bin/propolis-server/src/lib/stats.rs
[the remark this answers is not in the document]
gjcolombo: Thanks! Don't spend too much time on it; I was mostly just curious about how this had shaken out in practice.

bin/propolis-server/src/lib/vm/ensure.rs
bnaecker: The `create_kstat_sampler()` method returns an `Option<KstatSampler>`. That's `None` if either the `producer_registry` is `None`, or if we just fail to create the thing. I'll rename this variable to make that more clear!

That obviates your second question, about memory consumption, but the answer is "no". The `KstatSampler` requires one to choose a limited buffer size, used in a ringbufffer to maintain uncollected samples. The memory consumption is bounded, even if it's never polled for data.

bin/propolis-server/src/lib/initializer.rs
bnaecker: Nice catch! This is definitely wrong, we should be `continue`ing here instead. I'll update that, thanks.

bin/propolis-server/src/lib/vm/ensure.rs
gjcolombo: That makes sense--thanks for the explanation!

bnaecker:
> Does this leave in place the statistics that the crucible upstairs logs as well?

Yeah, the existing `crucible_upstairs:*` stats are still there. Those are currently used for the web console's view of instance stats. I think we should switch that over to these data, and then we can separately decide to remove the original data if we want.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -11,13 +11,16 @@ use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::serial::Serial;
+use crate::stats::virtual_disk::VirtualDiskProducer;
 use crate::stats::virtual_machine::VirtualMachine;
+use crate::stats::{create_kstat_sampler, track_vcpu_kstats};
```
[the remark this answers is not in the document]
bnaecker: Done in 15327f2

bin/propolis-server/src/lib/initializer.rs
bnaecker: Switched to a `continue` in 15327f2

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -1041,13 +1086,43 @@ impl<'a> MachineInitializer<'a> {
...
+        if let Some(sampler) = kstat_sampler.as_ref() {
+            track_vcpu_kstats(&self.log, sampler, self.properties).await;
+        }
         Ok(())
     }
+
+    /// Create an object used to sample kstats.
+    ///
+    /// This object is currently used to generate vCPU metrics, though guest NIC
+    /// metrics will be included soon.
```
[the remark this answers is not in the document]
bnaecker: Done in 15327f2cd7d9401c4de97c8dda4c4afff37f04d1

bin/propolis-server/src/lib/stats.rs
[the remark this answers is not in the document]
bnaecker: I think this is in a reasonable place as-is. Cleaning up the naming and the initialization process makes it much clearer that the `ServerStats` type is used to store metrics about the API server or requests it services. We could add HTTP latency tracking here for example, if we cared to know that. Thanks for flagging it!

gjcolombo (approved):

lib/propolis/src/block/mod.rs
```diff
@@ -224,6 +225,13 @@ pub trait Device: Send + Sync + 'static {
     /// Complete processing of result
     fn complete(&self, res: Result, id: ReqId);
 
+    /// Attach a callback to be run on completion of I/Os.
+    ///
+    /// Returns whether there was a previously-registered callback.
+    fn on_completion(&self, _cb: Box<dyn CompletionCallback>) -> bool {
+        false
```
pfmooney, line 232: I'm not wild about there being a default implementation here.  It would be misleading for a block device to accept setting a completion callback, only to never call it for any of its completed requests.

lib/propolis/src/block/tracking.rs
```diff
@@ -46,18 +63,44 @@ struct TrackingEntry<T> {
...
+        dev: Weak<dyn Device>,
+        cb: impl CompletionCallback,
+    ) -> Self {
+        Self::new_raw(dev, Some(Box::new(cb)))
+    }
+
+    fn new_raw(
+        dev: Weak<dyn Device>,
+        cb: Option<Box<dyn CompletionCallback>>,
+    ) -> Self {
```
pfmooney, line 80: Not required if `with_completion_callback` is eliminated.

lib/propolis/src/block/mod.rs
bnaecker: That's fair. How about returning something like an `anyhow::Result<bool>`? That would return whether the callback existed previously in the `Ok(_)` variant, or fail if the device chose not to support it.

pfmooney: > if the device chose not to support it.

All of the existing devices do support it, so I wouldn't worry about that sort of thing for now.  If things change in the future, we can deal with it then.

bnaecker: So to be concrete, you're only arguing that we remove the default impl? Otherwise everything else remains unchanged about this interface?

pfmooney: > you're only arguing that we remove the default impl

yeah, the rest is fine as-is

bnaecker: Removed the default impl in 8565b0c

lib/propolis/src/block/tracking.rs
bnaecker: Removed in 8565b0c

lib/propolis/src/block/tracking.rs
```diff
@@ -46,18 +63,44 @@ struct TrackingEntry<T> {
...
+        Self::new_raw(dev, None)
+    }
+
+    /// Create a new tracking object with a completion callback.
+    pub fn with_completion_callback(
+        dev: Weak<dyn Device>,
+        cb: impl CompletionCallback,
+    ) -> Self {
+        Self::new_raw(dev, Some(Box::new(cb)))
+    }
```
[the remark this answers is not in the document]
bnaecker: Removed in 8565b0c

pfmooney (approved):

bnaecker:
Just a note on the sampling interval. I talked in chat with @leftwo, and we decided that the current sampling rate of 10s is sufficient. In this PR, the sampling rate of these statistics is defined by the rate at which oximeter polls -- Propolis only generates sample on a request from oximeter. If we wanted to change this, we have two choices. We could lower that polling rate from oximeter, or we could decouple the rate at which we internally generate samples from the block-dev producer and the rate at which oximeter asks Propolis for samples. That's an architectural change, since we'd need a task that continually generates samples from the block-dev. It's also fairly easy to do, and not necessary right now. We can decide that later without any breaking changes downstream.

leftwo (approved):

Cargo.toml
bnaecker: Done in 3382496

Cargo.toml
bnaecker: Done in 3382496

Issue #742, opened 2024-08-15
Enable Guest/Instance network interface(s) metrics 


zeeshanlakhani:
- [x] Omicron PR with new `instance-network-interface.toml` schema and addition of interface_id/nic_id to `NetworkInterfaceRequest` => https://github.com/oxidecomputer/omicron/pull/6414
- [x] Have `viona` driver expose kstats directly (related: https://code.illumos.org/c/illumos-gate/+/3630, cc @pfmooney )
- [x] Plumb interface_id/nic_id through to VirtioNic Device => https://github.com/oxidecomputer/propolis/pull/747
- [x] Add Kstat sampler target for network interface `instance` stats (in addition to current sampler) => https://github.com/oxidecomputer/propolis/pull/747

PR #747, opened 2024-08-23
[network metrics] instance network interface metrics

zeeshanlakhani:
Overview
========

This closes the ongoing work related to [plumbing instance/guest metrics](https://github.com/oxidecomputer/propolis/issues/742) through to an oximeter producer in propolis.

Includes:

* An update to `NetworkInterfaceRequest` to include an interface_id, which is part of the request on the [sled instance in Omicron](https://github.com/oxidecomputer/omicron/pull/6414).
* Within initialization, with concrete types, we create interface_ids mapping nic UUIDs => `minor` number of the device instance of viona.
  - The device instance number matches the Kstat we'll look for.
  - We track this mapping via a type alias `InterfaceIdentifiers`, which is passed through from Machine initialization to VM objects (and VM objects shared), the latter of which we can use as part of registering the oximeter producer.
* A new stats module for collecting network_interface metrics.
  - In `to_samples` we generate multiple network interface targets based on vnics per instance, appropriately updating the `interface_id`, which is tracked in `interface_ids`.
* Move kstat_types out of virtual machine-specific stats and use its mocking for other stats tests

Dependencies
------------

- [x] https://github.com/oxidecomputer/omicron/pull/6414
- [x] https://code.illumos.org/c/illumos-gate/+/3630
- [ ] **Note**: Will need to update the commit hash in https://github.com/oxidecomputer/falcon once all this is finished.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,7 +157,7 @@ dependencies = [
 [[package]]
 name = "api_identity"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "omicron-workspace-hack",
  "proc-macro2",
@@ -754,7 +754,7 @@ checksum = "248e3bacc7dc6baa3b21e405ee045c3047101a49145e7e9eca583ab4c2ca5345"
 [[package]]
 name = "crucible"
 version = "0.0.1"
-source = "git+https://github.com/oxidecomputer/crucible?rev=d037eeb9a56a8a62ff17266f340c011224d15146#d037eeb9a56a8a62ff17266f340c011224d15146"
+source = "git+https://github.com/oxidecomputer/crucible?rev=2aee0bc0fea61f26474b04c88d2c150fa0d35f56#2aee0bc0fea61f26474b04c88d2c150fa0d35f56"
 dependencies = [
  "aes-gcm-siv",
  "anyhow",
@@ -807,7 +807,7 @@ dependencies = [
 [[package]]
 name = "crucible-client-types"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/crucible?rev=d037eeb9a56a8a62ff17266f340c011224d15146#d037eeb9a56a8a62ff17266f340c011224d15146"
+source = "git+https://github.com/oxidecomputer/crucible?rev=2aee0bc0fea61f26474b04c88d2c150fa0d35f56#2aee0bc0fea61f26474b04c88d2c150fa0d35f56"
 dependencies = [
  "base64 0.22.1",
  "crucible-workspace-hack",
@@ -820,7 +820,7 @@ dependencies = [
 [[package]]
 name = "crucible-common"
 version = "0.0.1"
-source = "git+https://github.com/oxidecomputer/crucible?rev=d037eeb9a56a8a62ff17266f340c011224d15146#d037eeb9a56a8a62ff17266f340c011224d15146"
+source = "git+https://github.com/oxidecomputer/crucible?rev=2aee0bc0fea61f26474b04c88d2c150fa0d35f56#2aee0bc0fea61f26474b04c88d2c150fa0d35f56"
 dependencies = [
  "anyhow",
  "atty",
@@ -850,7 +850,7 @@ dependencies = [
 [[package]]
 name = "crucible-protocol"
 version = "0.0.0"
-source = "git+https://github.com/oxidecomputer/crucible?rev=d037eeb9a56a8a62ff17266f340c011224d15146#d037eeb9a56a8a62ff17266f340c011224d15146"
+source = "git+https://github.com/oxidecomputer/crucible?rev=2aee0bc0fea61f26474b04c88d2c150fa0d35f56#2aee0bc0fea61f26474b04c88d2c150fa0d35f56"
 dependencies = [
  "anyhow",
  "bincode",
@@ -869,7 +869,7 @@ dependencies = [
 [[package]]
 name = "crucible-smf"
 version = "0.0.0"
-source = "git+https://github.com/oxidecomputer/crucible?rev=e58ca3693cb9ce0438947beba10e97ee38a0966b#e58ca3693cb9ce0438947beba10e97ee38a0966b"
+source = "git+https://github.com/oxidecomputer/crucible?rev=88c45eebf34027c3bc80e60a5f527738bcfaf16a#88c45eebf34027c3bc80e60a5f527738bcfaf16a"
 dependencies = [
  "crucible-workspace-hack",
  "libc",
@@ -1151,7 +1151,7 @@ dependencies = [
 [[package]]
 name = "dns-service-client"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "anyhow",
  "chrono",
@@ -1648,7 +1648,7 @@ dependencies = [
 [[package]]
 name = "gateway-client"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "base64 0.22.1",
  "chrono",
@@ -2111,7 +2111,7 @@ source = "git+https://github.com/oxidecomputer/opte?rev=76878de67229ea113d70503c
 [[package]]
 name = "illumos-utils"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -2211,7 +2211,7 @@ dependencies = [
 [[package]]
 name = "internal-dns"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "anyhow",
  "chrono",
@@ -2700,7 +2700,7 @@ dependencies = [
 [[package]]
 name = "nexus-client"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "chrono",
  "futures",
@@ -2724,7 +2724,7 @@ dependencies = [
 [[package]]
 name = "nexus-sled-agent-shared"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "omicron-common",
  "omicron-passwords",
@@ -2742,7 +2742,7 @@ dependencies = [
 [[package]]
 name = "nexus-types"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "anyhow",
  "api_identity",
@@ -3028,7 +3028,7 @@ dependencies = [
 [[package]]
 name = "omicron-common"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "anyhow",
  "api_identity",
@@ -3070,7 +3070,7 @@ dependencies = [
 [[package]]
 name = "omicron-passwords"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "argon2",
  "omicron-workspace-hack",
@@ -3084,7 +3084,7 @@ dependencies = [
 [[package]]
 name = "omicron-uuid-kinds"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "newtype-uuid",
  "paste",
@@ -3284,7 +3284,7 @@ dependencies = [
 [[package]]
 name = "oximeter"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "anyhow",
  "chrono",
@@ -3303,7 +3303,7 @@ dependencies = [
 [[package]]
 name = "oximeter-instruments"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "cfg-if",
  "chrono",
@@ -3321,7 +3321,7 @@ dependencies = [
 [[package]]
 name = "oximeter-macro-impl"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "omicron-workspace-hack",
  "proc-macro2",
@@ -3332,7 +3332,7 @@ dependencies = [
 [[package]]
 name = "oximeter-producer"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "chrono",
  "dropshot",
@@ -3353,7 +3353,7 @@ dependencies = [
 [[package]]
 name = "oximeter-schema"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "anyhow",
  "chrono",
@@ -3374,7 +3374,7 @@ dependencies = [
 [[package]]
 name = "oximeter-timeseries-macro"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "omicron-workspace-hack",
  "oximeter-schema",
@@ -3387,7 +3387,7 @@ dependencies = [
 [[package]]
 name = "oximeter-types"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "bytes",
  "chrono",
@@ -3406,7 +3406,7 @@ dependencies = [
 [[package]]
 name = "oxlog"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "anyhow",
  "camino",
@@ -3431,7 +3431,7 @@ dependencies = [
 [[package]]
 name = "oxql-types"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "anyhow",
  "chrono",
@@ -5122,7 +5122,7 @@ dependencies = [
 [[package]]
 name = "sled-hardware-types"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#758818aea3f0d375b6bbf9bd87f713a5da09ffbd"
+source = "git+https://github.com/oxidecomputer/omicron?branch=zl/instance-network-interface#5bb97ae7a432a34ec426add21fb6d9eb54e3a44e"
 dependencies = [
  "illumos-utils",
  "omicron-common",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,7 +157,7 @@ dependencies = [
 [[package]]
 name = "api_identity"
 version = "0.1.0"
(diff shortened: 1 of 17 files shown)


zeeshanlakhani:
This is still awaiting the finalization of work in https://code.illumos.org/c/illumos-gate/+/3630, so it can be tested in practice.

Cargo.toml
```diff
@@ -164,3 +165,22 @@ tracing-subscriber = "0.3.14"
 usdt = { version = "0.5", default-features = false }
 uuid = "1.3.2"
 zerocopy = "0.7.34"
+
+
+#
+# It's common during development to use a local copy of various complex
```
zeeshanlakhani, line 171: keeping this around b/c it's useful for patching.

bin/propolis-server/src/lib/stats.rs
```diff
@@ -10,31 +10,38 @@ use oximeter::{
...
 
 // Interval on which we ask `oximeter` to poll us for metric data.
 const OXIMETER_STAT_INTERVAL: tokio::time::Duration =
     tokio::time::Duration::from_secs(30);
 
-// Interval on which we produce vCPU metrics.
+/// Interval on which we sample instance/guest network interface metrics.
+#[cfg(all(not(test), target_os = "illumos"))]
+const NETWORK_INTERFACE_SAMPLE_INTERVAL: std::time::Duration =
+    std::time::Duration::from_secs(10);
```
pfmooney, line 42: Why are the three different stat intervals different?

bin/propolis-server/src/lib/stats.rs
```diff
@@ -201,34 +208,175 @@ pub async fn register_server_metrics(
...
+    nics: InstanceNetworkInterfaces,
     virtual_machine: VirtualMachine,
 ) -> Option<KstatSampler> {
-    let kstat_limit =
-        usize::try_from(virtual_machine.n_vcpus() * KSTAT_LIMIT_PER_VCPU)
-            .unwrap();
+    let n_interfaces = nics.interface_ids.len() as u32;
+    let kstat_limit = usize::try_from(
+        (virtual_machine.n_vcpus() * KSTAT_LIMIT_PER_VCPU)
+            + (n_interfaces * KSTAT_LIMIT_PER_VCPU),
```
pfmooney, line 217: PER_CPU limit used for network interfaces?

bin/propolis-server/src/lib/stats.rs
```diff
@@ -201,34 +208,175 @@ pub async fn register_server_metrics(
...
+    )
+    .unwrap();
     match KstatSampler::with_sample_limit(log, kstat_limit) {
         Ok(sampler) => {
```
pfmooney, line 221: This might be a little less tangled if it went something along the lines of:
```rust
let sampler = KstatSampler::with_sample_limit(log, kstat_limit).map_err(|e| {
    slog::error!(
        log,
        "failed to create KstatSampler, \
        vCPU and link stats will be unavailable";
        "error" => ?e,
    );
    e
}).ok()?;
let vcpu_details =
// ...
```

bin/propolis-server/src/lib/vm/services.rs
```diff
@@ -51,16 +51,24 @@ impl VmServices {
         vm_properties: &InstanceProperties,
         ensure_options: &super::EnsureOptions,
     ) -> Self {
+        let vm_objects = vm_objects.lock_shared().await;
```
bnaecker, line 54: If we do land this prior to https://github.com/oxidecomputer/propolis/pull/746, could we copy the interface identifiers here and pass them as owned values? Then we don't need to hold a read lock for the entire time we're setting up the oximeter producer, creating the kstats, etc.

crates/propolis-api-types/Cargo.toml
```diff
@@ -9,6 +9,7 @@ doctest = false
 
 [dependencies]
 crucible-client-types.workspace = true
+illumos-utils.workspace = true
```
bnaecker, line 12: I don't see this used anywhere, is this a development change?

crates/propolis-api-types/src/lib.rs
```diff
@@ -4,10 +4,10 @@
 
 //! Definitions for types exposed by the propolis-server API
 
-use std::{fmt, net::SocketAddr};
-
+use illumos_utils::zone::PROPOLIS_ZONE_PREFIX;
```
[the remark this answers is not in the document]
bnaecker: Ah Patrick beat me to it :)

crates/propolis-api-types/src/lib.rs
```diff
@@ -159,12 +159,12 @@ pub struct InstanceSpecGetResponse {
 
 #[derive(Clone, Deserialize, Serialize, JsonSchema)]
 pub struct InstanceStateMonitorRequest {
-    pub gen: u64,
+    pub r#gen: u64,
```
bnaecker, line 162: Is this a 2024-edition-preparation thing?

crates/rgb-frame/src/lib.rs
```diff
@@ -154,22 +154,22 @@ impl Frame {
 #[repr(u32)]
 #[strum(serialize_all = "UPPERCASE")]
 pub enum FourCC {
-    /// x:R:G:B [31:0] little endian
+    /// x:R:G:B `\[` 31:0 `\]` little endian
```
bnaecker, line 157: What do these changes do?

crates/viona-api/src/lib.rs
```diff
@@ -65,6 +66,14 @@ impl VionaFd {
         Ok(vers as u32)
     }
 
+    /// Retrieve the minor number of the viona device instance.
```
bnaecker, line 69: Maybe a note about why we added this?

lib/propolis/src/cpuid.rs
```diff
@@ -116,7 +116,7 @@ impl From<Set> for Vec<bhyve_api::vcpu_cpuid_entry> {
     }
 }
 
-/// Convert a [vcpu_cpuid_entry](bhyve_api::vcpu_cpuid_entry) into an ([Ident],
+/// Convert a [vcpu_cpuid_entry] into an ([Ident],
```
bnaecker, line 119: Ok, there are enough of these changes that I'm curious. Did you do them manually, or did you run some `cargo doc` command that fixed them?

bnaecker (commented):
Looks pretty good overall, thanks for doing this! I've got some related changes in https://github.com/oxidecomputer/propolis/pull/746 that will conflict, as a head's up. I think merging those would make this change a good bit simpler, so that might be worthwhile. Thanks!

bin/propolis-cli/src/main.rs
```diff
@@ -33,6 +25,13 @@ use propolis_client::{
     },
     Client,
 };
+use slog::{o, Drain, Level, Logger};
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio_tungstenite::tungstenite::{
+    protocol::{frame::coding::CloseCode, CloseFrame},
+    Message,
+};
+use uuid::Uuid;
```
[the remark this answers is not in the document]
zeeshanlakhani: yeah. Seemed to be, but since I ended up not touching it, I'll revert.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -37,7 +33,14 @@ use propolis_api_types::instance_spec::{self, v0::InstanceSpecV0};
 use propolis_api_types::InstanceProperties;
 use slog::info;
 
-// Arbitrary ROM limit for now
+use crate::serial::Serial;
+use crate::stats::VirtualMachine;
+use crate::vm::BlockBackendMap;
+use crate::vm::CrucibleBackendMap;
+use crate::vm::DeviceMap;
+use crate::vm::InterfaceIdentifiers;
```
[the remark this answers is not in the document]
zeeshanlakhani: It's because some of the imports were split, and some were not, so I was trying to be consistent.

bin/propolis-server/src/lib/stats.rs
zeeshanlakhani: They are different. The interface one is similar to what we do with sled metrics vs what was set for vcpu beforehand.

rustfmt.toml
```diff
@@ -3,3 +3,4 @@
 # ---------------------------------------------------------------------------
 max_width = 80
 use_small_heuristics = "max"
+edition = "2024"
```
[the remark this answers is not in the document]
zeeshanlakhani: Yeah. I was the newest Rust and ran into editor issues here. I'll separate this.

openapi/propolis-server.json
```diff
@@ -1354,6 +1354,10 @@
       "NetworkInterfaceRequest": {
         "type": "object",
         "properties": {
+          "interface_id": {
```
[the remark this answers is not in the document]
zeeshanlakhani: Yep @pfmooney. This should be done post-merge w/ commit no? I had it in the description.

zeeshanlakhani:
> Looks pretty good overall, thanks for doing this! I've got some related changes in #746 that will conflict, as a head's up. I think merging those would make this change a good bit simpler, so that might be worthwhile. Thanks!

Yeah, I looked at that PR and see how you moved the kstat pieces up. I'm ok updating once that's in.

lib/propolis/src/cpuid.rs
zeeshanlakhani: They threw doc errors on `cargo doc` run of CI.

crates/rgb-frame/src/lib.rs
zeeshanlakhani: Escape the slashes. `cargo doc` rendered these with warnings.

crates/propolis-api-types/src/lib.rs
zeeshanlakhani: Yep. I'll do this in another PR.

bin/propolis-server/src/lib/stats/virtual_machine.rs
```diff
@@ -26,124 +31,31 @@ use self::virtual_machine::{
...
-
 /// A wrapper around the `oximeter::Target` representing a VM instance.
 ///
 /// This is used to combine the "real" target,
 /// `self::virtual_machine::VirtualMachine`, with some additional fields to help
 /// collect data via kstats. It's not currently possible to attach fields like
 /// this to the code generated by the `oximeter::use_timeseries!()` macro.
 #[derive(Clone, Debug)]
-pub struct VirtualMachine {
+pub(crate) struct VirtualMachine {
```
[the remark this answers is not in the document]
zeeshanlakhani: Yep. I was just `pub(crate)`'ing the new stats module and realized this didn't need to be pub only.

crates/propolis-api-types/Cargo.toml
zeeshanlakhani: Removed.

bin/propolis-server/src/lib/stats/virtual_machine.rs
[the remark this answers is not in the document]
zeeshanlakhani: I'll bring this back to what it was.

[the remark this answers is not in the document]
zeeshanlakhani: I'll make notes after the updated push.

bin/propolis-server/src/lib/vm/services.rs
zeeshanlakhani: I'll bring in #746 to my branch and work with that.

bin/propolis-cli/src/main.rs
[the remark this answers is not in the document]
zeeshanlakhani: Reverted this (on next push).

lib/propolis/src/cpuid.rs
zeeshanlakhani: I'll push these out into a separate PR.

bin/propolis-server/src/lib/stats.rs
zeeshanlakhani: Fix incoming (was an oversight).

Cargo.lock
zeeshanlakhani, line 1: TODO: I will update the lock correctly once everything from omicron/main and the associated dependencies are in place.

Cargo.toml
```diff
@@ -96,7 +96,7 @@ bitflags = "2.4"
 bitstruct = "0.1"
 bitvec = "1.0"
 byteorder = "1"
-bytes = "1.1"
+bytes = "1.7.1"
```
zeeshanlakhani, line 99: Note: This matches omicron/main, as I ran into issues when using omicron/main patched without this update.

bin/propolis-server/src/lib/stats/network_interface.rs
```diff
@@ -0,0 +1,351 @@
...
+    KSTAT_TX_PACKETS,
+    KSTAT_RX_DROPS,
+];
+
+/// The name of the kstat module that contains the instance network interface metrics.
+const KSTAT_MODULE_NAME: &str = "viona";
+
+/// Helper function to extract the same kstat metrics from all link targets.
+///
+/// TODO: Match up the kstat names with the metrics they actually represent,
```
zeeshanlakhani, line 51: I will remove this once tested of course.

openapi/propolis-server.json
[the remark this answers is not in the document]
zeeshanlakhani: I've noted this in the PR description.

bin/propolis-server/src/lib/stats.rs
```diff
@@ -4,113 +4,125 @@
...
+mod network_interface;
 mod pvpanic;
-pub(crate) mod virtual_machine;
-pub use self::pvpanic::PvpanicProducer;
+mod virtual_machine;
 
-// Interval on which we ask `oximeter` to poll us for metric data.
+#[cfg(all(not(test), target_os = "illumos"))]
+use self::network_interface::InstanceNetworkInterfaces;
+pub(crate) use self::pvpanic::PvpanicProducer;
```
zeeshanlakhani, line 30: This structure didn't need to be just `pub`. I updated this to be consistent with the updates here in updating modules and making it cleaner than exposing it through the crate (as we already use `mod` here).

zeeshanlakhani:
@pfmooney @bnaecker. Great comments. I removed a bunch of the doc changes and NFCs. I'll move that into another PR. 

@bnaecker I slurpped-in your non-virtual-disk changes from https://github.com/oxidecomputer/propolis/pull/746 into this branch. This made it easier to update and use the newer abstraction(s) for initialization (and avoiding an update to the VM objects structure). We can discuss the best path to order those PRs and whatnot (I'm ok rebasing, w/e works).

zeeshanlakhani:
@pfmooney I think that https://code.illumos.org/c/illumos-gate/+/3630 landed, right? Noob Q: Is the next step an update to Helios?

openapi/propolis-server.json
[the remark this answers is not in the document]
zeeshanlakhani: Or, is there a spot where we do that here? Double-checking.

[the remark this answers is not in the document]
zeeshanlakhani: Ah, the openapi `.json` for falcon. Got it. And, it's done manually. Got it.

bin/propolis-server/src/lib/stats/network_interface.rs
```diff
@@ -0,0 +1,351 @@
...
+#[derive(Clone, Debug)]
+pub(crate) struct InstanceNetworkInterfaces {
+    /// The `oximeter::Target` itself, storing the target fields for the
+    /// timeseries.
+    ///
+    /// **NOTE**: While this struct represents multiple instance network interfaces,
+    /// they all share the same target fields.
+    ///
+    /// We default `interface_id` by generating a `uuid::Uuid::nil()` on first
+    /// creation, before creating multiple tragets in the `to_samples` method.
```
gjcolombo, line 115: nit: s/tragets/targets

bin/propolis-server/src/lib/vm/mod.rs
```diff
@@ -137,6 +140,9 @@ pub(crate) type CrucibleReplaceResult =
 pub(crate) type CrucibleReplaceResultTx =
     oneshot::Sender<CrucibleReplaceResult>;
 
+/// PCI device instance ID type.
+type DeviceInstanceId = u32;
```
gjcolombo, line 144: I had (am having?) a little trouble following exactly what this type was (or is). It looks like it's an opaque identifier for the kstat series for a particular device; is that right? If so I think I would rename a bit, maybe to `KstatInstanceId`, and update the doc comment to denote that this is just a disambiguating value that might exist for any per-component kstat (it doesn't just have to be a PCI device).

gjcolombo (commented):
Thanks for putting this together. The network interface ID plumbing generally makes sense to me, but I had a couple of comments about how things are named/documented that I think could help the concepts to hang together a little bit better. I'll look to do another lap once #746 merges and we rebase this (I think I have mostly kept straight which parts of the PR are from which change but will want to take another look once they're separated out again).

A note about the CI tests--although this is technically an API-breaking change, it's the rare sort of change that I think will *not* break the `phd-run-migrate-from-base` test run, because PHD doesn't currently support guest NICs. If that run does fail once all of the other CI tests are fixed up, we should take a look to make sure we understand what's broken.

bin/propolis-server/src/lib/stats.rs
bnaecker: @pfmooney `OXIMETER_STAT_INTERVAL` refers to the interval on which we ask `oximeter` to collect data from Propolis. The other intervals are used in the `KstatSampler` to determine how frequently to generate samples from the kstats themselves. Each item registered with the sampler defines its own sampling interval, since there's no one-size-fits-all.

I'm actually going to rename this for clarity in #746 to help address this.

pfmooney:
> @pfmooney I think that https://code.illumos.org/c/illumos-gate/+/3630 landed, right? Noob Q: Is the next step an update to Helios?

It is landed, yes.  The `stlouis` branch periodically does merges from upstream illumos-gate, at which point the new bits would be available in freshly built helios.

bin/propolis-server/src/lib/stats.rs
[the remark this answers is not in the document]
zeeshanlakhani: Yeah. This was stats previously, but I can do that in mine PR.

crates/propolis-api-types/src/instance_spec/components/devices.rs
```diff
@@ -96,6 +97,9 @@ pub struct VirtioNic {
     /// The name of the device's backend.
     pub backend_name: String,
 
+    /// The interface ID of the device.
```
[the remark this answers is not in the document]
zeeshanlakhani: It's at the [requested_nic](https://github.com/oxidecomputer/omicron/pull/6414/files#diff-e58fdced2e69108a490916cca62b391c44fdde81d564de6f86b8d4e1febc5998R694).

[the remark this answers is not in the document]
zeeshanlakhani: I'd say the latter.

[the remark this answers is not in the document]
zeeshanlakhani: I updated the comment.

zeeshanlakhani:
Ok, I've rebased against #746, but will wait to push when that's in. I've also got your review comments addressed in my next push @gjcolombo.

zeeshanlakhani:
> > @pfmooney I think that https://code.illumos.org/c/illumos-gate/+/3630 landed, right? Noob Q: Is the next step an update to Helios?
> 
> It is landed, yes. The `stlouis` branch periodically does merges from upstream illumos-gate, at which point the new bits would be available in freshly built helios.

Gotcha. Pinged @citrus-it on this to best get going w/ the testing. Thanks!

zeeshanlakhani:
Just drafting this until #746 is in.

bin/propolis-server/src/lib/stats/mod.rs
```diff
@@ -217,3 +239,149 @@ pub(crate) async fn track_vcpu_kstats(
...
+
+        fn as_i64(&self) -> Result<i64, Error> {
+            unimplemented!()
+        }
+
+        fn as_u64(&self) -> Result<u64, Error> {
+            if let NamedData::UInt64(x) = self {
+                Ok(*x)
+            } else {
+                panic!()
```
pfmooney, line 369: Why `panic` instead of emitting an error?

This is currently causing `stats::network_interface::test::test_kstat_to_samples` to fail on non-illumos

zeeshanlakhani: That was leftover from the previous code, but will update.

zeeshanlakhani: And, also update the test :).

pfmooney: The test may pass again if some of this shim code is updated to throw errors rather than panic.

bnaecker: The `panic!()` was intentional. It looks like the test you're referring to is failing because the code converting from the kstat to the oximeter sample is trying to convert a `NamedData::U32(_)` to a `u64`. We can return an error, but the code is wrong in either case. I wanted to loudly, quickly, and irrecoverably when testing conversion routines that expect some relationship between the types of the kstat and the oximeter sample it generates, in this case, that they're both `u64`s under the hood.

bnaecker:
@zeeshanlakhani Could you avoid force-pushing while the code is under review? GitHub makes it pretty difficult to follow changes in the code as things are overwritten. You can always amend the history as you see fit after the review is done and before merging, either in a new commit or at squash-and-merge time.

zeeshanlakhani:
In live testing with omicron-dev `ch-run`, `oximeter-standalone`, and running `propolis-server` and `propolis-cli` from [these docs], we see the updated time series here, containing `instance_network_interface:*`:

```
instance_network_interface:bytes_received
instance_network_interface:bytes_sent
instance_network_interface:errors_received
instance_network_interface:errors_sent
instance_network_interface:packets_dropped
instance_network_interface:packets_received
instance_network_interface:packets_sent
oximeter_collector:collections
oximeter_collector:failed_collections
virtual_machine:pv_panic_guest_handled
virtual_machine:pv_panic_host_handled
virtual_machine:reset
virtual_machine:vcpu_usage
```

These `instance_network_interface:*` series sample kstats that are available after a recent helios update from the [updating illumos-core from the st. Louis branch](https://github.com/oxidecomputer/illumos-gate/tree/stlouis/) (thank you @pfmooney for the work and @citrus-it for pushing his updates through). These are the representative kstats that are parsed:

```
viona:46:viona_stat:rx_bytes                                                                                                   
viona:46:viona_stat:rx_drops                                                                                                   
viona:46:viona_stat:rx_errors                                                                                                  
viona:46:viona_stat:rx_packets                                                                                                 
viona:46:viona_stat:tx_bytes                                                                                                   
viona:46:viona_stat:tx_drops                                                                                                   
viona:46:viona_stat:tx_errors                                                                                                  
viona:46:viona_stat:tx_packets
``` 

We can check what oximeter's collected with oxql:
```
0x> get instance_network_interface:bytes_received | last 1

instance_id: a5aa5bb1-4fa5-4bb3-8b06-9df335ebc28c
interface_id: 13149b56-8209-493e-b21a-a10367b8ce20
project_id: 33d039a6-d4fc-418b-a544-1f737791d559
silo_id: 6e8d2808-afcc-4d4d-9d05-415b04bff4f4
  [2024-08-29 17:19:55.279446436, 2024-08-29 18:50:18.094871292]: [4846]
```

which matches our `rx_bytes` kstat. 
```
⋊> zeeshan@rainbow ⋊> ~/o/o/o/o/schema on zl/instance-network-interface ⨯ kstat viona:46:viona_stat:rx_bytes           18:50:41
module: viona                           instance: 46
name:   viona_stat                      class:    misc
        rx_bytes                        4846
```

On a different run, let's double-check that the interface_ids match up:

```
network_devices: {
    "vnic-0.5.0": VirtioNic(
        VirtioNic {
            backend_name: "vnic-0.5.0-backend",
            interface_id: ca2dab31-bf60-413c-814c-d1fb78f32736,
            pci_path: PciPath {
                bus: 0,
                device: 5,
                function: 0,
            },
        },
    ),
},...
```
```
0x〉get instance_network_interface:bytes_received | last 1

instance_id: ff4bb3f9-9ce8-431d-a55f-00705f47fb55
interface_id: ca2dab31-bf60-413c-814c-d1fb78f32736
project_id: 1ddbd98a-9bee-42aa-9496-5b2c344e6dbf
silo_id: bea7cb74-1982-4416-8d08-75ffb6854f28
  [2024-08-29 18:57:48.589123332, 2024-08-29 19:01:28.598514660]: [3180]
```

And, yep, they do.

zeeshanlakhani:
> @zeeshanlakhani Could you avoid force-pushing while the code is under review? GitHub makes it pretty difficult to follow changes in the code as things are overwritten. You can always amend the history as you see fit after the review is done and before merging, either in a new commit or at squash-and-merge time.

Yep. I usually do. Apologies as I was keeping the temporary dependencies commit as the recent one.

bin/propolis-server/src/lib/stats/mod.rs
zeeshanlakhani: Yep. I quickly fixed up the code after changing NamedData type and moved to the error anyhow.

bin/propolis-server/src/lib/stats/mod.rs
```diff
@@ -171,8 +188,13 @@ pub fn start_oximeter_server(
 pub(crate) fn create_kstat_sampler(
     log: &Logger,
     n_vcpus: u32,
+    n_interfaces: u32,
```
gjcolombo, line 191: Optional: instead of having bare u32s it might be a bit cleaner to have this function take the instance properties/spec as an argument and peel off the values it needs (instead of giving callers the chance to make a transposition mistake).

bin/propolis-server/src/lib/vm/mod.rs
```diff
@@ -111,6 +111,9 @@ pub(crate) mod state_publisher;
 pub(crate) type DeviceMap =
     BTreeMap<String, Arc<dyn propolis::common::Lifecycle>>;
 
+/// Mapping of NIC identifiers to viona device instance IDs.
+pub(crate) type NetworkInterfaceIds = Vec<(uuid::Uuid, KstatInstanceId)>;
```
gjcolombo, line 115: Since this is semantically a map, should we just go ahead and make a `BTreeMap` out of it?

bin/propolis-server/src/lib/vm/ensure.rs
```diff
@@ -199,7 +199,7 @@ impl<'a> VmEnsureNotStarted<'a> {
         let ps2ctrl = init.initialize_ps2(&chipset)?;
         init.initialize_qemu_debug_port()?;
         init.initialize_qemu_pvpanic(properties.into())?;
-        init.initialize_network_devices(&chipset)?;
+        init.initialize_network_devices(&chipset, &maybe_kstat_sampler).await?;
```
gjcolombo, line 202: Instead of having to pass `maybe_kstat_sampler` as an arg to every init function that needs it, I'd strongly consider making it a `MachineInitializer` member at this point. (I think this would involve removing `MachineInitializer::create_kstat_sampler` and then just calling its callee free function directly from here; this shouldn't be too bad since we have all the MachineInitializer members in this function anyway.)

gjcolombo (commented):
Apologies for taking so long to do another lap here. This is coming together nicely; most of my feedback is minor style/organizational stuff. The main thing I want to make sure I have right is the description of the new `interface_id` field in the spec API.

bin/propolis-server/src/lib/stats/mod.rs
```diff
@@ -128,7 +145,7 @@ impl Producer for ServerStats {
 /// task, and will periodically renew that registration. The returned server is
 /// running, and need not be poked or renewed to successfully serve metric data.
 pub fn start_oximeter_server(
-    id: Uuid,
+    id: uuid::Uuid,
```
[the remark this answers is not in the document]
zeeshanlakhani: Ah no. There were just some inconsistencies around importing it or not. I'll choose the one path in this module.

bin/propolis-server/src/lib/spec/config_toml.rs
```diff
@@ -291,6 +291,9 @@ pub(super) fn parse_network_device_from_config(
 
     let device_spec = NetworkDeviceV0::VirtioNic(VirtioNic {
         backend_name: backend_name.clone(),
+        // We don't allow for configuration to specify the interface_id, so we
+        // generate a new one.
+        interface_id: uuid::Uuid::new_v4(),
```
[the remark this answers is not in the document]
zeeshanlakhani: Yeah, agreed, as I went back and forth on this. I had `nil` earlier and then went with a generated `v4`.

bin/propolis-server/src/lib/vm/mod.rs
zeeshanlakhani: Yeah, I talked to @bnaecker  on this earlier. Since we only really have 8 or some limited number, it made more sense just to use a Vec and iterate vs a map.

bin/propolis-server/src/lib/vm/ensure.rs
zeeshanlakhani: Yeah. I had a variation of this earlier before moving the sampler itself into initializer/ensure, as per @bnaecker's PR. I'll do this.

bin/propolis-server/src/lib/vm/mod.rs
zeeshanlakhani: I'll add a comment.

gjcolombo (approved):
Thanks for working in the feedback! This looks good from my end.

bin/propolis-server/src/lib/stats/mod.rs
```diff
@@ -217,3 +242,151 @@ pub(crate) async fn track_vcpu_kstats(
...
+    let details = oximeter_instruments::kstat::CollectionDetails::never(
+        NETWORK_INTERFACE_SAMPLE_INTERVAL,
+    );
+    let interface_id = nics.target.interface_id;
+    if let Err(e) = sampler.add_target(nics, details).await {
+        slog::error!(
+            log,
+            "failed to add network interface targets, \
+            network interface stats will be unavailable";
+            "network_interface_id" => %interface_id,
```
bnaecker, line 278: At this point, this UUID is `Uuid::nil()`, right? We probably want to print the list of actual UUIDs, in `nics.interface_ids`, I think.

bnaecker (approved):
Looks good, thanks for the back and forth! One suggestion about logging, but that's it. Thanks!

PR #751, opened 2024-08-28
server: populate disk backend names correctly

### LINKED ISSUE #750: `parse_disk_from_request` mishandles backend names
I borked this in #743. Relevant code: https://github.com/oxidecomputer/propolis/blob/88fbde76311cd6862721e3e085fe0093c2cbe6d9/bin/propolis-server/src/lib/spec/api_request.rs#L72-L109

The previous version of this code used to generate the backend name up front and then use it throughout the routine, i.e., what is now on line 94 of the snippet was actually up at line 75. This breaks disk request processing because the disk devices no longer point to the registered backends.

The completed instance spec refactor will help to avoid this problem by having just one collection for disk device/backend pairs instead of distinct collections for devices and backends. For now we need a fix to unblock the metrics work we merged in #746.

gjcolombo:
Fixes #750.

Tests: new cargo tests; dropped a package image into an Omicron dev cluster and verified that instances start correctly with it.

### DIFF
--- a/bin/propolis-server/src/lib/spec/api_request.rs
+++ b/bin/propolis-server/src/lib/spec/api_request.rs
@@ -73,13 +73,14 @@ pub(super) fn parse_disk_from_request(
     disk: &DiskRequest,
 ) -> Result<ParsedStorageDevice, DeviceRequestError> {
     let pci_path = slot_to_pci_path(disk.slot, SlotType::Disk)?;
+    let backend_name = format!("{}-backend", disk.name);
     let device_spec = match disk.device.as_ref() {
         "virtio" => StorageDeviceV0::VirtioDisk(VirtioDisk {
-            backend_name: disk.name.to_string(),
+            backend_name: backend_name.clone(),
             pci_path,
         }),
         "nvme" => StorageDeviceV0::NvmeDisk(NvmeDisk {
-            backend_name: disk.name.to_string(),
+            backend_name: backend_name.clone(),
             pci_path,
         }),
         _ => {
@@ -91,7 +92,6 @@ pub(super) fn parse_disk_from_request(
     };
 
     let device_name = disk.name.clone();
-    let backend_name = format!("{}-backend", disk.name);
     let backend_spec = StorageBackendV0::Crucible(CrucibleStorageBackend {
         request_json: serde_json::to_string(&disk.volume_construction_request)
             .map_err(|e| {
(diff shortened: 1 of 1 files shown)


jordanhendricks (approved):

bin/propolis-server/src/lib/spec/api_request.rs
```diff
@@ -152,3 +152,59 @@ pub(super) fn parse_nic_from_request(
...
+
+    fn check_parsed_storage_device_backend_pointer(
+        parsed: &ParsedStorageDevice,
+    ) {
+        let device_to_backend = match &parsed.device_spec {
+            StorageDeviceV0::VirtioDisk(d) => d.backend_name.clone(),
+            StorageDeviceV0::NvmeDisk(d) => d.backend_name.clone(),
+        };
+
+        assert_eq!(device_to_backend, parsed.backend_name);
```
iximeow, line 171: it seems like instead of having `ParsedStorageDevice.backend_name` and requiring it always matches what happens to also be in the device spec, that we'd be better off with a `StorageDeviceV0::backend_name()` helper that we use in places like `add_storage_device()`.  something like `.insert(device_spec.backend_name(), backend_spec)`?

do we expect to have storage devices that don't know their backend name? (that seems unlikely, but it's an obvious reason my thought wouldn't stick!)

(i also understand this change and why it fixes things, so +1 for landing to make things fixed and adjusting after if you're so inclined)

iximeow (approved):

bin/propolis-server/src/lib/spec/api_request.rs
gjcolombo: Ultimately my plan is to do away with `ParsedStorageDevice` entirely and replace it with a `Disk` type that contains both the device and backend halves of each of a VM's storage components. In this bug I got caught looking the wrong way halfway through the series of refactoring PRs that's supposed to make that happen ;)

I have the result of this wave of refactoring in [a branch in my fork](https://github.com/oxidecomputer/propolis/compare/master...gjcolombo:propolis:instance-spec-rework#diff-f8258ff7a3e14ba17fbb2c5c6655bea413975c40642622e5eeb3ba39cfa03729); here's what [the routine this PR fixes](https://github.com/gjcolombo/propolis/blob/2f65e8cd8f0d82ac09573d2a5edcf217798f9a49/bin/propolis-server/src/lib/spec/api_request.rs#L71-L102) will look like when that's done. This will (hopefully) foreclose on the sort of bug we hit here, since machine init won't be looking up named backends in a map anymore. (I think the code can be improved even more beyond what I have in the fork at this point, but am trying to stick to one renovation project at a time--insert the beaver-big-plans meme here...) 

There's even more context in #735 if you're looking for some light bedtime reading.

PR #752, opened 2024-08-28
Make xtask-prepush configurably verbose

pfmooney:


xtask/src/main.rs
```diff
@@ -38,7 +38,11 @@ enum Cmds {
     /// (Crudely) Check for appropriate license headers
     License,
     /// Preform pre-push checks (clippy, license, fmt, etc)
-    Prepush,
+    Prepush {
+        /// Suppress non-essential output
+        #[arg(short, long)]
+        quiet: bool,
```
hawkw, line 44: It looks like this changes the default behavior from quiet to verbose --- on `master` the command only prints output on errors, but now, it will print what it's doing _unless_ `--quiet` is passed. Is that the desired change, or would you prefer to make the flag turn _on_ extra output?

hawkw (approved):
No concerns from me!

xtask/src/main.rs
pfmooney: This was intentional.  I've had feedback from more than one person who through the command might be broken, because it was silently doing a bunch of clippy work.

hawkw: Ah, got it! I wasn't totally sure what the motivation for this change was, but that makes sense. In that case, looks good to me!

zeeshanlakhani (approved):

pfmooney:
Merged in 1c4ff1dd161b532f86167ad725aedbf832f7285d

PR #754, opened 2024-08-29
Wire up viona notify fast path

pfmooney:
The viona driver has the ability to hook an ioport (as specified by the userspace vmm) in order to more efficiently process notifications from the guest that virtqueues have new available entries.  This cuts down on the number of VM exits which must be processed by userspace.

### DIFF
--- a/lib/propolis/src/hw/pci/bar.rs
+++ b/lib/propolis/src/hw/pci/bar.rs
@@ -121,33 +121,35 @@ impl Bars {
         &mut self,
         bar: BarN,
         val: u32,
-    ) -> Option<(BarDefine, u64, u64)> {
+    ) -> Option<WriteResult> {
         let idx = bar as usize;
         let ent = &mut self.entries[idx];
-        let (def, old, new) = match ent.kind {
+        let (id, def, val_old, val_new) = match ent.kind {
             EntryKind::Empty => return None,
             EntryKind::Pio(size) => {
                 let mask = u32::from(!(size - 1));
                 let old = ent.value;
                 ent.value = u64::from(val & mask);
-                (BarDefine::Pio(size), old, ent.value)
+                (bar, BarDefine::Pio(size), old, ent.value)
             }
             EntryKind::Mmio(size) => {
                 let mask = !(size - 1);
                 let old = ent.value;
                 ent.value = u64::from(val & mask);
-                (BarDefine::Mmio(size), old, ent.value)
+                (bar, BarDefine::Mmio(size), old, ent.value)
             }
             EntryKind::Mmio64(size) => {
                 let old = ent.value;
                 let mask = !(size - 1) as u32;
                 let low = val & mask;
                 ent.value = (old & (0xffffffff << 32)) | u64::from(low);
-                (BarDefine::Mmio64(size), old, ent.value)
+                (bar, BarDefine::Mmio64(size), old, ent.value)
             }
             EntryKind::Mmio64High => {
                 assert!(idx > 0);
-                let ent = &mut self.entries[idx - 1];
+                let real_idx = idx - 1;
+                let id = BarN::from_repr(real_idx as u8).unwrap();
+                let ent = &mut self.entries[real_idx];
                 let size = match ent.kind {
                     EntryKind::Mmio64(sz) => sz,
                     _ => panic!(),
@@ -156,11 +158,11 @@ impl Bars {
                 let old = ent.value;
                 let high = ((u64::from(val) << 32) & mask) & 0xffffffff00000000;
                 ent.value = high | (old & 0xffffffff);
-                (BarDefine::Mmio64(size), old, ent.value)
+                (id, BarDefine::Mmio64(size), old, ent.value)
             }
         };
-        if old != new {
-            return Some((def, old, new));
+        if val_old != val_new {
+            return Some(WriteResult { id, def, val_old, val_new });
         }
         None
     }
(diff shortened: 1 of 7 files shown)


lib/propolis/src/hw/pci/device.rs
```diff
@@ -608,10 +651,9 @@ impl DeviceState {
         let attach = inner.attached();
         for n in BarN::iter() {
             if let Some((def, addr)) = inner.bars.get(n) {
-                let pio_en = inner.reg_command.contains(RegCmd::IO_EN);
-                let mmio_en = inner.reg_command.contains(RegCmd::MMIO_EN);
-
-                if (pio_en && def.is_pio()) || (mmio_en && def.is_mmio()) {
+                if (inner.pio_en() && def.is_pio())
+                    || (inner.mmio_en() && def.is_mmio())
```
hawkw, line 655: nit, take it or leave it: I wonder if there ought to be a method like
```rust
impl State {
    fn should_register_bar(&self, bar: &BarDefine) -> bool {
         (self.pio_en() && bar.is_pio()) || (self.mmio_en() && def.is_mmio())
    }
}
```
that  could be called both here and in `cfg_std_write`, above?

Not sure if it's worth factoring out or not, just noticed that it was repeated.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -70,11 +71,16 @@ enum VRingState {
...
+    iop_state: Option<u16>,
     vring_state: [VRingState; 2],
 }
 impl Inner {
     fn new() -> Self {
-        Self { poller: None, vring_state: [Default::default(); 2] }
+        Self {
+            poller: None,
+            iop_state: None,
+            vring_state: [Default::default(); 2],
```
hawkw, line 82: take it or leave it: this is identical to what a derived `Default` implementation would do...

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -648,6 +687,14 @@ impl VionaHdl {
         Ok(())
     }
 
+    fn set_notify_iop(&self, port: Option<u16>) -> io::Result<()> {
```
hawkw, line 690: Perhaps this is just because I don't have an illumos background, but it took a bit for me to realize that "IOP" here referred to "IO port"...I might prefer to expand it to `set_notify_io_port`? I note that we have a function above `notify_port_update` that calls this function, so perhaps we could call this `set_notify_port` to mirror that? But, if you prefer to more closely mirror the name of the ioctl, that's fair, too. :)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -475,6 +504,16 @@ impl PciVirtio for PciVirtioViona {
     fn pci_state(&self) -> &pci::DeviceState {
         &self.pci_state
     }
+    fn notify_port_update(&self, port: Option<u16>) {
```
hawkw, line 507: It might be worth a comment explaining the difference between `notify_port_update` and `set_notify_iop`, which this calls if the device is running? It could be helpful if we explained when this should be called, and when it's okay to call `set_notify_iop` unconditionally?

lib/propolis/src/lifecycle.rs
```diff
@@ -87,3 +89,96 @@ pub trait Lifecycle: Send + Sync + 'static {
...
+    pub fn halt(&self) {
+        self.state_change(IndicatedState::Halt);
+    }
+    /// Get the currently indicated state for the holder
+    pub fn state(&self) -> IndicatedState {
+        IndicatedState::from_usize(self.0.load(Ordering::SeqCst))
+    }
+
+    fn state_change(&self, new: IndicatedState) {
+        let old = self.0.swap(new as usize, Ordering::SeqCst);
```
hawkw, line 142: I don't think sequentially-consistent ordering is required here. I don't believe there is another memory location that we want to have a total ordering of sequentially-consistent stores/loads with, so `AcqRel` should be sufficient here?

I doubt the performance penalty for sequentially-consistent stores matters a whole lot here, though, so :woman_shrugging:

hawkw (commented):
This is a pretty cursory review, since I don't know a whole lot about the viona API --- probably worth getting a look from someone who does. Overall, though, this looks good, at least as far as I can tell.

lib/propolis/src/lifecycle.rs
```diff
@@ -95,3 +97,96 @@ pub trait Lifecycle: Send + Sync + 'static {
...
+impl IndicatedState {
+    const fn valid_transition(old: Self, new: Self) -> bool {
+        use IndicatedState::*;
+        match (old, new) {
+            (Init, Run) | (Init, Halt) => true,
+            (Run, Pause) => true,
+            (Pause, Run) | (Pause, Halt) => true,
+            _ => false,
+        }
+    }
```
iximeow: thinking out loud: this describes the valid state transitions for any `Lifecycle` implementer even if they don't use an `Indicator`. it's an explicit form of the doc comments on `trait Lifecycle`, and feels like we'd get useful error checking if every `Lifecycle` got a free `Indicator` riding along to tell us (panicfully) if a Propolis user gets the state machine wrong. i'm not sure how to spell this in a nice Rust-y way yet, but i agree with the value of `Indicator` and i'll make a followup issue for bringing the implict and explicit state machines together...

(this brought to you by me wondering what should happen when you `reset()`, looking at `Lifecycle`, thinking `Pause -> Run` is prooobably right, then following the transitions in `propolis-standalone` to be more sure.)

lib/propolis/src/hw/pci/bar.rs
```diff
@@ -121,33 +121,35 @@ impl Bars {
...
                 let mask = !(size - 1);
                 let old = ent.value;
                 ent.value = u64::from(val & mask);
-                (BarDefine::Mmio(size), old, ent.value)
+                (bar, BarDefine::Mmio(size), old, ent.value)
             }
             EntryKind::Mmio64(size) => {
                 let old = ent.value;
                 let mask = !(size - 1) as u32;
                 let low = val & mask;
```
iximeow: i really just noticed it while reading this code, but i think we should mask off the low four bits of `val` for `Mmio` and `Mmmio64` (i.e. the low BAR), 'cause it looks to me like if firmware wrote back the bits we offered we'd faithfully place the bar off-by-one for no good reason.

PR #756, opened 2024-09-03
teach propolis-server to understand configurable boot order

iximeow:
brings the existing fw_cfg-based boot order configuration that we already have in `propolis-standalone` over to `propolis-server` as well.

pairs with https://github.com/oxidecomputer/omicron/pull/6585 to actually configure instances' boot settings and get them to Propolis.

while we accept a list of boot options here, note that Nexus intentionally has a more reductive view, maintaining only a single `boot_disk`. this is heavily based on what i've learned writing tests for how OVMF+guests behave with different mixes of boot device orders and sources of boot configuration. [this conversation](https://github.com/oxidecomputer/rfd/pull/751#discussion_r1612632313) on RFD 470 gets at the crux of it - we don't have like a platform NVRAM device for guests to persist EFI variables in, so boot configuration is persisted in the EFI system partition (so, associated with a disk, not an instance). because we're not going to go modify the EFI system partition in user disks to effect boot order preferences, we're using the QEMU-style `fw_cfg` mechanism to indicate boot order. this, in turn, means that guest settings are blown away and replaced with what OVMF determines at boot time.

taken together, this isn't an ideal spot to be in if we were to support more complex boot behavior logic, and it's probably not how we want to work multi-device orderings into `propolis-server`. it also means that guest-managed boot ordering just don't reliably persist if the instance is configured with a specific boot device!

in the future with managed nonvolatile storage for guest firmware to persist settings in, we'll be in a somewhat better position to extend boot logic, and so this PR should leave us able to do so without much API hassle. that will probably be more _interesting_ on the control plane side, which is why i'm permissive on the Propolis side, but restrictive on the Nexus side.


phd-tests/tests/src/boot_order.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    cfg.data_disk(
+        "alpine-3-16",
+        DiskSource::Artifact("alpine-3-16"),
+        DiskInterface::Virtio,
+        DiskBackend::File,
+        24,
+    );
+
+    cfg.boot_order(vec!["alpine-3-20", "alpine-3-16"]);
```
iximeow, line 31: and finally this is where i'm *really* seeking input. i was _hoping_ to attach a few disks, or even the same disk a few times, vary the boot device order, and use tools in the guest image to verify which PCI slot the boot device is attached to.

in Alpine and other linuxes i could test this by looking for what disk backs `/`, then check which PCI device `/sys/class/block/<that_device>` is a symlink to. t(he perhaps-more-obvious `/dev/disk/by-path` wouldn't work because the minimal Alpine doesn't have udev.) - at first i thought testing this in minimal Linuxes would just be impossible, checking `/sys/class/` only struck me as i was writing this :)

then testing similar for Windows guests is probably doable with some `diskpart` incantations. so that would be great.

but the real wrinkle is that i'd kind of like to test arbitrary pairs of OSes here - whatever's first in the list should behave the same regardless of the second thing in the list, and more importantly if the first boot entry is not actually bootable, it would be a firmware bug if that made the second item fail to boot when we know it would boot on its own too. now, i realize that's really testing that OVMF works, so maybe it's a bit too overboard. opinions?

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -28,12 +28,18 @@ use propolis::hw::ibmpc;
...
+    fwcfg::{self, Entry},
+    ramfb,
+};
 use propolis::hw::uart::LpcUart;
 use propolis::hw::{nvme, virtio};
 use propolis::intr_pins;
 use propolis::vmm::{self, Builder, Machine};
-use propolis_api_types::instance_spec::{self, v0::InstanceSpecV0};
+use propolis_api_types::instance_spec::{
+    self, v0::BootDeclaration, v0::InstanceSpecV0,
```
iximeow, line 41: oh yes, one thing i'd not understood in our OpenAPI documents: why are some things named `v0::ThingV0` and some are... not? e.g. `v0::SerialPort` vs `v0::InstanceSpecV0`. would the new type be `v0::BootDeclarationV0`?

gjcolombo: My (possibly misguided) intention here was to separate the component definitions themselves (things like `SerialPort`) from the types that collect a bunch of components into a complete instance spec, so that defining a new "format" for a component collection wouldn't entail copying a bunch of component definitions into a new `v1` module (or having the `v1` module depend on component definitions from the `v0` module, which seemed confusing to me). 

At the risk of lulling you to sleep with more of my technical writing, there's a [theory statement](https://github.com/oxidecomputer/propolis/blob/5267be82e10d851a64196a8148893691b0b9f8ad/crates/propolis-api-types/src/instance_spec/mod.rs) in the instance spec portion of the types crate that might provide some additional color here.

The version-numbered types have the `V0` suffix because all of the types that contribute to a generated OpenAPI spec need to have distinct *names*, not just distinct Rust module paths; if you do try to refer to multiple identically-named types, Dropshot will pick one to include and discard the rest. (This is dropshot#383).

In this case, I think the boot order declaration is functionally a "component", so I don't think I would give it a suffix; it should live under the `components` part of the module tree and be referenced from within the spec structure from there. (Eventually I expect to collapse all the various component containers in `InstanceSpecV0` into a single bag of components so that it's clearer where to put new things.)

bin/propolis-standalone/src/main.rs
```diff
@@ -959,8 +959,13 @@ fn generate_smbios(params: SmbiosParams) -> anyhow::Result<smbios::TableBytes> {
     Ok(smb_tables.commit())
 }
 
-fn generate_bootorder(config: &config::Config) -> anyhow::Result<fwcfg::Entry> {
-    let names = config.main.boot_order.as_ref().unwrap();
+fn generate_bootorder(
+    config: &config::Config,
+) -> anyhow::Result<Option<fwcfg::Entry>> {
+    eprintln!("bootorder declared as {:?}", config.main.boot_order.as_ref());
```
pfmooney, line 965: This should probably pass in the `slog::Logger`  and emit it as something like debug level?

iximeow: (i tend to lean on eprintln/panic in Rust as easily-findable "reconsider this and make it worth actually landing" sigils. here and in other places it'll probably get sent through to `slog` if relevant bits aren't otherwise reworked)

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -994,6 +1001,121 @@ impl<'a> MachineInitializer<'a> {
...
+                        )
+                    });
+
+                bdf
+            };
+
+        for boot_entry in boot_names.iter() {
+            if boot_entry.first_boot_only {
+                continue;
+            }
```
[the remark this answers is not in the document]
gjcolombo: The machine initializer code only gets called once per instance ensure request--reboot requests don't re-invoke this path. I think we would want to handle the first-boot-only discipline in the fwcfg module instead. 

I haven't looked at the fwcfg code too much, so am spitballing a bit, `struct Fwcfg` impls the `Lifecycle` trait, particularly the `reset` function, so it already has a way to learn about reboots; we could, I think, have a special `Entry::BootOrder` directory entry type that remembers whether it's a one-shot, and on reset, we could go have `Fwcfg` rummage for it and remove the entry if it's marked as one-shot. This would also take care of things over a migration; once the initially-populated entry is removed, it won't be exported anymore, so it won't be present when `FwCfg::import` restores entries.

bin/propolis-standalone/src/main.rs
pfmooney: Sounds good

phd-tests/tests/src/boot_order.rs
gjcolombo: > now, i realize that's really testing that OVMF works, so maybe it's a bit too overboard. opinions?

I think that sort of test is well within bounds here. PHD (to my mind) is an integration test suite that's focused in particular on making sure the *guest abstraction* works the way that we expect. I think it's reasonable to test guest firmware behavior as part of that. It also fits naturally with the way we generate and consume EDK2 builds, where we would:

- make a change in our fork of the EDK2 repo
- wait for buildomat to build the `OVMF_CODE` images that Propolis consumes
- update references to those images in this repo to refer to the commit hash that buildomat processed in the previous step; this includes
  - the commit hash in the `package` directory that specifies what image should be included in the Propolis zone images that Omicron pulls in
  - the commit hash in the PHD artifacts TOML

The PR for this last change will run Propolis CI with the new firmware bits, which gives us signal about how they work "in context" with the rest of Propolis, booting real guest images, honoring settings from the fwcfg module, etc.

---

> but the real wrinkle is that i'd kind of like to test arbitrary pairs of OSes here

I think this might be tricky, or at least will require us to revisit some unwritten assumptions about how PHD artifacts are supposed to be used. Specifically, PHD assumes that each run selects one guest OS to run against (via the `--default-guest-os-artifact` parameter); that OS has to appear in the artifact TOML/artifact store, but it could be just one of the many possible selections there. My (so far unrealized) idea was to set up periodic test runs that run PHD multiple times, each time selecting a different guest OS of interest from a single artifact TOML that's checked into the repo; then we could see that, e.g., such-and-such change works fine for Debian 11 guests but breaks Windows Server 2019 guests, or works fine with this Linux distro but not that one, and so on.

The main thing is that, beyond the value passed in the `default-guest-os-artifact` switch, I hadn't really envisioned that tests would require any particular artifact name to exist in the store. That's so that folks can easily adapt PHD for their own local testing (e.g. by stashing extra test images on a local disk and referring to them in the artifact TOML) without requiring them to obtain a specific list of prerequisite artifacts beyond the ones they've chosen to test.

We could definitely revisit that strategy, but I think my vision for this particular test was that it would instantiate multiple disks backed by the chosen guest OS artifact, and then try to figure out which one actually got booted from. (If we want a non-bootable disk to test fallback, we can just paste a blank disk into the boot order, though this currently requires the use of a Crucible or in-memory disk.)

gjcolombo (commented):
@iximeow I've taken a stab at answering your other questions here; let me know if I can clarify any of those answers or get you any more info about how these parts of Propolis are (meant to be) arranged.

Thanks for putting this together and for asking thoughtful questions about the existing design! There are definitely some pre-existing things we can go improve here.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -994,6 +1001,121 @@ impl<'a> MachineInitializer<'a> {
...
+                let instance_spec::v0::NetworkDeviceV0::VirtioNic(vnic_spec) =
+                    spec;
+
+                &vnic_spec.backend_name
+            };
+
+            if let Some(device_spec) =
+                self.spec.devices.storage_devices.iter().find_map(
+                    |(_name, spec)| {
+                        if storage_backend(spec) == &boot_entry.name {
```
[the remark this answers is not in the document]
iximeow: oh, backend names are derived in Propolis! i re-found [`format!("{}-backend")`](https://github.com/oxidecomputer/propolis/blob/master/bin/propolis-server/src/lib/spec/api_request.rs#L76) that you'd fixed last week. at the very least this code is looking for backend names when it should be looking for device names.

FWIW i have a light preference towards using UUIDs as names here just to reduce the possibility of mishandling user input somewhere in, like, `sled-agent`. `1` we have to answer either way since we're turning one Nexus-provided name into two different identifiers (device vs backend), and `2`.. in a UUID-y world they probably _should_ be ephemeral-and-generated-on-start lest we have a category of device whose ID we _expect_ to match across all Propolis, versus the category we expect will always be unique globally.

i might have a stronger opinion about names here on the other end of this PR :) but until then i'm gonna put a pin in that thought.

phd-tests/tests/src/boot_order.rs
```diff
@@ -0,0 +1,673 @@
...
+// https://github.com/tianocore/edk2/blob/712797c/OvmfPkg/OvmfPkgIa32.fdf#L181
+// introduced in
+// https://github.com/tianocore/edk2/commit/16f26de663967b5a64140b6abba2c145ea50194c, note this
+// is the DXEFV entry.
+//
+// The *files* we'll care about in this test are identified by other GUIDs in the above *volume*.
+//
+// EFI Internal Shell: https://github.com/tianocore/edk2/blob/a445e1a/ShellPkg/ShellPkg.dec#L59-L60
+// UiApp:
+// https://github.com/tianocore/edk2/blob/a445e1a/MdeModulePkg/Application/UiApp/UiApp.inf#L13
```
iximeow, line 51: i was initially concerned went overboard here, but this turned out a very useful detail to be checking. i feel much better about this test version of the test now than i did when starting to write it last week :sweat_smile:

phd-tests/tests/src/boot_order.rs
```diff
@@ -0,0 +1,673 @@
...
+
+async fn run_long_command(
+    vm: &phd_framework::TestVm,
+    cmd: &str,
+) -> Result<String, Error> {
+    // Ok, this is a bit whacky: something about the line wrapping for long commands causes
+    // `run_shell_command` to hang instead of ever actually seeing a response prompt.
+    //
+    // I haven't gone and debugged that; instead, chunk the input command up into segments short
+    // enough to not wrap when input, put them all in a file, then run the file.
```
iximeow, line 297: @gjcolombo i figure i'll open an issue about this with phd and reference here, but leave this as a workaround for the time being. i couldn't exactly figure out where line wrapping was causing issues but judging from how Alpine and Ubuntu look over serial to my eyes i can certainly understand this being... difficult to automate robustly.

phd-tests/tests/src/boot_order.rs
```diff
@@ -0,0 +1,673 @@
...
+    // up will be "0004,0001,FFF0,FFFF,0003,0000".
+    //
+    // This almost makes sense, but with other devices in the mix I've seen reorderings like
+    // `0004,0001,<PCI 16.0>,0003,0000,FFF0,FFFF` turning into
+    // `0004,0001,FFF0,FFFF,<PCI 16.0>,0003,0000`. This is particularly strange in that the new
+    // options were reordered around some other PCI device. It's not the boot order we set!
+    //
+    // So, to at least confirm we *can* modify the boot order in a stable way, make a somewhat less
+    // ambitious change: insert the duplicate boot options in the order directly after the option
+    // they are duplicates of. This seems to not get reordered.
```
iximeow, line 625: i cannot make sense of this, but it doesn't seem to be a bug resulting from anything _we're_ doing. i'd describe the overall behavior as "kind of works, and probably not something to debug in Propolis, maybe something to better understand in OVMF"..

phd-tests/tests/src/boot_order.rs
```diff
@@ -0,0 +1,673 @@
...
+    let hex = run_long_command(vm, &cmd).await?;
+
+    Ok(unhex(&hex))
+}
+
+/// Write the provided bytes to the EFI variable `varname`.
+///
+/// For Linux guests: variables automatically have their prior attributes prepended. Provide only
+/// the variable's data.
+async fn write_efivar(
```
iximeow, line 339: boring but delayed realization: writing to NvVars only actually matters if the guest OS has an EFI partition to persist to. Alpine virt images, for example, do not, so these changes are always lost on reboot. The "guest can reconfigure boot order" test will fail because of this, and rightfully so because the guest image simply does not support reconfiguring the boot order via EFI.

i think that's all basically fine, since we'd expect Alpine to basically be fine for other functionality, but now i'm curious what CI thinks of this..

iximeow (commented):
somewhat closer to landing, almost all the changes here are to the new `boot_order.rs` tests as i've been checking that guests behave in a useful way. i've one last test i want to write on that front: what does a guest see if we configure a boot order via EFI variable, stop the VM, configure a different boot order via fw_cfg, then restart a VM?

phd-tests/tests/src/boot_order.rs
gjcolombo: My initial wild guess would be that the guest is inserting line control characters when it thinks a line has wrapped, which breaks `run_shell_command`'s logic for determining whether the input command was successfully echoed.

Enabling trace-level logging will make the serial buffering code dump its buffer whenever it checks to see if a wait-for-text request has been satisfied. Verbose as all get out, but should give us some idea why it doesn't think the waited-for text has been seen.

phd-tests/tests/src/boot_order.rs
gjcolombo: FWIW there's a `guest_os_has_read_only_fs` helper method on `TestVm` that's meant to help with cases like this--see `crucible::smoke::shutdown_persistence_test` for an example. (Of course, in that case you've just ended up not running the test, which makes CI pass but doesn't add much else of value.)

phd-tests/tests/src/boot_order.rs
iximeow: > We could definitely revisit that strategy, but I think my vision for this particular test was that it would instantiate multiple disks backed by the chosen guest OS artifact, and then try to figure out which one actually got booted from.

yup! that's where i ended up. "figuring out which one actually got booted from" was not immediately obvious, but parsing the EFI variables isn't too bad and gets us to the right spot. i believe none of the tests depend on anything about the guest image (other than these `boot_order.rs` tests all heavily depending on being able to read and write EFI variables)

phd-tests/framework/src/test_vm/config.rs
```diff
@@ -119,14 +123,29 @@ impl<'dr> VmConfig<'dr> {
...
+        // XXX: i thought that the implicit order would be *pci device order*, not disk creation
+        // order?
+        // XXX: figure out what to do if there is both an explicit boot order and a call to
+        // `.boot_disk`.
+
+        self.disks[0] = DiskRequest {
+            name: "boot-disk",
             interface,
             backend,
             source: DiskSource::Artifact(artifact),
```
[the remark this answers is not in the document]
iximeow: > the framework could also do something else by default, e.g. mark the first attached disk as the boot disk and allow tests to clear that designation or set it to something else. This might actually be a little more natural for test authors, now that I think about it; one of my design principles for PHD is to make tests as easy to write as possible, so we should do whatever we think will provide the least friction for test authors here.

i think this is effectively in line with the status quo as well: `.boot_disk` is _probably_ going to be the first PCI device, so it ends up the boot disk in all the tests anyway. it'll also match what i expect to end up being the control plane's configuration. writing tests for multi-device `fwcfg`-based boot orders has been.. informative.

phd-tests/tests/src/boot_order.rs
luqmana: I think this is because of ovmf trying to reconcile the EFI boot order with what's in the qemu fwcfg - https://github.com/oxidecomputer/edk2/blob/edd72391086a9a297bf9aa011fe53b449165e556/OvmfPkg/Library/QemuBootOrderLib/QemuBootOrderLib.c#L1911

luqmana: Ah, basically what Greg had [called out](https://github.com/oxidecomputer/rfd/pull/751#discussion_r1617885908) on the RFD

iximeow: yeah exactly. the specifically confusing thing about this, to me, is that the duplicate FFF0 and FFFF entries remain in the boot order at all. i would expect them to get dropped late in `SetBootOrderFromQemu`, when it calls into `BootOrderComplete` and gets [here](https://github.com/oxidecomputer/edk2/blob/propolis/edk2-stable202105/OvmfPkg/Library/QemuBootOrderLib/QemuBootOrderLib.c#L1784-L1789).

between this and fwcfg overriding guest-configured boot order generally, i'm leaning towards actual configurations Propolis receives only naming a single boot device. if any of this changes because of a change in OVMF at some point, i want the test to fail as an indicator that Something Changed, but it feels like we shouldn't be relying on specific semantics here

phd-tests/tests/src/boot_order.rs
iximeow: that'd need a related `fs_is_read_readonly()` helper or something, since EFI variables are on a different filesystem, which would be fine.. but the Alpine image that doesn't seem to persist NvVars _also_ has `/sys/firmware/efi/efivars` mounted rw. i'm not confident enough to assert this yet, but i suspect somewhere in `efivarfs` the actual persistence to disk is silently lost..?

openapi/propolis-server.json
```diff
@@ -662,6 +662,13 @@
             "additionalProperties": {
               "$ref": "#/components/schemas/StorageDeviceV0"
             }
+          },
+          "boot_order": {
+            "nullable": true,
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/BootDeclaration"
```
iximeow, line 670: @gjcolombo my current thinking is that we want to keep this as an array of structs for forward compatibility reasons, though that probably matters a lot less between Nexus and Propolis. wdyt? should it just be an array of string (or single string for the time being)

phd-tests/framework/src/guest_os/ubuntu24_04.rs
```diff
@@ -0,0 +1,28 @@
...
+        CommandSequence(vec![
+            CommandSequenceEntry::wait_for("ubuntu login: "),
+            CommandSequenceEntry::write_str("root"),
+            CommandSequenceEntry::wait_for(self.get_shell_prompt()),
+        ])
+    }
+
+    fn get_shell_prompt(&self) -> &'static str {
+        "root@ubuntu:~#"
+    }
```
iximeow, line 23: this was useful for my inconsistently-configured 24.04 test image, but probably not how any other 24.04 test image would be set up. or at least, it's different from 22.04 (which does not expect to be root). so i'll drop this from the PR and make sure we've got something like "how to get or assemble test images for PHD" written up.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -1069,6 +1076,123 @@ impl<'a> MachineInitializer<'a> {
...
+                &vnic_spec.backend_name
+            };
+
+            if let Some(device_spec) = self
+                .spec
+                .devices
+                .storage_devices
+                .iter()
+                .find_map(|(_name, spec)| {
+                    eprintln!("checking name \"{}\", spec {:?}", _name, spec);
```
hawkw, line 1139: should this `eprintln!' either be a log message or be removed?

phd-tests/framework/src/guest_os/ubuntu24_04.rs
hawkw: is 24.04 actually necessary for any particular reason? IIRC on CI we run these tests against an Alpine image, but if we need to start running certain tests against Ubuntu 24.04, presumably we can start doing that as well?

phd-tests/tests/src/boot_order/efi_utils.rs
```diff
@@ -0,0 +1,490 @@
...
+                );
+            }
+        }
+    }
+}
+
+impl EfiLoadOption {
+    // parsing here brought to you by rereading
+    // * https://uefi.org/specs/UEFI/2.10/10_Protocols_Device_Path_Protocol.html
+    // * https://uefi.org/specs/UEFI/2.10/03_Boot_Manager.html
```
hawkw, line 204: <3 thanks for including the references :)

phd-tests/tests/src/boot_order/efi_utils.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        // The `Description` field is a null-terminated string of char16.
+        let mut description_chars: Vec<u16> = Vec::new();
+
+        loop {
+            let c = bytes.read_u16::<LittleEndian>()?;
+            description_chars.push(c);
+            if c == 0 {
+                break;
+            }
```
hawkw, line 219: Huh, this appears to include the null terminator in the Rust string we turn this into, is that actually the desired behavior?

iximeow: ah, no, it's not really. i'd collected bytes and decided to turn them into `String` much later, this was an oversight along the way. good catch

phd-tests/framework/src/guest_os/ubuntu24_04.rs
iximeow: not at this point, this is partially an artifact of at one point using `efibootmgr` (which is not in the base Alpine image) for configuration.

i _did_ notice that `guest_can_adjust_boot_order` does not pass with the Alpine CI image, but does with my 24.04 image. i'm not _exactly_ sure why yet, but i don't think it works out in the direction of "need 24.04 in CI".

bin/propolis-server/src/lib/initializer.rs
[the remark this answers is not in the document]
iximeow: (i've walked back `first_boot_only` for this first PR. i _did_ look at wiring this up through `Lifecycle::reset` and that seems pretty workable, just not necessary right now)

phd-tests/framework/src/guest_os/ubuntu24_04.rs
gjcolombo: > make sure we've got something like "how to get or assemble test images for PHD" written up

I'd almost forgotten about it, but a ways back I created the [guest-os-image-configs repo](https://github.com/oxidecomputer/guest-os-image-configs) as a place to put stuff like this.

lib/propolis-client/src/instance_spec.rs
```diff
@@ -184,6 +184,35 @@ impl SpecBuilderV0 {
         }
     }
 
+    /// Sets a boot order. Names here refer to devices included in this spec.
+    ///
+    /// Permissible to not this if the implicit boot order is desired, but the implicit boot order
+    /// may be unstable across device addition and removal.
+    ///
+    /// XXX: talk about what happens if names are included that do not name real devices..?
```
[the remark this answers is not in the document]
iximeow: ok! now that the Propolis behavior is to return an error i've adjusted this comment, just for the sake of the docs being accurate. sufficiently scruple'd for the remainder of its existence.

phd-tests/tests/src/boot_order/efi_utils.rs
iximeow: (fixed in https://github.com/oxidecomputer/propolis/pull/756/commits/ed455ea695be0b6b03a7e39212ae6ec5b84b9fdc#diff-2a37b514b91da3540f6c06fa43cd50e589c319a328461ab05a7ea999a7bc162bL216)

bin/propolis-server/src/lib/initializer.rs
iximeow: `trace!` if it was going to stay, but i really don't see the value in it beyond my initial checking that it works at all, so it's removed

crates/propolis-api-types/src/instance_spec/v0/mod.rs
```diff
@@ -109,6 +109,11 @@ pub struct DeviceSpecV0 {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub qemu_pvpanic: Option<components::devices::QemuPvpanic>,
 
+    // same as above
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub boot_order: Option<Vec<BootDeclaration>>,
```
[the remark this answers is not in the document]
iximeow: oh, done as of https://github.com/oxidecomputer/propolis/pull/756/commits/ed455ea695be0b6b03a7e39212ae6ec5b84b9fdc

iximeow:
from bmat:
```
4349  2024-09-16T23:34:32.081Z  failures:
4350  2024-09-16T23:34:32.081Z      phd_tests::boot_order::boot_order_source_priority
4351  2024-09-16T23:34:32.081Z      phd_tests::boot_order::guest_can_adjust_boot_order
4352  2024-09-16T23:34:32.081Z      phd_tests::boot_order::unbootable_disk_skipped
4353  2024-09-16T23:34:32.081Z      phd_tests::boot_order::configurable_boot_order
```

this is somewhat surprising. i expected `guest_can_adjust_boot_order` to fail for  the same reason as mentioned [here](https://github.com/oxidecomputer/propolis/pull/756#discussion_r1757538536) but the rest all passed locally with the same Alpine image..

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -994,6 +1001,121 @@ impl<'a> MachineInitializer<'a> {
...
+
+                order.add_pci(bdf.location, "ethernet");
+            } else {
+                eprintln!(
+                    "could not find {:?} in {:?} or {:?}",
+                    boot_entry,
+                    self.spec.devices.storage_devices,
+                    self.spec.devices.network_devices
+                );
+                panic!("TODO: return an error; the device name doesn't exist?");
```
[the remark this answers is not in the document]
iximeow: (yes, this is forbidden in Nexus now. the attendant queries aren't too bad, in fact!)

openapi/propolis-server.json
iximeow: went with pretty much this, couldn't think of any downsides :) https://github.com/oxidecomputer/propolis/pull/756/commits/6a386b0806a9b5dcba6924619ac1afc541bc8715

phd-tests/tests/src/boot_order.rs
iximeow: lol: https://github.com/oxidecomputer/propolis/pull/756/commits/bcb682450ef362a98b5cb00025b392057dfaa447#diff-3a20a9ed692f02614280c3438c2e7368597cb587487dcc28be5c6bac8affc549R321-R340

the Alpine ISO 9660-backed virt image, for reference, looks like this:
```
localhost:~# mount | grep efivarfs 
efivarfs on /sys/firmware/efi/efivars type efivarfs (rw,nosuid,nodev,noexec,relatime)
```
note it even says `rw`!! i think this is because technically the filesystem could be reasonably written to even without a writable ESP - it _could_ be configuring real non-volatile storage. just, it's not here. it's unfortunate the write gets lost as a result, but it is what it is.

phd-tests/framework/src/test_vm/config.rs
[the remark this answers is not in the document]
iximeow: took this mostly along the lines of your third-and-a-half point: https://github.com/oxidecomputer/propolis/pull/756/commits/617dd0c870ce382f3208d1b5f42a6eb1c94df5bb - "boot_device()" replaces an existing disk by the same name, if one is there, and puts the new disk at the front of a boot order, creating one if there isn't one. this means that all tests by default have a boot order which will have at least `boot-disk` at the front.

`fn boot_order()` still exists to override this for the `boot_order.rs` tests that need to set it more specifically, and `fn clear_boot_order()` for cases where we really didn't want the default behavior. but i think this makes things better for the median test and gives us the right tools when we need them.

bin/propolis-server/src/lib/initializer.rs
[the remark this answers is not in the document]
iximeow: i have stronger opinions! as promised!

* if we use user-provided names anywhere, it feels much more intuitive for that to be a name visible in guests, rather than our own name for the Crucible volume backing a device by some name we invented. so i think we ought to flip `device_name` and `backend_name`.
* it'd be really nice to just get IDs from Nexus and not have to think about user-provided strings, lets figure out what needs doing for that.
* implicit component IDs probably should be defined by `propolis-server` unless we'd also have a runtime decoder ring of which ephemeral ID means what device for a given VM..

phd-tests/tests/src/boot_order.rs
iximeow: (gonna file an issue for this and look into it ~soon)

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -994,6 +1001,121 @@ impl<'a> MachineInitializer<'a> {
...
+            "generating bootorder with order: {:?}",
+            self.boot_order.as_ref()
+        );
+        let Some(boot_names) = self.boot_order.as_ref() else {
+            return Ok(None);
+        };
+
+        let mut order = fwcfg::formats::BootOrder::new();
+
+        let parse_bdf =
```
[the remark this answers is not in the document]
iximeow: rather than unwrapping here, i'd rather make `PciPath` implement `Into<Bdf>` (instead of `TryInto`), and pair it with an exhaustive test the the conversion is OK. we're doing the same checks in different words, for the two types. i'll do that as another followup.

phd-tests/tests/src/boot_order.rs
```diff
@@ -0,0 +1,533 @@
...
+    // Check for this condition and exit early if the guest OS configuration
+    // will not let us perform a useful test.
+    write_efivar(&vm, &bootvar(0xfff0), &boot_option_bytes).await?;
+    let reread = read_efivar(&vm, &bootvar(0xfff0)).await?;
+    if reread.is_empty() {
+        warn!(
+            "guest environment drops EFI variable writes! \
+            exiting test WITHOUT VALIDATING ANYTHING"
+        );
+        return Ok(());
```
gjcolombo, line 360: I might `phd_skip!` in this case so that the "skipped" bucket gets incremented instead of the "passed" bucket in the runner output.

phd-tests/framework/src/test_vm/config.rs
```diff
@@ -172,7 +207,29 @@ impl<'dr> VmConfig<'dr> {
             })
             .context("serializing Propolis server config")?;
 
-        let DiskSource::Artifact(boot_artifact) = self.boot_disk.source else {
+        // The first disk in the boot list might not be the disk a test *actually* expects to boot.
+        //
+        // If there are multiple bootable disks in the boot order, we'll assume they're all
+        // the same guest OS kind. So look for `boot-disk` - if there isn't a disk named
+        // `boot-disk` then fall back to hoping the first disk in the boot orrder is a bootable
```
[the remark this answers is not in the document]
iximeow: ow, ty

phd-tests/tests/src/boot_order.rs
iximeow: nice! yes, that's much better.

phd-tests/tests/src/boot_order.rs
```diff
@@ -0,0 +1,533 @@
...
+
+    let boot_num_bytes = read_efivar(&vm, BOOT_CURRENT_VAR).await?;
+
+    let boot_num: u16 = u16::from_le_bytes(boot_num_bytes.try_into().unwrap());
+
+    let boot_option_bytes = read_efivar(&vm, &bootvar(boot_num)).await?;
+
+    let mut cursor = Cursor::new(boot_option_bytes.as_slice());
+
+    let load_option = EfiLoadOption::parse_from(&mut cursor).unwrap();
```
hawkw, line 176: i wonder if this ought to be a function, 'cause it also shows up twice in the `configurable_boot_order` test?

phd-tests/tests/src/boot_order.rs
```diff
@@ -0,0 +1,533 @@
...
+    EDK2_FIRMWARE_VOL_GUID, EDK2_UI_APP_GUID,
+};
+
+pub(crate) async fn run_long_command(
+    vm: &phd_framework::TestVm,
+    cmd: &str,
+) -> Result<String, Error> {
+    // Ok, this is a bit whacky: something about the line wrapping for long
+    // commands causes `run_shell_command` to hang instead of ever actually
+    // seeing a response prompt.
```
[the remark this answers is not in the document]
iximeow: aha, that's a good approach too. so, that works because no individual line is long enough that we're never getting an unexpected `\n` back from the other end. it looks like the shell script also has no individual terribly-long lines, so that works.

i think this workaround is something other tests _could_ use, but i'd rather the serial listener so that it's not needed (destined to become a new issue in a bit here)

bin/propolis-server/src/lib/spec/builder.rs
```diff
@@ -110,6 +113,36 @@ impl SpecBuilder {
...
+
+        let is_disk =
+            self.spec.disks.values().any(|disk| match &disk.device_spec {
+                StorageDevice::Virtio(VirtioDisk { backend_name, .. }) => {
+                    backend_name == item.name.as_str()
+                }
+                StorageDevice::Nvme(NvmeDisk { backend_name, .. }) => {
+                    backend_name == item.name.as_str()
+                }
+            });
```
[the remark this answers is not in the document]
iximeow: yeah so it turns out the reason i thought so strongly that this matching is correct is because.. it was _for phd_. it was, of course, incorrect for disk requests as parsed from Nexus. this just matching on the `device_name` as of https://github.com/oxidecomputer/propolis/pull/756/commits/ba459299fac1e58cbbc88c5b1e589f80d65fbd9a, which also swaps how the disk name is used to match `parse_disk_from_request` on both `device_name` and `backend_name`.

in the process the gross matching is now gone. and it actually works now. yay :)

phd-tests/framework/src/test_vm/config.rs
```diff
@@ -119,31 +122,63 @@ impl<'dr> VmConfig<'dr> {
...
+        self.boot_order = Some(disks);
+        self
+    }
+
+    pub fn clear_boot_order(&mut self) -> &mut Self {
+        self.boot_order = None;
+        self
+    }
+
+    /// Add a new disk to the VM config, and add it to the front of the VM's boot order.
```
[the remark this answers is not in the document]
iximeow: i've gotta figure out how to make `set tw=80` and `set cc=80` live in a contextually-relevant place so i stop making Long Lines..

phd-tests/framework/src/test_vm/config.rs
```diff
@@ -172,7 +207,29 @@ impl<'dr> VmConfig<'dr> {
...
+        // TODO: theoretically we might want to accept configuration of a specific guest OS adapter
+        // and avoid the guessing games. So far the above supports existing tests and makes them
+        // "Just Work", but a more complicated test may want more control here.
+        let boot_disk =
+            self.disks.iter().find(|d| d.name == "boot-disk")
+                .or_else(|| if let Some(boot_order) = self.boot_order.as_ref() {
+                    boot_order.first().and_then(|name| self.disks.iter().find(|d| &d.name == name))
+                } else {
+                    None
+                })
```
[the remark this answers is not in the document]
iximeow: rehashing discussion elsewhere: it is actually important that the explicit boot order be ignored, since we now have tests that put a non-bootable disk up at the front of the list and expect to fall through.

if we had better support for one test having different OS kinds across its disks, it might make more sense to have an explicit setter for the guest OS kind and stop doing the guessing here. but that adds its own complication and even with the `boot_order` tests i've managed to avoid needing it.

that leaves this in the unfortunate limbo of "local minima of weird", i think

bin/propolis-cli/src/main.rs
```diff
@@ -517,6 +518,9 @@ async fn migrate_instance(
         // TODO: Handle migrating NICs
         nics: vec![],
         disks,
+        // TODO: Handle retaining boot settings? Or extant boot settings
+        // forwarded along outside InstanceEnsure anyway.
```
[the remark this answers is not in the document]
iximeow: this is something i was going to ask about actually, i'm not sure what `propolis-cli` is _for_ or exactly what the consequences of TODO'ing this are. ... but also `nics` is a similar TODO and losing networking as a result of migration seems a lot worse (if `propolis-cli` is just for our own testing, then i'm inclined to fix it separately from landing the change, given that)

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -1001,6 +1005,82 @@ impl<'a> MachineInitializer<'a> {
...
+                        // of 0, so do that here too. is that.. ok?
+                        order.add_nvme(bdf.location, 0);
+                    }
+                };
+            } else {
+                let message = format!(
+                    "Instance spec included boot entry which does not refer to an existing disk: `{}`",
+                    boot_entry.name.as_str(),
+                );
+                return Err(Error::new(ErrorKind::Other, message));
```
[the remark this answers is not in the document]
iximeow: sure, and agreed (also the string above it is wrapped reasonably now :grimacing: )

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -1001,6 +1005,82 @@ impl<'a> MachineInitializer<'a> {
...
+                    StorageDevice::Virtio(disk) => {
+                        disk.backend_name.as_str() == boot_entry.name.as_str()
+                    }
+                })
+                .map(|(_name, spec)| spec)
+            {
+                match &spec.device_spec {
+                    StorageDevice::Virtio(disk) => {
+                        let bdf = parse_bdf(&disk.pci_path)?;
+                        // TODO: check that bus is 0. only support boot devices
```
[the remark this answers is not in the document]
iximeow: yeah.. done as part of https://github.com/oxidecomputer/propolis/pull/756/commits/22b5a08ae09a1fb9fd8e7e9d76f9475a71f5b6cc

phd-tests/framework/src/test_vm/config.rs
```diff
@@ -257,10 +257,14 @@ impl<'dr> VmConfig<'dr> {
...
-            let device_name = format!("disk-device{}", idx);
+        for (req, hdl) in all_disks {
             let pci_path = PciPath::new(0, req.pci_device_num, 0).unwrap();
-            let (backend_name, backend_spec) = hdl.backend_spec();
+            let (device_name, backend_spec) = hdl.backend_spec();
+            // propolis-server uses a disk's name from the API as its device
+            // name. It then derives a backend name from the disk name. Match
+            // that name derivation here so PHD tests match the API as used by
+            // Nexus.
+            let backend_name = format!("{}-backend", device_name);
```
gjcolombo, line 276: This also requires an update to `VmSpec::refresh_crucible_backends` so that it can find the correct backend name for each disk handle in the resulting spec. (This is the cause of the most recent commits' CI failures--the Crucible migration tests are failing because the framework isn't correctly bumping the client generation numbers on the disks, which is happening because the backend elements in the instance spec don't have the names `refresh_crucible_backends` expects.)

I would probably plumb all this a bit differently so that disk objects return the backend names we expect:

- have separate `DiskDeviceName` and `DiskBackendName` newtypes, with an associated function on the former that generates the latter
- make `DiskFactory`'s functions take `DiskDeviceName`s instead of bare `String`s
- make `DiskConfig::backend_spec` return a `DiskBackendName`
- do whatever plumbing is needed in DiskFactory or its callees to do the correct device -> backend name conversion

iximeow: > cause of the most recent commits' CI failures

ah, i trying to figure out caused that outcome. ok, that all makes some sense

iximeow: rereading this, i see it's what you meant the entire time, but i ended up at https://github.com/oxidecomputer/propolis/pull/756/commits/3a3e753ca28a383f48eea33673f339f835a98546 - i'd like to just have whatever relation between `DeviceName`/`BackendName` be _the_ function between them instead of simply matching what `parse_disk_from_request` does, but i think that's much more invasive than is appropriate here..

hawkw:
you love to see it :tada:

bin/propolis-cli/src/main.rs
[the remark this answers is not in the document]
hawkw: seems reasonable...i don't know if i've ever actually used the CLI for anything, so i don't personally care about it! :)

bin/propolis-server/src/lib/initializer.rs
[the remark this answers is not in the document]
iximeow: (digested as https://github.com/oxidecomputer/propolis/issues/772)

PR #757, opened 2024-09-03
instance spec rework: remove most dependencies on `InstanceSpecV0` from propolis-server

gjcolombo:
(Part 6/X in the instance spec rework; see #735.)

Define propolis-server's internal representation of a VM component manifest (the `Spec` type) and switch almost everyone in the server over to it. (The exception is the live migration preamble; this will be dealt with in a future PR that moves migration compatibility checks from api-types to server.)

- Define the internal `spec::Spec` type and some helper types, like internal representations of a `Disk` and a `Nic`.
  - This PR starts to show some of the benefits of this series of changes (finally!): now a regular guest VNIC, which is only ever supposed to have a viona backend, is guaranteed by construction to have one, so the machine initializer doesn't need to handle the "oops wait you specified an invalid backend type here" case.
- Add conversion functions to/from the `InstanceSpecV0` public-API type. Converting to a versioned spec is infallible, but conversion from a versioned spec (which can be well-formed but not valid) is fallible.
- Represent Falcon devices with a little more care:
  - The parsed SoftNPU device collection now uses `Option`s for components (the main PCI device, the P9 device, and the P9 filesystem adapter) with a maximum cardinality of 1.
  - When enabling SoftNPU, explicitly denote in the instance spec that SoftNPU expects to use COM4 instead of excluding it from the spec entirely.

Known warts in this change that are on deck to be fixed later in this series:

* In the new internal spec type, serial devices don't have names yet (they still use port numbers as a key and convert these to implicit names).
* The new `Disk` and `Nic` types duplicate backend names: both the types themselves and their `device_spec` members contain `backend_name` fields, and these need to be identical or hilarity will ensue. Subsequent changes will remove the `backend_name` field from these types and just rely on the backend name that's in the device spec.

Tests: cargo test, PHD. (It probably wouldn't hurt for me to run this in an Omicron dev cluster, either. I can give that a try before merging.)

### DIFF
--- a/bin/propolis-server/src/lib/spec/api_spec_v0.rs
+++ b/bin/propolis-server/src/lib/spec/api_spec_v0.rs
@@ -7,14 +7,12 @@
 
 use propolis_api_types::instance_spec::{
     components::devices::{SerialPort as SerialPortSpec, SerialPortNumber},
-    v0::{InstanceSpecV0, NetworkDeviceV0, StorageDeviceV0},
+    v0::{InstanceSpecV0, NetworkBackendV0, NetworkDeviceV0, StorageDeviceV0},
 };
 use thiserror::Error;
 
 #[cfg(feature = "falcon")]
-use propolis_api_types::instance_spec::{
-    components::devices::SoftNpuPort as SoftNpuPortSpec, v0::NetworkBackendV0,
-};
+use propolis_api_types::instance_spec::components::devices::SoftNpuPort as SoftNpuPortSpec;
 
 #[cfg(feature = "falcon")]
 use crate::spec::SoftNpuPort;
@@ -34,6 +32,9 @@ pub(crate) enum ApiSpecParseError {
 
     #[error("backend {0} not used by any device")]
     BackendNotUsed(String),
+
+    #[error("network backend for guest NIC {0} is not a viona backend")]
+    GuestNicInvalidBackend(String),
 
     #[cfg(feature = "falcon")]
     #[error("network backend for device {0} is not a DLPI backend")]
@@ -69,10 +70,10 @@ impl From<Spec> for InstanceSpecV0 {
 
             assert!(_old.is_none());
 
-            let _old = spec
-                .backends
-                .network_backends
-                .insert(nic.backend_name, nic.backend_spec);
+            let _old = spec.backends.network_backends.insert(
+                nic.backend_name,
+                NetworkBackendV0::Virtio(nic.backend_spec),
+            );
 
             assert!(_old.is_none());
         }
@@ -183,6 +184,12 @@ impl TryFrom<InstanceSpecV0> for Spec {
                         device_name.clone(),
                     )
                 })?;
+
+            let NetworkBackendV0::Virtio(backend_spec) = backend_spec else {
+                return Err(ApiSpecParseError::GuestNicInvalidBackend(
+                    device_name,
+                ));
+            };
 
             builder.add_network_device(Nic {
                 device_name,
--- a/bin/propolis-server/src/lib/spec/api_spec_v0.rs
+++ b/bin/propolis-server/src/lib/spec/api_spec_v0.rs
@@ -44,38 +44,43 @@ pub(crate) enum ApiSpecError {
 impl From<Spec> for InstanceSpecV0 {
     fn from(val: Spec) -> Self {
         let mut spec = InstanceSpecV0::default();
-
         spec.devices.board = val.board;
+
+        // Internal instance specs (the inputs to this function) should assign
+        // a unique name to each component they describe. This invariant is
+        // enforced by the spec builder. Since component names are globally
+        // unique, they should never collide when inserted into an API spec's
+        // device and backend maps.
         for (disk_name, disk) in val.disks {
-            let _old = spec
+            let old = spec
                 .devices
                 .storage_devices
                 .insert(disk_name, disk.device_spec.into());
 
-            assert!(_old.is_none());
+            assert!(old.is_none(), "{old:?}");
 
-            let _old = spec
+            let old = spec
                 .backends
                 .storage_backends
                 .insert(disk.backend_name, disk.backend_spec.into());
 
-            assert!(_old.is_none());
+            assert!(old.is_none(), "{old:?}");
         }
 
         for (nic_name, nic) in val.nics {
-            let _old = spec
+            let old = spec
                 .devices
                 .network_devices
                 .insert(nic_name, NetworkDeviceV0::VirtioNic(nic.device_spec));
 
-            assert!(_old.is_none());
+            assert!(old.is_none(), "{old:?}");
 
-            let _old = spec.backends.network_backends.insert(
+            let old = spec.backends.network_backends.insert(
                 nic.backend_name,
                 NetworkBackendV0::Virtio(nic.backend_spec),
             );
 
-            assert!(_old.is_none());
+            assert!(old.is_none(), "{old:?}");
         }
 
         for (num, user) in val.serial.iter() {
@@ -87,18 +92,18 @@ impl From<Spec> for InstanceSpecV0 {
                     SerialPortNumber::Com4 => "com4",
                 };
 
-                let _old = spec
+                let old = spec
                     .devices
                     .serial_ports
                     .insert(name.to_owned(), SerialPortDesc { num: *num });
 
-                assert!(_old.is_none());
+                assert!(old.is_none(), "{old:?}");
             }
         }
 
         for (bridge_name, bridge) in val.pci_pci_bridges {
-            let _old = spec.devices.pci_pci_bridges.insert(bridge_name, bridge);
-            assert!(_old.is_none());
+            let old = spec.devices.pci_pci_bridges.insert(bridge_name, bridge);
+            assert!(old.is_none(), "{old:?}");
         }
 
         spec.devices.qemu_pvpanic = val.pvpanic.map(|pvpanic| pvpanic.spec);
@@ -109,7 +114,7 @@ impl From<Spec> for InstanceSpecV0 {
             spec.devices.softnpu_p9 = val.softnpu.p9_device;
             spec.devices.p9fs = val.softnpu.p9fs;
             for (port_name, port) in val.softnpu.ports {
-                let _old = spec.devices.softnpu_ports.insert(
+                let old = spec.devices.softnpu_ports.insert(
                     port_name.clone(),
                     SoftNpuPortSpec {
                         name: port_name,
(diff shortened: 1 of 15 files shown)


bin/propolis-server/src/lib/spec/mod.rs
```diff
@@ -2,44 +2,196 @@
...
+//! Instance specs describe how to configure a VM and what components it has.
+//!
+//! This module defines an "internal" spec type for the server's instance
+//! initialization code to use. This spec type is not `Serialize` and is not
+//! meant to be sent over the wire in API requests or the migration protocol.
+//! This allows the internal representation to change freely between versions
+//! (so long as it can consistently be converted to and from the wire-format
+//! spec in the [`propolis_api_types`] crate); this, in turn, allows this
+//! representation to take forms that might otherwise be hard to change in a
+//! backward-compatible way.
```
iximeow, line 15: this is definitely something i'd been thinking i was missing in the last week or so. glad to see it!

the paragraph here points heavily towards `struct Spec`, but it's largely applicable other types as well, like `StorageDevice`. what do you think about changing the mention of
> an "internal" spec type ...

to something like

> "internal" spec types ... which are used elsewhere via an enclosing [\`Spec\`]"

?

seems worth emphasizing that the properties here are intended for all members of this module, but that `Spec` really is special in being where users will get references to anything else in here.

iximeow (approved):

bin/propolis-server/src/lib/spec/mod.rs
gjcolombo: Good call. I've updated the prose here (and fixed up a couple additional doc comments) in 87324c2.

bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -0,0 +1,259 @@
...
+#[cfg(feature = "falcon")]
+use crate::spec::SoftNpuPort;
+
+use super::{
+    builder::{SpecBuilder, SpecBuilderError},
+    Disk, Nic, QemuPvpanic, SerialPortUser, Spec,
+};
+
+#[derive(Debug, Error)]
+pub(crate) enum ApiSpecParseError {
```
hawkw, line 26: suuuuuuper tiny, unimportant nit: most of these variants don't _really_ seem like "parse errors" --- referencing a backend that doesn't exist isn't a syntactic error in the spec, it's a semantic error...I don't know if this actually matters and i'm probably just being pedantic...

bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    #[error("backend {0} not used by any device")]
+    BackendNotUsed(String),
+
+    #[error("network backend for guest NIC {0} is not a viona backend")]
+    GuestNicInvalidBackend(String),
+
+    #[cfg(feature = "falcon")]
+    #[error("network backend for device {0} is not a DLPI backend")]
+    NotDlpiBackend(String),
```
hawkw, line 41: it might be nice if these indicated what the actual kind of backend we encountered was? but, not a big deal.

bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -0,0 +1,259 @@
...
+                .insert(nic_name, NetworkDeviceV0::VirtioNic(nic.device_spec));
+
+            assert!(_old.is_none());
+
+            let _old = spec.backends.network_backends.insert(
+                nic.backend_name,
+                NetworkBackendV0::Virtio(nic.backend_spec),
+            );
+
+            assert!(_old.is_none());
```
hawkw, line 78: tiny style nit: i don't think the `_` to suppress unused-variable warnings for `_old` is necessary here, since we use a regular `assert!` rather than a `debug_assert!` --- the value of `_old` actually *is* always used.
```suggestion
            let old = spec
                .devices
                .storage_devices
                .insert(disk_name, disk.device_spec.into());

            assert!(old.is_none());

            let old = spec
                .backends
                .storage_backends
                .insert(disk.backend_name, disk.backend_spec.into());

            assert!(old.is_none());
        }

        for (nic_name, nic) in val.nics {
            let old = spec
                .devices
                .network_devices
                .insert(nic_name, NetworkDeviceV0::VirtioNic(nic.device_spec));

            assert!(old.is_none());

            let old = spec.backends.network_backends.insert(
                nic.backend_name,
                NetworkBackendV0::Virtio(nic.backend_spec),
            );

            assert!(old.is_none());
```

hawkw: Also, I was going to ask whether it was correct to assert here, or if a duplicate name should be a user-facing error...but, looking at the `Spec` type, these are all stored in `HashMap`s, so there should never be duplicate names in the `Spec`. Perhaps it's worth documenting that, either with a comment or (preferably) in the assertion's panic message? It's probably obvious to you as the author of this code, but as a reader, I had to go check the `Spec` type...

Not a big deal though.

bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -0,0 +1,259 @@
...
+                    .serial_ports
+                    .insert(name.to_owned(), SerialPortDesc { num: *num });
+
+                assert!(_old.is_none());
+            }
+        }
+
+        for (bridge_name, bridge) in val.pci_pci_bridges {
+            let _old = spec.devices.pci_pci_bridges.insert(bridge_name, bridge);
+            assert!(_old.is_none());
```
hawkw, line 101:
```suggestion
                let _old = spec
                    .devices
                    .serial_ports
                    .insert(name.to_owned(), SerialPortDesc { num: *num });

                assert!(old.is_none());
            }
        }

        for (bridge_name, bridge) in val.pci_pci_bridges {
            let old = spec.devices.pci_pci_bridges.insert(bridge_name, bridge);
            assert!(old.is_none());
```

bin/propolis-server/src/lib/spec/api_spec_v0.rs
hawkw: Also, maybe we should write `assert_eq!(old, None)` so that the panic message includes the value of `old`, and/or include the duplicated name? In practice, it doesn't really matter, because this shouldn't ever happen...

bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -0,0 +1,259 @@
...
+                );
+
+                assert!(_old.is_none());
+
+                let _old = spec.backends.network_backends.insert(
+                    port.backend_name,
+                    NetworkBackendV0::Dlpi(port.backend_spec),
+                );
+
+                assert!(_old.is_none());
```
hawkw, line 127:
```suggestion
                let _old = spec.devices.softnpu_ports.insert(
                    port_name.clone(),
                    SoftNpuPortSpec {
                        name: port_name,
                        backend_name: port.backend_name.clone(),
                    },
                );

                assert!(old.is_none());

                let old = spec.backends.network_backends.insert(
                    port.backend_name,
                    NetworkBackendV0::Dlpi(port.backend_spec),
                );

                assert!(old.is_none());
```

bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -0,0 +1,259 @@
...
+    #[error("network backend for device {0} is not a DLPI backend")]
+    NotDlpiBackend(String),
+}
+
+impl From<Spec> for InstanceSpecV0 {
+    fn from(val: Spec) -> Self {
+        let mut spec = InstanceSpecV0::default();
+
+        spec.devices.board = val.board;
+        for (disk_name, disk) in val.disks {
```
hawkw, line 49: it occurs to me just now that we might want to stick something like this here:
```suggestion
        spec.devices.storage_devices.reserve(val.disks.len());
        spec.devices.storage_backends.reserve(val.disks.len());
        for (disk_name, disk) in val.disks {
```
Since we know the number of devices up front, we may as well preallocate enough capacity for them, rather than repeatedly growing the collections as we insert more devices. On the other hand, this is probably not a huge deal --- the improved performance is probably marginal at best, and I'm sure constructing these specs isn't what dominates instance-creation time anyway.

hawkw: (if you do decide to do this, we could also do it for the other loops as well)

hawkw (approved):
Overall, this looks great. I had some very minor nits, but none of them are major issues!

bin/propolis-server/src/lib/spec/mod.rs
```diff
@@ -143,9 +143,10 @@ pub struct Nic {
     pub backend_spec: VirtioNetworkBackend,
 }
 
+/// A kind of device to install as the listener on a COM port.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum SerialPortUser {
-    Standard,
+pub enum SerialPortDevice {
+    Uart,
```
hawkw, line 149: This is much nicer! :+1:

bin/propolis-server/src/lib/spec/api_spec_v0.rs
gjcolombo: Took me a couple of tries to get this where I wanted it, but this should be fixed in 02be2a2.

bin/propolis-server/src/lib/spec/api_spec_v0.rs
gjcolombo: I've left this alone for now because these variants are going to go away in a couple of PRs. (At least as I have things drafted the semantics of these variants will change from from "this NIC has a backend but its type is wrong" to "this NIC doesn't have a backend of the appropriate type".)

bin/propolis-server/src/lib/spec/api_spec_v0.rs
gjcolombo: I've also left this one alone for now since this type is going to get a substantial makeover very soon.

bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -0,0 +1,273 @@
...
+
+impl From<Spec> for InstanceSpecV0 {
+    fn from(val: Spec) -> Self {
+        // Inserts a component entry into the supplied map, asserting first that
+        // the supplied key is not present in that map.
+        //
+        // This assertion is valid because internal instance specs should assign
+        // a unique name to each component they describe. The spec builder
+        // upholds this invariant at spec creation time.
+        fn insert_component<T>(
```
hawkw, line 54: nit: maybe worth sticking a 
```suggestion
        #[track_caller]
        fn insert_component<T>(
```

gjcolombo: whoaaaaa

PR #759, opened 2024-09-05
move smbios table creation to propolis (out of server+standalone)

iximeow:
tiny cleanup i noticed alongside #756: the two `generate_smbios()` were similar but not quite the same, and it wasn't entirely clear what external inputs _were_ to the tables either would generate. so, take the `SmbiosParameters` idea from `propolis-standalone` and make `propolis-server` put its parameters in a similar struct.

nice benefit: anything we make configurable for `propolis-standalone` is at least progress towards configurability for `propolis- server`.

in the process i've added a few more fields to `SmbiosParameters` for values that either varied or had different defaulting behavior depending on the application: `rom_release_date`, `cpuid_vendor`, and the value as we use table 1's serial number and uuid, which i'm calling `system_id` - it's the instance's ID and just left as defaults in `propolis-standalone`.


bin/propolis-standalone/src/main.rs
```diff
@@ -1277,20 +1152,43 @@ fn setup_instance(
...
 
     // generate SMBIOS data and expose via fw_cfg
-    let smbios::TableBytes { entry_point, structure_table } =
-        generate_smbios(SmbiosParams {
+    let smbios::TableBytes { entry_point, structure_table } = {
+        let cpuid_ident = cpuid_ident
+            .unwrap_or_else(|| cpuid::host_query(cpuid::Ident(0x1, None)));
+        let cpuid_procname = cpuid_procname.unwrap_or_else(|| {
+            if cpuid::host_query(cpuid::Ident(0x8000_0000, None)).eax
+                >= 0x8000_0004
```
iximeow, line 1160: boring but funny note: the oldest CPUs that don't support cpuid leaves 0x8000_000[1..4] are, i think, Pentium II and K5. so as long as Propolis is running on a CPU newer than those, this arm is always taken?

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -982,100 +942,48 @@ impl<'a> MachineInitializer<'a> {
...
+        smb_tables
+            .add(phys_mem_array_handle, &smbios_params.table_type16())
+            .unwrap();
+        smb_tables
+            .add(
+                0x1700.into(),
+                &smbios_params.table_type17(phys_mem_array_handle),
+            )
+            .unwrap();
+        smb_tables.add(0x3200.into(), &smbios_params.table_type32()).unwrap();
```
iximeow, line 976: @pfmooney how's this seem? my thinking is that the `SmbiosParams::table_*` are "this is probably a reasonable way to put the tables together", and if we find value in changing any of the defaults we'd hoist them into fields of `SmbiosParameters`.

PR #761, opened 2024-09-12
instance spec rework: tighten up component naming

gjcolombo:
(7/X in the instance spec rework; see #735.)

Fix a few problems with the way components are named in propolis-server `Spec`s:

- Don't duplicate backend names in disk and NIC elements
- Use names as keys in the serial port map instead of deriving names from serial port numbers
- Properly register the names of bridges when adding them to the spec builder
- Make the builder reject requests to add a device/backend pair where the device and backend have the same name (and stop requesting this for cloud-init disks...)

Tests: cargo test, PHD.

### DIFF
--- a/bin/propolis-server/src/lib/spec/api_spec_v0.rs
+++ b/bin/propolis-server/src/lib/spec/api_spec_v0.rs
@@ -8,7 +8,7 @@
 use std::collections::HashMap;
 
 use propolis_api_types::instance_spec::{
-    components::devices::{SerialPort as SerialPortDesc, SerialPortNumber},
+    components::devices::SerialPort as SerialPortDesc,
     v0::{InstanceSpecV0, NetworkBackendV0, NetworkDeviceV0, StorageDeviceV0},
 };
 use thiserror::Error;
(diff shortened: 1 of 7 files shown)


hawkw:
What's the motivation for giving serial ports real names rather than hard-coding them to `com{1, 2, 3, 4}`?

gjcolombo:
> What's the motivation for giving serial ports real names rather than hard-coding them to `com{1, 2, 3, 4}`?

The (perhaps not-well-supported) premise is that clients should get to choose the monikers for the components they ask Propolis to create, and serial devices are just another kind of component. (The hard-coding may very well still happen, to be clear; it'd just be in the client.)

I don't think this has any tremendously profound technical benefit aside from that little bit of uniformity--it just seems a little odd to say that clients (who are using the instance spec API, at any rate; PHD is the only one today but I would like to try to make sled agent use it too) have to provide names for all their components, but some of the names actually get ignored because of the kind of the component in question.

hawkw:
Got it, yeah, that makes sense. May as well be consistent with every other thing!

bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -68,6 +68,7 @@ impl From<Spec> for InstanceSpecV0 {
         let mut spec = InstanceSpecV0::default();
         spec.devices.board = val.board;
         for (disk_name, disk) in val.disks {
+            let backend_name = disk.device_spec.backend_name().to_owned();
```
hawkw, line 71: it occurs to me just now that, since backend and device names never get mutated once they're parsed, we _could_ switch to using `Arc<str>` for backend names so that we don't memcpy the whole string and instead just increment its ref count. but, I don't know if this is worth the effort; each propolis-server process only has one spec, so the device/backend names probably don't represent a huge amount of memory, and we only do this on initialization, so we're not making a ton of clones in a hot loop...

hawkw (approved):
Looks good to me!

bin/propolis-server/src/lib/spec/api_spec_v0.rs
gjcolombo: Agreed--I don't think this is likely to be performance-critical. This conversion direction (internal spec to external wire-format spec) should only really be used during live migration, when the source has to rehydrate a wire spec to send to the target; that allocation should be transient, and all this work happens at the very beginning of migration (well before blackout), so probably doesn't need to be the very fastest thing in the world.

PR #766, opened 2024-09-17
instance spec rework: move migration compat checks to propolis-server

gjcolombo:
(Part 8/9 of the instance spec rework; see #735.)

Transfer Propolis's component-level live migration compatibility checks from the API types crate to propolis-server. This is the last big chunk of application code that lives in the api-types crate. This is not straight-up code movement because the internal `Spec` type doesn't have the same notion of "device" and "backend" collections that the wire type has. This means that the hierarchy of errors in the new propolis-server `compat.rs` is different (but hopefully simpler) than what was in the old propolis-api-types `migration.rs`. The actual compatibility rules should be the same as they were before, however, and I've tried to bring over wholesale the api-types crate's compatibility unit tests.

Also, adjust how compatibility checks work in the migration protocol. Before, the migration preamble sent a `DeviceSpecV0` and a list of backend component names; the destination checked that the devices were compatible and that the backend name sets matched (but, owing to the absence of the actual backend descriptors, did not check that the backends themselves were migration-compatible). With this change, the preamble contains the source's entire `InstanceSpecV0`, though the compatibility checks are still (currently) limited to the VM's devices. This breaks the protocol's dependency on the notion that instance specs necessarily have collections of "devices" and "backends" as opposed to simply having a set of *components*. This is a prerequisite for the next PR in this series, which removes this distinction from the `InstanceSpecV0` type itself.

This is a migration protocol-breaking change, so failures in the migrate-from-base tests are expected.

Tests: cargo test, PHD.


bin/propolis-server/src/lib/migrate/compat.rs
```diff
@@ -0,0 +1,791 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Checks for compatibility of two instance specs.
```
gjcolombo, line 5: this doc comment doesn't need to be 500 words or anything but I could probably elaborate a little bit here

bin/propolis-server/src/lib/migrate/compat.rs
```diff
@@ -0,0 +1,791 @@
...
+    ) -> Result<(), CollectionIncompatibility>;
+}
+
+#[derive(Debug, Error)]
+pub enum CompatibilityError {
+    #[error("specs have incompatible boards")]
+    Board(#[from] BoardIncompatibility),
+
+    #[error("specs have incompatible pvpanic settings")]
+    Pvpanic(#[from] PvpanicIncompatibility),
```
hawkw, line 40: take it or leave it: it seems to me that these _could_ just be `#[error(transparent)]`; I don't think "specs have incompatible boards" and then hiding the actual problem in the error's cause chain provides any more useful context over the inner `BoardIncompatibility` error's message?

bin/propolis-server/src/lib/migrate/compat.rs
```diff
@@ -0,0 +1,791 @@
...
+    #[error(
+        "disks have different device interfaces (self: {this}, other: {other})"
+    )]
+    Interface { this: &'static str, other: &'static str },
+
+    #[error("disks have different PCI paths (self: {this}, other: {other})")]
+    PciPath { this: PciPath, other: PciPath },
+
+    #[error(
+        "disks have different backend names (self: {this}, other: {other})"
```
hawkw, line 77: suuuuuper nitpicky, probably not actually important: names, being user-provided strings, should _probably_ get `fmt::Debug` in error messages suggesting they're not equal? although unlikely, it would be real sad if we encountered two names containing different Unicode code-points that *looked* the same, and the error message appeared to be showing the same string twice...
```suggestion
        "disks have different backend names (self: {this:?}, other: {other:?})"
```

bin/propolis-server/src/lib/migrate/compat.rs
```diff
@@ -0,0 +1,791 @@
...
+    ReadOnly { this: bool, other: bool },
+}
+
+#[derive(Debug, Error)]
+pub enum NicIncompatibility {
+    #[error("NICs have different PCI paths (self: {this}, other: {other})")]
+    PciPath { this: PciPath, other: PciPath },
+
+    #[error(
+        "NICs have different backend names (self: {this}, other: {other})"
```
hawkw, line 98: Same thing as above, these should maybe be `Debug`:
```suggestion
        "NICs have different backend names (self: {this:?}, other: {other:?})"
```

bin/propolis-server/src/lib/migrate/compat.rs
```diff
@@ -0,0 +1,791 @@
...
+    DownstreamBus { this: u8, other: u8 },
+}
+
+#[derive(Debug, Error)]
+pub enum PvpanicIncompatibility {
+    #[error("pvpanic presence differs (self: {this}, other: {other})")]
+    Presence { this: bool, other: bool },
+
+    #[error(
+        "pvpanic devices have different names (self: {this}, other: {other})"
```
hawkw, line 127: and here maybe:
```suggestion
        "pvpanic devices have different names (self: {this:?}, other: {other:?})"
```

bin/propolis-server/src/lib/migrate/compat.rs
```diff
@@ -0,0 +1,791 @@
...
+                    })
+                } else if this.spec.enable_isa != other.spec.enable_isa {
+                    Err(PvpanicIncompatibility::EnableIsa {
+                        this: this.spec.enable_isa,
+                        other: other.spec.enable_isa,
+                    })
+                } else {
+                    Ok(())
+                }
+            }
```
hawkw, line 255: nit, take it or leave it: i probably would've roll the `if` guards into the `match`, as in 
```suggestion
            (Some(this), Some(other)) if this.name != other.name => {
                Err(PvpanicIncompatibility::Name {
                    this: this.name.clone(),
                    other: other.name.clone(),
                })
            },
            (Some(this), Some(other)) if this.spec.enable_isa != other.spec.enable_isa => {
                Err(PvpanicIncompatibility::EnableIsa {
                    this: this.spec.enable_isa,
                    other: other.spec.enable_isa,
                })
            },
            (Some(_), Some(_)) => Ok(()),
```
you could do this, if you like, but i'll freely admit it doesn't actually matter :)

bin/propolis-server/src/lib/migrate/compat.rs
```diff
@@ -0,0 +1,791 @@
...
+        other: &Self,
+    ) -> Result<(), ComponentIncompatibility>;
+}
+
+trait CompatCollection {
+    fn is_compatible_with(
+        &self,
+        other: &Self,
+    ) -> Result<(), CollectionIncompatibility>;
+}
```
hawkw, line 32: not particularly important, but it occurs to me that this _could_ be one trait, like:
```suggestion
trait Compat {
    type Error;
    fn is_compatible_with(
        &self,
        other: &Self,
    ) -> Result<(), Self::Error>;
}
```

since the methods are otherwise the same. but, the current thing is also fine.

bin/propolis-server/src/lib/spec/mod.rs
```diff
@@ -186,6 +226,16 @@ pub struct SoftNpu {
     pub p9fs: Option<P9fs>,
 }
 
+#[cfg(feature = "falcon")]
+impl SoftNpu {
+    pub fn is_empty(&self) -> bool {
```
hawkw, line 231: considered pedantic, feel free to ignore: dunno if i love `is_empty()`, which kind of treats this as a generic collection; perhaps something more semantically meaningful like `is_softnpu_enabled()` or similar? :woman_shrugging:

bin/propolis-server/src/lib/migrate/compat.rs
```diff
@@ -0,0 +1,791 @@
...
+    #[error("pvpanic presence differs (self: {this}, other: {other})")]
+    Presence { this: bool, other: bool },
+
+    #[error(
+        "pvpanic devices have different names (self: {this}, other: {other})"
+    )]
+    Name { this: String, other: String },
+
+    #[error(
+        "pvpanic devices have different ISA settings (self: {this}, other: {other})"
```
hawkw, line 132: super pedantic, feel free to ignore: the QEMU docs always typeset it as "PVPANIC" (all caps), so the error messages I had added previously did so as well; perhaps we should here? i don't actually care about this though. :)

hawkw (approved):
Overall this looks great to me --- removing the distinction between "backends" and "devices" from the spec types is a nice simplification. I quite like the new compatibility errors, too: the messages seem a bit more helpful than the prior ones. I picked a few very minor nits but there's nothing terribly controversial here IMO.

bin/propolis-server/src/lib/migrate/compat.rs
gjcolombo: I'm gonna leave this one as-is--I know it's the PANIC DEVICE but IMO there's not really a need to shout ;) (If the name changes to the `Panic!` device I might change my opinion though.)

hawkw: yeah, i'm fine with that, i really don't care about typesetting it the same way as QEMU

PR #767, opened 2024-09-18
instance spec rework: flatten InstanceSpecV0

### LINKED ISSUE #735: instance specs could be improved

gjcolombo:
(9/9 in the instance spec rework; see #735.)

Finally, the point of all the prior refactoring PRs!

Remove the complex hierarchy of component types from `InstanceSpecV0` and instead turn V0 specs into a `Board` and a bag of `Component`s. This makes it much easier to define new component types flexibly: 

- New components all go into the same collection: no more deciding if you've defined a device or a backend or deciding what to do if you have something in between.
- New components always appear in a map and so always have names (their keys). A bonus is that having only one component map prevents keys from being reused in multiple maps.
- Because new components don't need fields, there's no need to remember to tag component fields with serde's `default` or `skip_serializing_if` attributes (though this may still be needed for component fields).
- It's also much harder to get painted into a backwards-compatibility corner. Suppose you define a new component, add a field for it, and make the field an `Option`. If you later want to support having more than one of that component, you need to add a new spec field with higher maximum cardinality to avoid breaking old specs. Now you have two fields for the same kind of component. Yuck.

The downside of all this is that there are now more wire specs that type check but that the server should reject; for example, you can now have a wire spec with multiple panic devices. The previous changes in this series mitigate this by converting wire specs into more rigorously organized internal specs (which may enforce e.g. cardinality limits) before propolis-server will actually use them.

Now that wire specs have a much simpler structure, the spec builder in the propolis-client lib pulls very little weight, so remove it. Incoming wire specs need to be validated on the server in any case, and two different server versions may disagree on whether a particular wire spec is acceptable (if one has features the other does not), so it's hard to have a single builder that checks invariants for all relevant server versions. Clients who want to have a mistake-catching/invariant-protecting builder are, of course, still free to define their own (and those who might not want one, like PHD, don't need to pull one in that they won't use).

Finally, remove the `#[cfg(feature = "falcon")]` guards from the api-types crate. Servers built without Falcon support will reject specs that contain Falcon components. The downside to this is that users of generated clients who never intend to access a Falcon server (e.g. sled agent) will start to see these types. The upside is that this means that propolis-server's OpenAPI definition no longer varies with its feature set, which means we no longer need `propolis-server-falcon.json` (which is easy to forget to update) or the Falcon variant of the generated client (whose Progenitor directives had to be manually kept in sync with the non-Falcon client).

This is (once again) a migration protocol-breaking change, so the migrate-from-base tests are (once again) expected to fail.

Tests: cargo test, PHD.

Closes #735.


bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -147,119 +171,144 @@ impl From<Spec> for InstanceSpecV0 {
...
+        let mut viona_backends: HashMap<String, VirtioNetworkBackend> =
+            HashMap::new();
+        let mut dlpi_backends: HashMap<String, DlpiNetworkBackend> =
+            HashMap::new();
+
+        for (name, component) in value.components.into_iter() {
+            match component {
+                ComponentV0::CrucibleStorageBackend(_)
+                | ComponentV0::FileStorageBackend(_)
+                | ComponentV0::BlobStorageBackend(_) => {
```
hawkw, line 188: i kinda wondered if it was worth having functions like `ComponentV0::is_storage_backend() -> bool` but, upon further consideration, we would only ever call them in one place, so probably not...

hawkw (approved):
Overall, this looks good to me --- I did try to  search for specific nits to pick and didn't really find any. It's a bit of a shame to do more of the instance-spec validation at runtime, but I agree that it's the right path forward given everything you wrote in the commit message.

gjcolombo:
Thanks for the review, @hawkw! I'm going to hold this until #756 lands to save us from having to spend time rebasing that change onto this one.

gjcolombo:
In addition to failing CI, my resolution of the conflicts from my force-push (to rebase on top of #756) is not as clean as it probably could be. I'll push another commit or two soon to clean things up.

crates/propolis-api-types/src/instance_spec/components/devices.rs
```diff
@@ -104,19 +104,44 @@ pub struct QemuPvpanic {
     // TODO(eliza): add support for the PCI PVPANIC device...
 }
 
+/// A set of settings that determines what Propolis tells guest firmware about
+/// how to boot a guest operating system.
```
iximeow, line 108:
```suggestion
/// how to boot a guest.
```

minorest quibble: the settings here are processed before the guest has landed in an OS for the most part. unless there's a way we'd expect to speak to guests via UEFI (which i genuinely would believe, and would believe if there's something specific to OSes that we anticipate and i dunno about)

bin/propolis-server/src/lib/spec/builder.rs
```diff
@@ -113,18 +116,51 @@ impl SpecBuilder {
...
+        for item in boot_options {
+            self.add_boot_option(item)?;
+        }
+
+        Ok(())
+    }
+
     /// Add a boot option to the boot option list of the spec under construction.
-    pub fn add_boot_option(
+    fn add_boot_option(
```
iximeow, line 148: fwiw i think this is of substantially less utility since you must have called `add_boot_order()` with at least an empty list for this to not panic on accessing `boot_settings`. as-is, most of this function, by vertical space, is just dancing around the fact that the caller just set `self.spec.boot_settings` to `Some`.

lets fold this into `add_boot_order` and be done with it?

iximeow (approved):
i forgot how much nicer components were going to be once #756 landed! two minor quibbles about boot settings though :)

crates/propolis-api-types/src/instance_spec/components/devices.rs
gjcolombo: I took this in a slightly different direction in 04fb943, but I think I've captured the gist of this (this is information for the firmware image to use, not--generally--the next-stage binary that it finds and transfers control to).

iximeow: oh yeah i like that a lot, thanks!

PR #774, opened 2024-10-01
PciPath to Bdf conversion is infallible; prove it and refactor

iximeow:
i initially tried to find `TryInto` cases by deleting the impl and seeing what broke, but clippy is much better at this than me. `cargo xtask clippy`, in fact, is something i forgot i can or should run.


gjcolombo (approved):
Nice!

pfmooney (approved):
Out of curiosity, how long does that test run on a modern CPU?

iximeow:
i'd mostly just noticed it didn't "feel slow", but on second look it's 1.87s on a 7950x. so maybe it's more "not slow if you've just also been waiting to rebuild propolis"..

pfmooney:
> i'd mostly just noticed it didn't "feel slow", but on second look it's 1.87s on a 7950x. so maybe it's more "not slow if you've just also been waiting to rebuild propolis"..

Seems reasonable enough.  If it becomes a problem in the future, we could rub some `quickcheck` (or similar) on it.

PR #777, opened 2024-10-02
server: add a first-class error type to machine init

gjcolombo:
Add a `MachineInitError` type and make `MachineInitializer` functions return it. This makes it more convenient to handle new logic errors in initializer functions. To handle the (many) cases where the initializer calls bhyve ioctls or filesystem functions that return `std::io:Error`, `MachineInitError` has a catchall `anyhow::Error` variant; this encourages calls that produce I/O errors to call `context` or `with_context` to better describe those errors.

Tested by running propolis-server ad hoc in a couple of invalid configurations (file backend not pointing to an actual file; too many vCPUs; more guest memory than the host has available) and verified that the error chains (and contexts, if applicable) show up in the Propolis logs.

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -4,7 +4,6 @@
 
 use std::convert::TryInto;
 use std::fs::File;
-use std::io::{Error, ErrorKind};
 use std::num::NonZeroUsize;
 use std::os::unix::fs::FileTypeExt;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


bin/propolis-server/src/lib/initializer.rs
```diff
@@ -1023,14 +1102,14 @@ impl<'a> MachineInitializer<'a> {
...
-    ) -> Result<Arc<ramfb::RamFb>, Error> {
+    ) -> Result<Arc<ramfb::RamFb>, MachineInitError> {
         let fwcfg = fwcfg::FwCfg::new();
         fwcfg
             .insert_legacy(
                 fwcfg::LegacyId::SmpCpuCount,
                 fwcfg::Entry::fixed_u32(u32::from(cpus)),
             )
-            .unwrap();
+            .map_err(|e| MachineInitError::FwcfgInsertFailed("cpu count", e))?;
```
iximeow, line 1112: oh that's nice along the way.

iximeow (approved):
:tada: thanks!

gjcolombo:
Thanks for the reviews!

PR #778, opened 2024-10-02
PHD: allow patched Crucible dependencies

### LINKED ISSUE #770: phd-runner build should allow patched crucible
When attempting to run `cargo xtask clippy` on the workspace while the `crucible` crate was patched to reference a local path, I encountered an error in the `phd-runner` build:
```
  Caused by:
      Crucible package should not be a workspace member, and therefore should have source metadata
```

The logic for querying the git hash for test artifact selection does not appear to tolerate a non-git dep for crucible.

hawkw:
`phd-runner` uses the dependency on Crucible specified in the workspace's Cargo.toml to determine which downstairs artifact to download from Buildomat. This only works when the Crucible dependency is a Git dep on the `oxidecomputer/crucible` GitHub repo. In some cases, developers may patch Crucible to a local checkout or a different repository, such as for testing changes that haven't been merged to `oxidecomputer/crucible`. Unfortunately, the phd-runner build script fails in this case, which makes running any PHD tests impossible.

This is unfortunate. PHD tests _can_ still be run when the Crucible dependency is patched, either by disabling the Crucible-based tests, or by manually providing a Crucible downstairs binary path. However, because the build script fails in this case, it's impossible to build `phd-runner`, which sucks. Instead, we should allow the build to succeed and just disable the `--crucible-downstairs-commit auto` CLI option when the dependency isn't pointed at the Crucible GitHub repo.

This commit does that. Now, we just log a warning and set the env var to a sentinel value that means we couldn't determine the SHA for auto mode. The `phd-runner` binary will then exit with an error if `--crucible-downstairs-commit auto` is selected when compiled without a known Git SHA.

To test this, I patched Crucible with a local checkout:

```toml
[patch."https://github.com/oxidecomputer/crucible"]
crucible = { path = "../crucible/upstairs" }
crucible-client-types = { path = "../crucible/crucible-client-types" }
```

Running `cargo build -p phd-runner` now succeeds:

```console
eliza@theseus ~/Code/oxide/propolis $ cargo build -p phd-runner
     Locking 4 packages to latest compatible versions
      Adding crucible v0.0.1 (/home/eliza/Code/oxide/crucible/upstairs)
      Adding crucible-client-types v0.1.0 (/home/eliza/Code/oxide/crucible/crucible-client-types)
      Adding crucible-common v0.0.1 (/home/eliza/Code/oxide/crucible/common)
      Adding crucible-protocol v0.0.0 (/home/eliza/Code/oxide/crucible/protocol)
   Compiling propolis-client v0.1.0 (/home/eliza/Code/oxide/propolis/lib/propolis-client)
   Compiling phd-runner v0.1.0 (/home/eliza/Code/oxide/propolis/phd-tests/runner)
warning: phd-runner@0.1.0: Crucible dependency is patched with a local checkout, so the `--crucible-downstairs-commit auto` flag will be disabled in this PHD build
   Compiling phd-framework v0.1.0 (/home/eliza/Code/oxide/propolis/phd-tests/framework)
   Compiling phd-testcase v0.1.0 (/home/eliza/Code/oxide/propolis/phd-tests/testcase)
   Compiling phd-tests v0.1.0 (/home/eliza/Code/oxide/propolis/phd-tests/tests)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 11.63s
```

Fixes #770

### DIFF
--- a/phd-tests/runner/src/config.rs
+++ b/phd-tests/runner/src/config.rs
@@ -272,8 +272,8 @@ impl RunOptions {
                     anyhow::bail!(
                         "Because {reason}, phd-runner's build script could not determine \
                          the Crucible Git SHA, so the `--crucible-downstairs-commit auto` \
-                         option has been disabled. You can provide a local Crucible \
-                         binary using `--crucible-downstairs-command`.",
+                         option has been disabled.\n\tYou can provide a local Crucible \
+                         binary using `--crucible-downstairs-cmd`.",
                     )
                 }
 
--- a/phd-tests/runner/src/config.rs
+++ b/phd-tests/runner/src/config.rs
@@ -265,9 +265,18 @@ impl RunOptions {
                 // The Git revision of Crucible we depend on is determined when building
                 // `phd-runner` by the build script, so that the `phd-runner` binary can
                 // be run even after moving it out of the Propolis cargo workspace.
-                let commit = env!("PHD_CRUCIBLE_GIT_REV").parse().context(
-                "PHD_CRUCIBLE_GIT_REV must be set to a valid Git revision by the build script",
-            )?;
+                let commit = match env!("PHD_CRUCIBLE_GIT_REV") {
+                    "CANT_GET_YE_CRUCIBLE_SHA" => anyhow::bail!(
+                        "Crucible upstairs dependency was patched with a \
+                        local path when PHD was compiled, so the \
+                        `--crucible-downstairs-commit auto` option has been \
+                         disabled."
+                    ),
+                    commit => commit.parse().context(
+                        "PHD_CRUCIBLE_GIT_REV must be set to a valid Git \
+                         revision by the build script",
+                    )?,
+                };
                 Ok(Some(CrucibleDownstairsSource::BuildomatGitRev(commit)))
             }
             None => Ok(None),
--- a/phd-tests/runner/src/config.rs
+++ b/phd-tests/runner/src/config.rs
@@ -265,18 +265,22 @@ impl RunOptions {
                 // The Git revision of Crucible we depend on is determined when building
                 // `phd-runner` by the build script, so that the `phd-runner` binary can
                 // be run even after moving it out of the Propolis cargo workspace.
-                let commit = match env!("PHD_CRUCIBLE_GIT_REV") {
-                    "CANT_GET_YE_CRUCIBLE_SHA" => anyhow::bail!(
-                        "Crucible upstairs dependency was patched with a \
-                        local path when PHD was compiled, so the \
-                        `--crucible-downstairs-commit auto` option has been \
-                         disabled."
-                    ),
-                    commit => commit.parse().context(
-                        "PHD_CRUCIBLE_GIT_REV must be set to a valid Git \
-                         revision by the build script",
-                    )?,
-                };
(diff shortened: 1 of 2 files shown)


phd-tests/runner/src/config.rs
```diff
@@ -265,9 +265,18 @@ impl RunOptions {
...
                 // be run even after moving it out of the Propolis cargo workspace.
-                let commit = env!("PHD_CRUCIBLE_GIT_REV").parse().context(
-                "PHD_CRUCIBLE_GIT_REV must be set to a valid Git revision by the build script",
-            )?;
+                let commit = match env!("PHD_CRUCIBLE_GIT_REV") {
+                    "CANT_GET_YE_CRUCIBLE_SHA" => anyhow::bail!(
+                        "Crucible upstairs dependency was patched with a \
+                        local path when PHD was compiled, so the \
+                        `--crucible-downstairs-commit auto` option has been \
+                         disabled."
```
gjcolombo, line 273: If we can't or don't want to go the "try to infer where the binary should be based on the path to the checkout" route, perhaps this message (and/or the one in build.rs) should advertise the `--crucible-downstairs-cmd` option so that folks building a local Crucible know they can run PHD with it?

gjcolombo (approved):
Off-the-wall idea here, but: is it possible to get `cargo xtask phd` to notice that the Crucible path has been patched and to build a downstairs binary from that path? We could then try to pipe it into a (separate) environment var and have `--crucible-downstairs-commit auto` try to pick that up: the git rev isn't valid, but I have a proposed path to a Crucible downstairs on the local machine, so let me try that instead.

If that's intractable for some reason (and I could totally believe it is, because e.g. we don't really know if the Crucible dep is local or we don't want to build other projects like that while building this one), then this LGTM as-is.

phd-tests/runner/src/config.rs
hawkw: Yeah, that's a good idea.

hawkw:
> Off-the-wall idea here, but: is it possible to get `cargo xtask phd` to notice that the Crucible path has been patched and to build a downstairs binary from that path? We could then try to pipe it into a (separate) environment var and have `--crucible-downstairs-commit auto` try to pick that up: the git rev isn't valid, but I have a proposed path to a Crucible downstairs on the local machine, so let me try that instead.

Hmm, that's a good idea, and I'm sure it's possible, although it might be a bit annoying. I have a slight preference for merging this change and *then* going and doing that, though, because the current state of the world is totally broken if you've patched Crucible, and I'd like `cargo build` to work sooner rather than later.

phd-tests/runner/src/config.rs
hawkw: It now says:

```
Error: Because Crucible dependency is patched with a local checkout, phd-runner's build script could not determine the Crucible Git SHA, so the `--crucible-downstairs-commit auto` option has been disabled. You can provide a local Crucible binary using `--crucible-downstairs-command`.
```

gjcolombo (approved):
The new message looks great--thanks for tweaking it!

hawkw:
@gjcolombo I've opened #779 to track having `cargo xtask phd` build a downstairs if needed, that was a good idea!

PR #780, opened 2024-10-04
server: accept CPUID values in instance specs and plumb them to bhyve

gjcolombo:
Add support for specifying CPUID templates in instance specs. A "template" consists of a CPU vendor and a set of CPUID leaves and associated values. When a client passes an instance spec containing a template, the server runs a `propolis::cpuid::Specializer` over it and applies the resulting values to each vCPU during machine initialization. Add the appropriate sanitization logic to the API-spec-to-internal-spec layer and add migration compatibility checks.

This work is supported by some new helper crates and types:

- The `propolis_types` crate now has definitions of CPUID leaf identifiers, CPUID values, and supported CPUID vendors.
- The `cpuid-utils` crate adds:
  - Utility functions for querying host CPUID; some of these used to live in the main Propolis lib, but extracting them to a crate makes them available to PHD, where they're useful for writing CPUID tests
  - Helper impls for converting instance spec CPUID types to leaf/value maps

Finally, add two new PHD tests. One simply checks that CPUID values sent to the server properly round-trip back to the client if it asks for the server's instance spec. The other sets up a stripped-down AMD-compatible CPUID configuration with a test-specific brand string in leaves 0x80000002-0x80000004 and verifies that the string is visible in `/proc/cpuinfo`.


gjcolombo:
I think this is now in good enough shape to merge on its own, so I'm taking it out of draft. #782 refines this change a bit by removing some of the duplication between the `cpuid::Set` and `CpuidMap` types (and adding a bunch of documentation), but it produced a relatively big diff and this PR's contents are useful without it, so I figured I'd post the two changes separately.

pfmooney (commented):
Sorry I didn't have time to dig into this properly, but the broad strokes look good to me.

crates/cpuid-utils/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+    #[error("mismatched CPUID leaves {0:?}")]
+    MismatchedLeaves(Vec<CpuidIdent>),
+
+    #[error("mismatched values for leaves {0:?}")]
+    MismatchedValues(Vec<(CpuidIdent, CpuidValues, CpuidValues)>),
+}
+
+impl CpuidMap {
+    /// Returns `Ok` if `self` is equivalent to `other`; if not, returns a
+    /// [`CpuidMapMismatch`] describing the difference that caused the
```
gjcolombo, line 42: fractured comment

phd-tests/tests/src/cpuid.rs
```diff
@@ -0,0 +1,159 @@
...
+            .iter()
+            .map(|(leaf, value)| CpuidEntry {
+                leaf: leaf.leaf,
+                subleaf: leaf.subleaf,
+                eax: value.eax,
+                ebx: value.ebx,
+                ecx: value.ecx,
+                edx: value.edx,
+            })
+            .collect(),
```
gjcolombo, line 148: Hmm, I thought I had a `From` impl for this sort of thing...

crates/cpuid-utils/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+
+#[cfg(feature = "instance-spec")]
+mod instance_spec;
+
+#[cfg(feature = "instance-spec")]
+pub use instance_spec::*;
+
+/// A map from CPUID leaves to CPUID values.
+#[derive(Clone, Debug, Default)]
+pub struct CpuidMap(pub BTreeMap<CpuidIdent, CpuidValues>);
```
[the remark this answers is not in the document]
gjcolombo: This'll be addressed by #782 if we decide to go that route.

crates/cpuid-utils/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Utility functions and types for working with CPUID values.
+//!
+//! If this crate is built with the `instance-spec` feature, this module
+//! includes mechanisms for converting from instance spec CPUID entries to and
+//! from the CPUID map types in the crate. These conversions require that the
+//! input list of CPUID entries contains no duplicate leaf/subleaf pairs and
+//! that the specified leaves all fall in the standard or extended ranges.
```
hawkw, line 11: I guess this is feature flagged so that propolis-lib can depend on just the CPUID utilities without the instance-spec stuff?

crates/cpuid-utils/src/instance_spec.rs
```diff
@@ -0,0 +1,123 @@
...
+                return Err(CpuidMapConversionError::SubleafZeroAliased(leaf));
+            }
+
+            if let Some(0) = subleaf {
+                if map.contains_key(&CpuidIdent::leaf(leaf)) {
+                    return Err(CpuidMapConversionError::SubleafZeroAliased(
+                        leaf,
+                    ));
+                }
+            }
```
iximeow, line 77: ~if provided a `value` like ` vec![CpuidEntry { leaf: N, subleaf: None, .. }, CpuidEntry { leaf: N, subleaf: Some(1), .. }]` the aliasing logic wouldn't notice that and we'd behave in weird (but not obviously disastrous) ways at runtime - the `subleaf: None` would act like a fallback for all `ecx!=1` queries.~

~i don't think any leaves of cpuid need behavior like that, and i think such a situation would be a bug, right?~

https://github.com/oxidecomputer/propolis/pull/782/commits/2ddcd3386730884d1e37e95dc00983ee0c3509eb yay!

crates/cpuid-utils/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+        std::arch::asm!(
+            "push rbx",
+            "cpuid",
+            "mov {0:e}, ebx",
+            "pop rbx",
+            out(reg) res.ebx,
+            // select cpuid 0, also specify eax as clobbered
+            inout("eax") leaf.leaf => res.eax,
+            inout("ecx") leaf.subleaf.unwrap_or(0) => res.ecx,
+            out("edx") res.edx,
```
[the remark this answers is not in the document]
iximeow: eliza's comment below mentions the libcore intrinsic, that in turn was enough breadcrumb to learn the unfortunate story:

the move to preserve rbx was intentional in rust-lang's part: https://github.com/rust-lang/stdarch/pull/1121/files

and it's to work around an LLVM issue that was still open least as of six years ago: https://bugs.llvm.org/show_bug.cgi?id=17907

so... that's underwhelming. i'd go do a PR to include these notes but i don't want to  conflict with you - if you include them in the motion then great, or i'll add 'em once this lands.

phd-tests/tests/src/cpuid.rs
```diff
@@ -0,0 +1,159 @@
...
+        leaf.leaf <= 0x7
+            || (leaf.leaf >= 0x8000_0000 && leaf.leaf <= 0x8000_0001)
+    });
+
+    // Report that leaf 7 is the last available standard leaf.
+    host_cpuid.get_mut(&CpuidIdent::leaf(0)).unwrap().eax = 7;
+
+    // Mask off feature bits that the Oxide platform won't support. See RFD
+    // 314. These are masks (and not assignments) so that features the host CPU
+    // support won't appear in the guest's feature masks.
```
gjcolombo, line 93: s/support/doesn't support

bin/propolis-server/src/lib/migrate/compat.rs
```diff
@@ -793,4 +837,84 @@ mod test {
...
+        let mut set2 = cpuid::Set::new(CpuidVendor::Amd);
+
+        let v1 = CpuidValues { eax: 4, ebx: 5, ecx: 6, edx: 7 };
+        let v2 = CpuidValues { eax: 100, ebx: 200, ecx: 300, edx: 400 };
+        set1.insert(CpuidIdent::leaf(0), v1);
+        set2.insert(CpuidIdent::leaf(0), v2);
+        s1.cpuid = Some(set1);
+        s2.cpuid = Some(set2);
+        assert!(s1.is_migration_compatible(&s2).is_err());
+    }
```
gjcolombo, line 919: Would also be good to have a test for the subleaf-some/subleaf-none mismatch.

phd-tests/tests/src/cpuid.rs
gjcolombo: Done (hopefully without creating too many headaches for myself when I rebase #782).

crates/cpuid-utils/src/lib.rs
gjcolombo: Done.

crates/cpuid-utils/src/lib.rs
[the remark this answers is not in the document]
gjcolombo: Obsoleted by switching to the Rust intrinsic.

crates/cpuid-utils/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+        );
+    }
+
+    res
+}
```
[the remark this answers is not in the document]
gjcolombo: Done.

phd-tests/tests/src/cpuid.rs
```diff
@@ -0,0 +1,159 @@
...
+    let leaf_1 = host_cpuid.get_mut(&CpuidIdent::leaf(1)).unwrap();
+    leaf_1.ecx &= 0xF6F8_3203;
+    leaf_1.edx &= 0x178B_FBFF;
+    let leaf_7 = host_cpuid.get_mut(&CpuidIdent::leaf(7)).unwrap();
+    leaf_7.eax = 0;
+    leaf_7.ebx &= 0x219C_03A9;
```
hawkw, line 99: i would really _like_ it if these masks were constructed by ORing together named constants for the things we want to mask out, but...i won't block this over it.

phd-tests/tests/src/cpuid.rs
```diff
@@ -0,0 +1,159 @@
...
+    leaf_7.edx = 0;
+
+    // Report that leaf 0x8000_0004 is the last extended leaf and clean up
+    // feature bits in leaf 0x8000_0001.
+    host_cpuid.get_mut(&CpuidIdent::leaf(0x8000_0000)).unwrap().eax =
+        0x8000_0004;
+    let ext_leaf_1 =
+        host_cpuid.get_mut(&CpuidIdent::leaf(0x8000_0001)).unwrap();
+    ext_leaf_1.ecx &= 0x4440_01F0;
+    ext_leaf_1.edx &= 0x27D3_FBFF;
```
hawkw, line 110: similarly, i must admit that the magic numbers make me feel great sadness in my heart

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -939,45 +944,82 @@ impl<'a> MachineInitializer<'a> {
...
+        ]
+        .into_iter()
+        // This returns None if any of the input options were None (i.e. if any
+        // of the requested leaves weren't found). This implies that if the
+        // `collect` returns `Some`, there are necessarily three elements in the
+        // `Vec`, so `try_into::<[CpuidValues; 3]>` will always succeed.
+        .collect::<Option<Vec<_>>>()
+        .map(TryInto::try_into)
+        .transpose()
+        .expect("output array should always have three elements");
```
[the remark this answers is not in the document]
gjcolombo: any day where I get to use `transpose` is a good day

phd-tests/tests/src/cpuid.rs
gjcolombo: It probably wouldn't hurt to have consts for these for the benefit of whatever code ends up implementing virtual platforms. I'll give this a shot.

phd-tests/tests/src/cpuid.rs
gjcolombo: Oh, right, I forgot: PHD is pulling in the generated type from propolis_client here, which doesn't get the From impls from propolis_api_types. This is why the build broke; I'll back this out to patch it up.

phd-tests/tests/src/cpuid.rs
gjcolombo: Done.

phd-tests/tests/src/cpuid.rs
gjcolombo: Done.

crates/propolis-api-types/src/instance_spec/components/board.rs
```diff
@@ -46,16 +100,10 @@ pub struct Board {
...
-            cpus: 0,
-            memory_mb: 0,
-            chipset: Chipset::I440Fx(I440Fx { enable_pcie: false }),
-        }
-    }
+    /// The CPUID values to expose to the guest. If `None`, bhyve will derive
+    /// default values from the host's CPUID values.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub cpuid: Option<Cpuid>,
+    // TODO: Processor and  topology.
```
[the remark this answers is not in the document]
gjcolombo: Fixed (this used to refer to memory topology and I guess I deleted that somewhere along the line? Restored now though).

crates/propolis-api-types/src/instance_spec/components/board.rs
```diff
@@ -34,6 +36,58 @@ pub enum Chipset {
...
 
+/// A set of CPUID values to expose to a guest.
+#[derive(Clone, Deserialize, Serialize, Debug, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct Cpuid {
+    /// A list of CPUID leaves/subleaves and their associated values.
+    ///
+    /// The `leaf` and `subleaf` fields of each entry in the template must
+    /// be unique. The leaf must also be in the standard or extended ranges
+    /// (0 to 0xFFFF or 0x8000_0000 to 0x8000_FFFF).
```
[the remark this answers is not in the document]
gjcolombo: Having slept on it, this is sort of an awkward spot to refer to other Rust crates, since this appears in the API documentation. I switched this around to refer to the Intel/AMD manuals instead.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -1157,6 +1199,24 @@ impl<'a> MachineInitializer<'a> {
...
+                    .with_vcpuid(vcpu.id)
+                    .with_cache_topo()
+                    .clear_cpu_topo(propolis::cpuid::TopoKind::iter())
+                    .execute(set.clone())
+                    .map_err(|e| {
+                        MachineInitError::CpuidSpecializationFailed(vcpu.id, e)
+                    })?;
+
+                vcpu.set_cpuid(specialized).with_context(|| {
+                    format!("setting CPUID for vcpu {}", vcpu.id)
```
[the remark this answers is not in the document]
hawkw: yeah, that works for me!

phd-tests/tests/src/cpuid.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    // Report that leaf 7 is the last available standard leaf.
+    host_cpuid.get_mut(&CpuidIdent::leaf(0)).unwrap().eax = 7;
+
+    // Mask off bits that a minimum-viable Oxide platform doesn't support or
+    // can't (or won't) expose to guests. See RFD 314 for further discussion of
+    // how these masks were chosen.
+    //
+    // These are masks (and not assignments) so that, if the host processor
+    // doesn't support a feature contained in an `ALL_FLAGS` value, the
```
iximeow, line 98: this (much nicer) comment lost its tail end. `the feature is not incorrectly advertised`?

gjcolombo: argh

phd-tests/tests/src/cpuid.rs
```diff
@@ -0,0 +1,178 @@
...
+    ext_leaf_1.ecx &= (AmdExtLeaf1Ecx::ALT_MOV_CR8
+        | AmdExtLeaf1Ecx::ABM
+        | AmdExtLeaf1Ecx::SSE4A
+        | AmdExtLeaf1Ecx::MISALIGN_SSE
+        | AmdExtLeaf1Ecx::THREED_NOW_PREFETCH
+        | AmdExtLeaf1Ecx::DATA_ACCESS_BP
+        | AmdExtLeaf1Ecx::DATA_BP_ADDR_MASK_EXT)
+        .bits();
+    ext_leaf_1.edx &=
+        (AmdExtLeaf1Edx::ALL_FLAGS & !AmdExtLeaf1Edx::RDTSCP).bits();
```
hawkw, line 129: love the bitflags here, thank you <3

hawkw (approved):
Overall, this looks good to me. I had one last thing to call out regarding target arch cfgs that won't actually compile for 32-bit targets, but...it was already broken before this change and presumably we don't really care about it very much.

phd-tests/tests/src/cpuid.rs
gjcolombo: Fixed in c560647.

crates/cpuid-utils/src/lib.rs
```diff
@@ -80,23 +80,10 @@ pub const EXTENDED_LEAVES: RangeInclusive<u32> = 0x8000_0000..=0x8000_FFFF;
...
-            "cpuid",
-            "mov {0:e}, ebx",
-            "pop rbx",
-            out(reg) res.ebx,
-            // select cpuid 0, also specify eax as clobbered
-            inout("eax") leaf.leaf => res.eax,
-            inout("ecx") leaf.subleaf.unwrap_or(0) => res.ecx,
-            out("edx") res.edx,
-        );
+        core::arch::x86_64::__cpuid_count(leaf.leaf, leaf.subleaf.unwrap_or(0))
```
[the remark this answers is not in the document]
gjcolombo: Makes sense--I don't think 32-bit hosts are in scope for us at the moment :) I adjusted the cfg guard in 58ce359.

iximeow (approved):
seems good, thank you!

gjcolombo:
Thanks for the reviews!

PR #782, opened 2024-10-05
lib: move cpuid::Set to cpuid_utils; prevent semantic subleaf conflicts

gjcolombo:
Move the functionality in `propolis::cpuid::Set` into `cpuid_utils`, then tidy it up a bit by removing its `for_regs` and `into_inner` functions (the former was dead code and the latter is no longer needed).

Rework `CpuidMap` to prevent callers from using both `CpuidIdent { leaf: x, subleaf: None }` and `CpuidIdent { leaf: x, subleaf: Some(y) }` as keys in the same map (does CPUID with eax = x ignore the value passed in ecx or not?).[^1] This requires a fair bit of new code to handle insert/remove and iteration. Add tests for these cases, including a property test to verify a larger number of iteration patterns than we could hope to write by hand.

Finally, add a theory statement and some additional function documentation to the `cpuid_utils` crate.

[^1]: In practice the behavior depends on the order in which the conflicting entries appear in the `vcpu_cpuid_entry` array that Propolis passes to bhyve (the first one to match an incoming CPUID request wins); see bhyve's `cpuid_find_entry`.

### DIFF
--- a/crates/cpuid-utils/src/lib.rs
+++ b/crates/cpuid-utils/src/lib.rs
@@ -125,7 +125,7 @@ impl CpuidMap {
         .flatten()
     }
 
-    /// Retrives a mutable reference to the values associated with the supplied
+    /// Retrieves a mutable reference to the values associated with the supplied
     /// `ident`, or `None` if the identifier is not present in the map.
     pub fn get_mut(&mut self, ident: CpuidIdent) -> Option<&mut CpuidValues> {
         match ident.subleaf {
--- a/crates/cpuid-utils/src/lib.rs
+++ b/crates/cpuid-utils/src/lib.rs
@@ -261,6 +261,20 @@ impl CpuidMap {
     /// Passes each leaf number in the map to `f` and removes any leaf entries
     /// for which `f` returns `false`. If a removed leaf has subleaves, all
     /// their entries are removed.
+    //
+    // This function can be made to consider subleaves by changing `f` to take a
+    // `CpuidIdent` and then writing the call to `retain` with a `match`:
+    //
+    // - If the leaf has `Subleaves::Absent`, pass the leaf ID through to `f`
+    //   directly and return the result to `retain`.
+    // - If the leaf has `Subleaves::Present`, call `retain` on the subleaf map,
+    //   passing each subleaf to `f`, then return `!map.is_empty()` to the outer
+    //   call to `retain` (i.e. keep the leaf entry if the subleaf map still has
+    //   entries in it).
+    //
+    // The cost of doing this is that the function now needs to visit every
+    // subleaf entry in the map, even if the caller doesn't care about subleaf
+    // IDs. So it may be better to break this out into a separate function.
     pub fn retain_by_leaf<F>(&mut self, mut f: F)
     where
         F: FnMut(u32) -> bool,
--- a/crates/cpuid-utils/src/lib.rs
+++ b/crates/cpuid-utils/src/lib.rs
@@ -99,12 +99,19 @@ pub enum SubleafInsertConflict {
 /// Tracking CPUID values in a simple `BTreeMap` from [`CpuidIdent`] to
 /// [`CpuidValues`] allows a single leaf to use both options at once, because
 /// `CpuidIdent { leaf: x, subleaf: None }` is not the same as `CpuidIdent {
-/// leaf: x, subleaf: Some(y) }`.
+/// leaf: x, subleaf: Some(y) }`. To avoid this kind of semantic confusion, this
+/// type's `insert` method returns a `Result` that indicates whether inserting
+/// the requested leaf/subleaf identifier would produce a semantic conflict
+/// between a subleaf-bearing and subleaf-free entry.
 ///
-/// To avoid this kind of semantic confusion, this type's `insert` method
-/// returns a `Result` that indicates whether inserting the requested
-/// leaf/subleaf identifier would produce a semantic conflict between a
-/// subleaf-bearing and subleaf-free entry.
+/// This structure allows "holes" in a leaf's subleaf IDs; that is, the
+/// structure permits `leaf: 0, subleaf: Some(0)` and `subleaf: Some(2)` to
+/// appear in a map where `subleaf: Some(1)` is absent. This is mostly for
+/// simplicity (it saves the map type from having to check for discontiguous
+/// subleaf domains); the existing subleaf-having CPUID leaves in the Intel and
+/// AMD manuals all specify contiguous subleaf domains, and a client who
+/// specifies a discontiguous subleaf set may find itself with unhappy guest
+/// operating systems.
 #[derive(Clone, Debug, Default)]
 pub struct CpuidMap(BTreeMap<u32, Subleaves>);
 
--- a/crates/cpuid-utils/src/lib.rs
+++ b/crates/cpuid-utils/src/lib.rs
@@ -585,6 +585,8 @@ pub fn host_query_all() -> CpuidMap {
 
 #[cfg(test)]
 mod test {
+    use proptest::prelude::*;
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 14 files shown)


crates/cpuid-utils/src/lib.rs
```diff
@@ -4,14 +4,54 @@
 
 //! Utility functions and types for working with CPUID values.
 //!
+//! # The `CPUID` instruction
```
hawkw, line 7: this is how you know it's gonna be a good block comment...

crates/cpuid-utils/src/lib.rs
```diff
@@ -23,55 +63,475 @@ pub mod bits;
...
+/// `CpuidIdent { leaf: x, subleaf: None }` is not the same as `CpuidIdent {
+/// leaf: x, subleaf: Some(y) }`.
+///
+/// To avoid this kind of semantic confusion, this type's `insert` method
+/// returns a `Result` that indicates whether inserting the requested
+/// leaf/subleaf identifier would produce a semantic conflict between a
+/// subleaf-bearing and subleaf-free entry.
 #[derive(Clone, Debug, Default)]
-pub struct CpuidMap(pub BTreeMap<CpuidIdent, CpuidValues>);
+pub struct CpuidMap(BTreeMap<u32, Subleaves>);
```
hawkw, line 109: The implementation of this type is lovely --- the code is very straightforward and I didn't really have any meaningful concerns or nitpicks. Very nice!

crates/cpuid-utils/src/lib.rs
```diff
@@ -23,55 +63,475 @@ pub mod bits;
...
+    /// Removes all data for the supplied `leaf`. If the leaf has subleaves, all
+    /// their entries are removed.
+    pub fn remove_leaf(&mut self, leaf: u32) {
+        self.0.remove(&leaf);
+    }
+
+    /// Passes each leaf number in the map to `f` and removes any leaf entries
+    /// for which `f` returns `false`. If a removed leaf has subleaves, all
+    /// their entries are removed.
+    pub fn retain_by_leaf<F>(&mut self, mut f: F)
```
hawkw, line 264: It seems like it could, perhaps, be useful to also pass the value into `f` so that it could retain based on which subleaves are present --- but that would require making the `Subleaves` enum public, and perhaps we have no need for such an operation currently...

crates/cpuid-utils/src/lib.rs
```diff
@@ -23,55 +63,475 @@ pub mod bits;
...
+impl<'a> CpuidMapIterator<'a> {
+    fn new(map: &'a CpuidMap) -> Self {
+        Self { leaf_iter: map.0.iter(), subleaf_iter: None }
+    }
+}
+
+impl Iterator for CpuidMapIterator<'_> {
+    type Item = (CpuidIdent, CpuidValues);
+
+    fn next(&mut self) -> Option<Self::Item> {
```
hawkw, line 316: I appreciate the comments here --- the implementation makes sense, but they make it a lot clearer.

crates/cpuid-utils/src/lib.rs
```diff
@@ -23,55 +63,475 @@ pub mod bits;
...
+    }
+}
+
+/// A map from CPUID leaves to CPUID values that includes a vendor ID. Callers
+/// can use the vendor ID to interpret the meanings of any extended leaves
+/// present in the map.
+#[derive(Clone, Debug)]
+pub struct CpuidSet {
+    map: CpuidMap,
+    pub vendor: CpuidVendor,
```
hawkw, line 361: this being public rather than having an accessor allows code in other crates to mutate the vendor after constructing a map, which...is probably fine but _feels weird_? i suppose none of this type's methods here actually interprets the CPUID values based on the vendor ID, so it's not like any invariant could be violated due to reaching in and changing it from AMD to Intel or whatever...

lib/propolis/src/vcpu.rs
```diff
@@ -33,6 +35,18 @@ pub const MAXCPU: usize = bhyve_api::VM_MAXCPU;
...
+#[derive(Debug, Error)]
+pub enum GetCpuidError {
+    #[error("failed to read CPUID values from bhyve")]
+    ReadIoctlFailed(#[source] std::io::Error),
+
+    #[error("failed to build a map of CPUID entries")]
+    MapConversion(#[from] CpuidMapConversionError),
+
+    #[error("failed to convert CPUID values to vendor string: {0}")]
+    VendorConversionFailed(&'static str),
```
hawkw, line 47: should this maybe be reworded to something like
```suggestion
    #[error("unrecognized CPUID vendor string: {0}")]
    VendorConversionFailed(&'static str),
```

because, technically, we _did_ convert it to a vendor string, but the vendor string was just, like, "CyrixInstead" or some other weird thing that we don't care to support...

lib/propolis/src/vcpu.rs
```diff
@@ -1297,7 +1317,18 @@ pub mod migrate {
...
     impl VcpuReadWrite for CpuidV1 {
         fn read(vcpu: &Vcpu) -> Result<Self> {
-            Ok(vcpu.get_cpuid()?.into())
+            Ok(vcpu
+                .get_cpuid()
+                .map_err(|e| {
+                    std::io::Error::new(
+                        std::io::ErrorKind::InvalidData,
+                        format!(
+                            "error reading CPUID for vCPU {}: {:?}",
```
hawkw, line 1326: should the error be formatted using `fmt::Display` here?

crates/cpuid-utils/src/lib.rs
```diff
@@ -100,25 +560,269 @@ pub fn host_query(leaf: CpuidIdent) -> CpuidValues {
...
+                })
+        ]
+    }
+
+    proptest! {
+        /// Verifies that a [`CpuidMapIterator`] visits all of the leaf and
+        /// subleaf entries in a map and does so in the expected order.
+        ///
+        /// proptest will generate a set of 3-8 leaves for each test according
+        /// to the strategy defined in [`map_entry_strategy`].
```
hawkw, line 772: you love to see it!

hawkw (approved):
Very nice! I had a very small suggestions and questions, but nothing big.

crates/cpuid-utils/src/lib.rs
```diff
@@ -23,55 +63,475 @@ pub mod bits;
...
+            }),
+            Some(sl) => self.0.get(&ident.leaf).map(|ent| match ent {
+                Subleaves::Absent(_) => None,
+                Subleaves::Present(sl_map) => sl_map.get(&sl),
+            }),
+        }
+        .flatten()
+    }
+
+    /// Retrives a mutable reference to the values associated with the supplied
```
gjcolombo, line 128: "retrieves"

crates/cpuid-utils/src/lib.rs
gjcolombo: It turns out this is only mutated in one place--the `cpuid::Specializer` has a `with_vendor` option that will override the vendor specified in the input `CpuidSet`, but as far as I can tell this option is never used. It may have been meant for a world where a caller reads the host vendor and then configures the Specializer to override it, but with this change we've drifted toward just having callers set up the CpuidSet in the way they want from the get-go. In any case I've removed the `pub` and added a getter for this.

crates/cpuid-utils/src/lib.rs
```diff
@@ -23,55 +63,494 @@ pub mod bits;
...
+}
+
+/// A mapping from CPUID leaf/subleaf pairs to CPUID return values. This struct
+/// allows each registered leaf either to have or not to have subleaf values,
+/// but not both at once.
+///
+/// Some CPUID leaves completely ignore the subleaf value passed in ecx. Others
+/// pay attention to it and have their own per-leaf semantics that govern what
+/// happens if the supplied ecx value is out of the function's expected range.
+/// Tracking CPUID values in a simple `BTreeMap` from [`CpuidIdent`] to
```
iximeow, line 99: as we found out a bit earlier, it's also acceptable (if not well-defined) for callers to omit subleaves, leaving a hole in the range `0...max_ecx`. if not documented here, it feels like that's worth mentioning somewhere around this type... maybe just a variant in the proptest that asserts such a construction is OK?

iximeow (approved):
i echo the proptest enthusiasm and appreciation for nice docs here :pray:

crates/cpuid-utils/src/lib.rs
gjcolombo: Good call--amended the doc comment in 80b3cc6.

gjcolombo:
Thanks as always for the reviews! Will merge once CI is green.

PR #785, opened 2024-10-10
PHD: support guest-initiated reboot

### LINKED ISSUE #783: `boot_order_source_priority` fails with Debian 11 guest
found while digging into #781: repro steps are basically identical. get a Debian 11 image, get Propolis from any commit at or after `11371b0`.

run the test with an appropriate adaptation of a command like:
```
./target/debug/phd-runner run \
    --propolis-server-cmd ./target/release/propolis-server \
    --tmp-directory /tmp/phd \
    --artifact-toml-path ./artifacts-debian-bullseye.toml \
    --default-guest-artifact debian_bullseye \
    --include-filter boot_order_source_priority
```

this should exit with something like `test result: ok. 1 passed; 0 failed; 0 skipped; 0 not run; finished in 23.39s` - this is the result with an Alpine guest.

instead, with a Debian 11 guest, this hangs at `Waiting for serial output on serial console target="root@debian:~#"` and eventually times out.

the tell here is that the line just before is `Waiting for output on serial console target="reboot"`. that is from [this `reboot`](https://github.com/oxidecomputer/propolis/blob/master/phd-tests/tests/src/boot_order.rs#L472), where we're waiting for the new line after sending `reboot`. it turns out on Debian 11, `reboot` just reboots the machine immediately, but on more recent linuxes (recent Alpine, Ubuntu 22.04, etc) this is just a compatibility adapter for, it seems like, some systemd incantation that will reboot the machine in the near future.

this is why "`reboot` and wait for login" works as a strategy for more recent Linux images, but just hangs on Debian 11. on Debian 11, we even cycle all the way through to the login prompt, it's just not what we expected to see in response to `reboot`, so we keep waiting.

i figure i'll add a helper in the guest adaptations for "Perform graceful reboot initiated by the guest", which should fix this in the process.

iximeow:
While in many cases a procedure like "`reboot` and wait for login" is sufficient to reboot a guest, it is not sufficient for *all* cases. For Windows, the command is spelled `shutdown /r` instead. And on Debian (at least 11), `reboot` will immediately take the guest down instead of printing a new shell prompt first.

So, have guest OS adapters provide commands that impel a guest to reboot, and wait for that reboot to occur (rather than the naive initial approach of reboot, wait for shell prompt, wait for login prompt, proceed).

Fixes #783.

(since the boot order tests are pretty Linux-specific what with efivarfs, the Windows `graceful_reboot` implementations are for completeness. I don't have a good Windows guest to try running them with at the moment, even...)

### DIFF
--- a/phd-tests/framework/src/guest_os/alpine.rs
+++ b/phd-tests/framework/src/guest_os/alpine.rs
@@ -31,4 +31,12 @@ impl GuestOs for Alpine {
             crate::serial::BufferKind::Raw,
         )
     }
+
+    fn graceful_reboot(&self) -> CommandSequence {
+        // For Alpine guests we've looked at, `reboot` kicks off OpenRC behavior
+        // to reboot the system. We *could* wait for a new shell prompt at this
+        // point, but it's more reliable to wait for a guest to have fully
+        // rebooted and log back in.
+        self.shell_command_sequence("reboot")
+    }
 }
--- a/phd-tests/framework/src/guest_os/debian11_nocloud.rs
+++ b/phd-tests/framework/src/guest_os/debian11_nocloud.rs
@@ -24,4 +24,16 @@ impl GuestOs for Debian11NoCloud {
     fn read_only_fs(&self) -> bool {
         false
     }
+
+    fn graceful_reboot(&self) -> CommandSequence {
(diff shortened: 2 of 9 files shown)


gjcolombo (approved):

Issue #781, opened 2024-10-04
`FAILED: writing efi produced unexpected output: -bash: printf: write error: Invalid argument` in `guest_can_adjust_boot_order` test with Debian 11 guest


gjcolombo:
Repro steps: run boot order tests with a Debian 11 image against commit 1ae1ee3b125d7a7a1180c5d086c763fc2a9ac6c9.

Expected: `guest_can_adjust_boot_order` passes.

Observed: the guest doesn't like the test's attempts to manipulate the boot order through sysfs:

```
2024-10-04T04:28:12.135106Z  INFO phd_runner::execute: 129: test phd_tests::boot_order::guest_can_adjust_boot_order ... FAILED: writing efi produced unexpected output: -bash: printf: write error: Invalid argument                                                                                                         
                                                                                                                                                                                                                                                                                                                             
Stack backtrace:                                                                                                                                                                                                                                                                                                             
   0: anyhow::error::<impl anyhow::Error>::msg                                                                                                                                                                                                                                                                               
             at /home/greg/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.89/src/backtrace.rs:27:14                                                                                                                                                                                                         
   1: phd_tests::boot_order::efi_utils::write_efivar::{{closure}}                                                                                                                                                                                                                                                            
             at ./phd-tests/tests/src/boot_order/efi_utils.rs:385:9                                                                                                                                                                                                                                                          
   2: phd_tests::boot_order::guest_can_adjust_boot_order::{{closure}}::{{closure}}                                                                                                                                                                                                                                           
             at ./phd-tests/tests/src/boot_order.rs:353:61
   3: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
             at /home/greg/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tracing-0.1.40/src/instrument.rs:321:9
   4: phd_tests::boot_order::guest_can_adjust_boot_order::{{closure}}
             at ./phd-tests/tests/src/boot_order.rs:267:1
```

At the end of the serial console output I see the following:

```
root@debian:~# mount | grep efivarfs
efivarfs on /sys/firmware/efi/efivars type efivarfs (rw,nosuid,nodev,noexec,relatime)
root@debian:~# rm cmd
rm: cannot remove 'cmd': No such file or directory
[...]

PR #786, opened 2024-10-10
PHD: write efivars in one go

iximeow:
on Alpine and Ubuntu 22/24 the builtin `printf` seems to be enough to
get a write the full input to somewhere in efivarfs, but on Debian 11 a
`\n` in the output will cause an early partial write. If the data being
written is easily determined to be an invalid value by the kernel (5.10
in the Debian 11 case), that write will fail with EINVAL, and in fact
the UEFI variable will not be written at all.

From a brief read, I believe this to be more related to a change in Bash
between Debian 11 and Ubuntu 22, than a change on the kernel side. At
least on Debian 11 I observed `printf <data> > /sys/firmware/efi/...`
producing two write calls, rather than one, separated exactly at an
`0x0a` in the input. And I've not found any obviously-related kernel
changes since 5.10 that would make a pair of writes valid if they
occurred.

Ensure that we write a full UEFI variable's worth of data at once by
applying `dd` with output block size set high enough to hold variable's
new data.

With #785, this fixes https://github.com/oxidecomputer/propolis/issues/781

Tested with Ubuntu 22.04, Alpine, and Debian 11 guests.

Based off of #785 because the two together are what fixes #785, but a separate PR because I don't want these two distinct fixes to get squashed together :crying_cat_face:


phd-tests/tests/src/boot_order/efi_utils.rs
```diff
@@ -363,20 +363,32 @@ pub(crate) async fn write_efivar(
...
+    // printf "\xAA\xAA\xAA\xAA\xDD\xDD\xDD\xDD" | \
+    //   dd obs={inlen} of=/sys/firmware/efi/efivars/... status=none
     // ```
     // where AAAAAAAA are the attribute bytes and DDDDDDDD are caller-provided
     // data.
+    //
+    // notably do not printf directly to /sys/firmware/efi/efivars/*!! printf
+    // may flush output early if the data to write contains a `\n` (observed at
+    // least on Debian 11), and such a partial write to efivars may be rejected
+    // as invalid UEFI variable data.
```
gjcolombo, line 379: Is this something that's induced by our current long-command handling? I'm trying to decide how much I should try to suppress the urge to overhaul PHD's serial console and `run_shell_command` abstractions (in my copious free time, you understand).

gjcolombo (approved):

phd-tests/tests/src/boot_order/efi_utils.rs
iximeow: fully unrelated! i'm really curious why `printf` is now sufficient, or what made it insufficient in the past, but it's definitely a shell behavior that's flushing a partial line. for a very bad time, you can see shadows of this in search results about people having issues with `printf blah > /<efivarfs>/whatever` returning `invalid argument` sporadically, and lots of very strange suggestions on how to "fix the error".. i got very "how did this ever work" about it once i figured out the bug.

the long-command handling just sneaks a long shell line into the guest without prompting a `\n` back, so it'd still happen even if that were fixed, and i don't think we'd use much different commands in that world either. but the `run_long_command` stuff is very much a gross hack i'm urged to fix also.

Issue #787, opened 2024-10-10
firmware: want auto-added boot options to always be before EFI shell


iximeow:
i think i can't add issues to oxidecomputer/EDK2 because it's a fork rather than a repo in its own right, so filing this here instead.

this is, fundamentally, the misbehavior that is noted in [Omicron#5112](https://github.com/oxidecomputer/omicron/issues/5112): a boot device's description changed, its previous entry was removed, a new entry was added, and that new entry was after the EFI shell in boot order. we should make sure the `EFI Internal Shell` option is close to the last option, which will help keep guests from auto-discovering their way into not booting.

it seems extremely unlikely that someone would want to end up at the EFI shell specifically - in almost every case someone ends up in the EFI shell rather than their expected OS, it comes with a question of, "oh no, what did i break, and can i fix it?". now that users can specify a boot disk, and that boot order takes precedence over existing UEFI boot option variables, it is at least fixable! but it seems like a misbehavior to end up in that situation in the first place. if a user _did_ want to get to the EFI shell, it probably would be through a procedure like booting a guest OS, modifying the UEFI BootOrder variable to move the EFI shell up, and then rebooting. or booting an instance with no disks.

in practice this would probably be some change to EDK2 around where the [`EFI Internal Shell`](https://github.com/oxidecomputer/edk2/blob/edd72391086a9a297bf9aa011fe53b449165e556/OvmfPkg/Library/PlatformBootManagerLib/BdsPlatform.c#L1529-L1539) option is added, but i couldn't figure out an obvious tiny change this morning.

overall this seems like pretty low priority, but nice to have.

PR #788, opened 2024-10-10
add a test about boot option descriptions

iximeow:
ensuring we don't inflict [Omicron#5112](https://github.com/oxidecomputer/omicron/issues/5112) on instances due to our own changes: boot option descriptions are load-bearing, so if we've changed them make sure we think through the consequences. i'm pleasantly surprised that `EfiLoadOption` just.. worked here. yay.

the procedure that gets us names we see for guest disks today is a bit subtle, but i've tried to document it fully to pair with the expectation in this test.

there are also specific conditions where failing this assertion is fine - say, if we've done https://github.com/oxidecomputer/propolis/issues/787, or changed how we're processing UEFI boot variables. just, as things are right now, we're surprisingly dependent on dancing with OVMF in exactly the same way every time a guest boots.

### DIFF
--- a/phd-tests/tests/src/boot_order.rs
+++ b/phd-tests/tests/src/boot_order.rs
@@ -525,6 +525,9 @@ async fn nvme_boot_option_description(ctx: &Framework) {
     cfg.boot_order(vec!["boot-disk", "nvme-test-disk"]);
 
     let mut vm = ctx.spawn_vm(&cfg, None).await?;
+    if !vm.guest_os_kind().is_linux() {
+        phd_skip!("boot option description test depends on efivarfs");
+    }
     vm.launch().await?;
     vm.wait_to_boot().await?;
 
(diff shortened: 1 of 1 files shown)


gjcolombo (approved):
This is both a good test for us to have as we (or, well, I) change how NVMe device parameters work and a great description of how our firmware image actually constructs these IDs. Assuming this still passes with the latest HEAD we should merge it!

iximeow:
finally coming back to this: yep! gave this a run rebased on current `master` and the test still passes on Alpine, Debian, Ubuntu. ran it with a Windows guest on a lark and realized that this was missing a `phd_skip!()` for Windows guests, so i've added that. once CI's happy, in it oges

iximeow:
the `phd-build` failure in https://github.com/oxidecomputer/propolis/commit/392b4b439b0d6e64e4ca4bae655d1d80341876df is because... after i added the `is_linux()` clause to the new test, i rebased that small diff back onto this old branch so i'd have a nice neat tiny diff between what Greg looked at and what i'm gonna merge. `is_linux()`, however, is newer than this test.

so i've just rebased onto master, and it'll _actually_ pass now :sweat:

Issue #328, opened 2023-03-02
saw "TSC found unstable after boot, most likely due to broken BIOS"


jmpesp:
one of my guests is showing some dmesg messages relating to clock skew:

```
[ 8996.144253] clocksource: timekeeping watchdog on CPU2: Marking clocksource 'tsc' as unstable because the skew is too large:
[ 8996.144260] clocksource:                       'acpi_pm' wd_now: f4b7d wd_last: 6c50c5 mask: ffffff
[ 8996.144261] clocksource:                       'tsc' cs_now: 1ca324f62cac cs_last: 1ca0b7a50c00 mask: ffffffffffffffff
[ 8996.144274] tsc: Marking TSC unstable due to clocksource watchdog
[ 8996.144401] TSC found unstable after boot, most likely due to broken BIOS. Use 'tsc=unstable'.
[ 8996.144403] sched_clock: Marking unstable (8996101633485, 42749061)<-(8996259532949, -115136521)
[ 8996.145058] clocksource: Checking clocksource tsc synchronization from CPU 5.
[ 8996.146113] clocksource: Switched to clocksource acpi_pm
```

jmpesp:
guest is

```
Linux inst4 5.10.0-18-amd64 #1 SMP Debian 5.10.140-1 (2022-09-02) x86_64 GNU/Linux
```

and the propolis version is 6282016343573335026e1f1adb5d4dedbcd8b079

jmpesp:
if I grep out all lines in the dmesg that contain the string `clock`:

```
[    0.038996] clocksource: refined-jiffies: mask: 0xffffffff max_cycles: 0xffffffff, max_idle_ns: 7645519600211568 ns
[    0.113658] clocksource: tsc-early: mask: 0xffffffffffffffff max_cycles: 0x3259dcb07dd, max_idle_ns: 440795231854 ns
[    0.845735] clocksource: jiffies: mask: 0xffffffff max_cycles: 0xffffffff, max_idle_ns: 7645041785100000 ns
               * this clock source is slow. Consider trying other clock sources
[    0.954091] clocksource: Switched to clocksource tsc-early
[    0.973840] clocksource: acpi_pm: mask: 0xffffff max_cycles: 0xffffff, max_idle_ns: 2085701024 ns
[    1.378022] rtc_cmos 00:00: setting system clock to 2023-03-02T15:48:22 UTC (1677772102)
[...]

PR #791, opened 2024-10-14
lib: add a sink for components to register to handle RDMSR/WRMSR

gjcolombo:
Define a small wrapper around an `ASpace` that allows other Propolis components to register to handle RDMSR/WRMSR instructions directed to specific parts of the MSR "address space." Handlers have the option of fully handling an access, directing the caller to inject #GP, or returning an error indicating the trap couldn't be fully handled. This last option is useful when handling an MSR access requires the handler to call some other, fallible API, like a bhyve ioctl.

Create an `MsrSpace` in each `Machine`, give each `Vcpu` a reference to it, and have `Vcpu::process_vmexit` dispatch MSR operations to its MSR space. MSR handlers are the first exit type that can return a library error, so expand `process_vmexit` to return a `Result<Option<VmEntry>>`, which allows propolis-server/propolis-standalone to distinguish "lib didn't handle the exit" from "lib raised an error" and to handle these cases separately. For MSRs, the propolis-server discipline is to panic on errors, drop unhandled writes, and return 0s for unhandled reads.

Define a `VcpuId` newtype that (as you'd expect) represents a vCPU identifier to distinguish vCPU IDs from MSR IDs. The MSR handler callback uses this type, but its use in the rest of Propolis is left for another PR.

Tests: ran a propolis-server binary, enabled the USDT probes, booted a guest, and checked that the probes produced the expected results. I've also tested this logic with additional changes that use it to provide an enlightenment interface.

Related to #328 (our most likely paths to a paravirt clock require MSR handling to allow the guest to set up the enlightenment).

### DIFF
--- a/bin/propolis-server/src/lib/vcpu_tasks.rs
+++ b/bin/propolis-server/src/lib/vcpu_tasks.rs
@@ -124,7 +124,17 @@ impl VcpuTasks {
                 Ok(exit) => exit,
             };
 
-            entry = vcpu.process_vmexit(&exit).unwrap_or_else(|| {
+            let maybe_entry = match vcpu.process_vmexit(&exit) {
+                Ok(entry) => entry,
+                Err(e) => {
+                    panic!(
+                        "unhandled library error processing VM exit \
+                        {exit:?}: {e}"
+                    )
+                }
+            };
+
+            entry = maybe_entry.unwrap_or_else(|| {
                 match exit.kind {
                     VmExitKind::Inout(pio) => {
                         debug!(&log, "Unhandled pio {:x?}", pio;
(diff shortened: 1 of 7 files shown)


gjcolombo:
@pfmooney and I discussed this a bit offline, specifically whether using `ASpace` here is sandblasting a soup cracker as opposed to just having a `match` or a simple `if` statement. I'm mulling this over, but I think that's probably right. `ASpace` is nice for distributing I/O regions to devices, where the mapping of devices to I/O space depends on what devices exist and what properties they have. MSRs are not like that, though--there are likely to be very few handlers for them, and we're likely to know what they all are in advance.

I'm going to kick this back to Draft while I play around with it a bit more.

gjcolombo:
The main purpose of this was to enable hypervisor enlightenments, and that's since been done in other ways (see #846), so I'm closing this as obsolete.

PR #794, opened 2024-10-17
add marker trait to help check safety of guest memory reads

iximeow:
we noted that a pointer into guest memory must point to a properly-initialized T when read into Propolis, but there was no way to actually check that was a case. for example, it may be tempting to write an enum describing states of a guest device like:
```
enum MyCoolDevicePower {
  Off = 0,
  On = 1,
}
```
and read/write to guest memory using the convenient read/write helpers. but a devious guest could put a `2` at that address, where reading that into Propolis would be UB.

`zerocopy::FromBytes` happens to have the same requirements about its implementors as we need, that they're always valid to view from bytes, so use it to check that we can safely read a type out of guest memory. in our case we'll always copy those bytes to our own buffer, but `zerocopy::FromBytes` also comes with a great proc macro so we can [`#[derive(FromBytes)]`](https://docs.rs/zerocopy/latest/zerocopy/derive.FromBytes.html) on structs to be copied out.

### DIFF
--- a/lib/propolis/src/hw/nvme/bits.rs
+++ b/lib/propolis/src/hw/nvme/bits.rs
@@ -5,11 +5,12 @@
 #![allow(dead_code)]
 
 use bitstruct::bitstruct;
+use zerocopy::{FromBytes, FromZeroes};
 
 /// A Submission Queue Entry as represented in memory.
 ///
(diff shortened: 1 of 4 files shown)


pfmooney (approved):
Nice added safety

Issue #755, opened 2024-09-01
Propolis server panicked after hitting some `????????` rust_panic call stack


askfongjojo:
@augustuswm found a VM that had a propolis server panic. Here are the propolis log lines at the time:
```
18:41:28.222Z INFO propolis-server (vcpu_tasks): vCPU released from hold
    vcpu = 0
18:41:28.222Z INFO propolis-server (vm_state_worker): publishing new instance state
    gen = 9
    migration = InstanceMigrateStatusResponse { migration_in: None, migration_out: None }
    state = Running
18:41:28.222Z INFO propolis-server (vm_state_worker): State worker handled event
    outcome = Continue
18:41:28.222Z INFO propolis-server (vcpu_tasks): vCPU released from hold
    vcpu = 1
18:41:28.269Z INFO propolis-server: request completed
    latency_us = 55
    local_addr = [fd00:1122:3344:116::1:a6]:12400
    method = GET
    remote_addr = [fd00:1122:3344:116::1]:52233
    req_id = 8c5001b9-e06d-4606-acae-3fe5ab9cfdb0
    response_code = 200
    uri = /instance/state-monitor
18:41:28.637Z INFO propolis-server (ramfb): ramfb change
    config = Config { addr: 3198525440, fourcc: 875713112, flags: 0, width: 800, height: 600, stride: 3200 }
    state = valid
18:41:28.637Z INFO propolis-server (ramfb): notifying
18:41:28.637Z INFO propolis-server (vnc-server): pixel format set to fourcc=0x34325258
thread 'vcpu-0' panicked at bin/propolis-server/src/lib/vm/mod.rs:386:9:
vCPU 0: Unhandled VM exit: InstEmul(InstEmul { inst_data: [175, 175, 175, 175, 175, 175, 175, 175, 175, 175, 175, 175, 175, 175, 175], len: 15 })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ Aug 12 18:41:34 Stopping because all processes in service exited. ]
[ Aug 12 18:41:34 Executing stop method (:kill). ]
```
Both the core file and propolis log can be found in `/staff/rack3/BRM42220077/2024-08-29` (`system-illumos-propolis-server:default.log.1723488297` and `core.oxz_propolis-server_cdcb2537-a5ee-4e9d-97d8-d8a86b57d2dd.propolis-server.23389.1723488093`.

@leftwo got the stack from the core file:
```
[...]

Issue #335, opened 2023-03-09
Better error reporting mechanism needed


pfmooney:
When running a guest, there are certain conditions which may cause one or more of the vCPUs to exit from VM context with an event we are unable to properly handle.  Notable examples of this include #333 and #300, where the guests appear to have jumped off into "space", leaving the instruction fetch/decode/emulate machinery unable to do its job.

The course of action we choose in such situations can have certain trade-offs.  The current behavior of `abort`ing the propolis process has the advantage of attempting to preserve the maximum state of both the userspace emulation (saved in the core) as well as the kernel VMM and instance state (residing in the vmm device instance, as long as it is not removed).  This may be beneficial for developer applications, but for running hosts in production, it is likely less than ideal.

Consumers in production likely expect a VM encountering an error condition like that to reboot, as if it had tripped over something like a triple-fault on a CPU.  Rebooting the instance promptly at least allows it to return to service quickly.  In such cases, we need to think about what bits of state we would want preserved from the machine and the fault conditions so it can be used for debugging later.  In addition to the details about the vm-exit on the faulting vCPU(s), we could export the entire emulated device state (not counting DRAM) as if a migration were occurring.  Customer policy could potentially choose to prune that down, or even augment it with additional state from the guest (perhaps the page of memory underlying `%rip` at the time of exit?)

With such a mechanism in place, we could still preserve the abort-on-unhandled-vmexit behavior if it is desired by developer workflows, but default to the more graceful mechanism for all other cases.

hawkw:
There's work currently in progress in the control plane to allow Nexus to automatically restart instances whose `propolis-server` has crashed (if configured to do so). In particular, oxidecomputer/omicron#6455 moves instances to the `Failed` state when their VMM has crashed, and oxidecomputer/omicron#6503 will add a RPW for restarting `Failed` instances, if they have an "auto-restart" configuration set. 

Potentially, we could leverage that here and just allow `propolis-server`s that encounter this kind of guest misbehavior to crash and leave behind a core dump, knowing that the control plane will restart the instance if that's what the user wanted. On the other hand, this is potentially less efficient than restarting the guest within the same `propolis-server`, since it requires the control plane to spin up a whole new VMM and start the instance there. But, I figured it was worth mentioning!

pfmooney:
[...]

Issue #333, opened 2023-03-08
Just after boot a focal image panics propolis


leftwo:
Running latest Omicron (on sn21) I encountered a panic in propolis-server after a first boot of a newly created Ubuntu Focal instance.

The tail of the propolis-server log looks like this:
```
Dec 28 03:06:02.914 INFO piix3pm ignored cfg read, register: SmbusBase, offset: 0, dev: chipset
Dec 28 03:06:02.914 INFO piix3pm ignored cfg read, register: DevResA, offset: 0, dev: chipset
Dec 28 03:06:02.915 INFO piix3pm ignored cfg read, register: DevResB, offset: 0, dev: chipset
Dec 28 03:06:02.915 INFO piix3pm ignored cfg read, register: DevResC, offset: 0, dev: chipset
Dec 28 03:06:02.915 INFO piix3pm ignored cfg read, register: DevResI, offset: 0, dev: chipset
Dec 28 03:06:02.915 INFO piix3pm ignored cfg read, register: DevResJ, offset: 0, dev: chipset
Dec 28 03:06:03.125 INFO rdmsr, msr: 3221225705, vcpu: 0, component: vcpu_tasks
Dec 28 03:06:43.620 WARN [1] flow control start 
Dec 28 03:06:44.621 WARN [1] flow control sending more work
Dec 28 03:06:44.621 WARN [1] flow control end 
thread 'vcpu-2' panicked at 'vCPU 2: Unhandled VM exit: InstEmul(InstEmul { inst_data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], len: 0 })', bin/propolis-server/src/lib/vm/mod.rs:450:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ Dec 27 19:06:46 Stopping because all processes in service exited. ]
[ Dec 27 19:06:46 Executing stop method (:kill). ]
```

The stack of the thread that panic'd is:
```
-------------- thread# 140 / lwp# 140 [vcpu-2] ---------------                              
 fffff5ffef15863a _lwp_kill () + a                                                          
 fffff5ffef0ec58e raise (6) + 1e                                                            
 fffff5ffef0c5ff8 abort () + 58                                                             
 0000000001d2e7c9 panic_abort::__rust_start_panic::abort::hc1d5f3609ef6507d () + 9          
 0000000001d2e7b9 __rust_start_panic () + 9                                                 
[...]

PR #795, opened 2024-10-17
lib: log vCPU diagnostics on triple fault and for some unhandled exit types

gjcolombo:
Add a `propolis::vcpu::Diagnostics` type that captures and pretty-prints the register state of a vCPU. Log this state if a vCPU triple faults or (in propolis-server) if it raises a `Paging` or `InstEmul` unhandled exit, which are the failure modes seen in #300, #333, #340, and #755.

To avoid logging guest data outside of development environments, the `Diagnostics` type only captures data if the `dump-guest-state` feature is enabled. Enable this feature by default for propolis-server and -standalone, then modify the Omicron image build job to disable it in Omicron zone images for now (so that the feature will be disabled for customer VMs). In the fullness of time we may want a more flexible way to configure this option.

Finally, hand-implement `Debug` for the `InstEmul` exit context to redact this exit's instruction bytes if `dump-guest-state` is disabled.

This helps with #335. In a subsequent change I'd like to add the memory region near guest %rip to the diagnostics, but this requires some additional plumbing to translate from a GVA to a GPA before trying to read memory (there's an ioctl for this already; just needs some track-laying in Propolis).

Tests: added a diagnostics log line to the "reset" suspend exit so that I could force diagnostics to be captured on demand via bhyvectl, then checked both that they're logged with a binary built in the default configuration and elided in a binary built with the Omicron configuration.

### DIFF
--- a/.github/buildomat/jobs/image.sh
+++ b/.github/buildomat/jobs/image.sh
@@ -28,16 +28,12 @@ rustc --version
 
 banner build
 
-# This job produces the propolis-server zone image that the Omicron repository
-# imports to create full-blown Oxide software packages. Disable all default
-# features and opt into the "omicron-build" feature instead to build libraries
-# in their "production" configurations.
+# Enable the "omicron-build" feature to indicate this is an artifact destined
+# for production use on an appropriately configured Oxide machine
 #
 # The 'release' profile is configured for abort-on-panic, so we get an
 # immediate coredump rather than unwinding in the case of an error.
-ptime -m cargo build --release --verbose -p propolis-server \
-    --no-default-features \
-    --features omicron-build
+ptime -m cargo build --release --verbose -p propolis-server --features omicron-build
 
 banner image
 ptime -m cargo run -p propolis-package
--- a/.github/buildomat/jobs/image.sh
+++ b/.github/buildomat/jobs/image.sh
@@ -28,12 +28,16 @@ rustc --version
 
 banner build
 
-# Enable the "omicron-build" feature to indicate this is an artifact destined
-# for production use on an appropriately configured Oxide machine
+# This job produces the propolis-server zone image that the Omicron repository
+# imports to create full-blown Oxide software packages. Disable all default
+# features and opt into the "omicron-build" feature instead to build libraries
+# in their "production" configurations.
 #
 # The 'release' profile is configured for abort-on-panic, so we get an
 # immediate coredump rather than unwinding in the case of an error.
-ptime -m cargo build --release --verbose -p propolis-server --features omicron-build
+ptime -m cargo build --release --verbose -p propolis-server \
+    --no-default-features \
+    --features omicron-build
 
 banner image
 ptime -m cargo run -p propolis-package
--- a/bin/propolis-server/src/lib/vcpu_tasks.rs
+++ b/bin/propolis-server/src/lib/vcpu_tasks.rs
@@ -159,14 +159,36 @@ impl VcpuTasks {
                         VmEntry::Run
                     }
                     VmExitKind::Suspended(SuspendDetail { kind, when }) => {
+                        use propolis::vcpu::Diagnostics;
                         match kind {
                             exits::Suspend::Halt => {
+                                slog::info!(
+                                    &log,
+                                    "halt event on vcpu {}",
+                                    vcpu.id;
+                                    "state" => %Diagnostics::capture(vcpu)
+                                );
+
                                 event_handler.suspend_halt_event(when);
                             }
                             exits::Suspend::Reset => {
+                                slog::info!(
+                                    &log,
+                                    "reset event on vcpu {}",
+                                    vcpu.id;
+                                    "state" => %Diagnostics::capture(vcpu)
+                                );
+
                                 event_handler.suspend_reset_event(when);
                             }
                             exits::Suspend::TripleFault(vcpuid) => {
+                                slog::info!(
+                                    &log,
+                                    "triple fault on vcpu {}",
+                                    vcpu.id;
+                                    "state" => %Diagnostics::capture(vcpu)
+                                );
+
                                 if vcpuid == -1 || vcpuid == vcpu.id {
                                     event_handler
                                         .suspend_triple_fault_event(vcpu.id, when);
--- a/bin/propolis-server/src/lib/vcpu_tasks.rs
+++ b/bin/propolis-server/src/lib/vcpu_tasks.rs
@@ -215,13 +215,20 @@ impl VcpuTasks {
                         event_handler.unhandled_vm_exit(vcpu.id, exit.kind);
                         VmEntry::Run
                     }
+                    // Bhyve emits the `Bogus` exit kind when there is no actual
+                    // guest exit for user space to handle, but circumstances
+                    // nevertheless dictate that the kernel VMM should exit to
+                    // user space (e.g. a caller requested that all vCPUs be
+                    // forced to exit to user space so their threads can
+                    // rendezvous there).
+                    //
+                    // `process_vmexit` should always successfully handle this
+                    // exit, since it never entails any work that could fail to
+                    // be completed.
                     VmExitKind::Bogus => {
-                        error!(log,
-                               "lib returned bogus exit from vCPU {}",
-                               vcpu.id);
-
-                        event_handler.unhandled_vm_exit(vcpu.id, exit.kind);
-                        VmEntry::Run
+                        unreachable!(
+                            "propolis-lib always handles VmExitKind::Bogus"
+                        );
                     }
                     VmExitKind::Debug => {
                         error!(log,
--- a/bin/propolis-server/src/lib/vcpu_tasks.rs
+++ b/bin/propolis-server/src/lib/vcpu_tasks.rs
@@ -162,23 +162,9 @@ impl VcpuTasks {
                         use propolis::vcpu::Diagnostics;
                         match kind {
                             exits::Suspend::Halt => {
-                                slog::info!(
-                                    &log,
-                                    "halt event on vcpu {}",
-                                    vcpu.id;
-                                    "state" => %Diagnostics::capture(vcpu)
-                                );
-
                                 event_handler.suspend_halt_event(when);
                             }
                             exits::Suspend::Reset => {
-                                slog::info!(
-                                    &log,
-                                    "reset event on vcpu {}",
-                                    vcpu.id;
-                                    "state" => %Diagnostics::capture(vcpu)
-                                );
-
                                 event_handler.suspend_reset_event(when);
                             }
                             exits::Suspend::TripleFault(vcpuid) => {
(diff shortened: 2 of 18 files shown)


bin/propolis-server/src/lib/vcpu_tasks.rs
```diff
@@ -187,7 +195,69 @@ impl VcpuTasks {
...
+                        VmEntry::Run
+                    }
+                    VmExitKind::Bogus => {
+                        error!(log,
+                               "lib returned bogus exit from vCPU {}",
+                               vcpu.id);
+
+                        event_handler.unhandled_vm_exit(vcpu.id, exit.kind);
+                        VmEntry::Run
+                    }
```
pfmooney, line 225: Bogus is not an unhandled exit

lib/propolis/src/exits.rs
```diff
@@ -93,11 +93,27 @@ impl From<&bhyve_api::vm_exit_vmx> for VmxDetail {
...
     pub len: u8,
 }
+
+impl std::fmt::Debug for InstEmul {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        #[cfg(not(feature = "dump-guest-state"))]
+        let inst_data = "<dump-guest-state disabled>";
+
+        #[cfg(feature = "dump-guest-state")]
+        let inst_data = &self.inst_data;
```
pfmooney, line 108: Rather than gating this output on a feature, perhaps just censor it on the propolis-server side, depending on the configuration there?

bin/propolis-server/src/lib/vcpu_tasks.rs
gjcolombo: True, it's unconditionally handled in the lib. It previously fell into the `_` match arm, which called `unhandled_vm_exit`, but as long the lib is always supposed to handle this exit kind it's probably better just to panic directly here (since this is really more of an assertion failure than a truly unhandled exit).

pfmooney: I would use `unreachable!`, then.  I think handling like this is misleading to the reader.

gjcolombo:
The clippy failure is a new suggestion from 1.82. I'll fix that separately.

lib/propolis/src/exits.rs
gjcolombo: I considered that but didn't like the ergonomics of it; it would mean that everyone in server who might want to debug-print a `VmExitKind` needs to remember to check what variant it is first to avoid printing an `InstEmul`. We could force that to happen by declining to derive Debug for VmExitKind, but that would put a big `match` into functions like `unhandled_vm_exit` to deal with the one variant that doesn't have a Debug impl, which I also don't love the look of. (I don't think there are any other call sites that would be affected, fortunately.)

bin/propolis-server/src/lib/vcpu_tasks.rs
gjcolombo: Done in fba85be. I also tried to describe what the bogus exit is for in a comment. LMK if it's not accurate; I'm going mostly off what I could infer from reading around where VM_EXITCODE_BOGUS is used in bhyve.

.github/buildomat/jobs/image.sh
```diff
@@ -28,12 +28,16 @@ rustc --version
...
+# imports to create full-blown Oxide software packages. Disable all default
+# features and opt into the "omicron-build" feature instead to build libraries
+# in their "production" configurations.
 #
 # The 'release' profile is configured for abort-on-panic, so we get an
 # immediate coredump rather than unwinding in the case of an error.
-ptime -m cargo build --release --verbose -p propolis-server --features omicron-build
+ptime -m cargo build --release --verbose -p propolis-server \
+    --no-default-features \
+    --features omicron-build
```
iximeow, line 40: it took me a second to connect the dots but from #755 and elsewhere we're clearly still collecting cores from a crashed Propolis - all we've done by disabling the logging is prevented ourselves from collecting more timely information (that in the case of #755 helped us tell the guest had gone off to space)

i think a consistent posture here would be either:

A: Propolis stderr/stdout is as sensitive as guest memory; do not collect, if you must collect then keep secret, if you must share, then do so carefully
B: keep Propolis stderr/stdout clean of user data, and when operating Propolis, do not retain user data (cores that could include pages of user data)

i'm inclined towards A on the grounds that if you're looking at or copying a Propolis log around, you probably also can access a core from that same process. and it means we're actually helping a situation like #755 with this change.

i still think the _feature_ is useful, just that we can leave default features enabled in this job. _or_ if we should consider ourselves responsible for keeping Propolis lib/standalone/server's logs free of potentially-sensitive bits, it'd be worth a note to ourselves that we should audit other logs for anything untoward - migration state comes to mind.

i don't want to be excessively black/white on this, but i'm trying to make sure there's a clear line for what is and isn't OK for Propolis to say. what do you think?

gjcolombo: I keep finding myself drawn to approach B. At some point we're going to have a situation where we need to debug a Propolis issue using only information from operator-provided support bundles. That's going to be tricky if operators feel they can't ever include Propolis logs in those bundles because Propolis is too cavalier about logging guest data.

> i'm inclined towards A on the grounds that if you're looking at or copying a Propolis log around, you probably also can access a core from that same process.

I would put it a bit differently: if you can *extract a Propolis log from a sled*, you can probably also extract core dumps from that sled. Because Oxide personnel are regularly in this position today (because we access deployed racks over the technician port in order to update/service them), I agree that redacting diagnostic data in logs currently has very little practical upside, at least from the point of view of an operator who wants to limit support personnel access to guest VM state. The long term goal, of course, is to wean ourselves off of the tech port via automated update, support bundles, and the like, and once we've done that I think redaction will be more obviously useful.

#755 is the only case I know of where we've seen one of the diagnostic-generating faults in a production Omicron deployment. All the other cases I know of have been in development environments (ad hoc Propolis server/standalone binaries or `a4x2` deployments). So my lean was (and still is, I think) toward disabling this by default in Omicron images, though I think it's a somewhat close call (made even closer by the fact that #755 happened on an Oxide-managed rack and not a customer-operated rack).

With all that said: WWYT of taking some steps now to make this into a runtime-configurable setting? I think that the ideal world is one @pfmooney laid out in chat in yesterday's HV huddle: if this behavior can change dynamically, and Nexus knows how to control it, then an operator who can reproduce a problem in a "clean" VM can enable extra data collection for just that instance (or its project/silo/what-have-you) and provide us with the resulting "full-fat" logs without worry. In the short run (i.e. before Omicron learns about the new API), and in a dire pinch, support personnel accessing a rack through the tech port can access the API, enabling a "user starts clean VM -> support person sets option -> user repros problem" approach if we desperately needed it. I think I can adjust this PR to move in that direction pretty easily (and I suspect it might improve its ergonomics in the process); will probably go ahead and give that a try, at least.

gjcolombo:
I'm going to rebase to pick up #794 and #796 before trying to tweak this PR further (#796 to silence the clippy warning, #794 because I have some Ideas that I think would otherwise conflict with it).

.github/buildomat/jobs/image.sh
iximeow: > wean ourselves off of the tech port via automated update, support bundles, and the like, and once we've done that I think redaction will be more obviously useful.

ah, yeah, in that context i see what you mean. i was over-indexing on today's diagnostics!

definitely a fan of making it runtime-configurable, then, but i'm even more glad to have piped up on RFD 496 - seems like we need a more nuanced treatment of the relationship between cores and support bundles.

lib/propolis/src/common.rs
```diff
@@ -11,13 +11,25 @@ use crate::vmm::SubMapping;
 
 /// Controls whether items wrapped in a [`GuestData`] are displayed or redacted
 /// when the wrappers are printed via their `Display` or `Debug` impls.
+#[no_mangle]
```
iximeow, line 14: (definitely worth a note like `// The #[no_mangle] here is not for internal correctness, but to ensure that two cohabitant lib-propolis versions produce an error rather than ambiguously work`)

gjcolombo: Done in 7334b8d.

gjcolombo:
OK, I think this is ready for another review lap; I'll tag folks back in presently.

I haven't added any API support for toggling the logging flag in this PR, mostly because I don't know what the plumbing is going to look like in Nexus, and I'm reluctant to have an unused internal API sitting around for an unknown length of time. I think this would be very, very straightforward to add, though, either as an instance ensure parameter or as a separate API that controls the behavior of the server process writ large.

lib/propolis/src/vmm/mem.rs
```diff
@@ -798,11 +797,14 @@ pub struct MemCtx {
 }
 impl MemCtx {
     /// Reads a generic value from a specified guest address.
-    pub fn read<T: Copy + FromBytes>(&self, addr: GuestAddr) -> Option<T> {
+    pub fn read<T: Copy + FromBytes>(
+        &self,
+        addr: GuestAddr,
+    ) -> Option<GuestData<T>> {
```
iximeow, line 803: fwiw this diff misses the same thing i did in #794, which is that `do_read` in `bhyve-api` reads from guest memory too. i was figuring that'd be a relatively small change to chase through this week but i wanted to go read the crate and bhyve side a bit first. if you don't get to it here, that future PR will be me doing "bring `MemCtx::read` improvements to `do_read`", if you do change that here too then that's also fine!

iximeow (approved):
seems good to me, big fan of a bit we can toggle for this. i mentioned the `bhyve-api` part but same as #794 i'm not pressed about that happening exactly right now as long as one of us gets to it sooner than later (e.g. i'm happy to make a note to wire that up this week)

lib/propolis/src/vmm/mem.rs
gjcolombo: My understanding of `do_read` is that it's used to read state from the in-kernel VMM--things like the state of in-kernel emulated devices, bhyve-managed MSRs, and captured VMX/SVM state that might need to be migrated. 
Some of that could technically have come from the guest (via WRMSR or an I/O instruction), but to my mind none of it is really guest *application* data of the sort we're looking to protect in this PR. (I'm going off the list of device state information classes in `bhyve-api/sys/src/vmm_data.rs` to draw this conclusion; there are `VDC_REGISTER` and `VDC_FPU` classes, but AFAICT these haven't actually been implemented for this bhyve interface. Everything else looks to me like internal kernel VMM state.)

lib/propolis/src/exits.rs
gjcolombo: Following up on this--the latest bits should be much cleaner now; they get rid of all the new propolis-lib feature checks in favor of having a flag that controls whether things tagged as "came from the guest" get redacted, and propolis-server uses the `omicron-build` compile time flag to decide how to set that value.

Issue #792, opened 2024-10-16
phd: lspci_lifecycle_test passes incorrectly


iximeow:
disclaimer: this is because of a framework bug, not a Propolis bug. i noticed that `lspci_lifecycle_test` fails on an Ubuntu 22.04 guest image i'd put together, and at first thought it was wrong, but the truth is stranger than fiction...

in `lspci_lifecycle_test` we run both `lspci` and `lshw`:
https://github.com/oxidecomputer/propolis/blob/93ed767388c4fab11af8a98ad33fdbeac4098b0c/phd-tests/tests/src/hw.rs#L21-L29

on an Ubuntu guest, the `lshw` assert fails because before and after messages don't match. the difference in the (rather large) strings of output is only that the machine's `serial` does not match after being stopped and started. i double-checked on a real instance, and a Debian 11 guest's observed value for `serial` is in fact the instance's ID, and that ID is stable across a stop and start. again, test bug not real bug.

the immediate issue in the test framework is that in [the test](https://github.com/oxidecomputer/propolis/blob/93ed767388c4fab11af8a98ad33fdbeac4098b0c/phd-tests/tests/src/hw.rs#L22) we validate that `lshw` and `lspci` agree across a `StopAndStart`, but that action involves [spawning a successor VM](https://github.com/oxidecomputer/propolis/blob/93ed767388c4fab11af8a98ad33fdbeac4098b0c/phd-tests/framework/src/lifecycle.rs#L64-L69) which [makes a new `TestVm`](https://github.com/oxidecomputer/propolis/blob/93ed767388c4fab11af8a98ad33fdbeac4098b0c/phd-tests/framework/src/lib.rs#L292-L296) and in turn [gets a new id](https://github.com/oxidecomputer/propolis/blob/93ed767388c4fab11af8a98ad33fdbeac4098b0c/phd-tests/framework/src/test_vm/mod.rs#L189-L194).

- [ ] we probably should just keep the same `TestVm` if _all_ we're doing is a `StopAndStart`. seems like this is the only test using `StopAndStart`, so that's simple enough.

why in the world did this pass with Alpine or Debian images though? i'm glad you've asked!
```
# on Alpine:
localhost:~# lshw
-ash: lshw: not found
localhost:~#

# on Debian:
root@debian:~# lshw
-bash: lshw: command not found
root@debian:~#
```

[...]

PR #797, opened 2024-10-18
phd: check exit status of commands, have tests expect success/failure

iximeow:
moderately in the vein of [std::process::Command](https://doc.rust-lang.org/std/process/struct.Command.html), though i figure trying to separate out stdout/stderr is _very_ much not worth the effort.

typically if a command fails or status-es a subsequent assert will fail, but in https://github.com/oxidecomputer/propolis/issues/792 we fail through to success. in the boot order tests with a Windows guest, the test likely fails through `dd | xxd -` and errors about being unable to unhex the output (though some tests fail for other reasons earlier)

so, hopefully this doesn't seem like too much tedium to layer on top of shell commands to double-check test commands are doing what we expect. but i'm not deeply attached to the change one way or the other!


iximeow:
oh the Rust checks don't run until a PR is actually opened, that explains why this passed without the clippy/1.82.0 patch. i was wondering how that could be..

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -868,7 +913,21 @@ impl TestVm {
...
+    // TO REVIEWERS: it would be really nice to write this as a function
+    // returning a `struct ShellCommandExecutor` that impls
+    // `Future<Output=String>` where the underlying ShellOutput is automatically
+    // `.expect_ok()`'d. In such a case it would be possible for the struct to
+    // have an `.expect_err()` that replaces teh default `.expect_ok()`
+    // behavior, so that the likely case doesn't need any change in PHD tests.
+    //
+    // unfortunately I don't know how to plumb the futures for that, since we'd
+    // have to close over `&self`, so doing any Boxing to hold an
+    // `async move {}` immediately causes issues.
```
iximeow, line 928: my thinking is that it'd be nice if `vm.run_shell_command("foo").await` automatically did what `expect_ok` does in this change, but if you know to expect an error you could `vm.run_shell_command("foo").expect_err().await` to modify the future and _then_ run it. concretely it looked something like:

```rust
struct ShellCommandExecutor {
    fut: Pin<Box<impl Future<Output=Result<ShellOutput>> + Send + Sync>>
}

impl ShellCommandExecutor {
    fn expect_err(self) -> ShellCommandExecutorExpectErr { /* build the thing */ }
}
impl Future for ShellCommandExecutor {
    fn poll(&self, cx: &mut Context) -> Self::Output {
        match self.fut.poll(cx) {
            Poll::Ready(v) => Poll::Ready(v.expect_ok()),
            Poll::Pending => Poll::Pending
        }
    }
}
```

but i couldn't figure out how to `impl Future` in a way that doesn't conflict with `fut` holding `&self`. so this patch does the next best thing i could think of.

phd-tests/tests/src/crucible/smoke.rs
```diff
@@ -37,6 +37,7 @@ async fn guest_reboot_test(ctx: &Framework) {
     vm.launch().await?;
     vm.wait_to_boot().await?;
 
+    // XXX: use graceful_reboot() now.
```
iximeow, line 40: (got a followup patch for this)

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -142,6 +142,51 @@ enum VmState {
...
+    /// with status zero.
+    pub fn expect_err(self) -> Result<String> {
+        if self.status != 0 {
+            Ok(self.output)
+        } else {
+            Err(anyhow!("command unexpectedly succeeded"))
+        }
+    }
+
+    pub fn ignore_status(self) -> String {
```
hawkw, line 173: I'm on the fence about this name; I do think it's nice that it makes it clear that you're ignoring the status, but I think it might be better named `into_output` or something, because "ignore_status" makes it clear what you're ignoring, but not what you're _getting_, which I think is useful to make clear at the callsite...

Also, I can write something like this:
```rust
let output: ShellOutput = ...;
if output.status > 0 {
   // ... 
}
do_something_with_output(output.ignore_status())
```
and i'm not *really* "ignoring" the status at all, i'm just handling it separately.

YES I KNOW THIS IS INCREDIBLY PEDANTIC OF ME AND IT DOESN'T ACTUALLY MATTER

phd-tests/framework/src/test_vm/mod.rs
hawkw: This feels like it's what [`IntoFuture`](https://doc.rust-lang.org/stable/std/future/trait.IntoFuture.html) is supposed to be for, right? "I want to have a thing which is not actually a `Future` implementation, but knows how to turn itself into a future, so you can still `.await` it and it will go and turn itself into a future".

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -142,6 +142,51 @@ enum VmState {
...
+/// Both the output and status of a command.
+pub struct ShellOutput {
+    status: u16,
+    output: String,
+}
+
+impl ShellOutput {
+    /// Consume this [`ShellOutput`], returning the command's output as text
+    /// if the command completed successfully, or an error if it did not.
+    pub fn expect_ok(self) -> Result<String> {
```
hawkw, line 154: it's a bit weird to me that `expect_ok()` returns a `Result` instead of panicking --- i would _generally_ expect (haha) a function named `expect_foo` to panic if the expectation is violated. Perhaps these would make more sense as just `.ok()`/`.err()`? I dunno. Maybe this is clearer to others...

hawkw (commented):
Overall, this looks great! I had a couple of goofy and obnoxious naming nits, and I think you might be able to solve the problem you described in "TO REVIEWERS" by using the `IntoFuture` trait.

phd-tests/framework/src/test_vm/mod.rs
iximeow: hum yeah that seems pretty reasonable. this naming _probably_ made a little more sense in the first draft where i was hoping this would configure a future to _expect the result is ok_ rather than this `goofy-Result-to-Result` transformation. i'll see if i can get something nice with `IntoFuture` and if not i'll `s/expect_/g`

phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -926,7 +926,6 @@ impl TestVm {
     // unfortunately I don't know how to plumb the futures for that, since we'd
     // have to close over `&self`, so doing any Boxing to hold an
     // `async move {}` immediately causes issues.
-    #[must_use]
```
hawkw: yeah, the `Result` makes this already `must_use`...but perhaps there should be a type-level `#[must_use]` on the `ShellOutput` struct itself?

phd-tests/framework/src/test_vm/mod.rs
iximeow: ok! now that i got the more Future-ful approach together i think this all is _much_ cleaner (and incidentally resolves your other comments). it's kinda sickrad that this only comes with like 10 changes in tests now, since it's going along with the general assumption that commands run in guests _should_ be successful.

phd-tests/framework/src/test_vm/mod.rs
iximeow: there's still an `ignore_result` but it's configuring the command execution Future now where i think the relationship is a lot less ambiguous. gonna resolve this but if the functions on `ShellOutputExecutor` still seem a bit weird please check me.

phd-tests/framework/src/test_vm/mod.rs
iximeow: `ShellOutput` is now fully gone :saluting_face:

hawkw:
CI still looks sad?

iximeow:
and in the changes that landed on `master` in the last several months, a few did change tests in a way that semantically conflicted with this branch, so it's good to refresh that anyway. my apologies for https://github.com/oxidecomputer/propolis/pull/797/commits/9a151ed8910d7a141c035a5b8bac6bd5d47cd5a9 but there is much to do that is... not debugging dmesg. if this proves flaky after all i'm happy to admit defeat on checking Hyper-V is present in CI or something.

PR #799, opened 2024-10-19
phd: gate OS-specific tests, make others more OS-agnostic

iximeow:
some smoketests instigated a graceful reboot by sending `reboot\n` to the serial console, but Windows guests are rewarded (via Cygwin) with

```
bash: reboot: command not found
```

as of #785 there is now a TestVm helper `graceful_reboot` that does this in whatever way the guest adapter desires, so use that instead.

unfortunately `graceful_reboot` is only mostly right for Windows guests, so fix that here too: `shutdown` immediately terminates the cmd.exe session, at which point the SAC redraws the previous screen, which happens to have all the sigils we look for to detect that Windows has freshly booted. we then log back into the imminently-shutdown Windows, shutdown takes effect, and PHD becomes fully desynchronized from the guest state.

so, look for "BdsDxe: loading " as an outside-the-guest-OS sigil that tells us we're watching a fresh boot. this comes from OVMF, so a booted guest will have clobbered the message and not know to redraw it even if it seeks to recreate a previous display.

other tests depend on Linux-specific features like `efivarfs` or `mount -o ro`, so skip them on non-Linux guests.

validated this with both a Debian guest and a Windows Server 2022 guest - Debian was and still is good, the 2k22 guest fail fewer tests and not any that it didn't fail before

### DIFF
--- a/phd-tests/framework/src/guest_os/windows.rs
+++ b/phd-tests/framework/src/guest_os/windows.rs
@@ -37,6 +37,22 @@ pub(super) fn get_login_sequence_for<'a>(
     ));
 
     let mut commands = vec![
+        // Look for `BdsDxe:` as a sign that we're actually seeing a fresh boot.
+        // This is not terribly important in the case of a first boot, but in a
+        // case such as logging out and waiting for reboot, exiting a cmd.exe
+        // session causes Windows to redraw its previous screen - everything
+        // past `Computer is booting, ...` below.
+        //
+        // A test that tries to boot and wait for a new login sequence would
+        // then incorrectly identify the already-booted VM as the freshly-booted
+        // OS it was waiting for, log in again, and at some point later finally
+        // actually reboot.
+        //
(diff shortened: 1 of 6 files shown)


phd-tests/framework/src/guest_os/windows.rs
```diff
@@ -37,6 +37,22 @@ pub(super) fn get_login_sequence_for<'a>(
     ));
 
     let mut commands = vec![
+        // Look for `BdsDxe:` as a sign that we're actually seeing a fresh boot.
+        // This is not terribly important in the case of a first boot, but in a
+        // case such as logging out and waiting for reboot, exiting a cmd.exe
+        // session causes Windows to redraw its previous screen - everything
+        // past `Computer is booting, ...` below.
```
gjcolombo, line 44: Oooof. Nice catch here.

gjcolombo (approved):
This had been on my "meant to do" list for a long while--thanks for patching this up!

PR #801, opened 2024-10-25
manually impl Deserialize for PciPath for validation purposes

### LINKED ISSUE #800: `PciPath` deserialization allows paths rejected by `PciPath::new`
The propolis-types crate's `PciPath` struct has a `new` function that validates that the requested path's device and function numbers are within valid ranges (0-31 and 0-7 respectively). These checks don't apply when creating a `PciPath` via its derived `Deserialize` impl. This breaks the Propolis lib's assumption that all `PciPath`s can be converted to its internal `Bdf` structures.

gjcolombo:
Fixes #800.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4469,6 +4469,7 @@ version = "0.0.0"
 dependencies = [
  "schemars",
  "serde",
+ "serde_json",
  "serde_test",
 ]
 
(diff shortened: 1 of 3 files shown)


pfmooney (approved):

gjcolombo:
Just for fun, I also wrote and ran an exhaustive version of the deserialization test (for each bus/device/function in 0..256, call `PciPath::new`, then format a JSON version of that path and see if deserialization produces a result that matches what `new` returned). This takes about a minute to run on my Zen 2 dev machine, which is slow enough that I'm reluctant to check it in, especially given that this code doesn't change very often. But wanted to note that this seems to work for all otherwise-well-formed BDF JSON blobs.

PR #803, opened 2024-10-29
api: clean up Instance and InstanceProperties types

### LINKED ISSUE #723: Make use of or remove unused fields in `InstanceProperties` and `Instance` types
These structs are part of the API and have long-unused fields. In `InstanceProperties` the `image_id` and `bootrom_id` fields are unused:

https://github.com/oxidecomputer/propolis/blob/99fa5647fa1a0baa747344ff7abab9f6e9155a8c/crates/propolis-api-types/src/lib.rs#L198-L217

And in `Instance` (the type returned by GET operations) the `disks` and `nics` fields are unused:

https://github.com/oxidecomputer/propolis/blob/99fa5647fa1a0baa747344ff7abab9f6e9155a8c/crates/propolis-api-types/src/lib.rs#L226-L233

These fields should either be used or removed. (In the latter case I'm uncertain whether or how the control plane would use this information.)

gjcolombo:
Clean up unused fields and structures in the server API:

- Remove the unused `image_id` and `bootrom_id` fields from `InstanceProperties`.
- Remove the `disks` and `nics` fields from the `Instance` struct returned by an instance GET and all their supporting types. These are also not used today. Callers who are interested in finding out about an instance's disks and NICs can query the instance spec GET endpoint instead.
- Remove the CPU count and memory size fields from `InstanceProperties` and put them in `InstanceEnsureRequest` instead. This eliminates an extra copy of these data in the server (they were available in both the properties and the instance spec; now they're just available in the spec).

N.B. This is a client API breaking change; requiescat, migrate-from-base tests. The live migration protocol is unaffected.

Tests: cargo test; PHD; deployed this change to an Omicron dev cluster and booted some instances there; launched and migrated a VM between local servers using propolis-cli.

Fixes #723.

### DIFF
--- a/bin/mock-server/src/lib/lib.rs
+++ b/bin/mock-server/src/lib/lib.rs
@@ -232,8 +232,6 @@ async fn instance_get(
     let instance_info = api::Instance {
         properties: instance.properties.clone(),
         state: instance.state,
-        disks: vec![],
-        nics: vec![],
     };
     Ok(HttpResponseOk(api::InstanceGetResponse { instance: instance_info }))
 }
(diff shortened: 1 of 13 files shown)


hawkw (approved):
You love to see it!

PR #806, opened 2024-11-12
server: consolidate VM creation APIs

### LINKED ISSUE #772: Prefer component IDs to names
from [this conversation](https://github.com/oxidecomputer/propolis/pull/756#discussion_r1774107838) in #756:

`propolis-server` gets disk requests from Nexus with disks named by [their name](https://github.com/oxidecomputer/omicron/blob/0c7fb27/nexus/src/app/instance.rs#L1069-L1070), likewise for NICs. these names are technically mutable, though such mutation should be disallowed by Nexus any time `propolis-server` might need to know. but at the same time, `propolis-server` derives identifiers from these names that then are used to identify those devices' _backends_, as [in `parse_disk_from_request`](https://github.com/oxidecomputer/propolis/blob/11371b0f3743f8df5b047dc0edc2699f4bdf3927/bin/propolis-server/src/lib/spec/api_request.rs#L76). VNICs do a similar derivation but with [PCI paths](https://github.com/oxidecomputer/propolis/blob/11371b0f3743f8df5b047dc0edc2699f4bdf3927/bin/propolis-server/src/lib/spec/mod.rs#L282-L287).

for disks, since `name` is essentially user-defined, it would be really nice to use the disk's ID instead - that has a much more structured shape that we control ("UUID", rather than "a string that Nexus hopefully validated enough for Propolis to handle responsibly").

while _propolis-server_ could handle getting UUIDs from Nexus in the `name` field, this would be a breaking change for users. the disk name ends up as part of the identifier for paths like `/dev/disk/by-id`, so changing the guest-presented name from the name to a UUID would break paths in VMs that expect `/dev/disk/by-id` to be immutable. it may well be a bug that `/dev/disk/by-id` can be changed by renaming a disk - such an operation probably should not be functionally similar to changing a serial number of a device..

we could easily punt on the user-facing breakage here by separating the disk name, which could reasonably be presented to guests. so maybe as an interim step we should get both the disk name _and_ ID from Nexus, and use ID in all places that would not be directly presenting the disk name to guest OSes.

for NICs, the PCI path is at less arbitrary user-defined input, but it seems like it would be nice to use a stable UUID there as well, since a VNIC's slot _could_ change.

it doesn't seem to me that other devices have a similar issue with user-provided input being used for internal identifiers, at least.

relatedly: @gjcolombo raised that we should think through what the IDs should be for other components like serial ports. whatever IDs get used for disks and others we should make sure that [`GET /instance/spec`](https://github.com/oxidecomputer/propolis/pull/756#discussion_r1775626960) is informative for debugging (might already be, i just dunno tests might apply here)

### LINKED ISSUE #776: server: machine init is inconsistent about the names it puts in its DeviceMap
Observed while reading this code for other purposes.

`struct MachineInitializer` has a `DeviceMap` field into which it inserts the devices that will receive instance lifecycle notifications. These names are important in part because they are used as keys to associate components with each other during live migration. This means that the source and target must always agree on what gets put in the `instance_name` field of each `migrate::Device` that gets generated when a source serializes its device state.

Some `MachineInitializer` functions (e.g. `initialize_uart`) use the component names in the instance's spec as keys in the `DeviceMap`, but most of them (e.g. storage/network devices) generate a name from other properties of the component (e.g. its PCI BDF). The chosen names are derived from component properties like PCI BDFs that are kept stable across migrations, but it's still possible for the source and target to disagree on how these names should be derived, which breaks migration.

In a world where components an instance spec have strong unique identifiers (see #772) and where migration targets might get their device configurations from their sources, it seems like it would also be good to standardize on these IDs: the ID a component has in the spec is also the ID it uses when pairing payloads to components during a migration (with logging affordances to easily match IDs to logical devices if we're debugging a migration issue).

### LINKED ISSUE #804: two instance ensure APIs is one too many
Today the server API offers two distinct ways to initialize a new VM:

- The caller can pass an `InstanceEnsureRequest` that specifies some, but not all, of the components it wants the new VM to have. Propolis then supplements this configuration with some default devices (e.g. serial ports), converts some of the information in the ensure request into concrete configuration data (e.g. slot numbers to PCI BDFs), and adds any devices specified in the config TOML passed to the server on its command line.
- Alternatively, the caller can pass a `VersionedInstanceSpec` that fully specifies all the components the new VM should have. The server does not add any extra devices by default and ignores the device list in the config TOML (though the TOML is still used to specify the bootrom path). 

In both cases, the server's first task is to try to convert its ensure parameters into a server-internal `Spec` structure that describes the new VM's components and their configuration. If this succeeds, machine initialization uses the resulting `Spec` to set up the new VM.

Both of these APIs accept an optional request to initialize via live migration, which contains the address of the migration source Propolis. If this is supplied:

- The target creates its internal `Spec` as usual, but before creating any VM components, it starts the live migration protocol.
- At the beginning of the protocol, the source Propolis sends its `VersionedInstanceSpec` to the target Propolis.
- The target compares this spec to its own `Spec` to make sure the two specs describe "migration-compatible" VMs; that is, the two specs must describe the same guest-visible abstractions and configuration. If they don't, migration fails.
- If the compat check passes, the target creates its VM components, then runs the rest of the LM protocol to get a copy of the source's component state. The target imports this state into its components, then resumes the VM.

Two APIs is one too many. Having two APIs is a maintenance headache, especially as we continue adding new Propolis configuration options (boot order and CPUID landed recently; configurable MSR dispositions and hypervisor enlightenments are on the horizon). We should consolidate down to a single VM creation API.

---

I think a better API would look something like this:

```rust
enum ReplacementComponent {
    // one variant for each ComponentV0 variant that can be replaced
    // during migration; see below
}

enum InstanceInitMethod {
    New {
        // N.B. doesn't need to be versioned since the client, sled agent, is
        // updated in lockstep with Propolis
        spec: InstanceSpecV0
    },
    MigrationTarget {
        migration_id: Uuid,
        source_addr: SocketAddr,
        replacement_components: HashMap<InstanceSpecKey, ReplacementComponent>
    }
}

struct InstanceEnsureRequest {
    // same as the existing instance properties struct
    properties: InstanceProperties,
    init: InstanceInitMethod
}
```

To create a new VM, a client creates a spec for it and passes an `InstanceInitMethod::New` in its ensure request. This will require Omicron to switch from using instance ensure requests to instance specs. This should mostly be a matter of moving Propolis's ensure-request-to-instance-spec logic up into Nexus; it shouldn't require any database schema changes. One additional benefit of doing this (besides simplifying Propolis's API) is that putting this logic in Nexus makes it easier to update. RFD 505 talks about all the tradeoffs of this approach in more detail.

In the new API, the migration interface doesn't take a full instance spec. Instead, the target Propolis takes the source spec that it gets from the migration protocol, replaces any entries therein that have entries in the caller-supplied `replacement_components` map, and uses the amended spec to initialize the new VM. The replacement map allows Nexus to e.g. specify new Crucible volume construction requests with new client generation numbers, but does not allow it to replace any device configuration. This approach has a couple of big benefits:

1. Propolis no longer needs a huge migration-compatibility-check module: the target's configuration is compatible with the source's because it *is* the source's configuration. Removing this module makes adding new guest-visible features significantly easier.
2. Nexus doesn't need to remember how it originally configured a VM in order to migrate it. This is a nice bug-prevention mechanism: today, changes to how Nexus constructs an `InstanceEnsureRequest` (or to how Propolis amends it!) can prevent a VM from ever being able to migrate (if the changes deterministically produce incompatible VM configuration). Even in the absence of bugs, this reduces the amount of instance initialization logic Nexus has to carry around ("I started this VM at this version of the code, so I have to use this specific logic to produce the correct spec/ensure request for it").

The server no longer reads device information from config TOMLs (it only reads the bootrom path; we might also be able to eliminate this someday, but I'm declaring it out of scope for this issue). The `propolis-server-config` library will stick around, though, and will provide a way for clients to convert a config TOML to a set of instance spec components so that they can easily use the new interface.

I've drafted enough of the server side of these changes to be pretty confident this scheme can be made to work and that it will make the system nicer overall.

gjcolombo:
(**N.B.** I'm still seeing if I can spin parts of this off into separate PRs, but I'm putting this up as-is for now (a) in case folks want to take a look as-is, and (b) to have commit hashes to refer to in my falcon/omicron clones so I can stop using path references everywhere.)

Consolidate the two separate instance-ensure APIs into a single API using the approach described in #804. There are several related changes in this PR, which are:

- The server now has a single instance PUT endpoint that takes the revised version of the `InstanceEnsureRequest` described in #804.
- Instance specs now use a `SpecKey` enum for their component map keys. This enum has UUID and string variants. This allows the control plane to supply UUID keys when it's possible and fall back to strings when it's not, and it allows Propolis to store/compare UUIDs in binary form when the control plane supplies them.
- Components' spec keys are now used as identifiers in a VM's device, block backend, and Crucible backend maps.
- During live migration, the target initializes devices using an amended version of the source's instance spec. The amendments are supplied by the target's caller, which may need to replace Crucible VCRs and VNIC names to allow the VM to function properly after migration. One consequence of this is that device-level migration compatibility checks are no longer needed--the target's device list *is* the source's device list, so they're always compatible.
- propolis-server processes no longer use config TOMLs. Instead, they take a bootrom path (and optional bootrom version string) on the command line, and expect all devices to specified through the ensure endpoint. To allow existing users to continue using their config TOMLs, add a TOML-to-instance-spec library to the `propolis-config-toml` crate and teach `propolis-cli` to read config TOMLs and create specs from them.

The main *mechanical* challenge in this PR is that it's no longer possible to create the VMM tokio runtime before starting the state driver task, because the runtime's thread count depends on the guest's vCPU count, and during a migration that's not known until after the migration protocol has begun to run. This is not intractable but accounts for the most significant changes to `propolis-server/src/lib/vm`.

Tests (so far):

- cargo test
- PHD (including migration variations, which now use the replacement-component functionality in several places)
- created some ad hoc VMs using `propolis-cli` and verified they had the expected components and the expected number of tokio runtime threads
- drafted changes to Falcon and Omicron that depend on this new API
  - tested a two-node Falcon topology with a softnpu link between the nodes & verified that the expected components ended up attached to the expected VMs
  - ran `cargo test` in Omicron (exercises the stripped-down mock server)
  - deployed an Omicron dev cluster, booted a VM there, verified that its component list is correct

Fixes #804. Fixes #776. Fixes #772.


gjcolombo:
This is split into multiple PRs by the series beginning with #807.

PR #807, opened 2024-11-13
server: get migration target specs from their sources

gjcolombo:
Broken out from #806; related to #804.

Instead of checking the compatibility of migration source and target instance specs, have targets initialize from amended versions of their sources' specs. The benefits of doing this are that

- The control plane no longer has to remember how it initialized a migration source in order to produce a compatible spec to send to the target, and
- Most migration compatibility checks are no longer needed--the target's device definitions are compatible with the source's because they *are* the source's definitions.

Some components of the source spec, notably its Crucible and viona backends, do need to change during migration: Crucible backends need new generation numbers, and viona backends need the VNIC names that the target sled assigned to them. To accommodate this, the sync phase substitutes the target's Crucible and viona backend descriptors into the source's spec. The forthcoming API changes in the rest of #806 will formalize this more, but the current approach is sufficient to keep the PHD Crucible migration tests passing.

In the future, migration targets won't have a spec until the migration protocol has started. Prepare for this change as follows:

- Create the VMM tokio runtime just before creating VM components instead of creating it up front and running the entire VM initialization process on it. This is needed because the runtime's thread count depends on the number of vCPUs in the VM, and that information comes from the instance spec.
- Slightly adjust the `instance_spec_get` API to reflect that a spec might not be available if a VM is initializing via live migration.

Tests: cargo test, PHD.


bin/propolis-server/src/lib/migrate/compat.rs
hawkw, line 1: yay for deleting code!

bin/propolis-server/src/lib/migrate/preamble.rs
```diff
@@ -20,26 +22,68 @@ impl Preamble {
         Preamble { instance_spec, blobs: Vec::new() }
     }
 
-    /// Checks to see if the serialized spec in this preamble is compatible with
-    /// the supplied `other_spec`.
+    /// Given the instance spec in this preamble (transmitted from the migration
+    /// source), replace any Crucible and viona backends with their
```
hawkw, line 26: with their...?

bin/propolis-server/src/lib/migrate/preamble.rs
```diff
@@ -20,26 +22,68 @@ impl Preamble {
...
     ///
     /// This check runs on the destination.
-    pub fn check_compatibility(
+    pub fn get_amended_spec(
```
hawkw, line 29: nit, take it or leave it: this actually amends the spec, it doesn't just return an already existing amended spec...so maybe this should just be called `amend_spec`?

bin/propolis-server/src/lib/migrate/preamble.rs
gjcolombo, line 26: Fragmented comment, will fix.

hawkw (approved):
overall, this looks great! i had a couple tiny nits.

bin/propolis-server/src/lib/migrate/preamble.rs
gjcolombo: Fixed in 7db1183.

bin/propolis-server/src/lib/migrate/preamble.rs
gjcolombo: Fixed in 7db1183.

PR #810, opened 2024-11-15
server: make clients parse config TOMLs

gjcolombo:
Stacked on #809. Related to #804 and #806.

Remove config TOML processing from propolis-server. The only thing the config TOML is used for in Omicron is specifying a bootrom path (and possibly a version string); turn this into a propolis-server command-line argument instead. The instance spec-based ensure API already ignored the server's config TOML, and the non-spec ensure API is about to be removed anyway, so this is a good time to make this shift.

To avoid breaking existing config TOML users, move the TOML-to-spec processing logic into the propolis-config-toml crate, and have it output propolis-client's generated instance spec types. (This would have been a layering violation previously, but is OK now that propolis-server no longer uses this crate.) Teach propolis-cli to use this crate to process TOMLs and to use the instance spec-based ensure endpoint to create and migrate VMs.

Remove the config TOML from the Propolis zone image and change its command line invocation. Also change PHD's propolis-server invocations as needed.

Finally, tweak the propolis-server and propolis-cli READMEs to describe the new state of affairs.

Tests: cargo test, PHD, created and migrated ad-hoc VMs (including with propolis-cli's `--crucible-disks` option).


gjcolombo:
The migrate-from-base tests' unhappiness is expected here, since downlevel propolis-server won't grok the args uplevel PHD is passing it. (This is a case where even HTTP server API versioning wouldn't have saved us!)

crates/propolis-config-toml/src/spec.rs
hawkw, line 1: AFAICT, this code hasn't changed meaningfully, just moved --- but if there are changes here that are worth a closer look, let me know?

hawkw (approved):
This is a large-ish diff, but the actual *change* is pretty straightforward and seems to be mostly moving code around, as far as I can tell. No concerns from me!

PR #814, opened 2024-11-18
Wire up viona params from illumos#16738

pfmooney:
Now that [illumos#16738](https://www.illumos.org/issues/16738) has landed, it'd be nice if propolis could make use of that functionality.

In `propolis-server`, we set parameters in line with how we expect it to operate in the oxide rack.  `propolis-standalone` remains at the defaults, while allowing modification via the toml config.

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -713,10 +713,38 @@ impl<'a> MachineInitializer<'a> {
             info!(self.log, "Creating vNIC {}", device_name);
             let bdf: pci::Bdf = nic.device_spec.pci_path.into();
 
+            // Set viona device parameters if possible.
+            //
+            // The values chosen here are tuned to maximize performance when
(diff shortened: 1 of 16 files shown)


bin/propolis-server/src/lib/initializer.rs
```diff
@@ -713,10 +713,32 @@ impl<'a> MachineInitializer<'a> {
             info!(self.log, "Creating vNIC {}", device_name);
             let bdf: pci::Bdf = nic.device_spec.pci_path.into();
 
+            // Set viona device params if possible
```
gjcolombo, line 716:
```suggestion
            // Set viona device params if possible. These parameters are tuned
	    // to maximize performance when Propolis is used with OPTE in a full
	    // Oxide deployment, though viona devices should continue to
	    // operate normally even when OPTE is not in use. These parameters
	    // are not guest-visible.
```

gjcolombo (approved):

PR #816, opened 2024-11-19
api: allow use of UUIDs as component IDs in instance specs

### LINKED ISSUE #772: Prefer component IDs to names

### LINKED ISSUE #776: server: machine init is inconsistent about the names it puts in its DeviceMap

gjcolombo:
Stacked on #813. Related to #804.

**Reviewer note:** Most of this change is mechanical conversion of types, but there are some slightly more substantial functional changes to Crucible backend management in server.rs, initializer.rs, and state_driver.rs.

---

Today instance specs use `String`s to identify VM components. In many cases, components are associated with control plane objects (like Disk and NetworkInterface records) that have UUIDs. Allow components to be identified by UUID by defining a `SpecKey` type that deserializes as a `Uuid` when possible and as a `String` when not, then plumb it throughout Propolis. See the comments on the `SpecKey` type for more details, including notes on why the type is a UUID/String union and not just a UUID.

The main functional change in this PR is that propolis-server's component maps now use `SpecKey`s to identify components. The idea is to say that if a component has key "Foo" in a VM's spec, then subsequent API calls that want to act on that component should pass "Foo" as the component ID. This (hopefully) simplifies how Crucible-related APIs should designate the disk they're targeting: Omicron should pick an ID for each Crucible backend it requests (which ID can just be the control plane disk ID), then pass the same ID to subsequent snapshot and VCR replacement operations. Control plane disk names no longer appear in the Propolis API at all.

Tests: cargo test, PHD, manual creation/migration of VMs with propolis-cli, manually hit the snapshot and VCR replacement endpoints and verified these dispatched their operations to the correct Crucible backend.

Fixes #776. Fixes #772.


gjcolombo:
@leftwo You may want to take a look at the Crucible bits of this one (in server.rs, initializer.rs, and state_driver.rs). Instead of calling `Volume::get_uuid` to get a disk ID from the VCR to use in the `CrucibleBackendMap`, Propolis now expects that

- Omicron will choose a Crucible backend component ID for each Crucible disk when a VM starts
- This ID will be used as the key to the Crucible backend map
- Omicron will specify the same ID when it needs to request a snapshot or replace a VCR

I reckon Omicron will just use the disk ID as the Crucible backend ID to make this as smooth as possible in Nexus and sled agent. As far as I can tell, the disk ID *is* the ID that we were getting from `get_uuid` before, because that's how Omicron constructs its volume construction requests; this just makes the IDs at the Propolis layer a little more explicit/transparent.

LMK whether this seems like a reasonable approach.

crates/propolis-api-types/src/instance_spec/mod.rs
```diff
@@ -155,18 +155,152 @@
...
+// In some cases it's not desirable or possible to use UUIDs this way:
+//
+// - Some components (like the cloud-init disk) don't have their own rows in the
+//   database and so don't have obvious UUIDs to use.
+// - Some objects (like Crucible disks) require both a device and a backend
+//   component in the spec, and these can't share the same key.
+// - Propolis users outside the control plane may not have any component UUIDs
+//   at all and may just want to use strings to identify all their components.
+//
+// For these reasons, the key type is a union of a UUID and a String. This
```
hawkw, line 184: suuuuuuper pedantic, but i wouldn't call this a "union", since Rust [also has `union` types](https://doc.rust-lang.org/reference/items/unions.html) which are untagged unions.

hawkw:
This seems reasonable to me overall, but I'll defer to the Crucible folks

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -610,66 +617,67 @@ impl<'a> MachineInitializer<'a> {
...
+                if prev.is_some() {
+                    return Err(MachineInitError::DuplicateCrucibleBackendId(
+                        backend_id.clone(),
+                    ));
+                }
+
+                // If metrics are enabled and this Crucible backend was
+                // identified with a UUID-format spec key, register this disk
+                // for metrics collection, using the key as the disk ID.
+                if let (Some(registry), SpecKey::Uuid(disk_id)) =
```
iximeow, line 674: Nexus will never tell `propolis-server` about a Crucible disk with non-UUID ID, right? if so, the `SpecKey::Name` path here, where there just won't be metrics collected for the disk, _shouldn't_ be reachable except if we've crafted a weird instance request by hand? if this shouldn't ever be reachable, it seems worth rejecting the spec /more precisely typing IDs we know _need_ to be UUID.

i do remember disk _names_ ending up in here somewhere, rather than IDs.. gonna go reread how that plumbing works unless you remember more immediately.

my real concern here is that this is where Nexus had been giving what are now `SpecKey::Name`, and that we'll lose Crucible metrics here. do you know if we have a test that exercises that?

gjcolombo: > Nexus will never tell propolis-server about a Crucible disk with non-UUID ID, right? if so, the SpecKey::Name path here, where there just won't be metrics collected for the disk, shouldn't be reachable except if we've crafted a weird instance request by hand? if this shouldn't ever be reachable, it seems worth rejecting the spec /more precisely typing IDs we know need to be UUID.

This makes sense. I'm a little leery of it because I think it makes the API a little weird: the ID for a given component can be either a UUID a Name, *unless* it's this one specific kind of component, in which case it has to be a UUID. This isn't horribly broken or anything, it's just a slightly sharp edge (and we'd also have to account for it in the non-Omicron tools/processes that let users attach ad-hoc Crucible volumes to Propolis servers).

We could address this in a few other ways:

1. `warn!` if the spec key isn't a UUID but there's a producer registry available
2. put a `#[cfg(feature = "omicron-build")]` guard in that enforces the "Crucible keys must be UUIDs" requirement
3. go back to calling `Volume::get_uuid` in this path to get the UUID to use for stats reporting, and `warn!` if it differs from the spec key

I kind of like option 3 now that I've written it up: it limits our regression risk (at least where metrics are concerned) but keeps the API flexible. WDYT?

> i do remember disk names ending up in here somewhere, rather than IDs.. gonna go reread how that plumbing works unless you remember more immediately.

IIRC the disk name wasn't used in this specific path, but it *was* used in VCR replacement, which worked like this:

- Take the VM objects lock
- Look up the `CrucibleStorageBackend` component in the instance spec
  - the key for this was derived from the disk name
- Look up the Crucible backend object in the VM's `crucible_backends` map
  - the key for this turns out to be the disk ID; the initialization code path used to read it from the Crucible backend's `get_uuid` function, which would read it out of the disk's volume construction request; Omicron guaranteed that this ID matched the disk ID (by creating the VCR that way)
- send the replace operation to Crucible
- on success, write the updated VCR back to the instance spec

> my real concern here is that this is where Nexus had been giving what are now SpecKey::Name, and that we'll lose Crucible metrics here. do you know if we have a test that exercises that?

We don't have an end-to-end test for disk metrics, at least AFAIK.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -586,22 +593,22 @@ impl<'a> MachineInitializer<'a> {
             Nvme,
         }
 
-        for (disk_name, disk) in &self.spec.disks {
+        for (device_id, disk) in &self.spec.disks {
```
iximeow, line 596: i do think that `disk_name` was a good name for this one - it's a `SpecKey`, and it _could_ be a `Uuid` if something sent one, but Nexus currently sends a `Name` and changing it would (for better or worse) change `/dev/disk/by-id/` paths guest VMs should be able to depend on.

i recognize that you're not changing `disk_name` to something else (yet?), but i happen to _also_ think we should change this _name_ to an ID. if you agree, it'd be nice to change the name of this variable when we do it :grin:

gjcolombo: Great catch here, thanks! I *think* the disk serials will remain stable over migrations, at least, but you're right, without some care this will change the serial number strings we use for NVMe devices across a reboot.

One of my goals for this layer is to divorce components' keys from their behaviors--i.e. two devices with the same component specs should have the same guest-visible behavior even if they had different keys. In this particular case, we can achieve that by adding a `serial_number` field to the `NvmeDisk` component, and then populating it with whatever makes sense in the control plane (the actual disk name, I believe). If that sounds good, I can give this a try when I next make updates to this change.

iximeow: oh yes, love that idea!

bin/propolis-server/src/lib/initializer.rs
iximeow: yeah, after re-reading a bit and taking a second look, i'm also a fan of `Volume::get_uuid` here. i generally like the change to make `crucible` just `Option<Arc<...>>` though, so to double-check: you're thinking literally calling `crucible.get_uuid()` somewhere around this `if`, yeah? i think that should be pretty clean.

FWIW the `get_uuid()` call before was context'd and `?`, so it would bubble an error up rather than warn. that still seems like an appropriate error behavior here too.

more UUID tracing i haven't done: could `backend_id` be different from `crucible.get_uuid()`? i assume this is "technically possible but not really desired"

bin/propolis-server/src/lib/initializer.rs
gjcolombo: Fixed in 6e82d49.

I represented the `serial_number` as a `[u8; 20]` instead of a `String` to try to make clients be explicit about padding. Just taking a `String` leaves Propolis to decide whether to pad with 0s, spaces, or something else; this method prevents us from accidentally changing padding disciplines across a migration. Another approach would be to have a `serial_number: String` and a `serial_padding_byte: u8` in the spec itself; this is more verbose, but it's convenient enough for clients that I added a support function to `propolis-client` to generate serial number byte arrays from a `str` and padding byte.

I'm not strongly beholden to this approach--let me know if you think a different one makes more sense.

bin/propolis-server/src/lib/initializer.rs
gjcolombo: Should be fixed in d6602c1.

> more UUID tracing i haven't done: could backend_id be different from crucible.get_uuid()? i assume this is "technically possible but not really desired"

That's correct, at least as I understand it:

- the serialized `VolumeConstructionRequest` that gets passed to the API is expected to be a `VolumeConstructionRequest::Volume` (which will have subvolumes of type `Region`)
- `Volume::get_uuid` returns the ID field in that top-level `Volume`
- when Nexus constructs the VCR for a new disk, it sets that ID to the disk ID

When I update Omicron to pick up this change, I expect that Nexus/sled-agent will use disk IDs as the spec keys for any `CrucibleStorageBackend`s they specify. That will preserve this relationship so long as Nexus keeps inserting the disk ID into its `VolumeConstructionRequest::Volume`s the way it does today.

crates/propolis-api-types/src/instance_spec/mod.rs
gjcolombo: Fixed in f8ecc34.

bin/propolis-server/src/lib/initializer.rs
iximeow: (i do have thoughts, but i think on `disk_name`/id/serial we're in a good spot. thanks! i'll comment elsewhere as appropriate.)

phd-tests/framework/src/test_vm/config.rs
```diff
@@ -303,12 +304,40 @@ impl<'dr> VmConfig<'dr> {
...
+                        // Unfortunately for us, guest firmware may persist disk
+                        // identifiers to a VM's nonvolatile EFI variable store.
+                        // This means that changing from zero-padded to
+                        // space-padded serial numbers may break an existing
+                        // VM's saved boot order.
+                        //
+                        // Until we decide whether and how to preserve
+                        // compatibility for existing VMs that may have
+                        // preserved a zero-padded disk ID, continue to zero-pad
+                        // disk IDs in PHD to match previous behavior.
```
iximeow, line 338: i initially wondered if you expected to copy this comment to Omicron later, but the corresponding Omicron change will probably be a migration to persist the null-padded serials and switch to space-padded for new disks huh? then at least any old wrong data will be made static, rather than "fixed" without us expecting.

lib/propolis-client/src/support.rs
```diff
@@ -27,6 +27,18 @@ impl Default for Chipset {
     }
 }
 
+/// Generates a 20-byte NVMe device serial number from the bytes in a string
+/// slice. If the slice is too short to populate the entire serial number, the
+/// remaining bytes are filled with `pad`.
```
iximeow, line 32:
```suggestion
/// Generates a 20-byte NVMe device serial number from the bytes in a string
/// slice. If the slice is too short to populate the entire serial number, the
/// remaining bytes are filled with `pad`.
///
/// NVMe specification (particularly 1.2 and later, see section 1.7
/// `Conventions`) requires that serial numbers are left-justified and padded
/// with spaces. We allow padding with non-spaces for the time being as a
/// flagrant violation of the spec, but as a necessary one while we decide if or
/// how we will carry forward compatibility for VMs given null-padded serials.
```
mostly want to emphasize that left-justification isn't because we think it looks nice, it's to accord with the NVMe spec. could be a comment in the function or on the test cases instead?

lib/propolis-client/src/support.rs
```diff
@@ -572,4 +584,21 @@ mod test {
...
+
+    #[test]
+    fn test_nvme_serial_from_str() {
+        use super::nvme_serial_from_str;
+
+        let expected = b"hello world         ";
+        assert_eq!(nvme_serial_from_str("hello world", b' '), *expected);
+
+        let expected = b"enthusiasm!!!!!!!!!!";
+        assert_eq!(nvme_serial_from_str("enthusiasm", b'!'), *expected);
```
iximeow, line 596: it is very funny to me that you have a test for this but _not_ `nvme_serial_from_str("these_are_not_made", 0x00)`. it'll definitely work for null-padding as-is, but that's the one non-space case we _do_ expect :D

gjcolombo: Yeah, we definitely want to cover that one! Fixed in d6dd7c7.

lib/propolis-client/src/support.rs
gjcolombo: Fixed in d6dd7c7.

iximeow (approved):
thanks for working through this!

phd-tests/framework/src/test_vm/config.rs
gjcolombo: Cleaned up in d6dd7c7. The main thing I wanted to emphasize in this comment is that PHD behaves this way because Omicron will when I create the initial changes to sled-agent to get it to use the spec-based ensure API), and I don't want their behaviors to drift too far apart.

> the corresponding Omicron change will probably be a migration to persist the null-padded serials and switch to space-padded for new disks huh?

This is definitely the most straightforward approach, but I don't really *like* it, because it entails having Nexus have to keep track of the (hopefully small-and-getting-smaller) set of existing disks that need the special "pad with zeroes not spaces" compat quirk. I would be willing to consider some possibly-outlandish options (e.g. trying to do something clever in the boot order management logic in our EDK2 fork) to avoid adding a quirk column/table just for this.

It probably goes without saying, but I intend not to try to solve that problem in the Omicron change that will ingest these Propolis changes--the first order of business is to convert to the new API without actually changing or breaking anything :)

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -610,74 +617,93 @@ impl<'a> MachineInitializer<'a> {
...
+                        unreachable!("disk is known to be an NVMe disk");
+                    };
+
                     // Limit data transfers to 1MiB (2^8 * 4k) in size
                     let mdts = Some(8);
-                    let component = format!("nvme-{}", disk_name);
+                    let component = format!("nvme-{}", device_id);
                     let nvme = nvme::PciNvme::create(
-                        disk_name.to_owned(),
+                        &nvme_spec.serial_number,
```
leftwo, line 646: Does this mean the disk serial number is showing up somewhere?
Did this mean the disk name was also somewhere in the past?  
Would a change here effect existing instances that had booted with `disk_name` in the past?

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -610,74 +617,93 @@ impl<'a> MachineInitializer<'a> {
...
+
                     if let Err(e) = registry.register_producer(stats.clone()) {
                         slog::error!(
                             self.log,
                             "Could not register virtual disk producer, \
                             metrics will not be produced";
-                            "disk_id" => %disk_id,
+                            "disk_id" => %backend_id,
+                            "volume_id" => %volume_id,
                             "error" => ?e,
```
leftwo, line 707: What's the difference here between `volume_id` and `disk_id`?

leftwo (approved):
I don't see any obvious problems with crucible and this change.  It would be good to take it for a test drive and replace a downstairs just to be sure we connected the dots correctly.

bin/propolis-server/src/lib/initializer.rs
gjcolombo: The guest can get at the serial number via NVMe controller commands. Linux guests will use it (for example) to name the disk's devfs nodes (`/dev/disk/by-id/foo` where `foo` is derived in part from the serial number).

On the Omicron side, whoever fills out this `nvme_spec` will populate the `serial_number` field using the corresponding Omicron disk name (with whatever padding regime is needed to maintain compatibility). https://github.com/oxidecomputer/propolis/pull/816#pullrequestreview-2455445757 has some more context here.

bin/propolis-server/src/lib/initializer.rs
gjcolombo: The distinction here is meant to be between "the ID that was used for the disk element in the spec" and "the ID we got back from `Volume::get_uuid`," but in practice I think these are (currently) always the same, because Nexus constructs VCRs that way (i.e. it always puts the disk ID into the relevant `id` field in the top-level `VolumeConstructionRequest::Volume` that it writes to the DB).

gjcolombo:
Thanks @leftwo for taking a look! Before I merge I'll make sure to spin up a Propolis with this change and recheck that I can properly dispatch operations to the snapshot and VCR replacement APIs. (We'll want to check that that works on the Omicron side as well, but I'll ask about how to do that when the Omicron changes are ready.)

bin/propolis-server/src/lib/initializer.rs
leftwo: Great, so as long as we don't mess up the Omicron side :)

gjcolombo:
> Before I merge I'll make sure to spin up a Propolis with this change and recheck that I can properly dispatch operations to the snapshot and VCR replacement APIs.

I've checked this for both of these endpoints: the snapshot and replace operations indeed get directed to the relevant Crucible downstairs processes (which then summarily reject them because I didn't build with the right features, didn't specify a VCR in the correct format, etc., but at least they're making it to the right places!).

PR #823, opened 2024-12-05
Switch viona back to packet copying for now

pfmooney:
Unexplained pathological performance was observed when running with packet memory loaning.  Further investigation is required before changing it back.


leftwo (approved):
As tested on london, we saw much better performance with this hot patched.

Issue #825, opened 2024-12-12
server: consider delegating migration payload logging to individual components


gjcolombo:
The migration destination runner logs the entire serialized device state payload it got from the source: https://github.com/oxidecomputer/propolis/blob/220a6f367c18f2452dbc4fa9086f3fe73b961739/bin/propolis-server/src/lib/migrate/destination.rs#L483-L495

Depending on exactly what devices are attached and how they're configured, this message may include tens of kilobytes of serialized state, e.g. repeated messages about MSI-X interrupts that are allocated to a particular device but aren't currently in use.

It would be nice to have the migration logic log only top-level information about the component payloads it received--the component name, the kinds/versions of the payloads, and maybe the lengths of those payloads--and delegate the logging of the actual payloads to the components themselves. Then the components can deserialize the data and decide whether and how they'd like to log it.

Note that not all components have ready access to a `slog::Logger` when they're importing device state, so we'll need to do some plumbing to make sure they have access to one (either by adding a Logger to the relevant component structures or updating the `MigrateSingle`/`MigrateMulti` traits to supply one).

pfmooney:
> Note that not all components have ready access to a `slog::Logger` when they're importing device state, so we'll need to do some plumbing to make sure they have access to one (either by adding a Logger to the relevant component structures or updating the `MigrateSingle`/`MigrateMulti` traits to supply one).

The reason I've dragged `MigrateCtx` around in those traits is for exactly that sort of situation.

gjcolombo:
> The reason I've dragged MigrateCtx around in those traits is for exactly that sort of situation.

Oh yeah, good call--I've looked at that type half a dozen times this week and didn't put two and two together on using it for this. That'll work.

PR #826, opened 2024-12-12
lib/server: allow migration of in-memory block backends

### LINKED ISSUE #301: In-memory disk backends need better migration support

gjcolombo:
(N.B. Stacked on #816.)

Register in-memory block backends for lifecycle callouts and teach them to export and import their contents.

Truthfully, this may be somewhat overengineered. Today, in-memory storage backends are only used for the cloud-init drive, which Nexus always marks as read-only. Since the guest can never change the drive's contents, and #807 ensures that propolis-server migration targets always inherit their initial in-memory backend contents from their sources, it is technically legal for the backend just to return `Migrator::Empty` if it's read-only (and return `Migrator::NonMigratable` if it's writable). However, this makes the Propolis lib assume things about the way propolis-server drives migrations, which seemed like a layering violation, so this change just does what's needed to migrate the backend's current data in all cases.

One downside of this approach is that the serialized backend contents show up in migration logs (as part of the giant blob logged in the destination's `device_state` function). It may make sense to adjust this logging so that the serialized device payloads are removed from this log message. See #825 for more thoughts on this. I think this change would be large enough that if pursued it should be in its own PR.

Add a PHD test for the new migration behavior. This test works on Debian 11, Ubuntu 22.04, and Windows Server 2022. It's skipped on Alpine guests because (for reasons I don't yet understand) the test's use of `iflag=direct` on that OS isn't sufficient to get reads to hit the actual in-memory backend.

Finally, tweak a few PHD bits:

- Add a helper to run Powershell cmdlets in Windows VMs.
- Change the Ubuntu 22.04 adapter's login sequence so that tests run as root. This can be cleaned up in the future by generating an Ubuntu 22.04 image where the default user is a superuser.

Tests: new PHD variation with Alpine, Debian 11, Ubuntu, and WS2022 guests.

Fixes #301.

### DIFF
--- a/lib/propolis/src/block/in_memory.rs
+++ b/lib/propolis/src/block/in_memory.rs
@@ -8,10 +8,16 @@ use std::sync::{Arc, Mutex};
 
 use crate::accessors::MemAccessor;
 use crate::block;
+use crate::migrate::{
+    MigrateCtx, MigrateSingle, MigrateStateError, Migrator, PayloadOffer,
+    PayloadOutput,
+};
 use crate::tasks::ThreadGroup;
 use crate::vmm::{MemCtx, SubMapping};
 
 use anyhow::Context;
+
+use super::Lifecycle;
 
 pub struct InMemoryBackend {
     state: Arc<WorkingState>,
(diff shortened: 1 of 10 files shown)


lib/propolis/src/block/in_memory.rs
```diff
@@ -8,11 +8,17 @@ use std::sync::{Arc, Mutex};
...
+use crate::migrate::{
+    MigrateCtx, MigrateSingle, MigrateStateError, Migrator, PayloadOffer,
+    PayloadOutput,
+};
 use crate::tasks::ThreadGroup;
 use crate::vmm::{MemCtx, SubMapping};
 
 use anyhow::Context;
 
+use super::Lifecycle;
```
pfmooney, line 20: style nit: use this out of `crate::common`, rather than the re-export from `block`

pfmooney (approved):
Looks reasonable (with one small nit)

phd-tests/tests/src/disk.rs
```diff
@@ -4,60 +4,136 @@
...
-    // its device is located at /dev/vda. If the by-path directory is present,
-    // try to check that the disk is located there and fail the test early if
-    // it's not. If the by-path directory is missing, put up a warning and hope
-    // for the best.
-    let dev_disk = vm.run_shell_command("ls /dev/disk").await?;
-    if dev_disk.contains("by-path") {
-        let ls = vm.run_shell_command("ls -la /dev/disk/by-path").await?;
-        info!(%ls, "guest disk device paths");
-        assert!(ls.contains("virtio-pci-0000:00:18.0 -> ../../vda"));
+    let device_path = if let Some(win_vm) = vm.get_windows_vm() {
```
iximeow, line 37: tiny but nice: if you added an `impl<'a> Deref for WindowsVm<'a> { type Target = TestVm<'a>; }` you could write this as `if let Some(vm) = vm.as_windows_vm()` and not have to know which "vm" is the right one for what you want to call. it'd all be `vm.run_powershell_command` and `vm.run_shell_command` in this arm.

phd-tests/tests/src/disk.rs
```diff
@@ -66,3 +142,73 @@ async fn in_memory_backend_smoke_test(ctx: &Framework) {
...
+    vm.run_shell_command(&format!(
+        "dd if=/dev/random of={device_path} oflag=sync bs=1K count=1"
+    ))
+    .await?;
+
+    // Read the scribbled data out to a file on the main OS disk.
+    vm.run_shell_command(&format!(
+        "dd if={device_path} of=/tmp/before iflag=direct bs=1K"
+    ))
+    .await?;
```
iximeow, line 185: idle thought, might be Alpine-inclusive while we don't know what's up with `dd iflag=direct`: `sync && echo 3 /proc/sys/vm/drop_caches` is (if memory serves) synchronous and will also ensure that written data has been flushed out. caches should then also be empty for reads later on.

but `O_DIRECT` is more obvious at time of use of course. so really more a "if you didn't think of it, might be interesting" - what's here seems good.

iximeow (approved):
seems good! just a few suggestions, take em or leave em

Issue #614, opened 2024-01-22
Update migration protocol to negotiate device version payloads


gjcolombo:
The Propolis library's migration framework supports composable versioned payloads through the `PayloadOffer` and `PayloadOutput` types in `propolis::migrate`. In this scheme, each chunk of migration data has a "kind" and a version, and each Propolis component (device or backend) can export one or more chunks. Because chunks are versioned independently, we can bump the version of a common subcomponent (e.g. the virtio state exported by all virtio devices) without having to re-version all of the independent versions that contain it.

Today the migration protocol has no affordance for the source and destination to agree on the component versions the source should send. (Note that it's not a given that the destination understands more versions than the source, since the migration may be a downgrade.) The protocol should include this, probably as part of the "preamble" phase that otherwise determines whether a migration from the source to the target will be accepted.

PR #830, opened 2025-01-10
server: introduce a serial console backend

### LINKED ISSUE #650: server serial task's send-to-listeners future is not cancel-safe
Discovered this during some ad hoc testing with multiple connections to a single guest's serial console, where I observed some data being sent a second time to a listener that had already received it.

If the serial console task has read some data from the guest that needs to be written to listeners, it generates a future to do that writing and then includes it in the main `select!` that determines the task's next action (lines 116-125 below where `cur_output` is `Some`): https://github.com/oxidecomputer/propolis/blob/1b34075ee5dd4235dc7eb46061e1e648fb4d121b/bin/propolis-server/src/lib/serial/mod.rs#L109-L130

`cur_output` is only set to `None` if this future is the one chosen by `select!`: https://github.com/oxidecomputer/propolis/blob/1b34075ee5dd4235dc7eb46061e1e648fb4d121b/bin/propolis-server/src/lib/serial/mod.rs#L226-L230

But if this branch is not chosen, `for_each_concurrent` may have sent the current output bytes to some listeners and not to others. (The underlying websocket `send` may not be cancel-safe either but I haven't looked at its implementation.) This will cause the same data to be re-sent in the next loop iteration.

An approach along the lines suggested in RFD 400 section 5.2 might help here (and other approaches may work too):

- If there's already a send-to-clients future active, try to resolve it
- Else if there's no such future but there is some data waiting to be sent, create a new future to try to send that data
- Else try to read more data from the guest

gjcolombo:
Rework parts of `propolis-server`'s serial console logic to achieve the following goals:

- Make the serial console history buffer (which is transferred in the live migration protocol) use the versioned payload helpers supplied by the main Propolis library.
- Provide the primitives necessary to implement the policies described in RFD 491: it should be possible to distinguish read-write and read-only clients of a serial console and apply different policies (around the number of clients and what happens if they fall behind when reading) to each of them.
- Simplify the logic for dealing with multiple serial console clients to avoid cancellation safety issues.

To do this, split the existing serial console task into two layers. The `ConsoleBackend` layer deals with I/O to and from a Propolis character device. The `SerialConsoleManager` layer creates a task for each websocket connection to a serial device and registers it as a client of the relevant `ConsoleBackend`. The backend layer allows the manager to declare that specific clients have read-only access to the console and to decide what happens if a client becomes unable to accept new bytes from a client; the manager implements the policies that determine how many clients can exist concurrently and what backend policies apply to them.

Update the serial API endpoint to allow users to request a writable connection to the console. Add parameters to the serial console helper types in `propolis-client` that govern how this is set. This will require a small update to Nexus the next time Omicron picks up a Propolis build to add the extra parameter. End users don't see this aspect of the serial API (Nexus serves as a proxy for connections to VMs' serial consoles), so this should be a non-breaking change for control plane API users.

`propolis-cli serial` defaults to a writable connection to match previous behavior, but it now supports a `--readonly` flag to create read-only connections.

As currently written, this PR implements the "only one read-write client at a time" policy described in RFD 491. I can easily disable this part of the change if we don't want to do this yet (should just need to remove a few lines from the console manager's client-attach function).

Tests: cargo test, PHD, ad hoc testing with local propolis-server processes. This change changes the order of messages sent in the migration protocol and so will break the migrate-from-base tests; this remains OK for now since live migration isn't being used in production yet.

Fixes #650. Related to #614.


bin/propolis-server/src/lib/serial/backend.rs
```diff
@@ -0,0 +1,397 @@
...
+pub(super) struct ClientHandle {
+    /// The client's ID.
+    id: ClientId,
+
+    /// A reference to the backend to which the client is connected.
+    backend: Arc<ConsoleBackend>,
+
+    /// The client's backend access rights.
+    permissions: Permissions,
+}
```
hawkw, line 76: considered nitpicky, take it or leave it: personally, since the `ClientHandle` type represents the external API for consumers of this module, while the `Client` type is an internal-only implementation detail, I might've given the more consicse name to the interface type --- perhaps `Client` and `ClientState`, or something?

this isn't a big deal, though; if renaming it is too much of a pain, feel free to ignore me.

bin/propolis-server/src/lib/serial/backend.rs
```diff
@@ -0,0 +1,397 @@
...
+
+    /// The backend should close this client's connection.
+    Close,
+}
+
+/// An individual client of a character backend.
+struct Client {
+    /// Bytes read from the character device should be sent to the client on
+    /// this channel.
+    tx: mpsc::Sender<u8>,
```
hawkw, line 59: `mpsc::Sender<u8>` is a pretty unfortunate way to represent an asynchronous stream of bytes, from a performance standpoint: this will allocate a node in a linked list for each individual byte. If you want an asynchronous byte pipe, I would consider using something like [`tokio::io::simplex`](https://docs.rs/tokio/latest/tokio/io/fn.simplex.html) , which is an asynchronous byte pipe implemented as a single contiguous in memory buffer. There's also [`DuplexStream`](https://docs.rs/tokio/latest/tokio/io/struct.DuplexStream.html), which is the same thing but bi-directional.

bin/propolis-server/src/lib/serial/backend.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// its device, so when this function returns, it is not guaranteed that all
+    /// of the written bytes have actually reached the guest.
+    ///
+    /// # Return value
+    ///
+    /// - `Ok(bytes)` if the write succeeded; `bytes` is the number of bytes
+    ///   that were written.
+    /// - `Err(ErrorKind::PermissionDenied)` if this handle does not grant write
+    ///   access to the device.
+    /// - `Err(ErrorKind::ConnectionAborted)` if this handle's client was
```
hawkw, line 104: nitpicky, considered unimportant: this doc comment is *technically* lying; this doesn't return an `ErrorKind`, but a whole `io::Error`...

bin/propolis-server/src/lib/serial/backend.rs
```diff
@@ -0,0 +1,397 @@
...
+    loop {
+        let bytes_read = select! {
+            biased;
+
+            _ = &mut done_rx => {
+                return;
+            }
+
+            res = buf.read(bytes.as_mut_slice(), source.as_ref()) => {
+                res.unwrap()
```
hawkw, line 342: nit: do we expect that this will never be an error? it might be nice to `expect` this with a string explaining why...

bin/propolis-server/src/lib/serial/backend.rs
```diff
@@ -0,0 +1,397 @@
...
+/// Character backend state that's protected by a lock.
+struct Inner {
+    /// A history of the bytes read from this backend's device.
+    buffer: HistoryBuffer,
+
+    /// A table mapping client IDs to clients.
+    clients: BTreeMap<ClientId, Client>,
+
+    /// The ID to assign to the next client to attach to this backend.
+    next_client_id: u64,
```
hawkw, line 165: Rather than a `BTreeMap` and counter of IDs, I might consider using the [`slab`](https://crates.io/crates/slab) crate, which implements a collection of "things" indexed by integer ID, but stored in an array (so both iteration and lookups are likely much more efficient). Do note that `slab` will reuse indices of removed elements, in case that's an issue here, but AFAICT, it wouldn't actually be.

bin/propolis-server/src/lib/serial/mod.rs
```diff
@@ -2,400 +2,533 @@
...
+    done_tx: oneshot::Sender<()>,
+}
 
-    #[error("Mismatched websocket streams while closing")]
-    MismatchedStreams,
+/// The collection of [`ClientTask`]s belonging to a console manager.
+#[derive(Default)]
+struct ClientTasks {
+    /// A table mapping task IDs to task structs.
+    tasks: BTreeMap<ClientId, ClientTask>,
```
hawkw, line 89: similarly, i might consider whether this could be a `slab`.

bin/propolis-server/src/lib/serial/mod.rs
```diff
@@ -2,400 +2,533 @@
...
+                console_rx,
+                control_rx,
+                start_rx: task_start_rx,
+                done_rx: task_done_rx,
+                client_tasks: self.client_tasks.clone(),
+                client_id,
+            };
+
+            let task = ClientTask {
+                hdl: tokio::spawn(async move { serial_task(ctx).await }),
```
hawkw, line 213: style nit: calling an async function, such as `serial_task`, returns a `Future`, so the `async move { ....await }` is unnecessary; this could just be: 
```suggestion
                hdl: tokio::spawn(serial_task(ctx)),
```

bin/propolis-server/src/lib/serial/mod.rs
```diff
@@ -2,400 +2,533 @@
...
-            // Poll in the order written
+        let event = select! {
+            // The priority of these branches is important:
+            //
+            // 1. Requests to stop the client take precedence over everything
+            //    else.
+            // 2. New bytes written by the guest need to be processed before any
+            //    other requests: if a guest outputs a byte while a read-write
+            //    client is attached, the relevant vCPU will be blocked until
+            //    the client processes the byte.
```
hawkw, line 368: thank you for documenting this <3

gjcolombo:
Thanks as always for the thorough review, @hawkw! I'm going to rebase this as-is onto #834 and then will start tightening things up.

bin/propolis-server/src/lib/serial/backend.rs
gjcolombo: `simplex` definitely seems like a much nicer tool here. The one gotcha with it is that `Sender` is `Clone`, but `SimplexStream` and its `WriteHalf` aren't, so the code in `read_task` that captures channels under the lock and then drops the lock before sending (so that it can guiltlessly block as needed) won't work as-is.

Now that I think about it, I wonder if it'd be better to rewrite `read_task` so that it doesn't need to look at the client map at all. (I think there's even an outside chance I could get rid of the client map altogether...) I'm going to give this a try to see how it shakes out.

gjcolombo:
@hawkw: I think I've addressed all your feedback in the newest round of commits. The new changes use `tokio::io::simplex` (thank you for suggesting this!), but this required some architectural adjustments to account for differences between a `WriteHalf` and an `mpsc::Sender` (the former isn't `Clone`, and writes won't return an error if the reader is dropped). For the most part, what this amounts to is that the `ConsoleBackend`'s read task no longer reads the backend's client table directly; instead the backend feeds it a set of `Reader` objects, and readers and their clients share a `watch` so they can mutually shut down without having to hold references to each other.

There's a new theory statement in `serial/backend.rs` that tries to explain this a bit more. Let me know what you think.

PR #832, opened 2025-01-11
add JSON output format to cpuid-gen

gjcolombo:
Add a JSON output format to `cpuid-gen`, which prints output suitable for inserting into an instance spec that can be fed to `propolis-cli`.

This is useful for ad hoc testing of hypervisor enlightenments. Enlightenments are exposed through special CPUID leaves in the hypervisor leaf range (0x4000_0000 to 0x4000_FFFF). To expose enlightenments, Propolis thus needs to pass the correct CPUID values to bhyve. However, if Propolis starts setting CPUID values for the hypervisor leaves, it will also need to set them for the standard and extended leaves. This implies that, to enable enlightenments, Propolis's instance-ensure API (or propolis-standalone's VM setup logic) will need a complete set of standard/extended CPUID values to apply. For `propolis-server`, the easiest way to do this is to use `propolis-cli new --spec`. The goal of this change is to make it easy to get a CPUID JSON blob to paste into this command's input spec.

While dorking around with this change, I discovered that the `propolis-cli new --spec` switch was not properly marked as exclusive with the `--config-toml` switch. Fix that up as well.

Finally, if `propolis-server` fails to convert an API spec to its internal instance spec type, and the resulting error has a source, print both the error and the source. This allows the display of messages like "leaf 7 is specified both with and without a subleaf" when appropriate.

Tested by booting some ad hoc propolis-server VMs using propolis-cli and the new output from cpuid-gen.

### DIFF
--- a/bin/propolis-utils/src/bin/cpuid-gen.rs
+++ b/bin/propolis-utils/src/bin/cpuid-gen.rs
@@ -266,22 +266,45 @@ fn print_json(results: &BTreeMap<CpuidKey, Cpuid>) {
     use propolis_api_types::instance_spec::components::board::{
         Cpuid, CpuidEntry,
     };
+
+    // Returns `true` if `key` has subleaf data and its leaf index is `leaf`.
+    fn key_matches_leaf_and_has_subleaves(key: &CpuidKey, leaf: u32) -> bool {
+        let CpuidKey::SubLeaf(l, _) = key else {
+            return false;
+        };
+
+        leaf == *l
+    }
+
     let cpuid = Cpuid {
         entries: results
             .iter()
-            .map(|(key, value)| {
+            .filter_map(|(key, value)| {
+                // propolis-server will complain if it receives a CPUID entry
+                // with no subleaf and a CPUID entry with a subleaf for the same
+                // leaf. Avoid emitting these by filtering out subleaf-less
+                // entries for any leaf that has an entry with subleaf data.
                 let (leaf, subleaf) = match key {
-                    CpuidKey::Leaf(leaf) => (*leaf, None),
+                    CpuidKey::Leaf(leaf) => {
+                        if results.keys().any(|k| {
+                            key_matches_leaf_and_has_subleaves(k, *leaf)
+                        }) {
+                            return None;
+                        }
+
+                        (*leaf, None)
+                    }
                     CpuidKey::SubLeaf(leaf, subleaf) => (*leaf, Some(*subleaf)),
                 };
-                CpuidEntry {
+
+                Some(CpuidEntry {
                     leaf,
                     subleaf,
                     eax: value.eax,
                     ebx: value.ebx,
                     ecx: value.ecx,
                     edx: value.edx,
-                }
+                })
             })
             .collect(),
         vendor: propolis_api_types::instance_spec::CpuidVendor::Amd,
@@ -292,9 +315,14 @@ fn print_json(results: &BTreeMap<CpuidKey, Cpuid>) {
 
 #[derive(Default, Clone, Copy, Debug)]
 enum OutputFormat {
+    /// Print a human-readable plain-text representation.
     #[default]
     Text,
+
+    /// Print TOML suitable for use in a propolis-standalone config file.
     Toml,
+
+    /// Print JSON suitable for inclusion in a propolis-server instance spec.
     Json,
 }
 
--- a/bin/propolis-utils/src/bin/cpuid-gen.rs
+++ b/bin/propolis-utils/src/bin/cpuid-gen.rs
@@ -267,6 +267,35 @@ fn print_json(results: &BTreeMap<CpuidKey, Cpuid>) {
         Cpuid, CpuidEntry,
     };
 
+    let vendor = {
+        use propolis_api_types::instance_spec::{CpuidValues, CpuidVendor};
+        match results.get(&CpuidKey::Leaf(0)) {
+            None => {
+                eprintln!("no result for leaf 0, selecting default vendor");
+                CpuidVendor::Amd
+            }
+            Some(values) => {
+                let values = CpuidValues {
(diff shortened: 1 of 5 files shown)


bin/propolis-utils/src/bin/cpuid-gen.rs
```diff
@@ -261,15 +262,125 @@ fn print_toml(results: &BTreeMap<CpuidKey, Cpuid>) {
...
+                }
+            }
+        }
+    };
+
+    // Returns `true` if `key` has subleaf data and its leaf index is `leaf`.
+    fn key_matches_leaf_and_has_subleaves(key: &CpuidKey, leaf: u32) -> bool {
+        let CpuidKey::SubLeaf(l, _) = key else {
+            return false;
+        };
```
pfmooney, line 303: style nit q:

Could this be worded as
```rust
match key {
    CpuidKey::SubLeaf(l, _) if *l == leaf => true,
    _ => false,
}
```
and then be inlined into the `any()` call below?

bin/propolis-utils/src/bin/cpuid-gen.rs
```diff
@@ -261,15 +262,125 @@ fn print_toml(results: &BTreeMap<CpuidKey, Cpuid>) {
...
 #[derive(clap::Parser)]
 struct Opts {
     /// Elide all-zero entries from results
     #[clap(short)]
     zero_elide: bool,
 
-    /// Emit toml instead of text
-    #[clap(short)]
-    toml_output: bool,
+    /// Emit output in the specified format
```
pfmooney, line 381: If clap isn't already magically printing the options for this, perhaps include them in a parenthetical?

bin/propolis-utils/src/bin/cpuid-gen.rs
gjcolombo: We can go one better and use `matches!` (I didn't realize at first that it let you append a conditional, but in retrospect it makes sense that it would). Fixed in 41fc83815d2d.

bin/propolis-utils/src/bin/cpuid-gen.rs
gjcolombo: It's not; fixed in 41fc83815d2d.

bin/propolis-utils/src/bin/cpuid-gen.rs
```diff
@@ -261,15 +262,116 @@ fn print_toml(results: &BTreeMap<CpuidKey, Cpuid>) {
...
+        entries: results
+            .iter()
+            .filter_map(|(key, value)| {
+                // propolis-server will complain if it receives a CPUID entry
+                // with no subleaf and a CPUID entry with a subleaf for the same
+                // leaf. Avoid emitting these by filtering out subleaf-less
+                // entries for any leaf that has an entry with subleaf data.
+                let (leaf, subleaf) = match key {
+                    CpuidKey::Leaf(leaf) => {
+                        if results.keys().any(|k| {
```
pfmooney, line 309: If N was more than dozens, at most, I'd say to use `range()` here to constrain the haystack, but it's not worth the effort.

pfmooney (approved):

bin/propolis-utils/src/bin/cpuid-gen.rs
```diff
@@ -261,15 +262,116 @@ fn print_toml(results: &BTreeMap<CpuidKey, Cpuid>) {
...
+fn print_json(results: &BTreeMap<CpuidKey, Cpuid>) {
+    use propolis_api_types::instance_spec::components::board::{
+        Cpuid, CpuidEntry,
+    };
+
+    let vendor = {
+        use propolis_api_types::instance_spec::{CpuidValues, CpuidVendor};
+        match results.get(&CpuidKey::Leaf(0)) {
+            None => {
+                eprintln!("no result for leaf 0, selecting default vendor");
```
hawkw, line 274: super nitpicky, but perhaps this should actually say who the default vendor is?
```suggestion
                eprintln!("no result for leaf 0, selecting default vendor (AMD)");
```
or similar?

bin/propolis-utils/src/bin/cpuid-gen.rs
```diff
@@ -261,15 +262,116 @@ fn print_toml(results: &BTreeMap<CpuidKey, Cpuid>) {
...
+            "toml" => Self::Toml,
+            "json" => Self::Json,
+            _ => {
+                return Err(
+                    "invalid output format, must be text, toml, or json",
+                )
+            }
+        })
+    }
+}
```
hawkw, line 364: nitpicky: If we just derived Clap's `ValueEnum` trait for this (see: https://docs.rs/clap/latest/clap/_derive/_tutorial/chapter_3/index.html#enumerated-values), it would do this for us, and also show the possible values in the help text, which is kinda nice...

hawkw (approved):
i had a few very minor suggestions you may find useful, but this looks good to me!

bin/propolis-utils/src/bin/cpuid-gen.rs
gjcolombo: We can go one better here: since `CpuidKey`'s `Ord` implementation guarantees that `CpuidKey::Leaf` precedes `CpuidKey::SubLeaf` provided that the leaf indices are equal, we only need to compare to the next entry in the map to see if there's a conflict. I implemented this in a274706.

bin/propolis-utils/src/bin/cpuid-gen.rs
gjcolombo: Done in 4aaff74.

bin/propolis-utils/src/bin/cpuid-gen.rs
gjcolombo: This is much, much nicer. Thanks for pointing it out! Done in 4aaff74.

PR #833, opened 2025-01-15
Wire up initial support for AMD perf counters

pfmooney:
Proposed counterpart to [illumos#16906](https://www.illumos.org/issues/16906) ([CR3865](https://code.illumos.org/c/illumos-gate/+/3865))


iximeow (approved):

gjcolombo (approved):
I'm trying to think of a good PHD test or two to write for this, but this is straightforward enough that I don't think it's worth blocking the PR for. I'll file an issue if I come up with something good.

PR #834, opened 2025-01-16
build: explicitly install libsqlite3-dev in CI

gjcolombo:
Shamelessly stolen from crucible#1607. Will hopefully resolve the build issue seen in #833.


pfmooney (approved):
Thanks for fixing this

PR #836, opened 2025-01-17
build: upgrade tokio to 1.40.0

gjcolombo:
This version includes some useful new features like `tokio::io::simplex`.

Both this version and 1.43.0 pass PHD tests, but this change opts for 1.40.0 because the Omicron repo has been on that version for several weeks and has (evidently) not had any problems with it.

Tests: cargo test, PHD.


hawkw (approved):
🆙

Issue #835, opened 2025-01-16
propolis-server should specify explicit CPUID values for all VMs, even if the spec contained none


gjcolombo:
Today the `cpuid` field of an instance spec's `Board` is an `Option`: https://github.com/oxidecomputer/propolis/blob/96f1bbd39f99f21b3291f192f6680fb6de006733/crates/propolis-api-types/src/instance_spec/components/board.rs#L108-L112

Specifying `None` gives the VM some sensible defaults for whatever host it happens to be running on. This is useful for running Propolis VMs on non-Oxide hardware (for ad hoc testing, in CI, as part of an Omicron dev cluster on a commodity machine, etc.), where the user may not have a CPUID profile ready to use in advance.

propolis-server handles a `None` CPUID value by doing nothing and letting bhyve's CPUID emulation decide what values to supply. However, we're close to needing something a bit more sophisticated:

- To fix #328 (and also to enable x2APIC support), Propolis needs to set appropriate values for a VM's hypervisor leaves (0x4000_0000-FFFF)
- To set a VM's hypervisor CPUID leaves, Propolis also needs to set its standard/explicit leaves (there's no way to use bhyve-supplied defaults for some leaves but not others)
- To set those leaves, Propolis needs to be told what values to use, i.e. the `cpuid` field in the instance spec must be `Some`

The problem is that it's not obvious what values to use in software that might run on arbitrary hardware, like PHD or an Omicron dev cluster. For ad hoc servers, it's possible to work around this by using the `cpuid-gen` utility in the repo to extract host CPUID information to put into a hand-crafted instance spec, but this is at least mildly annoying to do.

I propose changing the semantics of `cpuid: None` in `propolis-server` from "use bhyve's legacy emulation" to "extract the values bhyve's legacy emulation would provide and feed them back in as explicit CPUID values"--i.e., do the `cpuid-gen` trick, but do it automatically. This would allow PHD to test enlightenments without committing to any particular set of CPUID values in the framework (and without requiring some other way to specify default CPUID values to PHD out-of-band, which just moves the problem back up to whoever is running PHD). It would also make our CI VMs behave more like Omicron production VMs (in the sense that both environments will use the "explicit CPUID" code in bhyve, even though they'll get their CPUID values in different ways).

This would be very nice to have before doing #328. 

[...]

PR #843, opened 2025-01-28
crates: move host CPUID queries from cpuid-gen to cpuid-utils

### LINKED ISSUE #793: `cpuid_boot_test` goes out to lunch on Windows Server 2022 guest
i'm using a freshly-built Windows Server 2022 image, but running PHD more manually than i probably should:
```
../../target/debug/phd-runner run \
    --propolis-server-cmd ../../target/release/propolis-server \
    --tmp-directory /tmp/phd \
    --artifact-toml-path /rpool/devel/images/windows-server-2022.toml \
    --default-guest-artifact windows \
    --include-filter cpuid_boot_test
```
where that toml is as follows:
```
root@helios:/rpool/devel/oxide# cat ../images/windows-server-2022.toml
remote_server_uris = ["https://oxide-omicron-build.s3.amazonaws.com"]

[artifacts.windows]
filename = "windows-server-2022.bin"
[artifacts.windows.kind]
guest_os = "windowsserver2022"
[artifacts.windows.source.local_path]
path = "/rpool/devel/images"

[artifacts.ovmf]
filename = "OVMF_CODE.fd"
kind = "bootrom"
[artifacts.ovmf.source.buildomat]
repo = "oxidecomputer/edk2"
series = "image_debug"
commit = "6d92acf0a22718dd4175d7c64dbcf7aaec3740bd"
sha256 = "29813374b58e3b77fb665f2d95cb3bab37d44fdd2c4fce2a70de9d76a3512a4f"
```

the `cpuid_boot_test` runs and then hangs waiting for `Computer is booting, SAC started and initialized`. it turns out the guest has deviated from the normal boot path with:
```
Windows failed to start. A recent hardware or software change might be the cause. [...snip...] Status: 0xc0000017

Info: There isn't enough memory available to create a ramdisk device
```

OK, maybe Windows sees something in the new cpuid bits that requires more than the default 512MiB of memory? i added a `cfg.memory_mib(2048);` to `cpuid_boot_test`, rebuilt the tests, and ran again. with this, Windows prints `Preparing Automatic Repair Loading files ...` and a progress bar, reboots again, and loops six times this way before finally hanging.

i presume something in the cpuid bits has Windows convinced it needs to repair itself, setting up the repair context is what needs the extra memory, and then the final hang is maybe a safeguard against a bootloop trying to get to automatic repair?

----

since this test tries to `cat /proc/cpuinfo` anyway the test will certainly, so i'll probably just `phd_skip!` this one given a Windows guest, but the failure mode still seems noteworthy.

gjcolombo:
The `cpuid-gen` utility contains logic that reads and pretty-prints bhyve's default CPUID values (or the host processor's raw CPUID outputs). Move the logic that reads these values into the `cpuid-utils` crate, re-express it in terms of `cpuid-utils`'s CPUID types, and make `cpuid-gen` call the new logic. Future changes will make propolis-server use this logic to set default CPUID values for VMs whose instance specs contain no CPUID settings.

Fix a bit shifting bug in `collect_cpuid`'s handler for standard leaf D: this leaf outputs 64-bit values across two output registers, and the old logic wasn't shifting edx into the high 32 bits before OR'ing it with its counterpart register.

Finally, simplify PHD's `cpuid_boot_test`: since it can now ask bhyve for CPUID settings that bhyve thinks are valid, it no longer needs to try to stitch together its own minimum viable CPUID profile. This test now passes with a Windows Server 2022 guest.

Tests: cargo test; PHD with Debian 11 and WS2022 guests; manually compared text and TOML output from the old and new cpuid-gen binaries.

Fixes #793. Related to #835.

### DIFF
--- a/crates/cpuid-utils/src/host.rs
+++ b/crates/cpuid-utils/src/host.rs
@@ -1,0 +1,225 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use bhyve_api::{VmmCtlFd, VmmFd};
+use propolis_types::{CpuidIdent, CpuidValues, CpuidVendor};
+use thiserror::Error;
+
+use crate::{
+    bits::{
+        AmdExtLeaf1DCacheType, AmdExtLeaf1DEax, Leaf1Ecx, EXTENDED_BASE_LEAF,
+        STANDARD_BASE_LEAF,
+    },
+    CpuidSet, SubleafInsertConflict,
+};
+
+#[derive(Debug, Error)]
+pub enum GetHostCpuidError {
+    #[error("failed to insert into the CPUID map")]
+    CpuidInsertFailed(#[from] SubleafInsertConflict),
+
+    #[error("CPUID vendor not recognized: {0}")]
+    VendorNotRecognized(&'static str),
+
+    #[error("I/O error from bhyve API")]
+    BhyveError(#[from] std::io::Error),
+}
+
+/// A wrapper around a handle to a bhyve VM that can be used to query bhyve's
+/// default CPUID values.
+struct Vm(bhyve_api::VmmFd);
+
+impl Vm {
+    fn new() -> Result<Self, GetHostCpuidError> {
+        let name = format!("cpuid-gen-{}", std::process::id());
+        let mut req = bhyve_api::vm_create_req::new(name.as_bytes())
+            .expect("valid VM name");
+
+        let ctl = VmmCtlFd::open()?;
+        let _ = unsafe { ctl.ioctl(bhyve_api::VMM_CREATE_VM, &mut req) }?;
+
+        let vm = match VmmFd::open(&name) {
+            Ok(vm) => vm,
+            Err(e) => {
+                // Attempt to manually destroy the VM if we cannot open it
+                let _ = ctl.vm_destroy(name.as_bytes());
+                return Err(e.into());
+            }
+        };
+
+        Ok(Self(vm))
+    }
+
+    fn query(
+        &self,
+        vlc_eax: u32,
+        vlc_ecx: u32,
+    ) -> Result<CpuidValues, GetHostCpuidError> {
+        let mut data = bhyve_api::vm_legacy_cpuid {
+            vlc_eax,
+            vlc_ecx,
+            ..Default::default()
+        };
+        unsafe { self.0.ioctl(bhyve_api::VM_LEGACY_CPUID, &mut data) }?;
+        Ok(CpuidValues {
+            eax: data.vlc_eax,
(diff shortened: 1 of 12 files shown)


crates/cpuid-utils/src/host.rs
```diff
@@ -0,0 +1,225 @@
...
+}
+
+impl Drop for Vm {
+    fn drop(&mut self) {
+        let _ = self.0.ioctl_usize(bhyve_api::VM_DESTROY_SELF, 0);
+    }
+}
+
+/// Queries the supplied CPUID leaf on the caller's machine.
+#[cfg(target_arch = "x86_64")]
```
pfmooney, line 81: nit: could use x86 here, rather than demanding x86-64, but it's not the first thing the cpuid area to be like that

pfmooney (approved):
Nice clean-up

phd-tests/tests/src/cpuid.rs
```diff
@@ -63,77 +63,34 @@ async fn cpuid_boot_test(ctx: &Framework) {
...
-    host_cpuid.retain_by_leaf(|leaf| {
-        leaf <= 0x7 || (0x8000_0000..=0x8000_0001).contains(&leaf)
-    });
+    // This test works by injecting a fake brand string into extended leaves
+    // 0x8000_0002-0x8000_0004 and seeing if the guest observes that string. For
+    // this to work those leaves need to be present in the host's extended CPUID
+    // leaves.
+    let Some(ext_leaf_0) = host_cpuid.get(CpuidIdent::leaf(EXTENDED_BASE_LEAF))
+    else {
+        phd_skip!("cpuid_boot_test requires extended CPUID leaves");
```
iximeow, line 85: idle suggestion: worth making this fail if we can't get at the brand string leaves? these go back to like... the 586, so if they're not present something spectacularly strange is going on that's probably worth exploding over

crates/cpuid-utils/src/host.rs
gjcolombo: I gave this a try but discovered that `rustc` appears to have no 32-bit illumos target! 

```
$ rustc --version
rustc 1.84.0 (9fc6b4312 2025-01-07)
$ rustc --print=target-list | grep illumos
aarch64-unknown-illumos
x86_64-unknown-illumos
```

I don't think we're actually querying host CPUID in any platform-agnostic tests that one might want to run against an i686 target, so I left this as-is. (I think supporting x86 would also require defining these functions twice, since the CPUID intrinsic is in `core::arch::x86` when compiling against that target, unless there's some subtlety that I'm missing.)

phd-tests/tests/src/cpuid.rs
gjcolombo: Good idea, done in 5f7f101.

gjcolombo:
Thanks @pfmooney and @iximeow for the reviews! Will merge first thing tomorrow and (hopefully) will soon have a follow-up PR to put this code to work in propolis-server.

PR #846, opened 2025-01-30
server/lib: add generic hypervisor enlightenment interface

gjcolombo:
Implement some basic infrastructure to support hypervisor enlightenments. The general idea is as follows:

- propolis-lib defines `Enlightenment` and `EnlightenmentDevice` traits that are implemented by components that provide a particular guest-hypervisor interface.
- Library users instantiate their chosen enlightenment stack and pass the resulting `Arc<dyn EnlightenmentDevice>` to the lib's `machine::Builder`. The resulting `Machine` contains vCPU objects that hold a reference to the selected `Enlightenment`. This allows them to forward RDMSR/WRMSR exits to the enlightenment stack for (possible) processing.
- Library users can call `Enlightenment::add_cpuid` to allow hypervisor interfaces to inject CPUID identifiers into the `CpuidSet`s that are used to configure the VM's vCPUs.
- Finally, library users can upcast an `EnlightenmentDevice` to an `Arc<dyn Lifecycle>` and use that reference to dispatch lifecycle and migration notifications to the active enlightenment stack.

Implement a `BhyveGuestInterface` enlightenment stack that implements these interfaces but does nothing beyond exposing its vendor string in CPUID leaf 0x4000_0000. This matches the hypervisor-related CPUID and MSR behavior guests would have gotten before this change. More interesting hypervisor interfaces are left for future PRs.

Add an instance spec field to allow propolis-server clients to select an enlightenment stack. If not specified this defaults to the bhyve interface; this provides compatibility when migrating to/from servers built without this code.

Tests: cargo test; PHD w/Alpine, Debian 11, and WS2022 guests; verified in an ad hoc VM that leaf 0x40000000 returns the correct data to a Debian guest.

Related to #328.


lib/propolis/src/vmm/machine.rs
```diff
@@ -230,6 +246,15 @@ impl Builder {
...
     }
 
+    /// Sets the guest hypervisor interface for the machine.
+    pub fn guest_hypervisor_interface(
+        mut self,
+        guest_hv: Arc<dyn EnlightenmentDevice>,
+    ) -> Self {
+        self.guest_hv_interface = Some(guest_hv);
+        self
+    }
```
gjcolombo, line 256: Putting this method on the builder itself turns out to be inconvenient for enlightenment stacks that want to access guest memory. Those enlightenments will want to hold onto a `MemAccessor`, but there's no way to get one to pass here, because the "root" `MemAccessor` for the VM gets created in `finalize`.

I'm mulling over alternative approaches. One possibility is to make this routine take a function that gets a `MemAccessor` as input and produces a `Result<Arc<dyn EnlightenmentDevice>>`. Then `finalize` can call that function after it's created the other VM objects.

Another possibility is to add a layer of indirection between the vCPUs and the enlightenment stack. In my original prototype this was done by defining an "MSR dispatcher" type that was created here, stored in the `Machine`, and passed to each `Vcpu`. Later in machine init, propolis-server would create its chosen enlightenment and register it with the MSR dispatcher. vCPUs that took RDMSR/WRMSR exits would send them to the dispatcher, which would route them to the enlightenment stack (presuming the MSR IDs are in the hypervisor range).

An advantage of having a separate MSR dispatcher is that it can be taught to swallow accesses to MSRs that we know we don't care about (and don't want to present to the library user as being totally unhandled). I was hoping not to have to write that code here, but maybe it's the most efficient way to hook all this up.

gjcolombo: b4fcbf6 takes care of this (I've layered my in-progress Hyper-V emulation changes onto it and believe it unblocks those).

lib/propolis/src/enlightenment/bhyve.rs
```diff
@@ -0,0 +1,58 @@
...
+                    edx: 0x20657679,
+                },
+            )
+            .map_err(|_| {
+                anyhow::anyhow!("reserved leaf 0x4000_0000 already in map")
+            })?;
+
+        if old.is_some() {
+            anyhow::bail!("reserved leaf 0x4000_0000 already in map");
+        }
```
hawkw, line 46: I was a bit unsure why these both returned the same error, so I had to go re-read the `CpuidSet` implementation to figure this out. It's because `insert` returns an error if there's a conflict between the inserted values and whether or not an already-present value would have subleaves, and here, if we ever see that error, it implies that there's a value present, and we expect there not to be. So, using the same error message makes sense upon consideration, but it _might_ be nice to have a comment here writing that down for future readers?

lib/propolis/src/enlightenment/mod.rs
```diff
@@ -0,0 +1,95 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Components that implement enlightenments: mechanisms that allow guest
```
hawkw, line 5: suuuuuper nitpicky, feel free to totally disregard me: since we are introducing the term "enlightment" here, i might italicize it?
```suggestion
//! Components that implement _enlightenments_: mechanisms that allow guest
```

lib/propolis/src/vmm/machine.rs
gjcolombo: Fixed more permanently in 37b708c.

lib/propolis/src/enlightenment/bhyve.rs
gjcolombo: Fixed in 616ed70.

lib/propolis/src/enlightenment/mod.rs
gjcolombo: Fixed in c185b4b.

lib/propolis/src/enlightenment/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+//! software to cooperate directly with its host hypervisor.
+//!
+//! Although the high-level point of a virtual machine is to allow guest
+//! software to behave as though it was running on its own physical hardware,
+//! there are (as with any abstraction) places where a virtual machine is bound
+//! to behave differently than a "real" computer. For example, accessing a piece
+//! of chipset functionality (like a hardware timer) might be relatively
+//! inexpensive on real hardware, but in a virtual machine it requires an
+//! expensive VM exit. Similarly, guest OSes may generally assume that hardware
+//! timer interrupts will be delivered and serviced very promptly--promptly
```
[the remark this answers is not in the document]
gjcolombo: Fixed in c185b4b.

lib/propolis/src/enlightenment/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+//!    identify the host's hypervisor type and its capabilities.
+//! 2. The host can trap the RDMSR and WRMSR instructions to allow the guest
+//!    to communicate via synthetic model-specific registers.
+//! 3. The host and guest can agree to use pages of guest physical memory to
+//!    share data.
+//! 4. The guest can execute special instructions (VMCALL or VMMCALL) that exit
+//!    to the host with a special exit type that indicates that the guest
+//!    executed a "call hypervisor" instruction. Since the guest CPU's registers
+//!    are saved on exit, the host can read them and interpret them as
+//!    parameters to a virtual function call.
```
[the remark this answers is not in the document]
gjcolombo: I rewrote chunks of this in ddaff15 to try to draw a clearer distinction here. Let me know what you think!

FWIW: QEMU is happy, at least on KVM, to pretend to be Hyper-V, but it does that by [delegating CPUID selection to a KVM ioctl that returns the Hyper-V leaves describing the features KVM supports](https://github.com/qemu/qemu/blob/871af84dd599fab68c8ed414d9ecbdb2bcfc5801/target/i386/kvm/kvm.c#L1109). In the longer run, we could do something similar in bhyve (along the lines of asking it what Hyper-V-esque features it supports), and we might very well want to do that for something like (say) a spinlock yield enlightenment, where we want to respond to a guest's signal by changing which vCPUs we're scheduling. But for now it's simplest just to do everything up in Propolis. RFD 517 gets into all this in a little more detail.

lib/propolis/src/vmm/machine.rs
```diff
@@ -238,17 +268,24 @@ impl Builder {
 
         let bus_mmio = Arc::new(MmioBus::new(MAX_PHYSMEM));
         let bus_pio = Arc::new(PioBus::new());
-
         let acc_mem = MemAccessor::new(map.memctx());
         let acc_msi = MsiAccessor::new(hdl.clone());
 
+        let guest_hv_interface = match self.guest_hv_interface.take() {
+            Some(func) => {
+                func(acc_mem.child(Some("guest_hv_interface".to_string())))?
```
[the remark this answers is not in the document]
gjcolombo: I thought I'd done this but it turns out it was only in my Hyper-V branch. Will pull that back into this series.

lib/propolis/src/enlightenment/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    msr::{MsrId, RdmsrOutcome, WrmsrOutcome},
+};
+
+pub mod bhyve;
+
+/// A supertrait describing an enlightenment stack that can also receive
+/// instance lifecycle notifications. Automatically implemented for any
+/// component that implements both [`Enlightenment`] and [`Lifecycle`], which
+/// includes all the enlightenment stacks in this module's submodules.
+pub trait EnlightenmentDevice: Enlightenment + Lifecycle {
```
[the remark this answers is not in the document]
gjcolombo: Under the previous initialization method (where `Machine` held a trait object that the initializer needed to upcast) I couldn't find a way to do this without running afoul of an object safety rule somewhere or another. Fortunately, there's a better way to initialize the enlightenment stack, which I've implemented in 37b708c, along with a fix for this.

lib/propolis/src/vmm/machine.rs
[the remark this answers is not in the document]
gjcolombo: Ah, I know what happened: we're not actually calling `Accessor::child` anymore in this path with 37b708c in place, so no one is actually supplying a child accessor name here. The Hyper-V branch does call this routine (actually it calls `Accessor::adopt`), but this problem (should) be fixed over there when I push those changes.

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -1231,21 +1246,16 @@ impl MachineInitializer<'_> {
...
-                    CpuidIdent::leaf(HYPERVISOR_BASE_LEAF),
-                    HYPERVISOR_BHYVE_VALUES
-                )
-                .expect("no hypervisor subleaves")
-                .is_none(),
-                "CPUID set should have no hypervisor leaves"
-            );
+            hv_interface
+                .add_cpuid(&mut set)
+                .expect("CPUID in spec should have no hypervisor leaves");
```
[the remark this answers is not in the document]
gjcolombo: Fixed in caa8cdf.

lib/propolis/src/enlightenment/bhyve.rs
```diff
@@ -0,0 +1,60 @@
...
+};
+
+use crate::{
+    accessors::MemAccessor,
+    common::{Lifecycle, VcpuId},
+    enlightenment::Enlightenment,
+    msr::{MsrId, RdmsrOutcome, WrmsrOutcome},
+};
+
+pub struct BhyveGuestInterface;
```
pfmooney, line 21: This should probably get a comment about why it's here: just as a placeholder for the old CPUID leafs (providing whatever little value they hold), and nothing more.  This, given that there are no actual enlightenment features from the "bhyve brand" of hypervisor

lib/propolis/src/enlightenment/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// with the supplied [`MemAccessor`] at the root of its memory accessor
+    /// hierarchy.
+    ///
+    /// Enlightenment stacks that want to access guest memory should call
+    /// [`MemAccessor::new_orphan`] when they are created, and then should call
+    /// [`MemAccessor::adopt`] from this function.
+    ///
+    /// Users of an enlightenment stack must guarantee that this function is
+    /// called only once per instance of that stack.
+    fn set_parent_mem_accessor(&self, parent: &MemAccessor);
```
pfmooney, line 89: I would just name it something like `attach` (still passing in the `MemAccessor`), to make it a bit more generic, and in case we choose to wire up other bits (MMIO, etc) for other enlightments.

lib/propolis/src/enlightenment/mod.rs
```diff
@@ -48,36 +48,34 @@ use std::sync::Arc;
...
```
hawkw, line 89: It occurs to me that there _could_ be a way to guarantee this, by having this function take `self` and return an `Arc<Self>`. Which is a bit weird, as it means the implementation of `attach` must also be responsible for `Arc`ing the enlightenment stack, and the `machine::Builder::guest_hypervisor_interface` would take the enlightenment stack by value --- so we wouldn't be able to pass an `Arc` clone elsewhere before we build the `Machine`, which we might need to do?

Just a thought, not sure if it's a good idea.

hawkw: > Can you do that in a trait function?

[You can](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6e4d39b8be2eb00ae23879f3c345d21a), but upon further consideration I've remembered that the `self` receiver will make the trait no longer object-safe, which I think is a non-starter here.

> Regardless, we have the `attach` pattern in a number of places around the stack. The expectations expressed in the comment are not unreasonable, IMO.

Yeah, I agree --- I was just spitballing with this, I think it's probably not actually a good idea.

pfmooney: >  I've remembered that the `self` receiver will make the trait no longer object-safe, which I think is a non-starter here.

Ah yeah, it was object safety I was thinking of.

gjcolombo: I'm not totally sure we can get away with this, but I'd have to try it and see, I think.

> Can you do that in a trait function?

You can if you're willing to put a `Sized` bound on `Self`, i.e. it would be written as

`fn attach(self, mem_acc: &MemAccessor) -> Arc<Self> where Self: Sized;`

The caveat (AIUI) is that callers have to be able to call this function on a concrete type that impls `Enlightenment` (for `dyn` compatibility reasons). That's a problem in the current initialization path because `Builder::guest_hypervisor_interface` currently takes a trait object, and we don't want it to take a concrete `BhyveGuestInterface` because we'll soon have other types of interfaces. We could avoid *that* by turning `Builder` into `Builder<T: Enlightenment>` and then rewriting `guest_hypervisor_interface` to take a `T`, but I'm not sure what the blast radius of adding that generic would be.

The other challenge is that we need to be able to turn the enlightenment stack into a `Arc<dyn Lifecycle>` that can be added to the VM's device map. Once the `Machine` has converted the builder's stack into an `Arc<dyn Enlightenment>` it's too late--we can't call the `as_lifecycle` upcasting function through a `dyn Enlightenment` trait object because of the `Self: Sized` bound. We could get around all *that* by going back to the blanket upcast impl for anything that's `Enlightenment + Lifecycle`, but...

My inclination is to leave this as-is for this PR--there's only one code path that deals with VM component initialization in propolis-server, so I think the marginal value of enforcing the "must call exactly once" discipline for this type is lower than it would be if we were dealing with a type that we construct ubiquitously.

With all that said, I would give myself a C-grade grasp of the dyn compatibility rules, so I could be missing something about what you're proposing--let me know if that's the case.

gjcolombo:
@pfmooney - thanks for taking a look! 20f1204 should address all your comments.

lib/propolis/src/enlightenment/mod.rs
```diff
@@ -104,3 +105,62 @@ pub trait Enlightenment: Lifecycle + Send + Sync {
...
+        // appear either with or without subleaf entries (but not both). Its
+        // `insert` method returns `Err` if an insertion would violate this
+        // invariant and `Ok(Some)` if the insertion replaced an existing entry.
+        // If either of these cases arises, the input map contained a
+        // conflicting hypervisor entry, which is a caller error.
+        //
+        // Note that because `to_add` is itself a `CpuidSet`, it cannot contain
+        // a leaf/subleaf conflict or a duplicate leaf entry. Therefore, any
+        // conflicts of this kind must originate with the contents of
+        // `to_modify`.
```
hawkw, line 155: Thanks for documenting this!

lib/propolis/src/enlightenment/mod.rs
```diff
@@ -94,7 +95,7 @@ pub trait Enlightenment: Lifecycle + Send + Sync {
     /// CPUID leaves from 0x4000_0000 to 0x4000_00FF are reserved for the
     /// hypervisor's use. On entry, the caller must ensure that `cpuid` does not
     /// contain any leaf entries in this range.
-    fn add_cpuid(&self, cpuid: &mut CpuidSet) -> anyhow::Result<()>;
+    fn add_cpuid(&self, cpuid: &mut CpuidSet) -> Result<(), AddCpuidError>;
```
hawkw, line 98: take it or leave it: perhaps the `add_cpuid` method ought to still return an `anyhow::Result`, so that future implementations could return other errors as well as an `AddCpuidError`? i could maybe imagine an enlightenment stack which sets a more complicated set of hypervisor leaves wanting to return other errors if the enlightenment stack is misconfigured?

i dunno if this is actually something we want to do --- maybe it's better to constrain this to only return `AddCpuidError`s specifically to force other errors in the enlightenment stack's CPUID configuration to be returned elsewhere? what do you think?

hawkw (approved):

lib/propolis/src/enlightenment/mod.rs
pfmooney: I'd be inclined to leave it for now.  Since it's largely an internal interface, we're free to revisit later if need be.

gjcolombo: I also sort of expect enlightenments not to fail this call for reasons other than a conflict (though my imagination might be limited). If a user asks to configure an enlightenment such that it has no valid CPUID values, I'd hope to catch that in the enlightenment's constructor. (So far so good with the Hyper-V prototype, anyway!)

hawkw: Yup, makes sense!

PR #848, opened 2025-02-04
lib: don't send CPUID settings in vCPU device state payload

### LINKED ISSUE #847: `Vcpu::get_cpuid` always returns an empty `CpuidSet`
Observed during ad hoc testing.

The problem is that `get_cpuid` uses `Vec::with_capacity` to create an entry array, but never sets its actual length: https://github.com/oxidecomputer/propolis/blob/c0bb4cd800917f3ccd059c969adf309212a9a684/lib/propolis/src/vcpu.rs#L249-L256

This breaks live migration of CPUID settings because the empty CPUID set gets applied during the device state phase even if the target applied different CPUID settings when processing the source's instance spec.

This can be fixed trivially by calling `set_len` in this path. That said, this entire code path might be obsolete now: with #844 in place, the instance specs sent during live migration always have the correct CPUID settings, so exporting and importing them duplicates work. (See the discussion in #844 for more details.) Given this it might make even more sense just to remove CPUID settings from the vCPU device state payload entirely.

gjcolombo:
Remove the CPUID portion of the vCPU device state payload that gets sent during live migration. #847 prevented this payload from being constructed correctly. With #844 in place, however, the instance specs that targets receive during live migrations will always contain the source's explicitly-applied CPUID settings, even if the VM's "original" instance spec (the one used to create it) contained no explicit CPUID values. This means there's no longer any need to transfer these settings in the device state phase.

Upgrade PHD's `cpuid_boot_test` to detect this issue and verify the fix.

This change breaks the migrate-from-base tests because migration targets with the fix no longer accept all of the payload offers from sources without the fix. (In the future, when this sort of change is no longer allowed, targets that no longer care about a segment of a source device's payload offering will have to consume that segment explicitly but then ignore its contents.)

Fixes #847.

### DIFF
--- a/phd-tests/tests/src/cpuid.rs
+++ b/phd-tests/tests/src/cpuid.rs
@@ -3,11 +3,13 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use cpuid_utils::{CpuidIdent, CpuidValues, CpuidVendor};
+use phd_framework::test_vm::MigrationTimeout;
 use phd_testcase::*;
 use propolis_client::types::{
     CpuidEntry, InstanceSpecStatus, VersionedInstanceSpec,
 };
 use tracing::info;
+use uuid::Uuid;
 
 fn cpuid_entry(
     leaf: u32,
@@ -48,21 +50,11 @@ async fn cpuid_instance_spec_round_trip_test(ctx: &Framework) {
 }
 
 #[phd_testcase]
-async fn cpuid_boot_test(ctx: &Framework) {
+async fn cpuid_boot_and_migrate_test(ctx: &Framework) {
     use cpuid_utils::bits::*;
 
-    // This test verifies that plumbing a reasonably sensible-looking set of
-    // CPUID values into the guest produces a bootable guest.
-    //
-    // The definition of "reasonably sensible" is derived from the suggested
-    // virtual-Milan CPUID definitions in RFD 314. Those are in turn derived
-    // from reading AMD's manuals and bhyve's source code to figure out what
-    // host CPU features to hide from guests.
-    //
-    // To try to make this test at least somewhat host-agnostic, read all of the
-    // host's CPUID leaves, then filter to just a handful of leaves that
-    // advertise features to the guest (and that Linux guests will check for
-    // during boot).
+    // Use the current host's CPUID values to try to make this test
+    // host-agnostic.
     let mut host_cpuid = cpuid_utils::host::query_complete(
         cpuid_utils::host::CpuidSource::BhyveDefault,
     )?;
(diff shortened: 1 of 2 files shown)


phd-tests/tests/src/cpuid.rs
```diff
@@ -48,21 +50,11 @@ async fn cpuid_instance_spec_round_trip_test(ctx: &Framework) {
 }
 
 #[phd_testcase]
-async fn cpuid_boot_test(ctx: &Framework) {
+async fn cpuid_boot_and_migrate_test(ctx: &Framework) {
```
hawkw, line 53: should we, perhaps, leave the previous test in place, and add an additional test for the migration stuff? it seems like it's maybe worthwhile to have separate tests for "booting a VM with CPUID values works" and "CPUID values are not eaten on live migration" so we can detect those two failures separately?

on the other hand, setting up a PHD test is fairly expensive compared to e.g. typical unit tests, so maybe combining it into one test case is useful...I just felt like it was worth asking about.

phd-tests/tests/src/cpuid.rs
```diff
@@ -48,21 +50,11 @@ async fn cpuid_instance_spec_round_trip_test(ctx: &Framework) {
...
-    //
-    // The definition of "reasonably sensible" is derived from the suggested
-    // virtual-Milan CPUID definitions in RFD 314. Those are in turn derived
-    // from reading AMD's manuals and bhyve's source code to figure out what
-    // host CPU features to hide from guests.
-    //
-    // To try to make this test at least somewhat host-agnostic, read all of the
-    // host's CPUID leaves, then filter to just a handful of leaves that
-    // advertise features to the guest (and that Linux guests will check for
-    // during boot).
```
hawkw, line 65: this seems like information that we should hang onto here for future readers? or is it no longer true?

phd-tests/tests/src/cpuid.rs
hawkw: also, N.B. that the `phd_skip!` macro invocation a few lines down refers to the test by its previous name, maybe this should be updated? (can't comment on that line directly as it's not part of the diff)

hawkw (approved):
Looks good to me! I left a couple comments on the test, but they're not blockers.

phd-tests/tests/src/cpuid.rs
gjcolombo: Good catch--I'll rearrange this a bit.

> (can't comment on that line directly as it's not part of the diff)

someday...

phd-tests/tests/src/cpuid.rs
gjcolombo: It's obsolete as of #844, but I forgot to remove it in that PR. The test used to try to cobble together its own CPUID values that it thought would produce a bootable VM; now it just asks bhyve for the defaults that it would otherwise have supplied.

phd-tests/tests/src/cpuid.rs
gjcolombo: 5298e2a should help with this.

Issue #482, opened 2023-08-04
want instrumentation/"get my state" methods for Propolis entities


gjcolombo:
Sometimes when debugging live Propolises we'd like to be able to examine the contents of an entity's data structures (e.g. NVMe or Crucible upstairs queues) even when the entity is stuck or otherwise at rest (such that it won't cross any USDT probes or log lines in the course of its continued operation).

One possible way to do this:

- add a `trace` function to the `Entity` trait
- each implementation takes locks, examines internal data structures, etc. and logs output and/or fires USDT probes to describe the current state of the system
- Propolis server exposes an API endpoint that visits each entity in turn and calls the tracing callback

The user can then specify which probes to enable, what filtering rules to apply when reading the logs, etc. to get only information from the relevant entities or subsystems.

An alternative approach is to let the user specify exactly which entities should get tracing callbacks. This cuts runtime expenses (no visiting entities you don't care about) in exchange for forcing users to understand how entities were named.

pfmooney:
In the limit, I'd like to see category-specific interfaces (such as common to all block devices, for example) for this sort of things, but in the mean time, a generic "trace" interface seems like a good balance to strike.

leftwo:
How does the `trace` function work?  Tell me more :)

pfmooney:
Do we want those things to emit via their logger, or some output explicitly passed into the `trace()` endpoint?  I would lean toward the latter, so when someone is gathering trace data, they can do so without having to sift it out of the logs.

gjcolombo:
I was thinking the latter as well--we can even create a child logger per entity to ensure that all events emitted this way identify who emitted them (without making the `trace` implementations remember to do it):

```rust
for entity in self.entities.iter() {
    // Create a child logger with a KV entry identifying the entity for filtering purposes.
    // this way no one can forget to log exactly which instance of an entity they are;
    // e.g. all your VNICs will have distinct entity names that can be filtered on here.
    //
    // Could also add a KV entry identifying this as an event emitted by this path.
    let logger = self.log.new(slog::o!("entity" => entity.name()));
    entity.trace(logger);
}
```

luqmana:
I do like the idea of some way of having entities dump out their state. But this is still kind of a cooperative mechanism, e.g. relying on being able to take internal locks. For normal operations that's probably fine but for an abnormal case (which I imagine is when this kind of info might be most useful) we might be stuck?

[...]

PR #849, opened 2025-02-05
lib: emulate Hyper-V enlightenment stack

gjcolombo:
Add the beginnings of a Hyper-V enlightenment stack to complement the bhyve enlightenment stack. This change aims to provide the absolute minimum functionality needed for a hypervisor to claim to expose a Hyper-V-compatible guest-hypervisor interface. These minimum requirements are described in the doc comments throughout this change.

Add PHD tests that boot VMs with the Hyper-V enlightenment stack enabled and (if the guest supports it) that try to query the hypervisor interface from within the guest.

Add a simple `propolis-cli` option to enable the Hyper-V stack. Once the stack has some actual features, this may be better expressed as a table in the `propolis-config-toml` crate, but this gives users a quick and easy way to enable the new interface.

Tests: PHD with Debian 11, Ubuntu 22.04, Alpine, and Windows Server 2022 guests; repeated the Debian 11 and WS2022 runs with the Hyper-V enlightenment stack enabled by default for all PHD VMs. Tested with an ad hoc Debian 11 VM, verifying that it detected the Hyper-V hypervisor, that the guest wrote appropriate values to the Hyper-V MSRs, and (through mdb-bhyve) that the correct instruction sequence appears on the guest's hypercall page.

Related to #328.


lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+        // a day.
+        let outcome = if new.enabled() {
+            let memctx = self
+                .acc_mem
+                .access()
+                .expect("guest memory is always accessible during wrmsr");
+
+            let region = GuestRegion(new.gpa(), PAGE_SIZE);
+            if let Some(mapping) = memctx.writable_region(&region) {
+                write_overlay_page(&mapping, &hypercall_page_contents());
```
gjcolombo, line 130: FWIW, a minor incompatibility here: the TLFS also says that once a page is donated this way, writes to it should produce a #GP, but I'm not sure bhyve gives us the ability to rewrite the protection masks for an individual page in a previously-allocated guest memory region. But I'm not inclined to lose much sleep over it, since the guest will need to map this page as executable, and if it's wise it will hopefully not map it as WX...

lib/propolis/src/vmm/mem.rs
```diff
@@ -910,6 +910,12 @@ impl MemCtx {
         let mapping = self.region_covered(region.0, region.1, Prot::READ)?;
         Some(mapping)
     }
+    pub fn read_write_region(
+        &self,
+        region: &GuestRegion,
+    ) -> Option<SubMapping> {
+        self.region_covered(region.0, region.1, Prot::RW)
+    }
```
gjcolombo, line 918: This is dead code (used when I had a more complicated overlay page construct). Will remove.

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+        // leaving the hypercall page untouched (as would happen if the guest
+        // manually cleared this bit).
+        if value == 0 {
+            info!(&self.log, "guest cleared HV_X64_MSR_GUEST_OS_ID");
+            inner.msr_hypercall_value.clear_enabled();
+        } else {
+            info!(
+                &self.log,
+                "guest set HV_X64_MSR_GUEST_OS_ID";
+                "value" => value
```
hawkw, line 82: nitpick: this logs the value in base-10, which is probably not the most useful format for us?

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+        //
+        // The TLFS describes this page as an "overlay" that "covers whatever
+        // else is mapped to the GPA range." The spec is ambiguous as to whether
+        // this means that the page's previous contents must be restored if the
+        // page is disabled or its address later changes. Empirically, most
+        // guest OSes don't appear to move the page after creating it,
+        // and at least some other hypervisors don't bother with saving and
+        // restoring the old page's contents. So, for simplicity, simply write
+        // the hypercall instruction sequence to the requested page and call it
+        // a day.
```
hawkw, line 121: The term "overlay" sure does _seem_ to suggest that the previous contents of the page should come back, but if guests don't seem to mind clobbering it, we can always come back and fix this if it turns out to cause problems. This sure feels like one of those situations where the spec _meant_ one thing but enough implementations did the other thing that now it's the _de facto_ correct bheavior...

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+                .access()
+                .expect("guest memory is always accessible during wrmsr");
+
+            let region = GuestRegion(new.gpa(), PAGE_SIZE);
+            if let Some(mapping) = memctx.writable_region(&region) {
+                write_overlay_page(&mapping, &hypercall_page_contents());
+                info!(
+                    &self.log,
+                    "established hypercall page at GPA {:#x}",
+                    new.gpa().0
```
hawkw, line 134: not something we need to change in _this_ PR, but it kinda feels to me like `GuestAddr` ought to have a manual `Debug` implementation that always formats itself in hex, instead of deriving `Debug`...I can't really imagine when we would ever want it to format itself in decimal...

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+        } else {
+            WrmsrOutcome::Handled
+        };
+
+        // Commit the new MSR value if the write was handled.
+        if outcome == WrmsrOutcome::Handled {
+            inner.msr_hypercall_value = new;
+        }
+
+        outcome
```
hawkw, line 155: nit, take it or leave it: it seems like this logic could be simpler if we used an early return in the #GP case?

```suggestion
        if new.enabled() {
            let memctx = self
                .acc_mem
                .access()
                .expect("guest memory is always accessible during wrmsr");

            let region = GuestRegion(new.gpa(), PAGE_SIZE);
            if let Some(mapping) = memctx.writable_region(&region) {
                write_overlay_page(&mapping, &hypercall_page_contents());
                info!(
                    &self.log,
                    "established hypercall page at GPA {:#x}",
                    new.gpa().0
                );

            } else {
                info!(
                    &self.log,
                    "guest requested illegal hypercall GPA {:#x}",
                    new.gpa().0
                );
                return WrmsrOutcome::GpException;
            }
        }

        // Commit the new MSR value if the write was handled.
        inner.msr_hypercall_value = new;

        WrmsrOutcome::Handled
```
feel free to disagree, though --- maybe the explicit outcome in every branch is clearer to you...

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+            }
+            _ => RdmsrOutcome::NotHandled,
+        }
+    }
+
+    fn wrmsr(&self, _vcpu: VcpuId, msr: MsrId, value: u64) -> WrmsrOutcome {
+        match msr.0 {
+            HV_X64_MSR_GUEST_OS_ID => self.handle_wrmsr_guest_os_id(value),
+            HV_X64_MSR_HYPERCALL => self.handle_wrmsr_hypercall(value),
+            HV_X64_MSR_VP_INDEX => WrmsrOutcome::GpException,
```
hawkw, line 211: since we're about to fault the guest, should we log this maybe?

phd-tests/tests/src/hyperv.rs
```diff
@@ -0,0 +1,77 @@
...
+        // these distros, it's easiest to use `systemd-detect-virt` to determine
+        // whether the guest thinks it's running on a Hyper-V-compatible
+        // hypervisor. (Whether any actual enlightenments are enabled is another
+        // story, but those can often be detected by other means.)
+        let out = vm.run_shell_command("systemd-detect-virt").await?;
+        if out.contains("systemd-detect-virt: not found") {
+            warn!(
+                "guest doesn't support systemd-detect-virt, can't verify it \
+                detected Hyper-V support"
+            );
```
hawkw, line 24: Would it make sense to `phd-skip` out of the test in this case?

phd-tests/tests/src/hyperv.rs
```diff
@@ -0,0 +1,77 @@
...
+        } else {
+            assert_eq!(out, "microsoft");
+        }
+    } else if vm.guest_os_kind().is_windows() {
+        // Windows is good about giving signals that it's running in a Hyper-V
+        // *root partition*, but offers no clear signal as to whether it has
+        // detected a Hyper-V host when it's running as a non-root guest. (There
+        // are methods for detecting whether Windows is running as a guest, but
+        // these don't identify the detected hypervisor type.)
+        warn!("running on Windows, can't verify it detected Hyper-V support");
```
hawkw, line 34: similarly, should we `phd_skip` here?

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+                        .to_string(),
+                ));
+            }
+
+            let Some(mapping) = ctx
+                .mem
+                .writable_region(&GuestRegion(hypercall_msr.gpa(), PAGE_SIZE))
+            else {
+                return Err(MigrateStateError::ImportFailed(
+                    "couldn't map GPA in hypercall MSR".to_string(),
```
hawkw, line 283: nit, take it or leave it: is it worth including the specified GPA here? we're `to_string`ing anyway so we may as well `format!`...

hawkw (approved):
Overall, this looks great to me, and I didn't note any blocking concerns. Left a bunch of minor suggestions, hope they're helpful

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+        let old = inner.msr_hypercall_value;
+
+        // Per the spec, this MSR is immutable once the Locked bit is set.
+        if old.locked() {
+            return WrmsrOutcome::Handled;
+        }
+
+        // If this MSR is written when no guest OS ID is set, the Enabled bit is
+        // cleared and the write succeeds.
+        let mut new = MsrHypercallValue(value);
```
iximeow, line 103: we're taking the guest's provided bits 2 through 11 here, which don't get used but i'm not sure we really _want_ to retain? i suppose the TLFS doesn't say anything other than guests _should_ preserve these bits, and #GP if any of the reserved bits are touched might be too dramatic.

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+            let Some(mapping) = ctx
+                .mem
+                .writable_region(&GuestRegion(hypercall_msr.gpa(), PAGE_SIZE))
+            else {
+                return Err(MigrateStateError::ImportFailed(
+                    "couldn't map GPA in hypercall MSR".to_string(),
+                ));
+            };
+
+            write_overlay_page(&mapping, &hypercall_page_contents());
```
iximeow, line 287: so i realize this is pretty unlikely but with the _should_-but-doesn't #GP behavior for the page itself, at this point a guest could have modified the contents of the page which we'll be clobbering with a new hypercall page. can we check on export that the page is unmodified, and only successfully export the state if we know the recipient can reproduce the same contents? or should we just move the whole hypercall page?

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        // TLFS section 3.13 says that the hypercall page "becomes disabled" if
+        // the guest OS ID register is cleared after the hypercall register is
+        // set. It also specifies that attempts to set the Enabled bit in that
+        // register will be ignored if the guest OS ID is zeroed, so handle this
+        // case by clearing the hypercall MSR's Enabled bit but otherwise
+        // leaving the hypercall page untouched (as would happen if the guest
+        // manually cleared this bit).
+        if value == 0 {
+            info!(&self.log, "guest cleared HV_X64_MSR_GUEST_OS_ID");
```
pfmooney, line 76: Do we really want this at `info` level?

lib/propolis/src/enlightenment/hyperv/hypercall.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Returns whether the hypercall page location is locked. Once locked, the
+    /// value in `MSR_HYPERCALL` cannot change until the system is reset.
+    pub fn locked(&self) -> bool {
+        (self.0 & 2) != 0
+    }
+
+    /// Indicates whether the hypercall page is enabled.
+    pub fn enabled(&self) -> bool {
+        (self.0 & 1) != 0
+    }
```
pfmooney, line 42: Add some const definitions for these magic numbers?

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: I'd like to be able to walk up to a system where enlightenments don't seem to be working properly and find out what state it's in, but I think we can do that without allowing the guest to spam things into the logs. I've at least switched this to a USDT probe in 5bf277f7; this would also be a good thing to include in #482.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: Fixed in 5bf277f7 by switching this to a USDT probe.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: Added a note about these bits in 5bf277f7.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: Sooooo I gave this another look, and it turns out that the TLFS *does* specify this explicitly. It just does it in a completely different section (5.2.1) that's not cross-referenced from section 3.13. 

"If an overlay page is disabled or is moved to a new location in the GPA space, the underlying GPA page is
“uncovered”, and an existing mapping becomes accessible to the guest."

This suggests I need to bring back the "overlay page" abstraction that I previously discarded from this PR (particularly because the TSC page is also an overlay page). However, I'm a little inclined to do that in a follow-up PR, since (a) it will add a fair bit of code, and (b) I think we can get away with not saving the original hypercall page contents in the very short run, since at least one other implementation seems not to and no guest seems to care about relocating the page. I added a comment about this in c79ad3e; if this approach sounds good to you I'll also file a follow-up issue.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: I've removed this particular `info!` statement in favor of a USDT probe but plan to leave `GuestAddr`'s `Debug` impl for another day.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: Cleaned up in 5bf277f.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: I'd be a little inclined to add a probe for this sort of thing at the rdmsr/wrmsr handler level in the vCPU code, but since this is indeed the defined outcome of trying to write to HV_X64_MSR_VP_INDEX I've left this alone for now.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: Fixed in 5bf277f7.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: I'll have a look at this as part of the overlay page adjustments I described up above (since that's also going to change the way we handle migration of overlays).

lib/propolis/src/vmm/mem.rs
gjcolombo: Fixed in 12db242e (though we'll see if it has to come back later).

phd-tests/tests/src/hyperv.rs
gjcolombo: I considered this but think it's useful for the smoke tests in this module to produce a "Pass" result to show that they successfully launched a guest in a Hyper-V-emulating Propolis, even if the guest didn't offer a means to detect the hypervisor type. (In an earlier version of this PR I had a bug that made the enlightenment stack panic on shutdown...) I've added a comment about this in c721e11e.

phd-tests/tests/src/hyperv.rs
gjcolombo: Fixed in c721e11e.

phd-tests/tests/src/hyperv.rs
hawkw: Ah, that's a fair point, thanks for adding the comment!

hawkw:
Thanks for addressing the review feedback, the new changes look good!

lib/propolis/src/enlightenment/hyperv/hypercall.rs
```diff
@@ -0,0 +1,81 @@
...
+
+//! Support for hypercalls and their related MSRs.
+
+use crate::common::{GuestAddr, PAGE_SHIFT, PAGE_SIZE};
+
+const LOCKED_BIT: u64 = 1;
+const LOCKED_MASK: u64 = 1 << LOCKED_BIT;
+const ENABLED_BIT: u64 = 0;
+const ENABLED_MASK: u64 = 1 << ENABLED_BIT;
+const GPA_MASK: u64 = !((1 << PAGE_SHIFT) - 1);
```
pfmooney, line 13: Could use `PAGE_MASK` here, or in place of `GPA_MASK` all together.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: Robust overlay page management is big enough to merit its own PR. I've filed #850 to track that work. I don't think it should block this PR, because in practice guests seem not to care about disabling or relocating their hypercall pages, and in any case we're not enabling this enlightenment for Omicron-managed VMs yet, but we'll definitely want to fix this before we flip that switch.

lib/propolis/src/enlightenment/hyperv/hypercall.rs
gjcolombo: Done in dffcd6b.

Issue #850, opened 2025-02-06
lib: want better Hyper-V overlay page management


gjcolombo:
#849 takes a very cavalier approach to Hyper-V overlay pages; in particular it does not ensure that a guest physical page's original contents are restored when all overlays are removed. This breaks the semantics described in section 5.2.1 of the Hyper-V TLFS.

The only overlay used in #849 is the hypercall page overlay, and empirically, most guest OSes seem not to care about disabling or relocating this page once they've established it. Nevertheless, Propolis should follow the TLFS's rules for these pages.

Needed for #328.

PR #851, opened 2025-02-07
lib: add better management of Hyper-V overlay pages

gjcolombo:
Add to the Hyper-V module some supporting types to support overlay pages. A Hyper-V overlay page is a sort of auxiliary page of guest memory that "covers up" or temporarily "replaces" a specific physical page in the guest's physical memory, such that accesses to the overlaid GPA end up accessing the overlay instead. See the doc comments throughout `overlay.rs` for many more details.

To avoid creating a 2,500-line patch, this PR omits direct migration of overlay manager state. Instead, the target Hyper-V stack re-creates all overlays during import based on the contents of its imported MSRs. This is somewhat dangerous: a single guest page is allowed to have multiple overlays, and there's nothing in this change that guarantees that the source and target will agree on which overlay is active at migration time. The danger is mitigated here by having only one overlay page in the whole VM (the hypercall page), but this will need to be addressed in a follow-up PR before adding any more overlay kinds (like a reference TSC page).

Tests: cargo tests, including new unit tests of the overlay manager; PHD runs with Alpine and Debian 11 guests; manually migrated a VM and made sure the hypercall page was preserved properly over migration.

Related to #850 (but doesn't close it since migration work is still pending).

### DIFF
--- a/lib/propolis/src/enlightenment/hyperv/hypercall.rs
+++ b/lib/propolis/src/enlightenment/hyperv/hypercall.rs
@@ -4,7 +4,7 @@
 
 //! Support for hypercalls and their related MSRs.
 
-use crate::common::{GuestAddr, PAGE_MASK, PAGE_SIZE};
+use crate::common::{GuestAddr, PAGE_MASK, PAGE_SHIFT, PAGE_SIZE};
 
 const LOCKED_BIT: u64 = 1;
 const LOCKED_MASK: u64 = 1 << LOCKED_BIT;
--- a/lib/propolis/src/enlightenment/hyperv/hypercall.rs
+++ b/lib/propolis/src/enlightenment/hyperv/hypercall.rs
@@ -4,7 +4,10 @@
 
 //! Support for hypercalls and their related MSRs.
 
-use crate::common::{GuestAddr, PAGE_MASK, PAGE_SHIFT, PAGE_SIZE};
+use crate::{
+    common::{GuestAddr, PAGE_MASK, PAGE_SHIFT, PAGE_SIZE},
+    vmm::Pfn,
+};
 
 const LOCKED_BIT: u64 = 1;
 const LOCKED_MASK: u64 = 1 << LOCKED_BIT;
@@ -40,8 +43,8 @@ impl std::fmt::Debug for MsrHypercallValue {
 impl MsrHypercallValue {
     /// Yields the guest page number (the PFN) at which the guest would like the
     /// hypercall page to be placed.
-    pub fn gpfn(&self) -> u64 {
-        self.0 >> PAGE_SHIFT
+    pub fn gpfn(&self) -> Pfn {
+        Pfn::new(self.0 >> PAGE_SHIFT).unwrap()
     }
 
     /// Returns the guest physical address at which the guest would like the
--- a/lib/propolis/src/enlightenment/hyperv/hypercall.rs
+++ b/lib/propolis/src/enlightenment/hyperv/hypercall.rs
@@ -38,6 +38,12 @@ impl std::fmt::Debug for MsrHypercallValue {
 }
 
 impl MsrHypercallValue {
+    /// Yields the guest page number (the PFN) at which the guest would like the
+    /// hypercall page to be placed.
+    pub fn gpfn(&self) -> u64 {
+        self.0 >> PAGE_SHIFT
+    }
+
     /// Returns the guest physical address at which the guest would like the
     /// hypercall page to be placed.
     pub fn gpa(&self) -> GuestAddr {
--- a/lib/propolis/src/enlightenment/hyperv/mod.rs
+++ b/lib/propolis/src/enlightenment/hyperv/mod.rs
@@ -252,7 +252,24 @@ impl Lifecycle for HyperV {
 
     fn reset(&self) {
         let mut inner = self.inner.lock().unwrap();
-        *inner = Inner::default();
+
+        // Create a new overlay manager that tracks no pages. The old overlay
+        // manager needs to be dropped before any of the overlay pages that
+        // referenced it, so explicitly replace the existing manager with a new
+        // one (and drop the old one) before default-initializing the rest of
+        // the enlightenment's state.
+        let new_overlay_manager = Arc::new(OverlayManager::default());
+        let _ = std::mem::replace(
+            &mut inner.overlay_manager,
+            new_overlay_manager.clone(),
+        );
+
+        *inner = Inner {
+            overlay_manager: new_overlay_manager,
+            ..Default::default()
+        };
+
+        inner.overlay_manager.attach(&self.acc_mem);
     }
 
     fn halt(&self) {
--- a/lib/propolis/src/enlightenment/hyperv/mod.rs
+++ b/lib/propolis/src/enlightenment/hyperv/mod.rs
@@ -14,13 +14,16 @@
 //! that intends to implement a Hyper-V-compatible interface:
 //! https://github.com/MicrosoftDocs/Virtualization-Documentation/blob/main/tlfs/Requirements%20for%20Implementing%20the%20Microsoft%20Hypervisor%20Interface.pdf
 
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
 
 use cpuid_utils::{CpuidIdent, CpuidSet, CpuidValues};
+use overlay::{
+    OverlayContents, OverlayError, OverlayKind, OverlayManager, OverlayPage,
+};
 
 use crate::{
     accessors::MemAccessor,
-    common::{GuestRegion, Lifecycle, VcpuId, PAGE_SIZE},
+    common::{Lifecycle, VcpuId},
     enlightenment::{
         hyperv::{
             bits::*,
@@ -33,11 +36,11 @@ use crate::{
         PayloadOutput,
     },
     msr::{MsrId, RdmsrOutcome, WrmsrOutcome},
-    vmm::SubMapping,
 };
 
 mod bits;
 mod hypercall;
+mod overlay;
 
 #[usdt::provider(provider = "propolis")]
 mod probes {
@@ -49,13 +52,28 @@ mod probes {
 
 const TYPE_NAME: &str = "guest-hyperv-interface";
 
-#[derive(Debug, Default)]
 struct Inner {
+    /// This enlightenment's overlay manager.
+    overlay_manager: Arc<OverlayManager>,
+
     /// The last value stored in the [`bits::HV_X64_MSR_GUEST_OS_ID`] MSR.
     msr_guest_os_id_value: u64,
 
     /// The last value stored in the [`bits::HV_X64_MSR_HYPERCALL`] MSR.
     msr_hypercall_value: MsrHypercallValue,
+
+    hypercall_overlay: Option<OverlayPage>,
+}
+
+impl Default for Inner {
+    fn default() -> Self {
+        Self {
+            overlay_manager: OverlayManager::new(),
+            msr_guest_os_id_value: Default::default(),
+            msr_hypercall_value: Default::default(),
+            hypercall_overlay: None,
+        }
+    }
 }
 
 pub struct HyperV {
@@ -87,6 +105,7 @@ impl HyperV {
         // manually cleared this bit).
         if value == 0 {
             inner.msr_hypercall_value.clear_enabled();
+            inner.hypercall_overlay.take();
         }
 
         inner.msr_guest_os_id_value = value;
@@ -122,36 +141,43 @@ impl HyperV {
         // the guest.
         if !new.enabled() {
             inner.msr_hypercall_value = new;
+            inner.hypercall_overlay.take();
             return WrmsrOutcome::Handled;
         }
 
-        let memctx = self
-            .acc_mem
-            .access()
-            .expect("guest memory is always accessible during wrmsr");
-
-        let region = GuestRegion(new.gpa(), PAGE_SIZE);
-
-        // Mapping will fail if the requested GPA is out of the guest's physical
-        // address range. The TLFS specifies that this should raise #GP.
-        let Some(mapping) = memctx.writable_region(&region) else {
-            probes::hyperv_wrmsr_hypercall_bad_gpa!(|| new.gpa().0);
-            return WrmsrOutcome::GpException;
+        // Ensure the overlay is in the correct position.
+        let res = if let Some(overlay) = inner.hypercall_overlay.as_mut() {
+            overlay.move_to(new.gpfn())
+        } else {
+            match inner.overlay_manager.add_overlay(
+                new.gpfn(),
+                OverlayKind::Hypercall,
+                OverlayContents(Box::new(hypercall_page_contents())),
+            ) {
+                Ok(overlay) => {
+                    inner.hypercall_overlay = Some(overlay);
+                    Ok(())
+                }
+                Err(e) => Err(e),
+            }
         };
 
-        // Write the hypercall instruction sequence to the requested GPA.
-        //
-        // TODO: TLFS section 5.2.1 specifies that when an overlay is removed,
-        // "the underlying GPA page is 'uncovered', and an existing mapping
-        // becomes accessible to the guest." Empirically, at least some other
-        // Hv#1 implementations don't appear to follow this rule, and most
-        // common guest OSes don't rely on being able to disable or remove the
-        // hypercall page. Nevertheless, Propolis should eventually follow this
-        // rule.
-        write_overlay_page(&mapping, &hypercall_page_contents());
-
-        inner.msr_hypercall_value = new;
-        WrmsrOutcome::Handled
+        match res {
+            Ok(()) => {
+                inner.msr_hypercall_value = new;
+                WrmsrOutcome::Handled
+            }
+            Err(OverlayError::AddressInaccessible(_)) => {
+                WrmsrOutcome::GpException
+            }
+            // There should only ever be one hypercall overlay at a time, and
+            // guest memory should be accessible in the context of a VM exit, so
+            // (barring some other invariant being violated) adding an overlay
+            // should always succeed if the target PFN is valid.
+            Err(e) => {
+                panic!("unexpected error establishing hypercall overlay: {e}")
+            }
+        }
     }
 }
 
@@ -225,15 +251,9 @@ impl super::Enlightenment for HyperV {
 
     fn attach(&self, mem_acc: &MemAccessor) {
         mem_acc.adopt(&self.acc_mem, Some(TYPE_NAME.to_owned()));
+        let inner = self.inner.lock().unwrap();
+        inner.overlay_manager.attach(&self.acc_mem);
     }
-}
-
-fn write_overlay_page(mapping: &SubMapping<'_>, contents: &[u8; PAGE_SIZE]) {
-    let written = mapping
-        .write_bytes(contents)
-        .expect("overlay pages are always writable");
-
-    assert_eq!(written, PAGE_SIZE, "overlay pages can be written completely");
 }
 
 impl Lifecycle for HyperV {
@@ -244,6 +264,20 @@ impl Lifecycle for HyperV {
     fn reset(&self) {
         let mut inner = self.inner.lock().unwrap();
         *inner = Inner::default();
+    }
+
+    fn halt(&self) {
+        let mut inner = self.inner.lock().unwrap();
+
+        // Create a new overlay manager and drop the reference to the old one.
+        // This should be the only active reference to this manager, since all
+        // overlay page operations happen during VM exits, and the vCPUs have
+        // all quiesced by this point.
+        //
+        // This ensures that when this object is dropped, any overlay pages it
+        // owns can be dropped safely.
+        assert_eq!(Arc::strong_count(&inner.overlay_manager), 1);
+        inner.overlay_manager = OverlayManager::new();
     }
 
     fn migrate(&'_ self) -> Migrator<'_> {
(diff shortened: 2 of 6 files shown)


lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -0,0 +1,754 @@
...
+}
+
+impl Drop for OverlayPage {
+    fn drop(&mut self) {
+        // This overlay is already registered and has a valid PFN, so the only
+        // reason the manager could fail to remove it is if it failed to access
+        // guest memory entirely. The parent enlightenment stack prevents this
+        // by holding a strong reference to the manager until the VM is halted.
+        //
+        // Once the VM *is* halted, no one will access guest memory again,
```
gjcolombo, line 151:
```suggestion
        // Once the VM *is* halted, no one will access guest memory again, so
```

oops, typo

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -0,0 +1,754 @@
...
+//! module, which are described further below.
+//!
+//! TLFS section 5.2.1 describes the semantics of overlay pages in more detail.
+//! Importantly, the TLFS specifies that when multiple overlays exist for a
+//! single GPA, their logical ordering is implementation-defined (and not, for
+//! example, based on the order in which the overlays were requested).
+//!
+//! The TLFS doesn't explicitly specify the sizes of its overlay pages, but all
+//! the overlays this module cares about are 4 KiB, because the guest specifies
+//! their locations using 52-bit PFNs.
```
hawkw, line 41: I think it could be useful if this comment also included a high-level summary of our implementation of this, as well as discussing the HyperV concepts. In particular, there are a number of comments about the relationship between the lifecycles of the `OverlayManager` and the `OverlayPage`s it creates, but it would be nice if there was an introductory "theory of operation" type comment. I think it would be a bit easier to wrap ones head around those inline comments if there was an overview.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -0,0 +1,754 @@
...
+struct OverlaySet {
+    /// The contents of guest memory at this set's PFN at the time an overlay
+    /// first became active here.
+    original_contents: OverlayContents,
+
+    /// The kind of overlay that is currently active for this set's PFN.
+    active: OverlayKind,
+
+    /// The overlays that are waiting to be applied to this PFN.
+    pending: BTreeMap<OverlayKind, OverlayContents>,
```
hawkw: It occurs to me that we may not *really* need a whole `BTreeMap` for this, depending on the number of `OverlayKind`s we expect to have and whether any of them will ever contain data. We _could_ just add a `#[repr(u8)]` attribute to `OverlayKind`, and represent this as an `[Option<OverlayContents>; NUM_OVERLAY_KIND_VARIANTS]`. That would require replacing the `Test(u8)` variant with `Test1`, `Test2`, `Test...` or similar. There's even an [`enum_map` crate](https://docs.rs/enum-map/latest/enum_map/struct.EnumMap.html) with macros for generating this kind of thing, though I've not used it personally.

I dunno if this is worth the effort. If the number of overlay pages that will be pending at any time is small (and I suspect it will be), the overhead of the `BTreeMap` is probably not significant, and I don't think adding overlay pages is a hot operation W.R.T. guest performance? But, it was just something that occurred to me, so I figured I'd mention it.

lib/propolis/src/enlightenment/hyperv/overlay.rs
hawkw: ...oh, silly me --- i later discovered that `OverlayManager` has a comment explaining this. personally, i might consider moving the struct definition up in the file so that it's encountered _before_ the code on `ManagerInner`  and `OverlayPage` that references this stuff...

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -0,0 +1,754 @@
...
+            if set.get().pending.is_empty() {
+                mapped.write_page(&set.get().original_contents);
+                set.remove_entry();
+            } else {
+                let set = set.get_mut();
+
+                // TLFS section 5.2.1 specifies that when there are multiple
+                // overlays for a given page, the order in which they are
+                // applied is implementation-defined, so any method of choosing
+                // a new overlay from the pending set is sufficient.
```
hawkw: oh, _phew_

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -0,0 +1,754 @@
...
+        // tables yet, since applying the new overlay might still fail if there
+        // is an existing overlay of this kind at the target PFN.
+        //
+        // It's possible to do this check now, modify the tables in-place, and
+        // then assert that adding the new overlay succeeded, but this sequence
+        // allows this function to reuse the existing overlay removal code. This
+        // is unlikely to be a bottleneck in practice because guests change
+        // their overlay locations so infrequently (and any guest that spends
+        // all its time moving overlays around has, in some ways, asked for
+        // it...).
```
hawkw: Reasonable!

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -0,0 +1,754 @@
...
+/// will attempt to remove themselves when they are dropped and assert that they
+/// can do so successfully. This will trigger a panic during VM shutdown if an
+/// overlay page with an active manager is dropped after the VM's memory context
+/// has been released. To avoid this, a Hyper-V stack that owns an overlay
+/// manager can drop its reference to the manager in its [`Lifecycle::halt`]
+/// callout; at that point no vCPUs are active or will become active, so this
+/// should remove the last strong reference to the manager, after which it is
+/// safe to drop the pages it produced.
+///
+/// [`Lifecycle::halt`]: crate::lifecycle::Lifecycle::halt
```
hawkw, line 386: This is the comment that I mentioned above that I would have really liked to encounter earlier in the file. Not a huge deal, though.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -0,0 +1,754 @@
...
+        // it...).
+        let contents = if from_set.active == kind {
+            let mut from_mapping = from_pfn
+                .map(&memctx)
+                .expect("active overlay PFNs can be mapped");
+            let mut contents = OverlayContents::default();
+            from_mapping.read_page(&mut contents);
+            contents
+        } else if let Some(pending_contents) = from_set.pending.get(&kind) {
+            pending_contents.clone()
```
hawkw: Hm, cloning a `Box<[u8; PAGE_SIZE]>` memcpys the whole page, which seems like a bit of a bummer since we are removing the overlay from the other `OverlaySet` anyway. It seems like we ought to be able to change this to `from_set.pending.remove(&kind)` to take ownership of the page contents and just reuse the existing pointer when we call `add_overlay_using_mapping`. The one note is that the call to `remove_overlay` would have to be conditional on the moved overlay being the active overlay in `from_set`, but i _think_ we could just move the `remove_overlay` call into the `if` block above after we call `read_page`?

Again, this may not be worth the effort in light of your observation a couple lines ago about how a guest is generally not going to do a ton of overlay page churning...it just bummed me out a little.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Fixed in 1bcf2eb

lib/propolis/src/enlightenment/hyperv/overlay.rs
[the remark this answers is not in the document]
gjcolombo: Gave this a try in 1bcf2eb.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -0,0 +1,754 @@
...
+    use crate::{
+        common::{GuestAddr, GuestRegion, PAGE_MASK, PAGE_SHIFT, PAGE_SIZE},
+        vmm::SubMapping,
+    };
+
+    use super::{OverlayContents, OverlayError};
+
+    /// A 52-bit physical page number.
+    #[derive(Clone, Copy, Ord, PartialOrd, Eq, PartialEq)]
+    pub(super) struct Pfn(u64);
```
hawkw, line 463: somewhat surprised we didn't have this already! might it be worth defining the `Pfn` type in the `mem` module so it could more easily be used in other code?

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -0,0 +1,754 @@
...
+
+    impl std::fmt::Debug for Pfn {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(f, "Pfn({:x})", self.0)
+        }
+    }
+
+    impl std::fmt::Display for Pfn {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(f, "{:x}", self.0)
```
hawkw, line 473: nitpicky: if we instead wrote
```suggestion
            fmt::LowerHex::fmt(self.0, f)
```
we would just call into the `fmt::LowerHex` implementation directly, and it would honor any additional format specifiers (e.g. `#`,  leading zeros, alignment, etc) that this was called with. 

the current implementation eats that, so if i wrote something like:
```rust
write!("{:#016}", Pfn(42));
```
and expected to get
```
0x0000000000002a
```
i would actually get
```
2a
```
instead. it _might_ be worth handling that nicer?

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -308,48 +330,36 @@ impl ManagerInner {
...
-        };
+        // If the overlay exists in its old location, then it can definitely be
+        // moved to its new one, so it's now safe to try removing the old
+        // overlay.
+        let mut contents = Some(MaybeUninit::uninit());
+        self.remove_overlay(from_pfn, kind, acc_mem, contents.as_mut())?;
+
+        // Safety: `remove_overlay` guarantees that the output contents will be
+        // initialized on success.
+        let contents = unsafe { contents.unwrap().assume_init() };
```
hawkw, line 359: Hm, I don't think I see why all this `MaybeUninit` business is necessary. Naively, it _looks_ to me like `remove_overlay` should be able to just return an `OverlayContents` which we could use here?

hawkw: Ah, I see, the intent is that whether or not the `Option<&mut MaybeUninit>` is passed in dictates whether or not we return the previous contents, so that we can avoid reading the old overlay page contents if we're just removing it rather than moving it. Hm. That does make sense although I have to admit I'm not terribly enamored with it...

lib/propolis/src/enlightenment/hyperv/overlay.rs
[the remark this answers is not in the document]
gjcolombo: ...and reverted in bede51e.

There are two important invariants that the OverlaySet has to uphold for any PFN with an overlay:

- there is exactly one active overlay for that PFN
- an overlay of a given kind can only be registered once per PFN

The first invariant is more important than the second: having an overlay set with multiple overlays of the same kind is at least comprehensible (though maybe indicative of a programming error); having an overlay set with a pending overlay but no active overlay is a serious issue (why isn't the pending overlay active now?). Switching to a single map makes the second invariant easier to uphold at the cost of making it harder to enforce the first one.

Having a single map to contend with does ever-so-slightly simplify the logic needed to export/import overlay manager state, but on reflection I don't think that benefit outweighs the drawbacks.

The other part of 1bcf2eb (which got rid of some unnecessary cloning) is reimplmeneted in f09684e.

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -49,13 +52,28 @@ mod probes {
...
+
+impl Default for Inner {
+    fn default() -> Self {
+        Self {
+            overlay_manager: OverlayManager::new(),
+            msr_guest_os_id_value: Default::default(),
+            msr_hypercall_value: Default::default(),
+            hypercall_overlay: None,
+        }
+    }
```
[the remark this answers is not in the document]
gjcolombo: Fixed in 98647e9.

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -122,36 +141,43 @@ impl HyperV {
...
+                new.gpfn(),
+                OverlayKind::Hypercall,
+                OverlayContents(Box::new(hypercall_page_contents())),
+            ) {
+                Ok(overlay) => {
+                    inner.hypercall_overlay = Some(overlay);
+                    Ok(())
+                }
+                Err(e) => Err(e),
+            }
```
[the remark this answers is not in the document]
gjcolombo: Done in ebf83bd.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Done in 094bd27.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: I still hope the guest has better things to do than spam its overlay MSRs, but this comment is gone with f09684e, along with cloning pending overlay contents :)

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: It's a huge relief!

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: I doubt we're going to have that many overlay kinds for a while--certainly no more than two to start out with.

I also expect the `pending` map to be empty for virtually all overlaid GPAs, because I suspect most guests aren't going to assign multiple overlays to a single GPA. After all, if they do that, they'll only see one of the pages they requested! This still needed for correctness, of course, because pages *are* allowed to alias if the guest asks, but in practice I think we'll find that these maps are basically always empty.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Fixed in 094bd27.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Fixed in f09684e.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Great suggestion, thanks! Done in c8ccb8b.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: I reimplemented this in f09684e with slightly different semantics (caller passes a flag, routine returns an `Option`). Turned out a bit nicer, I think--now the caller will panic if the callee forgot to return contents instead of invoking UB, and it's not really any less economical from a lines-of-code perspective than the MaybeUninit version.

gjcolombo:
Thanks for the review, @hawkw! I think this is ready for another lap at your convenience. I'll work on rebasing #853 onto this.

lib/propolis/src/enlightenment/hyperv/overlay.rs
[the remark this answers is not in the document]
hawkw: Yeah, upholding the invariant that only one overlay is active should be a priority. I agree with your conclusion that not being able to represent multiple active overlays is more important than making the import/export code a bit simpler. Good call!

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -308,48 +339,36 @@ impl ManagerInner {
...
-            pending_contents.clone()
-        } else {
-            return Err(OverlayError::NoOverlayOfKind(from_pfn.into(), kind));
-        };
+        // Removing the old overlay might fail if it doesn't exist in
+        // `from_pfn`, but if it succeeds it will always produce the previous
+        // page contents.
+        let contents = self
+            .remove_overlay(from_pfn, kind, acc_mem, RemoveReturnContents::Yes)?
+            .expect("asked for old contents");
```
hawkw, line 427: having to `expect` this is a bit of a bummer, but overall, it still seems less grody than the previous thing with `MaybeUninit`...

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -39,6 +41,54 @@
...
+//!   its manager reference.
+//! - When a manager's owner receives a [`Lifecycle::halt`] callout, it releases
+//!   all its strong references to its `OverlayManager`. (This is generally
+//!   possible because a halting VM's vCPUs are permanently quiesced and so
+//!   can't write any more data that would create or move another overlay.)
+//!
+//! This pattern allows a `HyperV` instance to "run down" its manager by
+//! simply dropping all references to it. After that, it can drop its
+//! `OverlayPage`s at any time without fear of them trying and failing to access
+//! guest memory.
```
hawkw, line 89: this is lovely, thanks for adding it!

lib/propolis/src/vmm/mem.rs
```diff
@@ -1063,6 +1065,61 @@ impl<T: Copy + FromBytes> Iterator for GuestData<MemMany<'_, T>> {
...
+        value.0
+    }
+}
+
+impl Pfn {
+    /// Creates a new PFN wrapper for the supplied physical page number. Returns
+    /// `None` if the page number cannot correctly be shifted to form the
+    /// corresponding 64-bit address.
+    pub(crate) fn new(pfn: u64) -> Option<Self> {
+        if pfn > (PAGE_MASK as u64 >> PAGE_SHIFT) {
```
hawkw, line 1096: nitpicky, feel free to ignore me: personally I like to define max values for things like this as a public associated constant in case it needs to be referenced in other code --- though this is more important in a public library API, and since this is crate private, we could always just go and add a constant for `PAGE_MASK as u64 >> PAGE_SHIFT` later if we need it. so...maybe that doesn't matter.

lib/propolis/src/vmm/mem.rs
```diff
@@ -1059,6 +1065,61 @@ impl<T: Copy + FromBytes> Iterator for GuestData<MemMany<'_, T>> {
...
 }
 
+/// A 52-bit physical page number, i.e., the ordinal index of a 4 KiB page of
+/// guest memory.
+#[derive(Clone, Copy, Ord, PartialOrd, Eq, PartialEq)]
+pub(crate) struct Pfn(u64);
+
+impl std::fmt::Debug for Pfn {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Pfn({:x})", self.0)
```
hawkw, line 1075: nitpicky: perhaps we should stick a 
```suggestion
        write!(f, "Pfn({:#x})", self.0)
```
in here to get a leading `0x` to make it clearer that it's hex?

lib/propolis/src/enlightenment/hyperv/hypercall.rs
```diff
@@ -39,6 +39,12 @@ impl std::fmt::Debug for MsrHypercallValue {
 }
 
 impl MsrHypercallValue {
+    /// Yields the guest page number (the PFN) at which the guest would like the
+    /// hypercall page to be placed.
+    pub fn gpfn(&self) -> u64 {
+        self.0 >> PAGE_SHIFT
+    }
```
hawkw, line 46: nit: now that the `Pfn` newtype is crate-public, perhaps this should be:
```suggestion
    /// Yields the guest page number (the PFN) at which the guest would like the
    /// hypercall page to be placed.
    pub fn gpfn(&self) -> Pfn {
        Pfn::new(self.0 >> PAGE_SHIFT)
    }
```

hawkw: Also, perhaps the `Debug` impl above that currently includes the GPA for `MsrHypercallValue` ought to also include the PFN?

hawkw (approved):
The new changes look good to me! Perhaps worth getting a second opinion?

gjcolombo:
I found a bug during ad hoc testing: `HyperV::reset` creates a new overlay manager, but doesn't attach its (new) memory accessor to the accessor hierarchy. It needs to do that (and `reset` needs to be careful to drop things in the correct order; it does so today, but only by the grace of the drop ordering rules for structs...).

I'm going to merge #849 and rebase this branch before pushing a fix and working on other PR feedback.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Yeah, I don't think there's really a better way out of this one, but at least there's no `unsafe`.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Thanks!

lib/propolis/src/vmm/mem.rs
gjcolombo: Fixed in 6e61e6e.

lib/propolis/src/vmm/mem.rs
gjcolombo: I've left this as-is. As you mention, this is internal-only, and I suspect this is a case where callers aren't especially likely to care about the maximum possible 4K PFN (they're more likely to care about, say, whether a PFN is within the bounds of the guest's physical memory, but that depends on how much guest memory there is).

lib/propolis/src/enlightenment/hyperv/hypercall.rs
gjcolombo: I switched to using the wrapper in more places in d378e33. I left the `Debug` impl as-is, though; anyone who wants to get a 4K PFN from a logged GPA can just put a thumb over the low three nibbles :)

Issue #811, opened 2024-11-15
server: failed migrations are reported as having completed before the source VM is fully resumed


gjcolombo:
Originally seen in #809; see excerpt below.

---

The PHD failure here is weird. The failure is in the `running_process::import_failure` test, which is what I'd expect this change to break if it broke anything. But the failure mode is odd: the first migration fails as expected, but then the source VM immediately gets a *stop request*, which it immediately handles, rendering it unable to serve as a migration source for the second migration:

```
{"msg":"migration out failed, resuming","v":0,"name":"propolis-server","level":30,"time":"2024-11-13T23:28:13.434800658Z","hostname":"buskin","pid":1013,"component":"vm_state_driver","error":"RemoteError(Destination, \"failed to migrate device state: failed to apply deserialized device state: you have no chance to survive, make your time\")"}
{"msg":"publishing new instance state","v":0,"name":"propolis-server","level":30,"time":"2024-11-13T23:28:13.434818339Z","hostname":"buskin","pid":1013,"component":"vm_state_driver","migration":"InstanceMigrateStatusResponse { migration_in: None, migration_out: Some(InstanceMigrationStatus { id: d5a22278-1907-4188-beb1-dbaab29b33a3, state: Error }) }","state":"Running","gen":12}
{"msg":"state driver handled event","v":0,"name":"propolis-server","level":30,"time":"2024-11-13T23:28:13.434839439Z","hostname":"buskin","pid":1013,"component":"vm_state_driver","outcome":"Continue"}
{"msg":"request completed","v":0,"name":"propolis-server","level":30,"time":"2024-11-13T23:28:13.459488929Z","hostname":"buskin","pid":1013,"uri":"/instance","method":"GET","req_id":"2b877526-cd31-4935-81e0-ab86c6740d17","remote_addr":"127.0.0.1:37727","local_addr":"127.0.0.1:9000","latency_us":173,"response_code":"200"}
{"msg":"requested state via API","v":0,"name":"propolis-server","level":30,"time":"2024-11-13T23:28:13.460691072Z","hostname":"buskin","pid":1013,"component":"vm_state_driver","state":"Stop"}
{"msg":"Queuing external request","v":0,"name":"propolis-server","level":30,"time":"2024-11-13T23:28:13.460710913Z","hostname":"buskin","pid":1013,"component":"request_queue","component":"vm_state_driver","disposition":"Enqueue","request":"Stop"}
[...]

PR #853, opened 2025-02-10
lib: migrate overlay manager state explicitly

### LINKED ISSUE #850: lib: want better Hyper-V overlay page management

gjcolombo:
(N.B. Stacked on #851.)

Explicitly migrate the state of a Hyper-V stack's `OverlayManager` instead of having the stack reapply its own overlays. The manager's `OverlaySet`s are exported and imported more or less as-is, though with some transformations between types for serialization purposes (e.g. `Box<u8; PAGE_SIZE>` is serialized as `Vec<u8>`) or for compatibility reasons (`enum MigratedOverlayKind` is distinct from `enum OverlayKind` so that the latter can change without rendering the manager unable to recognize overlay kinds from old Propolis versions).

Importing manager state produces a set of `OverlayPage` registrations that the higher-level Hyper-V import logic can audit for correctness (there should be exactly as many entries as MSRs with active overlays, and they should be associated with the correct PFNs).

Tests: cargo test and PHD; examined the migration logs from an ad hoc migration and verified that the expected overlay state was migrated.

Fixes #850.

### DIFF
--- a/lib/propolis/src/enlightenment/hyperv/overlay.rs
+++ b/lib/propolis/src/enlightenment/hyperv/overlay.rs
@@ -570,8 +570,10 @@ impl ManagerInner {
             .collect::<Result<BTreeMap<Pfn, OverlaySet>, OverlayError>>()?;
 
         // Now that the tracking tables have been reconstructed, create a set of
-        // overlay page descriptors to return to the caller. The caller will
-        // audit these to make sure they align with other enlightenment state.
+        // overlay page descriptors to return to the caller.
+        //
+        // The caller can check the kinds and PFNs of these pages to make sure
+        // that they line up with the enlightenment's imported MSR values.
         let pages: Vec<OverlayPage> = self
             .overlays
             .iter()
--- a/lib/propolis/src/enlightenment/hyperv/overlay.rs
+++ b/lib/propolis/src/enlightenment/hyperv/overlay.rs
@@ -97,7 +97,10 @@ use std::{
 
 use thiserror::Error;
 
-use crate::{accessors::MemAccessor, common::PAGE_SIZE, vmm::Pfn};
+use crate::{
+    accessors::MemAccessor, common::PAGE_SIZE, migrate::MigrateStateError,
+    vmm::Pfn,
+};
 
 use self::pfn::MappedPfn;
 
@@ -130,6 +133,27 @@ pub enum OverlayError {
     /// caller.
     #[error("PFN {0:#x} has no overlay of kind {1:?}")]
     NoOverlayOfKind(u64, OverlayKind),
+
+    #[error("imported PFN {0:#x} is too large to be a 4K page number")]
+    ImportedPfnTooLarge(u64),
+
+    /// An overlay manager import payload contained overlay page contents
+    /// that are not page-sized.
+    #[error("imported contents for pfn {pfn:#x} have invalid length {len}")]
+    ImportedContentsNotPageSized { pfn: u64, len: usize },
+
+    /// An overlay manager import payload contained an overlay set where a
+    /// single kind of overlay was both active and pending.
+    #[error("imported overlay set for pfn {0:#x} has duplicate kind {1:?}")]
+    ImportedSetDuplicateKind(u64, OverlayKind),
+}
+
+impl From<OverlayError> for MigrateStateError {
+    fn from(value: OverlayError) -> Self {
+        MigrateStateError::ImportFailed(format!(
+            "error importing overlay state: {value}"
+        ))
+    }
 }
 
 /// The contents of a 4 KiB page. These are boxed so that this type can be
@@ -143,6 +167,15 @@ impl Default for OverlayContents {
     }
 }
 
+impl TryFrom<Vec<u8>> for OverlayContents {
+    type Error = usize;
+
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        let len = value.len();
+        Ok(Self(Box::new(value.try_into().map_err(|_| len)?)))
+    }
+}
+
 impl std::fmt::Debug for OverlayContents {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_tuple("OverlayContents").field(&"<page redacted>").finish()
@@ -151,7 +184,7 @@ impl std::fmt::Debug for OverlayContents {
 
 /// A kind of overlay page.
 #[derive(Clone, Copy, Debug, Ord, PartialOrd, Eq, PartialEq)]
-pub(super) enum OverlayKind {
+pub enum OverlayKind {
     Hypercall,
 
     #[cfg(test)]
@@ -160,7 +193,6 @@ pub(super) enum OverlayKind {
 
 /// A registered overlay page, held by a user of this module. The overlay is
 /// removed when this page is dropped.
-#[derive(Debug)]
 pub(super) struct OverlayPage {
     /// A back reference to this page's manager. This is `Weak` so that pages
     /// can detect when the manager's Hyper-V stack has halted and can thereby
@@ -175,7 +207,24 @@ pub(super) struct OverlayPage {
     pfn: Pfn,
 }
 
+impl std::fmt::Debug for OverlayPage {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("OverlayPage")
+            .field("kind", &self.kind)
+            .field("pfn", &self.pfn)
+            .finish()
+    }
+}
+
 impl OverlayPage {
+    pub(super) fn kind(&self) -> OverlayKind {
+        self.kind
+    }
+
+    pub(super) fn pfn(&self) -> Pfn {
+        self.pfn
+    }
+
     /// Moves this overlay to a new PFN.
     ///
     /// # Panics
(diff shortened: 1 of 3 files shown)


lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -175,7 +207,24 @@ pub(super) struct OverlayPage {
     pfn: Pfn,
 }
 
+impl std::fmt::Debug for OverlayPage {
```
hawkw, line 210: nit, TIOLI: maybe worth a comment stating that we don't derive this because we don't want to include the `OverlayManager`?

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -175,7 +207,24 @@ pub(super) struct OverlayPage {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("OverlayPage")
+            .field("kind", &self.kind)
+            .field("pfn", &self.pfn)
+            .finish()
```
hawkw, line 215: take it or leave it: a technique I like for hand-implemented `Debug` impls is to exhaustively destructure `self`, explicitly skipping any fields we don't want to format, like this:
```suggestion
        let Self { kind, pfn, manager: _ } = self;
        f.debug_struct("OverlayPage")
            .field("kind", kind)
            .field("pfn", pfn)
            .finish()
```
This way, if a new field is added, the exhaustive destructuring will fail to compile, serving as a nice little reminder to either add the new field to the `Debug` output or explicitly decide to ignore it.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -508,6 +677,93 @@ impl OverlayManager {
...
+        pub(super) pending: BTreeMap<MigratedOverlayKind, Vec<u8>>,
+    }
+
+    impl std::fmt::Debug for OverlaySetV1 {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            f.debug_struct("OverlaySetV1")
+                .field("original_contents", &"<page redacted>")
+                .field("active", &self.active)
+                .field("pending", &self.pending.keys().collect::<Vec<_>>())
+                .finish()
```
hawkw, line 753: again, this is where I might like to write
```suggestion
            let Self { original_contents: _, active, pending } = self;
            f.debug_struct("OverlaySetV1")
                .field("original_contents", &"<page redacted>")
                .field("active", active)
                .field("pending", pending.keys().collect::<Vec<_>>())
                .finish()
```

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -508,6 +677,93 @@ impl OverlayManager {
...
+        pub(super) active: MigratedOverlayKind,
+        pub(super) pending: BTreeMap<MigratedOverlayKind, Vec<u8>>,
+    }
+
+    impl std::fmt::Debug for OverlaySetV1 {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            f.debug_struct("OverlaySetV1")
+                .field("original_contents", &"<page redacted>")
+                .field("active", &self.active)
+                .field("pending", &self.pending.keys().collect::<Vec<_>>())
```
hawkw, line 752: nitpicky: surprise allocations in a `Debug` impl are generally something i try to avoid...

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -427,6 +476,126 @@ impl ManagerInner {
...
+                        active: set.active.into(),
+                        pending,
+                    },
+                ))
+            })
+            .collect::<Result<BTreeMap<Pfn, OverlaySet>, OverlayError>>()?;
+
+        // Now that the tracking tables have been reconstructed, create a set of
+        // overlay page descriptors to return to the caller. The caller will
+        // audit these to make sure they align with other enlightenment state.
```
hawkw, line 572: Wait, sorry, what exactly are you referring to here with "the caller will audit these..."? Is that just the check that the hypercall page's PFN matches the hypercall MSR's PFN?

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -427,6 +476,126 @@ impl ManagerInner {
...
+                    kind: *kind,
+                    pfn: *pfn,
+                }));
+
+                entries
+            })
+            .reduce(|mut acc, mut e| {
+                acc.append(&mut e);
+                acc
+            })
```
hawkw, line 594: nit, take it or leave it:: I think the `map`/`reduce` could be simplified to `flat_map`?
```suggestion
            .flat_map(|(pfn, set)| {
               std::iter::once(OverlayPage {
                    manager: Arc::downgrade(manager),
                    kind: set.active,
                    pfn: *pfn,
                })
                    .chain(set.pending.keys().map(|kind| OverlayPage {
                        manager: Arc::downgrade(manager),
                        kind: *kind,
                        pfn: *pfn,
                    }))
            })
```

gjcolombo: Good call--fixed (and added a unit test) in ea80080.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: More or less (there's also a check for whether the Hyper-V stack and the overlay manager even agree that an overlay should exist). I clarified this comment a bit in 4f99c85.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Somehow I overlooked that `btree_map::Keys` impls `Debug`... Fixed in 2d23e4e.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Good call, thanks! I'm also generally very enthusiastic about using exhaustive destructuring for this sort of thing, but it didn't occur to me to do it here. Fixed in 2d23e4e.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Fixed in 2d23e4e.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Fixed in 2d23e4e.

gjcolombo:
The first PHD run on the most recent commit hit known test flake #811 (really need to go fix that...); I've asked for a re-run.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -1063,4 +1063,46 @@ mod test {
 
         ctx.assert_pfn_has_fill(pfn, 0x70);
     }
+
+    #[test]
+    fn export_import_round_trip() {
```
hawkw, line 1068: take it or leave it: this type of "round trip" test always feels like kind of a good candidate for a property test where we generate a bunch of random things and assert they look the same on both sides. but, i'm not sure if it's really going to gain that much from it, and writing the proptest code to generate arbitrary migration payloads might be too much work...just a thought!

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -427,6 +478,123 @@ impl ManagerInner {
...
+                })
+                .chain(set.pending.keys().map(|kind| {
+                    OverlayPage {
+                        manager: Arc::downgrade(manager),
+                        kind: *kind,
+                        pfn: *pfn,
+                    }
+                }))
+            })
+            .collect();
```
hawkw, line 594: glad this worked :)

hawkw (approved):
looks great! i might wait for @pfmooney's ack before merging?

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -508,6 +676,94 @@ impl OverlayManager {
...
+            }
+        }
+    }
+
+    #[derive(Serialize, Deserialize)]
+    pub struct OverlaySetV1 {
+        pub(super) original_contents: Vec<u8>,
+        pub(super) active: MigratedOverlayKind,
+        pub(super) pending: BTreeMap<MigratedOverlayKind, Vec<u8>>,
+    }
```
pfmooney, line 744: From a high level, I'm not sure I'm wild about migrating the overlay state like this.  We're essentially duplicating some state, vs what we have stored for the `msr_hypercall` data in `HyperVEnlightenmentV1` (that a hypercall page is overlaying a given address).  There is still the matter of the covered page data, but that could go along with the `msr_hypercall` bits, in theory (in terms of which part of the migration payload it should be attached to).

That way, it does less in the way of encoding the OverlayManager as part of the migration payload, rather than it being an implementation detail at the time.  (Partially thinking about if/when the kernel vmm makes overlay pages less of a hassle.)

Now, this isn't to say that I'm fundamentally opposed to how things are structured here.  If you think this is the right direction for all this, I'll defer.  I'm happy to chat about trade-offs as well.

gjcolombo:
#861 supersedes this. I'll close this PR once that merges and #856 is rebased.

gjcolombo:
856 is rebased, so closing this.

PR #855, opened 2025-02-11
Test oximeter metrics end to end

iximeow:
Other tests in Propolis validate that the Oximeter translation of kstats for a VM are coherent but do so from stubbed kstat readings. Those tests are fine (and quick!), but don't cover the risk of a translation error in Propolis reading from a real system, nor the possibility of reading kstats incorrectly in the first place.

... good news: test doesn't catch anything interesting! The bug I saw is somewhere else.

.... i also spent an embarrassingly long time trying to figure out how i'd stand up an Oximeter producer in the test before realizing i really needed to just plumb the flag for `propolis-server` to start _its_ Oximeter producer.


phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,488 @@
...
+            .fold(0, |total, vcpu_usage| total + vcpu_usage.metrics[state])
+    }
+
+    fn update_metric_times(&mut self, metric_time: DateTime<Utc>) {
+        self.oldest_time = std::cmp::min(self.oldest_time, metric_time);
+    }
+
+    /// Integrate a list of samples into this collection of virtual machine
+    /// metrics.
+    fn add_producer_result(&mut self, samples: &[Sample]) {
```
iximeow, line 114: this felt kind of awful to write! i hope it's less awful to read! there's a lot of enum type checking and crawling through the Oximeter types to extract the relevant details for this test.

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,495 @@
...
+
+    let full_emul_delta = (now_metrics.vcpu_state_total(&VcpuState::Emulation)
+        - start_metrics.vcpu_state_total(&VcpuState::Emulation))
+        as u128;
+
+    // Pick 100ms as a comically high upper bound for how much time might have
+    // been spent emulating instructions on the guest's behalf. Anecdotally the
+    // this is on the order of 8ms between the two samples. This should be very
+    // low; the workload is almost entirely guest user mode execution.
+    assert!(full_emul_delta < Duration::from_millis(100).as_nanos());
```
iximeow, line 482: this is basically the whole motivator for the test. we'd seen a chart where `emul` seemed to track closely but not perfectly with `run`, and now we can be a bit more confident that the system is producing coherent numbers at least, and what we saw there was an issue somewhere further up.

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,495 @@
...
+/// of samples is conceptually distinct but may still be interesting to
+/// tset. In `propolis-server`, the first list of samples will be
+/// `virtual_machine:vcpu_usage`, which may be blank of kstats have not been
+/// sampled since the last producer poll. The second list of samples
+/// will be `virtual_machine:reset`.
+///
+/// `VirtualMachineMetrics` collects these all back together into a single view
+/// to test against. See [`VirtualMachineMetrics::add_producer_result`] as the
+/// means to accumulate samples into this struct.
+struct VirtualMachineMetrics {
```
iximeow, line 95: i'm right on the fence for this being buried away in `stats.rs` or more formally part of PHD. probably depends on how much more we expect propolis-sever to provide?

phd-tests/tests/src/stats.rs
bnaecker: Forgive the naive question, by why would one expect `emul` to perfectly track `run`?

iximeow: ah, i wouldn't expect `emul` to perfectly track `run` either. if they were exactly the same, i would have just first looked for something to be mislabeled (anywhere between illumos and the console)

close-but-not-perfectly tells me that it's not just mislabeled data, which could be illumos calculating stats in a strange way, or something weird about the sampling, or something later at query time. so, test helps rule out the first two possibilities..

my actual expectation is that `emul` might count up a little more often when `run` is counting up, but as a side effect of having exit for instruction emulation. they ought to be orders of magnitude different relatively independent.

phd-tests/framework/src/test_vm/environment.rs
```diff
@@ -71,6 +84,7 @@ impl<'a> Environment<'a> {
                     .port_allocator
                     .next()
                     .context("getting Propolis server port")?;
+                let metrics_addr = builder.metrics_addr.clone();
```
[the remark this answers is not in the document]
iximeow: yeah, i'd `cargo xtask clippy`'d locally but forgot `--strict`, so the warning got lost in the build output. ugh

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,488 @@
...
+    async fn wait_for_producer(&self) {
+        loop {
+            {
+                let sampler = self.sampler.lock().unwrap();
+                if sampler.is_some() {
+                    return;
+                }
+            }
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        }
```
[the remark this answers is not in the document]
iximeow: initially i'd actually written it like this to do some "hey i'm still alive just waiting for the sampler" logging when i was getting the metrics address through phd to `propolis-server`, but yeah it looks pretty funky as-is.

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,488 @@
...
+                    self.reset.is_none(),
+                    "multiple virtual_machine:reset measurements for a \
+                     single Propolis?"
+                );
+
+                let datum = last_sample.measurement.datum();
+                let amount = if let Datum::CumulativeU64(amount) = datum {
+                    amount.value()
+                } else {
+                    panic!("unexpected reset value type");
```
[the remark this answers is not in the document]
iximeow: yup, seems prudent (https://github.com/oxidecomputer/propolis/pull/855/commits/53196223f371697652b9c2c2f1d83dbc223149b9)

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,488 @@
...
+                }
+            }
+
+            tokio::time::sleep(retry_delay).await;
+        }
+
+        panic!(
+            "propolis-server Oximeter stats unavailable? waited {:?}",
+            max_wait
+        );
```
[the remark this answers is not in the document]
iximeow: `backoff` felt a bit heavy-handed since i really just wanted "do this once a second for a bit" but it turned out less annoying than i feared (and so is done in https://github.com/oxidecomputer/propolis/pull/855/commits/53196223f371697652b9c2c2f1d83dbc223149b9)

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,488 @@
...
+        match name {
+            "emulation" => VcpuState::Emulation,
+            "run" => VcpuState::Run,
+            "idle" => VcpuState::Idle,
+            "waiting" => VcpuState::Waiting,
+            other => {
+                panic!("unknown Oximeter vpcu state name: {}", other);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
iximeow: eh, less annoying to keep the enum up to date this way and there's only one caller of `from_oximeter_state_name` who'd have to panic. [so it derives `strum::EnumStr` now](https://github.com/oxidecomputer/propolis/pull/855/commits/53196223f371697652b9c2c2f1d83dbc223149b9).

(only quibble is that `strum`'s ParseError wouldn't tell you the token that it couldn't make sense of, so you need to carry that around to do error reporting, unless i missed something? either way marginally better now ty)

phd-tests/tests/src/stats.rs
iximeow: probably, but the bit that can be generalized is probably `PropolisOximeterSampler` and `FakeNexusContext`, here, more than anything. i assume dendrite or crucible or whomever would have their own versions of `VirtualMachineMetrics` and walking through Oximeter samples to rebuild their expected reporting feels like it'd be an annoying day of plumbing or so for other tests like this.

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,501 @@
...
+        let metrics_url = {
+            let sampler = self.sampler.lock().unwrap();
+            let stats = sampler.as_ref().expect("stats url info exists");
+            format!("http://{}/{}", stats.addr, stats.uuid)
+        };
+        let res = reqwest::Client::new()
+            .get(metrics_url)
+            .send()
+            .await
+            .expect("can send oximeter stats request");
```
iximeow, line 247: hey @bnaecker how much of a crime _is_ this, in your Oximeter eyes? to me it ffeels kinda bad to embed how to query the Oximeter producer here in PHD, but at least this ought to break loudly enough if it stopped returning the expected stats.

Eliza was also talking about having something like an Oximeter test crate [here](https://github.com/oxidecomputer/propolis/pull/855#discussion_r1951654030), and having to make the metrics URL here is really what sells me on the idea, but maybe we expect this to be more stable than i'm giving it credit for.

phd-tests/tests/src/stats.rs
bnaecker: Depending on exactly what you're looking for, there _is_ a test `oximeter` collector crate. The main program can be run in "standalone" mode. In that case, it pretends to be Nexus, very similarly to what's done in this PR. It can also either actually insert data into ClickHouse, or simply print the results it collects if we don't want to stand up ClickHouse as well. One thing it does not currently do is push the samples to a controllable sink -- that is, you can't directly get access to the collected samples today, you have to fetch them from the database. If that functionality is desirable, we can add it to `oximeter-collector` without too much fuss.

In any case, that standalone mode uses the existing `oximeter-collector` crate's public API. I don't know what the constraints are here, but you can certainly use that in PHD if you wish.

phd-tests/tests/src/stats.rs
bnaecker: This URL is very stable, but I can see why one would neither know that nor want to rely on it. The producer server one stands up to emit metrics fixes the URL for you, there's zero client control over it today. But as I mentioned on the other thread, the only reason we don't already have the ability for you to extract the samples that oximeter collects already is because we haven't needed it yet. Printing or inserting into the DB has been sufficient.

I think adding functionality to give `oximeter` your own sink or channel or whatever, where it will forward any samples it collects, seems like a fine and and valuable tool to add.

phd-tests/tests/src/stats.rs
bnaecker: Thanks, that's helpful. I misunderstood your first comment as meaning that you expected run / emul to be very highly correlated. It sounds like you expect a weak positive correlation, which makes sense to me. Is the issue is that you see the values being much closer in absolute terms than you expect? How close are we talking about here?

I'm just curious about all this, I have very little intuition about what one _should_ see.

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,501 @@
...
+    let start_metrics =
+        fake_nexus.app_private().wait_for_propolis_stats().await;
+
+    // Measure a specific amount of time with guest vCPUs in the "run" state.
+    //
+    // We measure the "run" state using some fixed-size busywork because we
+    // can't simply say "run for 5 seconds please" - if we did, a combination of
+    // host OS or guest OS may leave the process timing itself descheduled for
+    // some or all of that time, so we could end up with substantially less than
+    // 5 seconds of execution and a flaky test as a result.
```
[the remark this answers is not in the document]
gjcolombo: One downside I forgot to mention is that it seems like the most straightforward Linux way of running a process with this sort of affinity is to use `taskset(1)`... which is now another entry on my list of things that aren't in our default Alpine image (though to be fair I haven't checked my other Linux test images for it yet). Maybe we could gin up something equivalent by manipulating cgroups ourselves?

phd-tests/framework/src/test_vm/metrics.rs
```diff
@@ -0,0 +1,200 @@
+use std::net::SocketAddr;
```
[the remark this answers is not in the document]
iximeow: https://github.com/oxidecomputer/propolis/pull/855/commits/be7f609de209badaddca5805a3b4ee5a7e7f530f

phd-tests/tests/src/stats.rs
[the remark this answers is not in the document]
iximeow: (in part of reworking the metrics glue here it uses a `tokio::sync::watch` of an Option and is generally much nicer. ty for bringing it up!)

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,501 @@
...
+use chrono::{DateTime, Utc};
+use dropshot::endpoint;
+use dropshot::ApiDescription;
+use dropshot::ConfigDropshot;
+use dropshot::HttpError;
+use dropshot::HttpResponseCreated;
+use dropshot::HttpServer;
+use dropshot::HttpServerStarter;
+use dropshot::RequestContext;
+use dropshot::TypedBody;
```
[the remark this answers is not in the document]
iximeow: consolidated (and moved!) in https://github.com/oxidecomputer/propolis/pull/855/commits/fe6ab8870c4c6d32af0cfa180efc8fd190d3ed0f

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,501 @@
...
+    //
+    // This is to say: there isn't a strong statement that we can make about
+    // idle time at this point other than that it is probably around a large
+    // enough value to fill the total time out to a mutiple of 5 seconds.
+    //
+    // The guesswork to validate that doesn't seem great in the face of
+    // variable-time CI. We'll validate idle time measurements separately,
+    // below.
+
+    // Idle time boundaries are a little differnt than running time boundaries
```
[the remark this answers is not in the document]
iximeow: (this is just what we'd talked about, noted here for reference):

while i totally agree on the mechanics here - this should work and measure time more precisely - i'd _really_ prefer this test to work across guest OSes if only because if we trust _our_ accounting it starts to double as a useful smoke test of guests themselves. _and_ even if we could make the intentional sleeps 1s or 0s, we're still beholden to the ~5s periods for `propolis-server` to scrape new vCPU kstats. the explicit sleeps are pretty rough, but each metrics sample is also an implicit up-to-5s delay.. we _could_ configure that, but at that point i'd rather just make this conditional and run only in CI :D

if the test itself ends up being long enough to be really annoying locally, splitting the test in two for idle/running time could help some, too.

as for being a useful test of guests, in restructuring the metrics glue i'd noticed that Windows is always checking the current time and so this test becomes _very_ sensitive to Windows using the PM timer vs TSC. so once your TSC enlightenment work is all in, we can and should strike [this condition](https://github.com/oxidecomputer/propolis/pull/855/files#diff-67993f8ca2908cc69d9c30896bf91659188664d9a0e98303edfd004c48bb351eR391-R396) - guests starting to do a bunch of emulated vCPU time while idle seems CI fail worthy, generally speaking!

phd-tests/tests/src/stats.rs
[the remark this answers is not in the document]
iximeow: (broadly same as above: mechanically i agree and i'd also thought about `taskset` at one point, before convincing myself that being OS-agnostic will be valuable to longer-term us. if i've not misunderstood and you agree, then i think the more precise tests are.. useful, but less immediately valuable than "are the orders of magnitude where they should be" that i'm going for here)

phd-tests/framework/src/test_vm/metrics.rs
```diff
@@ -0,0 +1,200 @@
...
+// Re-registration interval for tests. A long value here helps avoid log spew
+// from Oximeter, which will re-register after about 1/6th of this interval
+// elapses.
+const INTERVAL: Duration = Duration::from_secs(300);
+
+fn test_logger() -> Logger {
+    let dec = slog_term::PlainSyncDecorator::new(slog_term::TestStdoutWriter);
+    let drain = slog_term::FullFormat::new(dec).build().fuse();
+    Logger::root(drain, slog::o!("component" => "fake-cleanup-task"))
+}
```
gjcolombo, line 30: In CI runs this causes non-Bunyan-formatted output to be interleaved with the regular PHD logs (see e.g. lines 892-894 of [this log](https://buildomat.eng.oxide.computer/wg/0/artefact/01JMJ7QKRVNVR81EHNBVM26R94/7Ctp8PJX3OEe6HFhjnwvwKYhrS3x3vOIKtDzltjBOFwryFGa/01JMJ7RQWKS735NRX2HBHV9QS1/01JMJF7CMBZ33HS7SPMWWG0X4C/phd-runner.log?format=x-bunyan)).

It'd be nice to have this obey the logging rules from the PHD runner. It might be a bit of a headache to plumb the `emit_bunyan` setting all the way down here, but I think it's probably better than the alternatives (e.g. putting framework Oximeter logs into a separate log file). The way I did this for the propolis-server logging params is to pass them from the test runner to the `Framework`, which packs them into a `ServerProcessParameters` that get passed to `start_local_vm`.

phd-tests/framework/src/test_vm/metrics.rs
```diff
@@ -0,0 +1,200 @@
...
+
+// Re-registration interval for tests. A long value here helps avoid log spew
+// from Oximeter, which will re-register after about 1/6th of this interval
+// elapses.
+const INTERVAL: Duration = Duration::from_secs(300);
+
+fn test_logger() -> Logger {
+    let dec = slog_term::PlainSyncDecorator::new(slog_term::TestStdoutWriter);
+    let drain = slog_term::FullFormat::new(dec).build().fuse();
+    Logger::root(drain, slog::o!("component" => "fake-cleanup-task"))
```
gjcolombo, line 29: nit: I might pick a more descriptive component name; maybe `phd-oximeter-consumer`?

phd-tests/tests/src/stats.rs
```diff
@@ -0,0 +1,415 @@
...
+#[derive(Default, Debug)]
+struct VcpuUsageMetric {
+    metrics: HashMap<VcpuState, u64>,
+}
+
+/// A collection of the stats produced by `propolis-server`'s Oximeter producer.
+///
+/// Oximeter producers produce a series of lists of samples, where each list
+/// of samples is conceptually distinct but may still be interesting to
+/// tset. In `propolis-server`, the first list of samples will be
```
gjcolombo, line 38:
```suggestion
/// test. In `propolis-server`, the first list of samples will be
```

phd-tests/framework/src/test_vm/environment.rs
```diff
@@ -16,15 +16,31 @@ pub enum VmLocation {
...
+/// Specifies where test VMs should report metrics to, if anywhere.
+#[derive(Clone, Copy, Debug)]
+pub enum MetricsLocation {
+    /// Oximeter metrics should be reported to a server colocated with the test
+    /// VM to be started.
+    Local,
+    // When the time comes to support remote VMs, it will presumably be useful
+    // to have local and (perhaps multiple) remote VMs report metrics to the
+    // same server. But we don't support remote VMs yet.
+}
```
gjcolombo, line 28: Multi-host PHD is obviously not very well fleshed-out yet, but I think the way I would develop it might make this enum obsolete. My 30,000-foot sketch of this is

- of the several machines conducting a multi-host PHD run, one is primary and the others are secondary
- the primary machine runs `phd-runner` and executes the actual test cases
- the secondary machines run a `phd-agent` that exposes a Dropshot API that handles the tasks that need to be performed on remote machines (e.g. starting/stopping Propolis servers)
- when invoking the runner on the primary, one passes the addresses of the agent servers

IIUC, even with an architecture like this, the fake Oximeter server still needs to be "local," even if the metrics are coming from a secondary host: submitted samples need to show up in the *runner's* `Framework` so that they can be queried by the tests running in that process.

This is obviated somewhat by switching to Clickhouse, I think, since then (I presume) there would be a single Clickhouse server instance that all PHD processes (both runner and agent) could insert data into, and the framework on the runner would just query that server. Even so, it still seems possible in that case for remote Propolis processes just to submit their samples to the collector in the primary runner.

WDYT?

phd-tests/framework/src/test_vm/metrics.rs
```diff
@@ -0,0 +1,200 @@
...
+        assert!(
+            res.status().is_success(),
+            "failed to fetch stats from propolis-server"
+        );
+        trace!(?res, "got stats response");
+        Ok(res.json::<ProducerResults>().await?)
+    }
+}
+
+// Stub functionality for our fake Nexus that test Oximeter produces
```
gjcolombo, line 154:
```suggestion
// Stub functionality for our fake Nexus that test Oximeter producers
```

phd-tests/tests/src/stats.rs
gjcolombo: With the caveat that I haven't looked at as many Oximeter interfaces or internals as I probably should have, I think this split looks pretty reasonable. Someday we may want to see about streamlining this a bit (so that, if I do all the work to define a new metric series in Propolis, it just "lights up" in PHD without my having to do anything), but I think this is a really good place to start. Thanks for reorganizing these bits!

phd-tests/tests/src/stats.rs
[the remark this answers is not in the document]
gjcolombo: > so once your TSC enlightenment work is all in, we can and should strike this condition - guests starting to do a bunch of emulated vCPU time while idle seems CI fail worthy, generally speaking!

We can do this now! You'll need to copy the VM configuration logic from the Hyper-V tests for now, though. (I'm strongly considering making the Hyper-V config the default once the feature is available/on by default in Omicron, but need to do a smoke test or two first, and need to think about whether we lose any coverage by setting this default for all tests--I don't think we do but need to consider it a bit.)

gjcolombo (commented):
I like how this is coming together! I have a couple of nitpicky typo fixes, a thought about the "metrics location" enum, and a note about logging, but I'm really liking the direction here.

phd-tests/tests/src/stats.rs
iximeow: agh yeah, fixed in e42be6010125c87d4443f1d9918ff233e3e40674

phd-tests/framework/src/test_vm/metrics.rs
iximeow: definitely, this was not a great holdover name :) went with `phd-oximeter-consumer`.

phd-tests/framework/src/test_vm/metrics.rs
iximeow: i hadn't realized that we take `emit_bunyan` or plumb it anywhere in the first place! e88d07c638c9ef3c103ba9f34f41b44c1c458807 does the plumbing here, and it's.. not all that bad really!

the crux of it is that it feels like the format for log output probably should flow from what the caller of `phd-runner` wanted, where we take `--emit-bunyan`. as-is `propolis-server` and Crucible guess the right output formats based on `atty::is()`, but the plumbing for this preference through PHD itself isn't too bad.

after looking at the logs from `phd-run` output i think it'd be interesting and useful to have separate log files for the fake Oximeter server, but that should also come with reorganizing the logs into per-test directories so we could have `phd/migration_smoke/propolis.{stderr,stdout}.log` and `phd/migration_smoke/oximeter.log` and ... etc. bonus being that Crucible disks are just a top-level mash of `crucible-{uuid}-{port}` and have fun cross-referencing that with tests... and hey, we could plumb that per-test log root in `LogConfig`!

phd-tests/framework/src/test_vm/metrics.rs
iximeow: (included in e88d07c638c9ef3c103ba9f34f41b44c1c458807, ty)

phd-tests/tests/src/stats.rs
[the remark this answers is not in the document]
iximeow: unfortunately the TSC was not the sole source of Windows having more emulation time.. i've included the enlightenment for Linux to be consistently well-behaved, and refined the comment with my better understanding of the state of affairs now.

FWIW my inclination is that we probably want the defaults for PHD to hew towards defaults in Omicron? we'll definitely lose test coverage in a whole-system sense, like if the PM timer became panicful and we didn't know because everyone trusted the TSC enlightenment. but that seems more.. niche. feels like checking that guests start without enlightenments and are basically functional covers that?

phd-tests/framework/src/test_vm/environment.rs
iximeow: i was figuring in a case like that we'd still use `Local` to send `localhost:{oximeter_port}` instead of something fancier, but overall that would boil it down to "wants oximeter: yes/no". i do think the plumbing to express that here today is a bit more awkward.. we don't really have a clean "set up dependent services" -> "start `propolis-server`" split in PHD and that does get to be more reorganizing than i want to do here :sweat_smile:

gjcolombo (approved):

PR #856, opened 2025-02-12
lib: implement reference TSC enlightenment

### LINKED ISSUE #328: saw "TSC found unstable after boot, most likely due to broken BIOS"

gjcolombo:
Add a Hyper-V reference time enlightenment to the Hyper-V stack. See the new `tsc` module's doc comment for details about how the enlightenment operates.

Add PHD tests that try to verify both that the enlightenment is present (by querying the current clocksource) and that it keeps time roughly correctly (by comparing host and guest time readings). The latter test also takes readings over a migration to help get some coverage of the time data phase of live migration (though the migration is not currently cross-machine, so this is a little less interesting than it might otherwise be).

Tests: cargo test; PHD w/Alpine, Debian 11, WS2022 guests.

Fixes #328.

### DIFF
--- a/lib/propolis/src/enlightenment/hyperv/mod.rs
+++ b/lib/propolis/src/enlightenment/hyperv/mod.rs
@@ -14,7 +14,7 @@
 //! that intends to implement a Hyper-V-compatible interface:
 //! https://github.com/MicrosoftDocs/Virtualization-Documentation/blob/main/tlfs/Requirements%20for%20Implementing%20the%20Microsoft%20Hypervisor%20Interface.pdf
 
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, OnceLock};
 
 use cpuid_utils::{CpuidIdent, CpuidSet, CpuidValues};
 use overlay::{OverlayError, OverlayKind, OverlayManager, OverlayPage};
@@ -27,7 +27,7 @@ use crate::{
         hyperv::{
             bits::*,
             hypercall::MsrHypercallValue,
-            tsc::{MsrReferenceTscValue, ReferenceTscPage},
+            tsc::{MsrReferenceTscValue, ReferenceTsc},
         },
         AddCpuidError,
     },
@@ -69,44 +69,12 @@ struct HypercallOverlay(OverlayPage);
 /// Wrapper around a TSC overlay page.
 struct TscOverlay(OverlayPage);
 
-impl MsrReferenceTscValue {
-    /// If this MSR value would enable the reference TSC page, attempts to
-    /// create a reference TSC overlay at the MSR's specified PFN.
-    ///
-    /// This routine swallows errors from [`OverlayManager::add_overlay`].
-    /// Notably, this means it returns `None` if a reference TSC overlay already
-    /// exists at the specified PFN.
-    fn create_overlay(
-        &self,
-        manager: &Arc<OverlayManager>,
-        guest_freq: u64,
-    ) -> Option<TscOverlay> {
-        if !self.enabled() {
-            return None;
-        }
-
-        let page = ReferenceTscPage::new(guest_freq);
-
-        manager
-            .add_overlay(self.gpfn(), OverlayKind::ReferenceTsc(page))
-            .ok()
-            .map(TscOverlay)
-    }
-}
-
 /// A collection of overlay pages that a Hyper-V enlightenment stack might be
 /// managing.
 #[derive(Default)]
 struct OverlayPages {
     hypercall: Option<HypercallOverlay>,
     tsc: Option<TscOverlay>,
-}
-
-#[derive(Clone, Copy, Debug)]
-enum ReferenceTsc {
-    Disabled,
-    Uninitialized,
-    Present { msr_value: MsrReferenceTscValue, guest_freq: u64 },
 }
 
 struct Inner {
@@ -119,6 +87,7 @@ struct Inner {
     /// The last value stored in the [`bits::HV_X64_MSR_HYPERCALL`] MSR.
     msr_hypercall_value: MsrHypercallValue,
 
+    /// The state of this stack's reference TSC enlightenment.
     reference_tsc: ReferenceTsc,
 
     /// This enlightenment's active overlay page handles.
@@ -140,14 +109,16 @@ impl Inner {
         }
     }
 
+    /// Resets this enlightenment block's volatile values (e.g. MSR values) to
+    /// their initial values.
     fn reset(&mut self) {
         *self = Self {
             overlay_manager: self.overlay_manager.clone(),
             msr_guest_os_id_value: 0,
             msr_hypercall_value: MsrHypercallValue::default(),
             reference_tsc: match &self.reference_tsc {
-                ReferenceTsc::Present { guest_freq, .. } => {
-                    ReferenceTsc::Present {
+                ReferenceTsc::Enabled { guest_freq, .. } => {
+                    ReferenceTsc::Enabled {
                         guest_freq: *guest_freq,
                         msr_value: MsrReferenceTscValue::default(),
                     }
@@ -164,22 +135,16 @@ impl Inner {
             ReferenceTsc::Uninitialized => {
                 panic!("reference TSC MSR read before enlightenment was ready")
             }
-            ReferenceTsc::Present { msr_value, .. } => {
+            ReferenceTsc::Enabled { msr_value, .. } => {
                 RdmsrOutcome::Handled(msr_value.0)
             }
         }
     }
 
     fn handle_wrmsr_reference_tsc(&mut self, value: u64) -> WrmsrOutcome {
-        let (msr_value, guest_freq) = match &mut self.reference_tsc {
-            ReferenceTsc::Disabled => return WrmsrOutcome::GpException,
-            ReferenceTsc::Uninitialized => panic!(
-                "reference TSC MSR written before enlightenment was ready"
-            ),
-            ReferenceTsc::Present { msr_value, guest_freq } => {
-                (msr_value, *guest_freq)
-            }
-        };
+        if !self.reference_tsc.is_present() {
+            return WrmsrOutcome::GpException;
+        }
 
         let new = MsrReferenceTscValue(value);
         probes::hyperv_wrmsr_reference_tsc!(|| (
@@ -188,18 +153,21 @@ impl Inner {
             new.enabled()
         ));
 
+        // Unlike the hypercall MSR, writes to the reference TSC MSR always
+        // succeed without raising an exception, even if they try to enable the
+        // TSC overlay page at an invalid PFN.
         let old_overlay = self.overlays.tsc.take();
+        self.reference_tsc.set_msr_value(new);
         self.overlays.tsc = if new.enabled() {
             if let Some(mut overlay) = old_overlay {
                 overlay.0.move_to(new.gpfn()).ok().map(|_| overlay)
             } else {
-                new.create_overlay(&self.overlay_manager, guest_freq)
+                self.reference_tsc.create_overlay(&self.overlay_manager)
             }
         } else {
             None
         };
 
-        *msr_value = new;
         WrmsrOutcome::Handled
     }
 }
@@ -210,6 +178,7 @@ pub struct HyperV {
     features: Features,
     inner: Mutex<Inner>,
     acc_mem: MemAccessor,
+    vmm_hdl: OnceLock<Arc<VmmHdl>>,
 }
 
 impl HyperV {
@@ -228,7 +197,19 @@ impl HyperV {
             features,
             inner: Mutex::new(Inner::new(&features)),
             acc_mem,
+            vmm_hdl: OnceLock::new(),
         }
+    }
+
+    /// Returns a reference to this manager's VMM handle.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the handle has not been initialized yet, which occurs if this
+    /// routine is called before the enlightenment receives its `attach`
+    /// callout.
+    fn vmm_hdl(&self) -> &Arc<VmmHdl> {
+        self.vmm_hdl.get().unwrap()
     }
 
     /// Handles a write to the HV_X64_MSR_GUEST_OS_ID register.
@@ -318,6 +299,8 @@ impl HyperV {
         }
     }
 
+    /// Handles a read of the `HV_X64_MSR_TIME_REF_COUNT` register. See TLFS
+    /// section 12.4.
     fn handle_rdmsr_time_ref_count(&self) -> RdmsrOutcome {
         if !self.features.reference_tsc {
             return RdmsrOutcome::GpException;
@@ -480,11 +463,21 @@ impl super::Enlightenment for HyperV {
             let time_data = vmm::time::export_time_data(&vmm_hdl)
                 .expect("VMM time data is accessible during attach");
 
-            inner.reference_tsc = ReferenceTsc::Present {
+            // N.B. This guest TSC frequency may be overwritten by a future
+            // request to import state from a migration source. This is
+            // intentional; the migration protocol will configure the kernel VMM
+            // to apply hardware TSC scaling so that the guest observes the
+            // imported frequency.
+            inner.reference_tsc = ReferenceTsc::Enabled {
                 guest_freq: time_data.guest_freq,
                 msr_value: MsrReferenceTscValue::default(),
             }
         }
+
+        // `attach` should only called once on each enlightenment instance.
+        // `VmmHdl` doesn't implement `Debug`, so it's not possible to use
+        // `unwrap` or `expect` here.
+        assert!(self.vmm_hdl.set(vmm_hdl).is_ok());
     }
 }
 
@@ -543,8 +536,10 @@ impl Lifecycle for HyperV {
             .map(HypercallOverlay);
 
         let tsc_overlay = inner
-            .msr_reference_tsc_value
-            .create_overlay(&inner.overlay_manager, self.vmm_hdl());
+            .reference_tsc
+            .is_present()
+            .then(|| inner.reference_tsc.create_overlay(&inner.overlay_manager))
+            .flatten();
 
         inner.overlays =
             OverlayPages { hypercall: hypercall_overlay, tsc: tsc_overlay };
@@ -558,10 +553,7 @@ impl Lifecycle for HyperV {
         // `pause` before `reset`.
         assert!(inner.overlay_manager.is_empty());
 
-        *inner = Inner {
-            overlay_manager: inner.overlay_manager.clone(),
-            ..Default::default()
-        };
+        inner.reset();
     }
 
     fn halt(&self) {
@@ -583,7 +575,18 @@ impl MigrateSingle for HyperV {
         Ok(migrate::HyperVEnlightenmentV1 {
             msr_guest_os_id: inner.msr_guest_os_id_value,
             msr_hypercall: inner.msr_hypercall_value.0,
-            msr_reference_tsc: inner.msr_reference_tsc_value.0,
+            reference_tsc: match inner.reference_tsc {
+                ReferenceTsc::Disabled => None,
+                ReferenceTsc::Uninitialized => {
+                    panic!("asked to export an uninitialized TSC enlightenment")
+                }
+                ReferenceTsc::Enabled { msr_value, guest_freq } => {
+                    Some(migrate::ReferenceTscV1 {
+                        msr_value: msr_value.0,
+                        guest_freq,
+                    })
+                }
+            },
         }
         .into())
     }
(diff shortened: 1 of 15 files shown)


phd-tests/tests/src/hyperv.rs
```diff
@@ -83,3 +89,186 @@ async fn hyperv_lifecycle_test(ctx: &Framework) {
...
+    // The logic below takes pairs of time readings on the host and guest and
+    // compares them to check that the amount of elapsed time perceived by the
+    // guest is roughly equivalent to the "real" elapsed time on the host. The
+    // host and guest measurements can't be taken atomically, so some difference
+    // between these readings is expected (especially because the guest shell
+    // command needs to print its result to the serial port, and the harness has
+    // to read it).
+    //
+    // Accept a measurement error of up to 1% of the elapsed time between host
+    // readings. This is an extremely generous tolerance. It is chosen primarily
```
gjcolombo, line 231: Apparently it's not generous enough:

```
phd-runner: [TEST - EVENT] compared time readings
    after = Reading { taken_at: Instant { tv_sec: 1525, tv_nsec: 922518064 }, guest_ns: 2289389900 }
    before = Reading { taken_at: Instant { tv_sec: 1525, tv_nsec: 376576078 }, guest_ns: 1753546300 }
    diff = 10098386
    diff_pct = 0.018497177830173333
    file = phd-tests/tests/src/hyperv.rs
    guest_delta_ns = 535843600
    host_delta_ns = 545941986
    line = 208
    path = phd_tests::hyperv::hyperv_reference_tsc_elapsed_time_test
    target = phd_tests::hyperv
```

lib/propolis/src/enlightenment/hyperv/tsc.rs
```diff
@@ -0,0 +1,238 @@
...
+//! elapsed reference units
+//!     = TSC ticks * (reference frequency / TSC frequency)
+//!     = (TSC ticks * scaling factor) / 2^64
+//!     = (TSC ticks * scaling factor) >> 64
+//! ```
+//!
+//! There is one small catch: the scaling factor is nominally a 128-bit integer,
+//! but the x86-64 `IMUL` instruction's maximum operand size is 64 bits. There
+//! are several ways around this; Hyper-V's is to observe that if the host TSC
+//! frequency is greater than or equal to 10 MHz, then scaling factor will be
```
gjcolombo, line 70: nit: greater than, not greater than or equal to

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -358,16 +495,57 @@ impl MigrateMulti for HyperV {
...
+        // exported. Thus it's not necessary to change the reference TSC page's
+        // contents here. See the TSC module documentation for more details.
+        let msr_reference_tsc = MsrReferenceTscValue(data.msr_reference_tsc);
+        let tsc_overlay = overlays
+            .iter()
+            .position(|page| page.kind() == OverlayKind::ReferenceTsc)
+            .map(|pos| overlays.swap_remove(pos));
+
+        if msr_reference_tsc.enabled() {
+            let Some(overlay) = &tsc_overlay else {
```
gjcolombo, line 510: This is not actually an error if the GPA in the MSR is out of range. (This is different from the hypercall MSR case, where register values with an invalid GPA raise #GP instead of writing back to the MSR.) This might call for a mechanism that simply checks if an overlay *could be created* at a given GPA but that doesn't actually create one.

phd-tests/tests/src/hyperv.rs
gjcolombo: This passed with flying colors on a rerun, though. Instead of loosening the tolerance, I might adjust the logic so that this part of the test fails if all, or a majority, of the samples exceed the tolerance, since that implies a lower likelihood that it's all just noise. Will mull this over a bit more (and other suggestions are especially welcome!).

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -168,6 +208,75 @@ impl HyperV {
...
+
+        // `hrtime` is the current host high-resolution timer value (in
+        // nanoseconds), and `boot_hrtime` is the high-resolution timer value at
+        // which the VM was started, so it suffices to subtract the latter from
+        // the former and divide by 100 to get 100-nanosecond units.
+        //
+        // If this VM is migrated or otherwise saved/restored, the migration
+        // protocol is expected to set `boot_hrtime` for the restored VM to
+        // account for any hrtime differences between hosts and any time where
+        // the VM was paused.
```
jordanhendricks, line 228: I don't think it's true that it takes into account pause time explicitly; the adjustment on the propolis side for live migration is to reconcile the boot_hrtime on a new host (relative to a new host hrtime), and on the kernel side, the TSC is adjusted, but not the boot_hrtime. The time of migration is accounted for in these calculations, but not so explicity.

https://github.com/oxidecomputer/illumos-gate/blob/7633a05bff8c639f2df722d1fba7b889b2763d3d/usr/src/uts/intel/io/vmm/vmm.c#L4999-L5031

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -358,16 +495,57 @@ impl MigrateMulti for HyperV {
             }
         }
 
+        // This enlightenment assumes that if a VM is exported and imported, the
+        // caller asking to import will configure the bhyve VM so that it has
+        // the same apparent guest frequency and offset the VM had when it was
+        // exported. Thus it's not necessary to change the reference TSC page's
+        // contents here. See the TSC module documentation for more details.
```
jordanhendricks, line 502: is "TSC module" referring to the Hyper-V docs?

lib/propolis/src/enlightenment/hyperv/tsc.rs
```diff
@@ -0,0 +1,238 @@
...
+//! fractional part. That is, the idea is to re-express the frequency multiplier
+//! term in the conversion above as follows:
+//!
+//! ```text
+//! frequency multiplier
+//!     = reference frequency / TSC frequency
+//!     = ((reference frequency) * 2^N) / TSC frequency) * (1 / 2^N)
+//! ```
+//!
+//! In the present case, N = 64, and the `reference frequency * 2^N` term is a
```
jordanhendricks, line 51: Is this just an example? For intel/amd, the size of the fractional number is different (amd is 40 bits with 8int/32 frac, and intel 64 bits with 16 int / 48 frac)

jordanhendricks (commented):
initial pass; want to learn a bit more about this feature in depth and I'll take another one

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -62,23 +74,51 @@ struct Inner {
 
     /// The last value stored in the [`bits::HV_X64_MSR_HYPERCALL`] MSR.
     msr_hypercall_value: MsrHypercallValue,
-
     hypercall_overlay: Option<OverlayPage>,
+
+    /// The last value stored in the [`bits::HV_X64_MSR_REFERENCE_TSC`] MSR.
+    msr_reference_tsc_value: MsrReferenceTscValue,
+    reference_tsc_overlay: Option<OverlayPage>,
```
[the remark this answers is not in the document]
gjcolombo: I tried this out, and there are definitely a couple of sort of smelly places where we have to acquire the mutex transiently just to call `is_some()` on the reference TSC `Option`. `handle_rdmsr_time_ref_count` is one; `add_cpuid` is another surprising one (that function needs to check the feature flag to figure out what to advertise in HV leaf 3). This also didn't work wonders for the ergonomics of `handle_wrmsr_reference_tsc` (I ended up having to add a lot of `as_mut`s to avoid holding a mutable reference to the reference TSC state while using an immutable reference to access the overlay manager).

I think a more structurally pleasing option is to put the reference TSC state under its own `Mutex` and wrap *that* in an `Option`, but that also turns out to be something of a pain, because the overlay manager lives under the other lock, and now callers have to reason about lock ordering. It might be possible not to have the overlay manager live under the mutex, but it's really convenient to be able to reset it (and invalidate its `OverlayPage`s) by just swapping its reference out for a reference to a new manager.

Anyway. I think it's a somewhat close call; it's nice not to use extra storage when the enlightenment is off, and it's also nice not to be able to forget to check for the feature before futzing with the enlightenment state. But it makes all the handling code fiddly enough that I think I like the current arrangement a bit better.

phd-tests/tests/src/hyperv.rs
gjcolombo: 6330dff takes a swing at this.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: Fixed in 908237b

lib/propolis/src/enlightenment/hyperv/tsc.rs
gjcolombo: Fixed in dce0275.

lib/propolis/src/enlightenment/hyperv/tsc.rs
```diff
@@ -0,0 +1,238 @@
...
+/// Represents a value written to the [`HV_X64_MSR_REFERENCE_TSC`] register.
+///
+/// [`HV_X64_MSR_REFERENCE_TSC`]: super::bits::HV_X64_MSR_REFERENCE_TSC
+#[derive(Clone, Copy, Default)]
+pub(super) struct MsrReferenceTscValue(pub(super) u64);
+
+impl std::fmt::Debug for MsrReferenceTscValue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("MsrReferenceTscValue")
+            .field("raw", &self.0)
```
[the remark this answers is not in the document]
gjcolombo: Fixed in b73925d

lib/propolis/src/enlightenment/hyperv/mod.rs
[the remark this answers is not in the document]
gjcolombo: My pleasure!

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -168,6 +208,75 @@ impl HyperV {
...
+        ));
+
+        // The reference TSC MSR can always be written even if the guest OS ID
+        // MSR is 0. TLFS section 12.7.1 specifies that if the selected GPA is
+        // invalid, "the reference TSC page will not be accessible to the
+        // guest," but the MSR write itself does not #GP.
+        let mut inner = self.inner.lock().unwrap();
+        if !new.enabled() {
+            inner.reference_tsc_overlay.take();
+        } else if let Some(mut overlay) = inner.reference_tsc_overlay.take() {
```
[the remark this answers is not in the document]
gjcolombo: I think we still need a `new.enabled()` check to avoid moving the overlay if the whole thing is now disabled, but I tightened this up a bit in 9958696.

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -358,16 +495,57 @@ impl MigrateMulti for HyperV {
             }
         }
 
+        // This enlightenment assumes that if a VM is exported and imported, the
+        // caller asking to import will configure the bhyve VM so that it has
+        // the same apparent guest frequency and offset the VM had when it was
+        // exported. Thus it's not necessary to change the reference TSC page's
```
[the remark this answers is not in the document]
gjcolombo: Fixed in dce0275.

lib/propolis/src/enlightenment/hyperv/tsc.rs
```diff
@@ -0,0 +1,238 @@
...
+//!     = TSC ticks * (1 / TSC frequency) * reference frequency
+//!     = TSC ticks * (reference frequency / TSC frequency)
+//! ```
+//!
+//! (This calculation assumes that the TSC does in fact tick at a constant
+//! frequency. This is often the case on modern processors, but it is not
+//! guaranteed, and system software is expected to check CPUID to see if the CPU
+//! advertises such an "invariant" TSC before doing this kind of calculation.)
+//!
+//! KVM and Linux use nanoseconds as the reference time unit and so have a
```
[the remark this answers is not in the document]
gjcolombo: Fixed in dce0275.

lib/propolis/src/enlightenment/hyperv/tsc.rs
```diff
@@ -0,0 +1,238 @@
...
+//! reference frequency of 1e9 ticks/sec. Windows and Hyper-V use 100ns units
+//! and a frequency of 1e7 ticks/sec. Because these frequencies are expressed in
+//! cycles per second, using simple integer divisions to convert ticks to
+//! seconds or to scale frequencies will lose all sub-second precision, which
+//! precision is of course the point of having a high-resolution timekeeping
```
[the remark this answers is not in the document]
gjcolombo: YOU CAN'T CUT FUNDING FOR MY ARCHAIC USAGE OF NOUNS IN RELATIVE CLAUSES! YOU WILL REGRET THIS!

Fixed in dce0275.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: I knocked this language out in dce0275. I had in mind a case where one has a running VM and chooses to "suspend" it so that its state is saved and it no longer perceives the passage of time. In a case like that I think you'd need to adjust `boot_hrtime` on "resume" to give the illusion that no time passed while the VM was paused. But that's not a use case we have today, so I've simplified this comment just to talk about migration.

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: Fixed in dce0275.

lib/propolis/src/enlightenment/hyperv/tsc.rs
gjcolombo: Substantially reworked in dce0275 to try to clarify that the 64.64 fixed-point fraction format is chosen by the enlightenment and isn't strictly required to do this kind of math.

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -225,6 +338,16 @@ impl super::Enlightenment for HyperV {
                 let id: u32 = vcpu.into();
                 RdmsrOutcome::Handled(id as u64)
             }
+            HV_X64_MSR_TIME_REF_COUNT => self.handle_rdmsr_time_ref_count(),
+            HV_X64_MSR_REFERENCE_TSC => {
+                if self.features.reference_tsc {
+                    RdmsrOutcome::Handled(
+                        self.inner.lock().unwrap().msr_reference_tsc_value.0,
+                    )
```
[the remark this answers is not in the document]
gjcolombo: I'd actually be sort of surprised if guests even bother reading this before writing it, though I suppose they're technically supposed to in order to preserve the value of any reserved bits. In any case I doubt anyone's going to read it more than once, since it'll just yield the value the guest previously set.

lib/propolis/src/enlightenment/hyperv/mod.rs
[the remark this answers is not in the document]
hawkw: Yeah, I think that's all fair. I agree that it would be better not to have a separate `Mutex` for it if we also need to lock the overlay manager...

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -168,6 +208,78 @@ impl HyperV {
...
+            new.enabled()
+        ));
+
+        // The reference TSC MSR can always be written even if the guest OS ID
+        // MSR is 0. TLFS section 12.7.1 specifies that if the selected GPA is
+        // invalid, "the reference TSC page will not be accessible to the
+        // guest," but the MSR write itself does not #GP. This means that if the
+        // overlay is enabled, it suffices to try to create an overlay (or move
+        // the existing one) to the requested location and, if this succeeds,
+        // store the resulting overlay page.
```
gjcolombo, line 256: nit: this last sentence could use some polish

gjcolombo: fixed in 24be46e

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -168,6 +208,78 @@ impl HyperV {
...
+        // which the VM was started, so it suffices to subtract the latter from
+        // the former and divide by 100 to get 100-nanosecond units.
+        //
+        // This module assumes that if its user migrates this VM, the migration
+        // procedure will adjust `boot_hrtime` on the target to account for any
+        // hrtime differences between the source and target VM hosts.
+        let ns_since_boot: u64 = time_data
+            .hrtime
+            .checked_sub(time_data.boot_hrtime)
+            .expect("overflow/underflow while calculating reference uptime")
```
iximeow, line 231: thanks for bearing with me kicking the tires on the arithmetic here - if you want to keep the `.expect()`s, how do you feel about including a note that overflow implies the VM was up `2**63 / 1e9 / 3600 / 24 / 356 years` aka >240 years?

phd-tests/tests/src/hyperv.rs
```diff
@@ -83,3 +89,216 @@ async fn hyperv_lifecycle_test(ctx: &Framework) {
...
+            Box::pin(async move {
+                if !check_clocksource {
+                    let echo = target
+                        .run_shell_command(
+                            "echo clocksource queries not supported",
+                        )
+                        .await
+                        .unwrap();
+
+                    assert_eq!(echo, "clocksource queries not supported");
```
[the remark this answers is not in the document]
gjcolombo: Good call. Done in 7e12738.

phd-tests/tests/src/hyperv.rs
```diff
@@ -83,3 +89,216 @@ async fn hyperv_lifecycle_test(ctx: &Framework) {
...
+        .migrate_from(&vm, Uuid::new_v4(), MigrationTimeout::default())
+        .await?;
+
+    let after_migration = Reading::take_from(&target).await?;
+    let diff_pct = after_migration.compare_with_earlier(&before_migration);
+    if diff_pct < TOLERANCE {
+        good_diffs += 1;
+    } else {
+        bad_diffs += 1;
+    }
```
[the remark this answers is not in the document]
gjcolombo: Good call here, too. I reworked this in 97b15e6: now the test takes six readings (for five diffs) per pass and does them both before and after the live migration.

I also bumped the tolerance up to 2.5% to try to deflake CI even further. That number feels very large to me--we really shouldn't have that much error under normal conditions--but I think it's still small enough to catch really egregious errors, like "I shifted by 63 places and not 64 when I calculated the scaling factor."

lib/propolis/src/enlightenment/hyperv/mod.rs
gjcolombo: Annotated in b3d3a74. I also added a USDT probe to this path to verify that I could get a guest to hit it (and I can: Linux will hammer this MSR if you offer the TSC enlightenment but write a 0 to the `sequence` field of the reference TSC page).

iximeow: ah this is lovely, thanks!

gjcolombo:
I've rebased this to the latest master. This was unfortunately a bit of a mess, because #861 obsoleted a fair amount of the migration logic that was present previously. It may help to look at just the [last five commits](https://github.com/oxidecomputer/propolis/pull/856/files/5265e0e594d75a5eb89c4cd08898ca3f0ccc94d5..9b3c29486e64685e7b78a676a08be5d344ff6cdc) to get a sense of what's actually changed (beyond merge conflict resolutions during the rebase).

Now that the Hyper-V stack doesn't migrate overlay page contents directly, it needs to be more careful to ensure that it generates TSC overlays that are based on the correct guest frequency. The challenges here are

- we can't read the guest frequency during enlightenment creation because we don't have a `VmmHdl` yet
- during import, we *do* have a `VmmHdl`, but we'd like to avoid assuming that the time-data phase of migration has already happened (which has to happen for us to read the correct guest frequency out of the kernel VM)

Most of the new code involves a new `ReferenceTsc` helper type that keeps track of whether the enlightenment is on, whether it's initialized, and what guest frequency it's supposed to use when constructing the reference page. A lot of the newest code is meant to work with this type and to make sure its configuration data doesn't get blown away during VM reset.

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -358,16 +495,57 @@ impl MigrateMulti for HyperV {
...
+                    overlay.pfn()
+                )));
+            };
+        } else {
+            if tsc_overlay.is_some() {
+                return Err(MigrateStateError::ImportFailed(
+                    "reference TSC overlay present but page is disabled"
+                        .to_string(),
+                ));
+            }
```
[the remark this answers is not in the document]
gjcolombo: These presence checks were obsolete and are now (or should now be) gone in the latest version of the migration logic.

hawkw (approved):
Hm, the introduction of the `Uninitialized` state etc is not my _favorite_ thing, but I agree that this is necessary due to the changes to how we migrate overlays, and I can't think up a solution that avoids it. This seems good to me!

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -82,11 +87,89 @@ struct Inner {
...
+            },
+            overlays: OverlayPages::default(),
+        }
+    }
+
+    fn handle_rdmsr_reference_tsc(&self) -> RdmsrOutcome {
+        match self.reference_tsc {
+            ReferenceTsc::Disabled => RdmsrOutcome::GpException,
+            ReferenceTsc::Uninitialized => {
+                panic!("reference TSC MSR read before enlightenment was ready")
```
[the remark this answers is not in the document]
iximeow: worse than our fault, should be entirely impossible to get here if we've gone through `guest_hv_interface.attach()`. worth making this an `unreachable!`?

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -533,7 +575,18 @@ impl MigrateSingle for HyperV {
         Ok(migrate::HyperVEnlightenmentV1 {
             msr_guest_os_id: inner.msr_guest_os_id_value,
             msr_hypercall: inner.msr_hypercall_value.0,
-            msr_reference_tsc: inner.msr_reference_tsc_value.0,
+            reference_tsc: match inner.reference_tsc {
+                ReferenceTsc::Disabled => None,
+                ReferenceTsc::Uninitialized => {
+                    panic!("asked to export an uninitialized TSC enlightenment")
```
iximeow, line 581: i can tell this is unreachable: to migrate you have to have a `VmObjects`, but to have gotten a `VmObjects` you have to have gotten through `machine::Builder::finalize` which would have moved this out of `Uninitialized`. i don't quite see what would happen if we were asked to migrate out while still `Creating`, but i assume it's either an error or blocks until the VM is created?

either way, seems like this could be `unreachable!` with some gesturing at the state machine in a comment here too

iximeow (approved):
more nits about the panics :) seems good, very excited for this!

lib/propolis/src/enlightenment/hyperv/mod.rs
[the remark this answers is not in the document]
gjcolombo: I think I mildly prefer `panic!` here for layering reasons: this path is reachable through no fault of the library's if the library consumer (`propolis-server` or `propolis-standalone`) messes up and fails to call `attach` before starting vCPUs with this enlightenment. The panic message could stand to be a little clearer about this; I'll tweak it a bit.

gjcolombo:
Thanks for re-reviewing! ae5cfac should take care of the latest batch of comments.

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -183,6 +189,12 @@ pub struct HyperV {
 
 impl HyperV {
     /// Creates a new Hyper-V enlightenment stack with the supplied `features`.
+    ///
+    /// The caller must call [`attach`] to finish initializing this
+    /// enlightenment stack before starting any VM components that depend on it.
+    /// Otherwise the stack may panic while the VM is running.
```
hawkw, line 195: mmmmmaybe this ought to have a big leading "WARNING:" on it? not a big deal though.

lib/propolis/src/enlightenment/hyperv/mod.rs
```diff
@@ -209,7 +221,10 @@ impl HyperV {
     /// routine is called before the enlightenment receives its `attach`
     /// callout.
     fn vmm_hdl(&self) -> &Arc<VmmHdl> {
-        self.vmm_hdl.get().unwrap()
+        self.vmm_hdl.get().expect(
+            "a fully-initialized Hyper-V enlightenment always has a \
+            VMM handle (did the library user remember to call `attach`?)",
```
hawkw, line 226: good call with these notes!

hawkw (approved):
new commentary looks lovely, thank you!

jordanhendricks (approved):
:ship:

PR #861, opened 2025-02-18
lib: tidy up overlay page migration & reduce memory usage

gjcolombo:
Tweak a few bits of Hyper-V overlay migration:

- Don't store page kinds separately from page contents. Instead, a page's kind *determines* its contents. Most overlays have static contents or contents that depend on only a couple of parameters (e.g. the TSC scale/offset for the TSC reference page), so this saves a fair bit of memory for most kinds of pending overlay pages.

- Change the overlay ordering discipline to depend entirely on the ordering of `OverlayKind`s. (Previously the first overlay assigned to a PFN would become active and then would remain active until it was removed, but pending overlays were promoted in order by kind.) This obviates the need to export/import a PFN's current active overlay during migration (or to remember the order in which overlays were applied), since a set of overlays at a single PFN will always have the same active overlay (...provided the ordering remains stable across Propolis versions; there's a comment about this caveat in the new code.)

- Continue to manually transfer the original contents of guest pages that are covered by an overlay (since these still need to be restored), but wrap them in an enum with a variant that specifies that the overlaid page contained all zeroes. Because VM memory is initially zeroed, and many guests will establish overlay pages early in boot over pages they haven't otherwise touched, this can significantly reduce the amount of serialized page data that the manager needs to send over a migration.

See https://github.com/oxidecomputer/propolis/pull/853#discussion_r1960595633 for a bit more context.

### DIFF
--- a/lib/propolis/src/enlightenment/hyperv/overlay.rs
+++ b/lib/propolis/src/enlightenment/hyperv/overlay.rs
@@ -97,13 +97,16 @@ use std::{
 
 use thiserror::Error;
 
-use crate::{accessors::MemAccessor, common::PAGE_SIZE, vmm::Pfn};
+use crate::{
+    accessors::MemAccessor, common::PAGE_SIZE, migrate::MigrateStateError,
+    vmm::Pfn,
+};
 
 use self::pfn::MappedPfn;
 
 /// An error that can be returned from an overlay page operation.
 #[derive(Debug, Error)]
-pub enum OverlayError {
+pub(super) enum OverlayError {
     /// The guest memory context can't be accessed right now. Generally this
     /// means that the caller is trying to create an overlay too early (i.e.,
     /// before the overlay manager is attached to the memory hierarchy) or too
@@ -130,6 +133,27 @@ pub enum OverlayError {
     /// caller.
     #[error("PFN {0:#x} has no overlay of kind {1:?}")]
     NoOverlayOfKind(u64, OverlayKind),
+
+    #[error(
+        "imported overlay contents have length {actual_len}, \
+        expected {expected_len}"
+    )]
+    OriginalContentsCountMismatch { actual_len: usize, expected_len: usize },
+
+    #[error("saved overlay contents PFN {0:#x} is not a valid 52-bit PFN")]
+    InvalidImportPfn(u64),
+
+    #[error(
+        "saved overlay contents for PFN {0:#x} are {1} bytes, \
+        expected PAGE_SIZE"
+    )]
+    InvalidSavedContentsLength(u64, usize),
+}
+
+impl From<OverlayError> for MigrateStateError {
+    fn from(value: OverlayError) -> Self {
+        MigrateStateError::ImportFailed(value.to_string())
+    }
 }
 
 /// The contents of a 4 KiB page. These are boxed so that this type can be
@@ -146,6 +170,15 @@ impl Default for OverlayContents {
 impl std::fmt::Debug for OverlayContents {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_tuple("OverlayContents").field(&"<page redacted>").finish()
+    }
+}
+
+impl TryFrom<Vec<u8>> for OverlayContents {
+    type Error = usize;
+
+    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+        let len = value.len();
+        Ok(Self(Box::new(value.try_into().map_err(|_| len)?)))
     }
 }
 
@@ -418,6 +451,37 @@ impl ManagerInner {
 
         Ok(())
     }
+
+    /// Visits each of the overlaid pages known to this manager and overwrites
+    /// their original contents with the data in the supplied `contents` map.
+    fn restore_original_contents(
+        &mut self,
+        contents: BTreeMap<u64, Vec<u8>>,
+    ) -> Result<(), OverlayError> {
+        if contents.len() != self.overlays.len() {
+            return Err(OverlayError::OriginalContentsCountMismatch {
+                actual_len: contents.len(),
+                expected_len: self.overlays.len(),
+            });
+        }
+
+        for (pfn, contents) in contents {
+            let pfn =
+                Pfn::new(pfn).ok_or(OverlayError::InvalidImportPfn(pfn))?;
+
+            let entry = self
+                .overlays
+                .get_mut(&pfn)
+                .ok_or(OverlayError::NoOverlaysActive(pfn.into()))?;
+
+            entry.original_contents = OverlayContents::try_from(contents)
+                .map_err(|len| {
+                    OverlayError::InvalidSavedContentsLength(pfn.into(), len)
+                })?;
+        }
+
+        Ok(())
+    }
 }
 
 /// An overlay tracker that adds, removes, and moves overlays.
@@ -497,6 +561,29 @@ impl OverlayManager {
             kind,
             &self.acc_mem,
         )
+    }
+
+    /// Saves the original page contents of each overlaid PFN registered with
+    /// this manager, returning a mapping from raw PFN values to Vecs containing
+    /// the relevant pages' contents.
+    pub(super) fn save_original_contents(&self) -> BTreeMap<u64, Vec<u8>> {
+        self.inner
+            .lock()
+            .unwrap()
+            .overlays
+            .iter()
+            .map(|(pfn, set)| {
+                (u64::from(*pfn), set.original_contents.0.to_vec())
+            })
+            .collect()
+    }
+
+    /// Attempts
+    pub(super) fn restore_original_contents(
+        &self,
+        contents: BTreeMap<u64, Vec<u8>>,
+    ) -> Result<(), OverlayError> {
+        self.inner.lock().unwrap().restore_original_contents(contents)
     }
 }
 
@@ -758,5 +845,21 @@ mod test {
             .unwrap_err();
 
         ctx.assert_pfn_has_fill(pfn, 0x70);
+    }
+
+    #[test]
+    fn restore_original_contents() {
+        let ctx = TestCtx::new();
+        let pfn = Pfn::new_unchecked(0x60);
+
+        let page = ctx
+            .manager
+            .add_overlay(pfn, OverlayKind::Test { index: 0, fill: 0x88 })
+            .unwrap();
+
+        let contents = BTreeMap::from([(0x60, vec![0x44; PAGE_SIZE])]);
+        ctx.manager.restore_original_contents(contents).unwrap();
+        drop(page);
+        ctx.assert_pfn_has_fill(pfn, 0x44);
     }
 }
--- a/lib/propolis/src/enlightenment/hyperv/overlay.rs
+++ b/lib/propolis/src/enlightenment/hyperv/overlay.rs
@@ -91,7 +91,7 @@
 //! [`Lifecycle::halt`]: crate::lifecycle::Lifecycle::halt
 
 use std::{
-    collections::{btree_map::Entry, BTreeMap},
+    collections::{btree_map::Entry, BTreeMap, BTreeSet},
     sync::{Arc, Mutex, Weak},
 };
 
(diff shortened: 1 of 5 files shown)


lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -407,26 +492,52 @@ impl ManagerInner {
...
-        self.add_overlay_using_mapping(&mut to_mapping, kind, contents)
+        self.add_overlay_using_mapping(&mut to_mapping, kind)
             .expect("already checked new PFN for an overlay of this kind");
 
         Ok(())
     }
+
+    /// Visits each of the overlaid pages known to this manager and overwrites
+    /// their original contents with the data in the supplied `contents` map.
+    fn restore_original_contents(
```
hawkw, line 516: I find the naming here a little bit confusing: when removing an overlay page, we call a function called `restore_overlaid_data` that actually puts back the previous contents of the guest page when an overlay was applied. This function, `restore_original_contents`, seems like it's called during an import (?), and sets the original contents for each page in the overlay manager to the values in the provided map --- it doesn't actually put back the original contents in the guest page, it just sets what the original contents that we _would_ restore when removing an overlay page would be. This surprised me. 

I might suggest not calling both of these operations "restore" --- this one seems more like `set_original_contents` or even just `import_original_contents` or something...

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -130,13 +134,45 @@ pub enum OverlayError {
...
+    /// A caller asked to restore the guest page underlying an invalid PFN.
+    #[error("saved overlay contents PFN {0:#x} is not a valid 52-bit PFN")]
+    InvalidImportPfn(u64),
+
+    /// A caller supplied guest data to use as the original contents of an
+    /// overlaid page, but that data was not page-sized.
+    #[error(
+        "saved overlaid guest page contents are {0} bytes, expected PAGE_SIZE"
+    )]
+    InvalidOverlaidGuestPageLength(usize),
```
hawkw, line 156: since this is the complete error message that will bubble up when an import fails, perhaps we ought to include the PFN in _all_ the errors to aid in debugging?

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -215,40 +300,69 @@ impl Drop for OverlayPage {
...
-    active: OverlayKind,
+    original_contents: OverlaidGuestPage,
 
-    /// The overlays that are waiting to be applied to this PFN.
-    pending: BTreeMap<OverlayKind, OverlayContents>,
+    /// The set of overlays that have been requested at this PFN.
+    ///
+    /// TLFS section 5.2.1 specifies that when there are multiple overlays for a
+    /// given page, the order in which they are applied is
+    /// implementation-defined. For simplicity
```
hawkw, line 356: for simplicity...?

hawkw (commented):
I think I like this approach: treating the set of overlays as ordered by "priority" based on what kind of overlay they are, and considering whatever is the highest-priority overlay in the set the "active" one feels like a nicer design than the previous approaches of having a separate field for tracking the active overlay, and having activeness be part of the map key. It's nice that (unless we screw up and reorder the overlay kind variants by mistake) we can deterministically reason about how we select the active overlay based on what kind of overlay it is, rather than the order that the overlays were created.

I also like the optimizations to reduce how many bytes we have to send during migration; in particular, I think you're almost certainly right that guests will likely be using zeroed pages for overlays most of the time and it seems goofy to migrate a bunch of zeros, so that's a good case to optimize for.

I assume that the rest of the migration plumbing will be added in this branch if we decide to actually go forwards with it?

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Fixed in eaa6f94.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Fixed in 9ecba4e.

lib/propolis/src/enlightenment/hyperv/overlay.rs
gjcolombo: Changed the verbs to "export" and "import" in 9ecba4e (in part to reflect that these are really only expected to be used during a migration).

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -508,6 +618,34 @@ impl OverlayManager {
...
+            .map(|(pfn, set)| {
+                (
+                    u64::from(*pfn),
+                    MigratedOverlaidGuestPage::from(&set.original_contents),
+                )
+            })
+            .collect()
+    }
+
+    /// Attempts
```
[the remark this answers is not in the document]
gjcolombo: Attempts!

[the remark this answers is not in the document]
gjcolombo: (Fixed in 9ecba4e.)

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -215,40 +300,69 @@ impl Drop for OverlayPage {
...
+impl std::fmt::Debug for OverlaidGuestPage {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::ZeroPage => write!(f, "ZeroPage"),
+            Self::Page(_) => {
+                f.debug_tuple("Page").field(&"<page redacted>").finish()
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
gjcolombo: Fixed in 578500d.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -149,13 +185,62 @@ impl std::fmt::Debug for OverlayContents {
...
+
+/// A kind of overlay page, annotated with any other information that may be
+/// needed to generate the page's contents.
+///
+/// WARNING: The relative ordering of variants in this enum must remain stable!
+/// This enum's `PartialOrd` impl determines the overlay the manager applies if
+/// a guest requests multiple overlays at the same PFN. While TLFS section 5.2.1
+/// allows the hypervisor to choose any ordering, it implies that once chosen
+/// for a particular set of overlays, the "top-most" overlay will not change
+/// unless an overlay is enabled, disabled, or moved.
```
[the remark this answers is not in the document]
gjcolombo: Fixed in ef98df6.

gjcolombo:
Thanks for the review! I've moved this out of draft since I think I'm inclined to move forward with it. (This will supersede #853.)

> I assume that the rest of the migration plumbing will be added in this branch if we decide to actually go forwards with it?

I'm not sure there's any additional plumbing to add, though I may well have missed something. I think the additional checks on the import path in #853 are covered implicitly in this design:

- Hypercall MSR says the overlay should be enabled, but there's no record for that PFN in the exported page contents: this will produce a length mismatch when importing those contents
- Hypercall MSR says the overlay shouldn't be enabled, but there's an entry for it in the exported page contents: same as above
- Source and target disagree on the PFN at which the hypercall overlay is enabled: this should also be caught by original-page import. The target will always apply the hypercall overlay exactly where the source told it to; if the source's overlay manager and MSR disagree about where that overlay is actually located (yikes!), import will fail because the target's manager won't find a set into which to import the original page contents

The `export` and `import` routines on `OverlayManager` in #853 are obsoleted by their counterparts in this PR.

The other big difference between this and #853 is that this PR doesn't represent the original overlay contents as a separately-versioned type with its own `Schema` impl. The risk of doing that is that if the overlay contents' format ever changes, we'll need to rev the entire `HyperVEnlightenmentV1` structure instead of just a component thereof. I think that's OK for a couple of reasons:

- the format of the map doesn't seem that likely to change: we're just mapping from PFNs to saved page contents and not encoding anything about the structure of the overlay manager's tables
- because only the Hyper-V stack exports/imports overlay state, we're not really limiting the number of things we have to update by giving it its own schema (cf. something like device-agnostic PCI state, where we want to be able to update the schema without having to revise the device state structures for every PCI device type in the system)

Let me know if you had something else in mind here.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -215,40 +323,60 @@ impl Drop for OverlayPage {
...
-    active: OverlayKind,
+    original_contents: OverlaidGuestPage,
 
-    /// The overlays that are waiting to be applied to this PFN.
-    pending: BTreeMap<OverlayKind, OverlayContents>,
+    /// The set of overlays that have been requested at this PFN.
+    ///
+    /// TLFS section 5.2.1 specifies that when there are multiple overlays for a
+    /// given page, the order in which they are applied is
+    /// implementation-defined. Here, the overlay with the lowest priority (as
```
gjcolombo, line 368: nit: might be clearer to say the "lowest *numerical* priority"

gjcolombo:
I've pushed one more commit, a0690e4, that takes a still somewhat different approach to the migration problem (but tries to keep some of the goodies from elsewhere in the PR).

The idea is to notice that the Hyper-V manager is now reconstituting *pending* overlays from their "parameters" during a migration--i.e., instead of sending a pending overlay's pending contents in the raw, it now just recreates the overlay from scratch on the target. a0690e4 extends this to *active* overlays: when the Hyper-V manager pauses, it drops all its active overlays, allowing all the original guest memory contents to be sent in the RAM transfer phase instead of being serialized and sent in the device state phase. If migration succeeds, the target will recreate all necessary overlays during device state import; if it fails, the source will reapply them when the enlightenment resumes.

The downside of this change is that it strictly requires vCPUs to pause before devices (otherwise a vCPU running after the enlightenment has paused can read from a page whose overlay is "missing"). Fortunately, this is already the ordering we use in both propolis-server and propolis-standalone. If we decide much later that we want to allow (some) devices to pause before vCPUs have stopped, we can always break pause into "before vCPUs" and "after vCPUs" phases.

hawkw:
> I'm not sure there's any additional plumbing to add, though I may well have missed something. I think the additional checks on the import path in #853 are covered implicitly in this design:

Nope, I just scrolled past the changes to the migration functions and thought you hadn't done them yet --- my bad. Disregard that comment from last night.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -149,24 +132,92 @@ impl std::fmt::Debug for OverlayContents {
...
+}
+
+impl OverlayKind {
+    /// Returns the "priority" of this page relative to overlay pages of other
+    /// kinds. For each PFN the page with the lowest priority value is active.
+    ///
+    /// WARNING: The priorities of existing overlay kinds should not be changed,
+    /// since this can cause a PFN's active overlay to change if a VM is
+    /// migrated from a Propolis using one ordering to a Propolis using a
+    /// different ordering.
```
hawkw, line 171: it occurs to me that we _could_ have a test that writes out the overlay kind priorities to a file and asserts the existing ones haven't changed, or something. not sure if this is worth bothering with.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -149,24 +132,92 @@ impl std::fmt::Debug for OverlayContents {
...
+        match self {
+            Self::HypercallReturnNotSupported => OverlayContents(Box::new(
+                super::hypercall::hypercall_page_contents(),
+            )),
+            #[cfg(test)]
+            Self::Test { index: _, fill } => {
+                OverlayContents(Box::new([*fill; PAGE_SIZE]))
+            }
+        }
+    }
```
hawkw, line 200: lovely, i like this factoring.

lib/propolis/src/enlightenment/hyperv/overlay.rs
pfmooney: A guest stacking overlays is considered bad behavior, and the TLFS basically describes it similar to UB, where the observed result is largely up to the hypervisor.  I don't think we need to bend over backwards to provide a good experience for guests doing such things.

lib/propolis/src/lifecycle.rs
```diff
@@ -51,13 +51,21 @@ pub trait Lifecycle: Send + Sync + 'static {
     /// The device is not required to finish pausing inline. Instead, its
     /// implementation of [`Lifecycle::paused`] should return a future that
     /// completes only when the device is paused.
+    ///
+    /// WARNING: This function may only be called after pausing all of a VM's
```
pfmooney, line 55: Thanks for adding the docs here.  The context is definitely nice for implementers.

pfmooney (approved):
I like how this looks from a high level, but have not done more thorough review.

lib/propolis/src/enlightenment/hyperv/overlay.rs
```diff
@@ -149,24 +132,92 @@ impl std::fmt::Debug for OverlayContents {
...
+            Self::Test { .. } => u16::MAX,
+        };
+
+        let low: u16 = match self {
+            #[cfg(test)]
+            Self::Test { index, .. } => *index as u16,
+            _ => 0,
+        };
+
+        ((high as u32) << 16) | (low as u32)
```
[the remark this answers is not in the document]
gjcolombo: I doubt it, to be honest--this just seemed like a convenient way to set up the nesting.

PR #864, opened 2025-02-20
various new 1.85 clippy lints

iximeow:
the packed structs almost certainly should be repr(C) (why does that default to repr(Rust)???)

the indentation change is unfortunate and doc_overindented_list_items hasn't really convinced me that this *would* introduce broken spacing in rendered docs, but i don't have it in me to allow the lint for a few lines like this.

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -1020,7 +1020,7 @@ impl MachineInitializer<'_> {
             .map(|ident| {
                 match ident.eax & 0xf00 {
                     // If family ID is 0xf, extended family is added to it
-                    0xf00 => (ident.eax >> 20 & 0xff) + 0xf,
+                    0xf00 => ((ident.eax >> 20) & 0xff) + 0xf,
                     // ... otherwise base family ID is used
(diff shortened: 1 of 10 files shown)


gjcolombo (approved):
Looks good. If you haven't already, it might be worthwhile to spin up a propolis-server and connect to its VNC port just to smoke-test the `repr` changes. Thanks for taking care of this!

iximeow:
@gjcolombo it turns out i'd been on the beta channel locally, so i'd updated to 1.86 (!) which does not have a lint present in 1.85 (???) about parenthesizing bitwise operations explicitly, so there are a few more changes here. clippy suggestions are all making the expected operations explicit so it's all pretty straightforward, but mind taking another look? (sorry for the runaround)

and yeah, i'll check out VNC while this goes through CI, good call!

gjcolombo (approved):

hawkw (approved):

iximeow:
update: after a slight adventure starting `propolis-server`, VNC looks all good (modulo https://github.com/oxidecomputer/propolis/issues/868, which after checking is also working as expected). reasonable framebuffer, mouse and keyboard still good.

PR #866, opened 2025-02-20
propolis-server should not crash when failing to start a VM

### LINKED ISSUE #838: failing to start a VM produces a "cannot drop a runtime in a context where blocking is not allowed" panic
Repro steps: start a VM with an invalid configuration (e.g. an attached NIC that refers to a nonexistent host VNIC).

Expected: initialization fails and returns a nice error message to the instance_ensure caller.

Observed: Propolis panics with a tokio "can't drop a runtime in this context" error.

---

The dropped runtime is the per-VMM runtime that gets created during VM initialization and used to initialize the VMM, run per-component tasks, and the like. This doesn't happen during orderly VM shutdown because `Vm::complete_rundown` calls `shutdown_background` on this runtime.

iximeow:
tripped over this while trying to figure out how it was that `propolis-server` is now failing to create VMs with a toml that was working fine just an hour ago. what was a crash with stdout having an  `INFO ... migration: InstanceMigrateStatusResponse { migration_in: None, migration_out: None}` (silently truncated early due to panic) now very helpfully does _not_ kill `propolis-server` and instead has an `ERRO ... failed to add low memory region: Not enough space`.

Fixes #838

### DIFF
--- a/bin/propolis-server/src/lib/vm/ensure.rs
+++ b/bin/propolis-server/src/lib/vm/ensure.rs
@@ -290,7 +290,26 @@ impl<'a> VmEnsureNotStarted<'a> {
                     kernel_vm_paused: false,
                 })
             }
-            Err(e) => Err(self.fail(e).await),
+            Err(e) => {
+                // In the happy path, `vmm_rt` is moved to the caller inside
+                // `VmEnsureObjectsCreated`. Here, we would `self.fail()` and
+                // implicitly drop `vmm_rt` before returning up the stack. This
+                // would result in a Tokio panic like `Cannot drop a runtime in
+                // a context where blocking is not allowed`, killing
+                // `propolis-server` and even resulting in API clients just
+                // losing connections instead of getting well-formed error
+                // responses back.
+                //
+                // So instead, spawn_blocking() to get a context where blocking
+                // is allowed, drop the `vmm_rt` there, and continue on with
+                // whatever unfortunate failure has transpired.
+                tokio::task::spawn_blocking(move || {
+                    std::mem::drop(vmm_rt);
+                })
+                .await
+                .expect("can drop vmm_rt");
+                Err(self.fail(e).await)
+            }
         }
     }
 
--- a/bin/propolis-server/src/lib/vm/ensure.rs
+++ b/bin/propolis-server/src/lib/vm/ensure.rs
@@ -256,9 +256,6 @@ impl<'a> VmEnsureNotStarted<'a> {
         // When the runtime is returned to this thread, it must not be dropped.
         // That means that the path between this result and returning an
         // `Ok(VmEnsureObjectsCreated)` must be infallible.
-        //
-        // `VmEnsureObjectsCreated` (and later state transitions) take care to
-        // `shutdown_background` the runtime.
         let result: InitResult = tokio::task::spawn_blocking(move || {
             // Create the runtime that will host tasks created by
             // VMM components (e.g. block device runtime tasks).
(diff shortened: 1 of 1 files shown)


hawkw (approved):
bit of a bummer we have to do this, but yeah, seems good to me!

bin/propolis-server/src/lib/vm/ensure.rs
```diff
@@ -226,7 +226,26 @@ impl<'a> VmEnsureNotStarted<'a> {
                     kernel_vm_paused: false,
                 })
             }
-            Err(e) => Err(self.fail(e).await),
+            Err(e) => {
```
gjcolombo, line 229: One minor caveat: this works in part because in today's code, no one ever actually drops a `VmEnsureObjectsCreated` (both code paths that instantiate one more or less immediately call `ensure_active` on it, which infallibly moves the runtime to `Vm::make_active`; after that point it's guaranteed that someone will call `shutdown_background` on the runtime before it gets dropped).

I think fixing this might be mildly involved (though maybe there's an elegant solution I haven't seen yet), so I don't think I'd hold up the whole PR for it, but it might be good to note this property in a comment somewhere around here.

gjcolombo (approved):

bin/propolis-server/src/lib/vm/ensure.rs
iximeow: so, from this nudging and our chat, i think being more intentional about creating and initializing tasks on `vmm_rt` is pretty workable! that's how i've got the PR now. at least this leaves us with a clearly-safe way to handle fallibility on the path to a `VmEnsureObjectsCreated`.

if we decided to do something fallible between a `VmEnsureObjectsCreated` and an `ensure_active`, well, that would be pretty unfortunate.

bin/propolis-server/src/lib/vm/ensure.rs
```diff
@@ -236,39 +236,70 @@ impl<'a> VmEnsureNotStarted<'a> {
...
+        // the fledgling VMM runtime itself. Dropping a Tokio runtime on a
+        // worker thread in a Tokio runtime will panic, so do all init in a
+        // `spawn_blocking` where this won't be an issue.
+        //
+        // When the runtime is returned to this thread, it must not be dropped.
+        // That means that the path between this result and returning an
+        // `Ok(VmEnsureObjectsCreated)` must be infallible.
+        //
+        // `VmEnsureObjectsCreated` (and later state transitions) take care to
+        // `shutdown_background` the runtime.
```
gjcolombo, line 261: nit: this isn't quite right, I think; once you have a `VmEnsureActive` you are assured that the runtime will be shut down in the background when it's dropped. You get that by calling `VmEnsureObjectsCreated::ensure_active`.

We might want to put a `WARNING:` comment on `VmEnsureObjectsCreated` indicating that it can't currently be dropped (or maybe just add a Drop impl for it that calls `shutdown_background`, though I would have to convince myself that this won't leak any tasks).

gjcolombo (approved):
Thanks for going the additional mile here!

bin/propolis-server/src/lib/vm/ensure.rs
iximeow: oh, i'd conflated `ActiveVm` and `VmEnsureActive`, yeah you're right.

i'll go with a stern warning. a Drop impl that calls `shutdown_background` _would_ immediately leak any object's tasks, but they're all also cancelled so they should stop "soon", so i think it's fine. but being in a situation where we're dropping `VmEnsureObjectsCreated` itself feels like a bug. _and_ if you innocently destructure `VmEnsureObjectsCreated` outside `ensure_active` you still have to be careful about the runtime again..

iximeow: (https://github.com/oxidecomputer/propolis/pull/866/commits/7053acf3b599981df0f471107bf18060145f329a)

PR #869, opened 2025-02-21
mock-server: add single-step API

### LINKED ISSUE #858: mock-server: want manual control of state transitions
in order to simulate cases where `propolis-server` gets stuck or takes a long time between instance state transitions (e.g. #841), it would be nice if there was a way for tests to control the state transitions simulated by the mock server. this incldues including preventing them from occurring or delaying them until the test requests that they proceed, or injecting failures. this would be useful for testing sled-agent interactions with Propolis, such as oxidecomputer/omicron#7548

hawkw:
For certain test scenarios, the `propolis-mock-server` ought to have a mechanism
for manual control of the mocked instance's progress through the state machine.
In particular, this is necessary for testing changes like
oxidecomputer/omicron#7548, which adds a timeout tracked by the sled-agent when
an instance is stopped. If Propolis is stuck and cannot progress, the sled-agent
will forcefully terminate it after that timeout...but testing this requires a
way to make the mock Propolis pretend to be stuck.

This commit adds the following new endpoints to the mock server which are not
part of the real `propolis-server` API:

- `PUT /mock/mode`: sets a mock mode, either `Run` (the normal behavior), or
  `SingleStep`, where state transitions only ocur when asked for by the test.
- `GET /mock/mode`: returns whether or not we are single-steppy
- `PUT /mock/step`: advances to the next queued generation

Testing: I've written [a test in Omicron][1] that uses this, and I can make
the mock propolis get wedged in the correct place. So that's nice.

Closes #858 

[1]: https://github.com/oxidecomputer/omicron/compare/eliza/single-steppy?expand=1

### DIFF
--- a/bin/mock-server/src/lib/lib.rs
+++ b/bin/mock-server/src/lib/lib.rs
@@ -22,6 +22,7 @@ use tokio_tungstenite::WebSocketStream;
 
 mod api_types;
 use api_types::types::{self as api, InstanceEnsureRequest};
+pub use api_types::MockMode;
 
 #[derive(Debug, Eq, PartialEq, Error)]
 pub enum Error {
@@ -452,7 +453,7 @@ async fn instance_serial_history_get(
 }]
 async fn mock_mode_get(
     rqctx: RequestContext<Arc<Context>>,
-) -> Result<HttpResponseOk<api_types::MockMode>, HttpError> {
+) -> Result<HttpResponseOk<MockMode>, HttpError> {
     let instance = rqctx.context().instance.lock().await;
     let instance = instance.as_ref().ok_or_else(|| {
         HttpError::for_internal_error(
@@ -460,9 +461,9 @@ async fn mock_mode_get(
         )
     })?;
     let mode = if instance.state_watcher_rx.borrow().single_step_gen.is_some() {
-        api_types::MockMode::SingleStep
+        MockMode::SingleStep
     } else {
-        api_types::MockMode::Run
+        MockMode::Run
     };
     Ok(HttpResponseOk(mode))
 }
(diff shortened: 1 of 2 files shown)


bin/mock-server/src/lib/lib.rs
```diff
@@ -275,11 +285,42 @@ async fn instance_state_monitor(
...
+                    slog::info!(
+                        rqctx.log,
+                        "instance state monitor: wait for single step...";
+                        "request_gen" => gen,
+                        "current_gen" => g,
+                    );
+                    None
+                }
+                // Otherwise, if we have stepped to the requested generation, or
+                // if we are not in single-step mode, just return the current
```
gjcolombo, line 310: This is subtly different from the real server's behavior, I think--should the generation number "floor" still apply in automatic mode?

gjcolombo (commented):
The mechanics of this generally look fine--just one question about a possible semantic difference between the mock and real servers.

bin/mock-server/src/lib/lib.rs
gjcolombo: Ah, I've misread this--if the requested `gen` hasn't been published yet, then the `get` on line 312 will return `None` and we'll end up getting the expected behavior.

I think there is still a (different) subtle difference here: if you ask this API for generation J, and the state machine has advanced to generation K > J, then the mock will return the state as it was at generation J, but the "real" state machine will return whatever data is latest. But I think this kind of difference is OK to have if it makes the test double more useful when writing sled-agent tests. (If/when we refactor so that the mock state machine is based on propolis-server's state driver, we might want to put this behavior into a different API instead of changing the semantics of `instance-state-monitor`, but that's a problem for another day.)

gjcolombo (approved):

Issue #841, opened 2025-01-25
want Crucible VCR replacement to be able to proceed before a VM has fully started


gjcolombo:
See https://github.com/oxidecomputer/omicron/pull/7398 for more background.

## Background

Instances with Crucible disks need to "activate" their connections to their Crucible downstairs servers before those servers will accept I/O requests. Each Crucible backend needs to be able to reach all three downstairs specified in the instance spec in order to activate. If a server is unreachable, the startup process hangs until it becomes reachable.

The control plane provides a solution to this problem. If a downstairs becomes degraded, it can set up a new downstairs to take its place and amend the volume configurations (i.e. the `VolumeConstructionRequest`s, or VCRs) of any affected disks so that they refer to the new downstairs. When a Crucible client connects to these three downstairs, they will observe that repair is needed, and the two healthy servers will send their data over to the newly-created server.

This process requires a Crucible client to coordinate the repair. When a disk is not attached to an instance, or its instance is stopped, the control plane spins up a Crucible pantry job that spawns such a client. But when a disk is attached to an active instance, there's already a Crucible client in the instance's Propolis process, so the control plane [sends a request directly to Propolis](https://github.com/oxidecomputer/propolis/blob/a3e035a315b945d5a8d0010c81b49e376c19e077/bin/propolis-server/src/lib/server.rs#L600) to ask it to send new volume configuration to its existing upstairs.

## The problem

Propolis queues requests to change a VCR to its state driver task. This is meant to minimize our exposure to tricky concurrency problems: there's no need to worry about, say, changing Crucible configuration during a migration if migrations and Crucible configuration changes are serialized by the state driver queue!

The problem here is that VM startup is *also* a state driver operation. While Crucible's `Volume` abstraction supports replacing a VCR on a volume that is still activating, if the state driver is wedged waiting for an activation to complete, there's no one available to issue that request.[^1]

The request here is for a better approach to handling VCR replacement that avoids concurrency pitfalls but still allows VCR replacements to go through when an instance is starting or stopping.

Note that this is not a problem for disk *snapshot* requests, because those don't get queued to the state driver. This is because snapshot requests don't mutate the VM's configuration (in a way that might conflict with a consumer of that configuration, like VM startup or VM live migration), whereas VCR replacement requests do change how the VM is configured.

[^1]: In fact the state driver's queuing logic currently rejects requests to queue a VCR replacement when a VM is Starting, but if it didn't do that the requests would just get stuck on its queue in this scenario.

PR #872, opened 2025-02-24
phd: add smoke test for VCR replacement

gjcolombo:
Add a smoke test for Crucible VCR replacement:

- Add a `CrucibleDisk` function to get a disk's current VCR.
- Add a `TestVm` framework to send a VCR replacement request.
- Fix a couple of bugs in Crucible disk setup:
  - The `id` field in the VCR's `CrucibleOpts` needs to be the same in the old and new VCRs, so use the disk ID for it instead of calling `Uuid::new_v4()` every time the VCR is generated.
  - When using a `Blank` disk source, properly account for the fact that the disk source size is given in bytes, not gibibytes.

Also add a couple of bells and whistles to allow this test to be transformed into a test of VCR replacement during VM startup:

- Make PHD's `VmSpec` type a public type and amend `Framework` to allow tests to create a VM from a spec. This gives tests a way to access a config's Crucible disks before actually launching a VM (and sending an instance spec to Propolis).
- Reorganize the `CrucibleDisk` types to wrap the disk innards in a `Mutex`, allowing them to be mutated through the `Arc` references that tests get. This will eventually be used to allow tests to override the downstairs addresses in a disk's VCRs before launching a VM that uses that disk, which will be used to test fixes for #841. In the meantime, use the mutex to protect the VCR generation number, which no longer needs to be an `AtomicU64`.

### DIFF
--- a/phd-tests/framework/src/disk/crucible.rs
+++ b/phd-tests/framework/src/disk/crucible.rs
@@ -8,7 +8,7 @@ use std::{
     net::{Ipv4Addr, SocketAddr, SocketAddrV4},
     path::{Path, PathBuf},
     process::Stdio,
-    sync::atomic::{AtomicU64, Ordering},
+    sync::Mutex,
 };
 
 use anyhow::Context;
@@ -62,9 +62,70 @@ impl Drop for Downstairs {
 /// An RAII wrapper around a Crucible disk.
 #[derive(Debug)]
 pub struct CrucibleDisk {
-    /// The name to use in instance specs that include this disk.
     device_name: DeviceName,
+    inner: Mutex<Inner>,
+}
 
+impl CrucibleDisk {
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn new(
+        device_name: DeviceName,
+        min_disk_size_gib: u64,
+        block_size: BlockSize,
+        downstairs_binary_path: &impl AsRef<std::ffi::OsStr>,
+        downstairs_ports: &[u16],
+        data_dir_root: &impl AsRef<Path>,
+        read_only_parent: Option<&impl AsRef<Path>>,
+        guest_os: Option<GuestOsKind>,
+        log_mode: ServerLogMode,
+    ) -> anyhow::Result<Self> {
+        Ok(Self {
+            device_name,
+            inner: Mutex::new(Inner::new(
+                min_disk_size_gib,
+                block_size,
+                downstairs_binary_path,
+                downstairs_ports,
+                data_dir_root,
+                read_only_parent,
+                guest_os,
+                log_mode,
+            )?),
+        })
+    }
+
+    /// Obtains the current volume construction request for this disk.
+    pub fn vcr(&self) -> VolumeConstructionRequest {
+        self.inner.lock().unwrap().vcr()
+    }
+
+    /// Sets the generation number to use in subsequent calls to create a
+    /// backend spec for this disk.
+    pub fn set_generation(&self, generation: u64) {
+        self.inner.lock().unwrap().generation = generation;
+    }
+}
+
+impl super::DiskConfig for CrucibleDisk {
+    fn device_name(&self) -> &DeviceName {
+        &self.device_name
+    }
+
+    fn backend_spec(&self) -> ComponentV0 {
+        self.inner.lock().unwrap().backend_spec()
+    }
+
+    fn guest_os(&self) -> Option<GuestOsKind> {
+        self.inner.lock().unwrap().guest_os()
+    }
+
+    fn as_crucible(&self) -> Option<&CrucibleDisk> {
+        Some(self)
+    }
+}
+
+#[derive(Debug)]
+struct Inner {
     /// The UUID to insert into this disk's `VolumeConstructionRequest`s.
     id: Uuid,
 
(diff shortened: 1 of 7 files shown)


phd-tests/framework/src/test_vm/mod.rs
```diff
@@ -603,6 +604,33 @@ impl TestVm {
...
+            .instance_issue_crucible_vcr_request()
+            .id(disk.device_name().clone().into_backend_name().into_string())
+            .body(body)
+            .send()
+            .await?;
+
+        anyhow::ensure!(
+            response_value.status().is_success(),
+            "VCR replacement request returned an error value: \
+            {response_value:?}"
```
hawkw, line 628: might it be worth including the VCR that was sent in this?

phd-tests/tests/src/crucible/smoke.rs
```diff
@@ -82,3 +86,34 @@ async fn shutdown_persistence_test(ctx: &Framework) {
...
+        DiskBackend::Crucible {
+            min_disk_size_gib: 1,
+            block_size: BlockSize::Bytes512,
+        },
+        5,
+    );
+
+    let spec = config.vm_spec(ctx).await?;
+    let disks = spec.get_crucible_disks();
+    let disk = disks[0].as_crucible().unwrap();
```
hawkw, line 111: `get_crucible_disks` returns the disks where `as_crucible()` returns `Some`...why not change `get_crucible_disks` to use `filter_map(|d| d.as_crucible())` and just return the crucible disks, so we don't have to unwrap here?

not a huge deal either way.

phd-tests/framework/src/disk/crucible.rs
```diff
@@ -62,9 +62,70 @@ impl Drop for Downstairs {
...
+
+    fn as_crucible(&self) -> Option<&CrucibleDisk> {
+        Some(self)
+    }
+}
 
+#[derive(Debug)]
+struct Inner {
     /// The UUID to insert into this disk's `VolumeConstructionRequest`s.
     id: Uuid,
```
hawkw, line 130: since the UUID will not change across VCRs, perhaps it ought _not_ be inside the mutex, forcing it to be immutable?

not actually a big deal as i don't _think_ we need to access it anywhere that we could avoid taking a lock, but it might express intent a bit better?

hawkw (approved):
this is pretty straightforward! i had a couple teensy nitpicks.

phd-tests/tests/src/crucible/smoke.rs
iximeow: the borrows get a bit funky 'cause `as_crucible()` discards the Arc from whence it came, so that makes `get_crucible_disks` only suitable when you know you don't want to get a new reference to the disk, i assume Greg's thinking was that `Arc<dyn DiskConfig>` is more like the canonical representation and `CrucibleDisk` is the last-mile thing you can turn it into when you really want to mess with a disk?

(both testing my read of the pieces and kicking the tires on your suggestion here)

my 2c here is that `disks[0].as_crucible()` seems ambiguous in a way that i think the comment above describes, so i'm kinda confused: if the boot disk is also backed by Crucible, wouldn't `disks[0]` pick the wrong one which the comment describes VCR replacements being disallowed for?

so my thought is, `spec.get_disk("vcr-replacement-target").unwrap().as_crucible().unwrap()` probably reads less funky and avoids the ambiguity. we'd just ... need a `fn get_disk` :D

iximeow (commented):
+1 to Eliza's comments, plus maybe bonus wrinkle in the new test?

phd-tests/framework/src/test_vm/mod.rs
gjcolombo: Added as a log line in 2907f70.

phd-tests/framework/src/disk/crucible.rs
gjcolombo: I've pulled some of this stuff out of `Inner` in 1f04b2f.

phd-tests/tests/src/crucible/smoke.rs
gjcolombo: I like the `get_disk` formulation; added in 7720200. (Technically, the `VmConfig` and `VmSpec` could/should ensure that the disks in a spec have unique device names, but I've left that improvement for another PR.)

leftwo (approved):
I think any questions I have are more applicable to #873, so I'll go review that now :)

phd-tests/tests/src/crucible/smoke.rs
```diff
@@ -82,3 +86,36 @@ async fn shutdown_persistence_test(ctx: &Framework) {
...
     assert_eq!(lsout, "foo.bar");
 }
+
+#[phd_testcase]
+async fn vcr_replace_test(ctx: &Framework) {
+    let mut config = ctx.vm_config_builder("crucible_vcr_replace_test");
+
+    // Create a blank data disk on which to perform VCR replacement. This is
+    // necessary because Crucible doesn't permit VCR replacements for volumes
+    // whose read-only parents are local files (which is true for artifact-based
```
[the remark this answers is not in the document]
gjcolombo: We're tracking this with https://github.com/oxidecomputer/crucible/issues/1658.

PR #873, opened 2025-02-26
server: improve behavior of starting VMs that are waiting for Crucible activation

### LINKED ISSUE #841: want Crucible VCR replacement to be able to proceed before a VM has fully started

gjcolombo:
Modify the state driver's VM startup procedure to allow the driver to process Crucible volume configuration changes while block backends are being activated. This fixes a livelock that occurs when starting a VM with a Crucible VCR that points to an unavailable downstairs: the unavailable downstairs prevents Crucible activation from proceeding; Nexus sends a corrected VCR that, if applied, would allow the upstairs to activate; but the state driver never applies the new VCR because it's blocked trying to activate using the broken VCR.

Since the state driver can now dequeue VM state change requests during startup, also teach it to abort startup if a stop request is received while block backends are starting. This is very slightly spicy, because it can cancel a block backend startup future in a way that was not possible before, but (a) the only affected backend is the Crucible backend, and (b) there should be no requests in flight anyway because, if this case is reached, the VM's vCPUs have not started.

Add PHD coverage of the new behaviors:

- Modify the PHD VCR replacement smoke test to check that start requests with a bad set of Crucible targets can be unblocked by replacing a bad target with a good target.
- Add a server state machine test that checks that a VM that is blocked waiting for Crucible activation can be explicitly stopped.

To assist with this, add a feature to PHD Crucible disks that allows a test to specify that a disk's generated VCRs should contain an invalid downstairs IP address. Starting a VM with a disk configured this way will cause activation to block until the disk's VCR is replaced with a corrected VCR.

Tests: cargo test, PHD w/Alpine and Debian 11 guests.

Fixes #841.

### DIFF
--- a/bin/propolis-server/src/lib/vm/request_queue.rs
+++ b/bin/propolis-server/src/lib/vm/request_queue.rs
@@ -88,10 +88,6 @@ impl std::fmt::Debug for StateChangeRequest {
 pub enum ComponentChangeRequest {
     /// Attempts to update the volume construction request for the supplied
     /// Crucible volume.
-    ///
-    /// TODO: Due to https://github.com/oxidecomputer/crucible/issues/871, this
-    /// is only allowed once the VM is started and the volume has activated, but
-    /// it should be allowed even before the VM has started.
     ReconfigureCrucibleVolume {
         /// The ID of the Crucible backend in the VM's Crucible backend map.
         backend_id: SpecKey,
@@ -356,7 +352,6 @@ impl ExternalRequestQueue {
                     "request" => ?request,
                     "reason" => %reason
                 );
-
                 return Err(reason);
             }
         }
@@ -795,7 +790,6 @@ mod test {
     fn mutation_disallowed_after_stopped() {
         let mut queue =
             ExternalRequestQueue::new(test_logger(), InstanceAutoStart::Yes);
-
         queue.notify_request_completed(CompletedRequest::Start {
             succeeded: true,
         });
--- a/bin/propolis-server/src/lib/vm/state_driver.rs
+++ b/bin/propolis-server/src/lib/vm/state_driver.rs
@@ -109,7 +109,7 @@ use crate::{
         destination::DestinationProtocol, source::SourceProtocol, MigrateRole,
     },
     spec::StorageBackend,
-    vm::state_publisher::ExternalStateUpdate,
+    vm::{state_publisher::ExternalStateUpdate, BlockBackendMap},
 };
 
 use super::{
@@ -139,6 +139,12 @@ enum HandleEventOutcome {
 pub(super) enum VmStartReason {
     MigratedIn,
     ExplicitRequest,
+}
+
+enum VmStartOutcome {
+    Succeeded,
+    Failed,
+    Aborted,
 }
 
 /// A kind of event the state driver can handle.
@@ -195,12 +201,17 @@ impl InputQueue {
         }
     }
 
-    /// Waits for an event to arrive on the input queue and returns it for
-    /// processing.
+    /// Waits for a new event to arrive on one of the queue's sub-queues and
+    /// dispatches it for processing.
     ///
-    /// External requests and guest events are stored in separate queues. If
-    /// both queues have events when this routine is called, the guest event
-    /// queue takes precedence.
+    /// The sub-queues, listed here in priority order, are:
+    ///
+    /// - Guest events: These are signals raised from the VM's vCPUs and
+    ///   devices (e.g. a request to reboot or halt the VM arising from a vCPU
+    ///   asserting a virtual chipset signal).
+    /// - External requests: These are state change requests received via the
+    ///   server API. See [`super::request_queue`] for more details about how
+    ///   these requests are queued.
     ///
     /// # Synchronization
     ///
@@ -478,10 +489,10 @@ impl StateDriver {
         info!(self.log, "state driver launched");
 
         let final_state = if migrated_in {
-            if self.start_vm(VmStartReason::MigratedIn).await.is_ok() {
-                self.event_loop().await
-            } else {
-                InstanceState::Failed
+            match self.start_vm(VmStartReason::MigratedIn).await {
+                VmStartOutcome::Succeeded => self.event_loop().await,
+                VmStartOutcome::Failed => InstanceState::Failed,
+                VmStartOutcome::Aborted => InstanceState::Destroyed,
             }
         } else {
             self.event_loop().await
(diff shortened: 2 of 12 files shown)


bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -520,20 +534,162 @@ impl StateDriver {
...
+        // be buffered and handled after the rest of the VM has started.
+        async fn start_block_backends(
+            log: slog::Logger,
+            backends: BlockBackendMap,
+        ) -> anyhow::Result<()> {
+            for (name, backend) in backends {
+                info!(log, "starting block backend {}", name);
+                let res = backend.start().await;
+                if let Err(e) = &res {
+                    error!(log, "startup failed for {}: {:?}", name, e);
```
hawkw, line 587: nit, take it or leave it: perhaps  we should include
```suggestion
                    error!(log, "startup failed for block backend {}: {:?}", name, e);
```
since we state it explicitly in the other log line...

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -520,20 +534,162 @@ impl StateDriver {
...
+        let block_backends = objects.block_backend_map().clone();
+        let block_backend_fut =
+            start_block_backends(self.log.clone(), block_backends);
+        tokio::pin!(block_backend_fut);
+
+        // Drop the VM object lock before proceeding to allow other API calls
+        // that simply read the VM to make progress. Again, note that the set of
+        // objects being started still can't change, not because the lock is
+        // held, but because the only entity that can change them is the current
+        // task, which can decide whether and how to buffer incoming requests.
```
hawkw, line 604: Thank you for documenting this, makes sense to me!

phd-tests/tests/src/crucible/smoke.rs
```diff
@@ -111,11 +112,22 @@ async fn vcr_replace_test(ctx: &Framework) {
...
+    disk.enable_vcr_black_hole();
 
     let mut vm = ctx.spawn_vm_with_spec(spec, None).await?;
     vm.launch().await?;
-    vm.wait_to_boot().await?;
 
+    // The VM is expected not to reach the Running state. Unfortunately, there's
+    // no great way to test that this is never going to happen; as a best-effort
+    // alternative, wait for a short while and assert that the VM doesn't reach
+    // Running in the timeout interval.
```
hawkw, line 123: Urgh...

phd-tests/tests/src/crucible/smoke.rs
```diff
@@ -88,8 +88,9 @@ async fn shutdown_persistence_test(ctx: &Framework) {
 }
 
 #[phd_testcase]
-async fn vcr_replace_test(ctx: &Framework) {
-    let mut config = ctx.vm_config_builder("crucible_vcr_replace_test");
```
hawkw, line 92: by replacing the previous VCR replacement test where the VM has already started, are we no longer testing some code paths? would it make sense to have kept the prior test as well as adding the new one for replacement during a start?

phd-tests/framework/src/disk/crucible.rs
```diff
@@ -107,6 +121,39 @@ impl CrucibleDisk {
...
+    pub fn enable_vcr_black_hole(&self) {
+        info!(disk = self.device_name.as_str(), "enabling vcr black hole");
+
+        // 100::/64 is the IPv6 discard prefix (per RFC 6666).
+        let address = SocketAddr::V6(SocketAddrV6::new(
+            Ipv6Addr::new(0x100, 0, 0, 0, 0, 0, 0, 0),
+            9000,
+            0,
+            0,
+        ));
```
hawkw, line 138: nice use of this :)

phd-tests/tests/src/crucible/smoke.rs
gjcolombo: I updated the test so that it attempts a second replacement after the VM starts running. That way we only have to wait for one guest boot sequence (and we also verify that we can replace VCRs more than once on a single instance).

phd-tests/tests/src/crucible/smoke.rs
```diff
@@ -125,9 +129,17 @@ async fn vcr_replace_during_start_test(ctx: &Framework) {
...
     disk.disable_vcr_black_hole();
     disk.set_generation(2);
     vm.replace_crucible_vcr(disk).await?;
-
     vm.wait_to_boot().await?;
+
+    assert_eq!(vm.get().await?.instance.state, InstanceState::Running);
+
+    // VCR replacements should continue to be accepted now that the instance is
+    // running.
```
hawkw, line 142: Good call adding this, thank you!

hawkw (approved):
New changes look great! I imagine you might want to wait for an approval from @leftwo as well?

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -211,6 +217,8 @@ impl InputQueue {
                 let mut guard = self.inner.lock().unwrap();
                 if let Some(guest_event) = guard.guest_events.pop_front() {
                     return InputQueueEvent::GuestEvent(guest_event);
+                } else if let Some(req) = guard.buffered_external.pop_front() {
```
mkeeter, line 220: Nit: the docstring should be updated to also include the `buffered_external` queue.

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -227,6 +235,13 @@ impl InputQueue {
         }
     }
 
+    /// Pushes an external request to the end of the buffered external request
+    /// list.
+    fn buffer_external_request(&self, req: ExternalRequest) {
+        let mut guard = self.inner.lock().unwrap();
+        guard.buffered_external.push_back(req);
```
mkeeter, line 242: Does this also need to ping the notifier?

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -520,20 +534,167 @@ impl StateDriver {
...
+                    //
+                    // It is possible, at least in theory, to drop the block
+                    // backend startup future (or cooperatively cancel it) and
+                    // then pause/halt the VM immediately. This requires block
+                    // backends' startup operations to be cancel-safe (in the
+                    // sense that it must be possible to pause and halt a
+                    // backend whose startup sequence was interrupted). Since
+                    // block backends aren't currently required to guarantee
+                    // this, just buffer the stop request.
+                    self.input_queue.buffer_external_request(req);
```
mkeeter, line 668: I'm worried about the interaction of this line and `self.input_queue.wait_for_next_event()` in the `select!` above.

Specifically, `wait_for_next_event()` will return events from the `buffered_external` queue, so I'm suspicious that if we send `ExternalRequest::Stop` while the machine is starting, we'll end up in a spin-loop of removing then re-adding the request to the `buffered_external` quee.

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -520,20 +534,167 @@ impl StateDriver {
...
+                    return res;
+                }
+
+                dequeued = self.input_queue.wait_for_next_event() => {
+                    dequeued
+                }
+            };
+
+            // The VM's vCPUs haven't been started yet, so there should be no
+            // way for the VM to produce any guest events to handle.
```
mkeeter, line 636: Is there a possibility for a race condition here?  I don't think so, because there are no yield points between starting the vCPUs and returning from the `block_backend_fut`, but want to double-check (and maybe suggest a comment to that effect).

phd-tests/framework/src/disk/crucible.rs
```diff
@@ -338,8 +386,11 @@ impl Inner {
     }
 
     fn vcr(&self, disk_id: Uuid) -> VolumeConstructionRequest {
-        let downstairs_addrs =
-            self.downstairs_instances.iter().map(|ds| ds.address).collect();
+        let downstairs_addrs = self
+            .downstairs_instances
+            .iter()
+            .map(|ds| ds.vcr_address())
```
mkeeter, line 392: Nit: I'm surprised Clippy hasn't suggested
```suggestion
            .map(Downstairs::vcr_address)
```
since it usually prefers point-free style.

phd-tests/tests/src/crucible/smoke.rs
```diff
@@ -88,8 +88,9 @@ async fn shutdown_persistence_test(ctx: &Framework) {
 }
 
 #[phd_testcase]
-async fn vcr_replace_test(ctx: &Framework) {
-    let mut config = ctx.vm_config_builder("crucible_vcr_replace_test");
+async fn vcr_replace_during_start_test(ctx: &Framework) {
```
mkeeter, line 91: Should we also have a test for stopping while starting, to exercise the `buffered_external` queue?

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: Fixed in bfb8b07.

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: Great catch here--the new test case I added for this in bfb8b07 fails due to this problem. That commit should also fix the issue (by waiting to self-request a stop until the attempt to start has actually finished).

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: Added a comment in bfb8b07: IIUC, this shouldn't race because vCPUs only start when the `block_backend_fut` branch of the `select!` is actually chosen, and once that happens the whole function returns, so if we're here then we haven't had an occasion to start the vCPUs yet.

phd-tests/framework/src/disk/crucible.rs
gjcolombo: Fixed in 65f862f.

phd-tests/tests/src/crucible/smoke.rs
gjcolombo: We absolutely should! Added one in bfb8b07, which immediately found several bugs that are also fixed in that commit.

gjcolombo:
@mkeeter: Thanks for the detailed review! I think I've addressed all your feedback in the latest push.

@hawkw: There are enough changes in the new commits that I'd appreciate another review from you when you have a moment. (In particular I've changed things so that Propolis now reports that a VM is "Starting" and not "Creating" before the component start sequence actually begins. IIRC, we discussed this in DMs recently and decided this would be OK to do, since Omicron collapses these two states into a single "Starting" VMM state anyway.)

gjcolombo:
Ran PHD, but forgot to run `cargo test` before pushing. I'll fix the request queue unit test to match the new expected behavior and push that tomorrow.

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -334,15 +328,30 @@ impl ExternalRequestQueue {
                 AllowedRequests { reboot: Disposition::Ignore, ..self.allowed }
             }
 
-            // Requests to stop the instance block other requests from being
-            // queued. Additional requests to stop are ignored for idempotency.
+            // Once the instance is asked to stop, further requests to change
+            // its state are ignored.
+            //
+            // Requests to change Crucible volume configuration can still be
+            // queued if they were previously alloewd. This allows the state
```
mkeeter, line 335:
```suggestion
            // queued if they were previously allowed. This allows the state
```

mkeeter:
@gjcolombo the new changes look pretty good!  I'm now wondering if `self_request` is necessary at all, since it only ever stores a single `ExternalRequest::Stop` – could we just push that to the normal `external_requests` queue instead?

bin/propolis-server/src/lib/vm/request_queue.rs
gjcolombo: Whoops. Fixed in e66e5deefd48.

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -520,20 +548,184 @@ impl StateDriver {
...
+                res = &mut block_backend_fut => {
+                    if res.is_ok() {
+                        let objects = &self.objects;
+                        objects.lock_exclusive().await.resume_vcpus();
+                        self.publish_steady_state(InstanceState::Running);
+
+                        // If a stop request was seen while in the startup
+                        // sequence, self-request a stop for the main event
+                        // loop to handle.
+                        if stopped_while_starting {
```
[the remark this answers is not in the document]
gjcolombo: It's OK to ignore this if startup failed--the caller (`run`) will detect the error, move the state machine to Failed, and then return from the state driver entirely, which triggers rundown and cleanup of the entire VM. (I think we also want to be careful about going through the pause-and-halt machinery in this case, because I don't think the server currently promises not to call `pause` or `halt` on a VM component that hasn't been fully started... but I'd have to look at who currently relies on what promises there.)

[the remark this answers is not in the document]
gjcolombo: Forgot to say that I added a comment in e66e5deefd48 that describes this expectation in a bit more detail.

gjcolombo:
> I'm now wondering if self_request is necessary at all, since it only ever stores a single ExternalRequest::Stop – could we just push that to the normal external_requests queue instead?

Interesting question. I think this actually illuminates a subtle problem with the queue dispositions we expose with this change:

1. VM begins to start
2. External requestor asks to stop the VM; this sets the queue disposition to ignore new stop requests (for idempotency), reject reboot/migration requests (since the VM will stop before these are processed), and allow VCR change requests (new in this PR)
3. `start_vm` pops the stop request and sets `stopped_while_starting`
4. Block backends finish starting up
5. `start_vm` calls `publish_steady_state(Running)`; this sets the queue dispositions to appropriate dispositions for a *running* VM (reboot/migration out requests are allowed again; VCR change requests are still allowed; stop requests are enqueued, not ignored)
6. the state driver processes its buffered stop request; this corrects the queue dispositions again (to reject everything but stop requests, which are ignored)

So the answer to your question is "we could do that," at least insofar as the request queue logic will allow a new stop request to be enqueued if we do that after marking the VM as Running in step 5. But that disposition change is itself a problem--the goal is to reject requests that won't be handled because the VM is going to stop first, and setting the "now running" dispositions allows that.

I'm going to mull this over a little bit more--the more I look at the request queue's complexity, the less I like it, and the more I want to try to see how much of it is really essential.

gjcolombo:
> I'm going to mull this over a little bit more--the more I look at the request queue's complexity, the less I like it, and the more I want to try to see how much of it is really essential.

So this blew up. (Shame I don't have a SoundCloud.) I've described this problem in more detail and proposed a fix in #879.

My current plan is to put up a request queue refactoring PR that will fix #879, then amend this PR to take advantage of it. I have a working version of this PR that uses the improved request queue API to ditch the "self request" queue (the improved API lets the state driver pop VCR replacement requests without popping or reordering "state change" requests like stop requests). Assuming the fix for #879 is acceptable, I'll merge it and then push the necessary amendments to this PR.

gjcolombo:
I've opened #880 to deal with the request queue improvements I discussed above and have changed this PR's merge base to that branch.

I also added commit 0fa8e37, which changes what happens when the state driver gets a stop request while block backends are still activating: instead of buffering the request, the driver simply returns and drops the block backend startup future. The only block backend that has an actually-asynchronous startup procedure is Crucible, and its only async call is to `activate`, so I *believe* this should be safe. I've added a warning label to the `Backend` trait describing this. (Note that, in any case, the "abort" path can only be reached if the VM's vCPUs haven't started yet, so there should be no outstanding I/O that could be damaged by unceremoniously dropping the future and the backends.)

Doing this eliminates the need for a "self-request" queue in the state driver, so the state driver logic turns out to be a bit simpler.

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -478,10 +489,10 @@ impl StateDriver {
...
 
         let final_state = if migrated_in {
-            if self.start_vm(VmStartReason::MigratedIn).await.is_ok() {
-                self.event_loop().await
-            } else {
-                InstanceState::Failed
+            match self.start_vm(VmStartReason::MigratedIn).await {
+                VmStartOutcome::Succeeded => self.event_loop().await,
+                VmStartOutcome::Failed => InstanceState::Failed,
+                VmStartOutcome::Aborted => InstanceState::Destroyed,
```
hawkw, line 495: it occurs to me that we _could_ just return `Result<(), InstanceState>` in `start_vm`...but I think `VmStartOutcome` is maybe a bit better at expressing the semantic meaning to the reader. Returning a `Result` would let `start_vm` use `?` if it `map_err`ed results to `InstanceState`s, but...i don't actually see a ton of places where we might want to do that...

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -524,32 +535,208 @@ impl StateDriver {
...
-            Ok(()) => {
-                self.external_state.update(ExternalStateUpdate::Instance(
-                    InstanceState::Running,
-                ));
+        // Send synchronous start commands to all devices.
+        for (name, dev) in objects.device_map() {
+            info!(self.log, "sending start request to {}", name);
+            let res = dev.start();
+            if let Err(e) = res {
+                error!(self.log, "startup failed for {}: {:?}", name, e);
```
hawkw, line 578: nit, take it or leave it: perhaps the device name ought to be a structured field on these logs to ease in using `looker` or similar?

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -524,32 +535,208 @@ impl StateDriver {
...
+            backends: BlockBackendMap,
+        ) -> anyhow::Result<()> {
+            for (name, backend) in backends {
+                info!(log, "starting block backend {}", name);
+                let res = backend.start().await;
+                if let Err(e) = &res {
+                    error!(
+                        log,
+                        "startup failed for block backend {}: {:?}", name, e
+                    );
```
hawkw, line 603: similarly, maybe we should use a structured slog field for the block backend name on these logs?

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -524,32 +535,208 @@ impl StateDriver {
...
+                // queue.
+                biased;
+
+                res = &mut block_backend_fut => {
+                    Selection::BackendFuture(res)
+                }
+
+                event = self.input_queue.wait_for_next_event() => {
+                    Selection::Event(event)
+                }
```
hawkw, line 645: `wait_for_next_event()` won't actually remove an event from the queue unless it completes, right? I assume this is the case but wanted to make extra sure that when block backend startup completes, we won't consume an event from the queue that _should_ still be there for the caller to consume once the VM is started.

bin/propolis-server/src/lib/vm/state_driver.rs
```diff
@@ -524,32 +535,208 @@ impl StateDriver {
...
+                    return VmStartOutcome::Aborted;
+                }
+                ExternalRequest::Component(
+                    ComponentChangeRequest::ReconfigureCrucibleVolume {
+                        backend_id,
+                        new_vcr_json,
+                        result_tx,
+                    },
+                ) => {
+                    let _ = result_tx.send(
```
hawkw, line 715: why would the receiver side of this have gone away? is that just a "the HTTP connection was closed before the request completed" type situation?

hawkw (approved):
the new state driver logic makes sense to me! i left some small notes, but no blocking concerns.

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: Yes, although I think this may not actually happen in practice, since IIUC Dropshot won't cancel the task that owns the receiver just because the client timed out. I'll add a comment here.

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: Correct, `wait_for_next_event` should be `select!`-safe: if it pops something from the queue, then it always returns the popped value immediately without awaiting again.

bin/propolis-server/src/lib/vm/state_driver.rs
gjcolombo: I do think we can cut back on the stuttering a little by putting a method on `VmStartOutcome` that yields the expected final state for that outcome (if there is one). That at least removes the repetition between the explicit-request and migrated-in startup paths.

mkeeter (approved):

Issue #698, opened 2024-05-02
Want CFS/NSNR and related NVMe Error Injection Capabilities


rmustacc:
In the field, there are a number of ways that NVMe drives can fail. There are few things that it would be great to simulate in Propolis. I'm going to call out a few of these here:

* We would like the ability to set the drive into a `CFS` state where by that bit is set and it will refuse to process additional I/O. There are two different paths forward we would like to have on the device:
    * On reset, the device continues to basically issue an NSNR until we format it. I realize that may be a little complicated as it requires some persistent state to be kept and tracked around. I think it'd be fine is this was only in-memory state.
    * On reset the CFS is cleared and the device begins processing again. The former is one we've seen more commonly on certain firmware revs, but this behavior seems useful to test.
* A device starting up in a read-only media mode
* Injecting asynchronous events that cover:
    * NVM subsystem reliability which results in a read-only media and related information in the SMART / health log. PArticularly, bit 3 is the most relevant in the critical warning field, but others may be interesting.
    * Transient / persistent internal errors, the latter which results in CFS being set.

Obviously with all these these are things we likely don't want to include in the product as a whole; however, the media read-only mode based on crucible going down to a single replica may be interesting.

PR #878, opened 2025-03-04
Use production propolis-server flags in PHD CI builds

### LINKED ISSUE #412: Set up PHD runs using binaries with production feature flags
Once #410 lands, the Propolis server images built in CI for inclusion in Omicron deployments and host OS images will have a subtly different feature set than "vanilla" Propolis binaries. We should update our CI tests to use these binaries to make our test signal as accurate as possible.

iximeow:
We use the dev (e.g. non-release) build of propolis-server in PHD to get debug assertions, which we want, but we should use the `omicron-build` feature to be in line with production binaries.

`omicron-build` had also acquired code related to failure injection, which we *also* want in CI so we can test migration failure codepaths. That is moved to a new `failure-injection` feature flag with a bit of ceremony to stop us early if we're ever inadvertently packaging a propolis-server with `omicron-build` and `failure-injection`.

Fixes #412. I assume the `failure-injection` feature will be useful for future testing enhancements like https://github.com/oxidecomputer/propolis/issues/698 

`cargo xtask phd run` still passes with an Alpine guest, waiting to see the same from `phd-run`.

### DIFF
--- a/.github/buildomat/jobs/phd-build.sh
+++ b/.github/buildomat/jobs/phd-build.sh
@@ -31,6 +31,8 @@ rustc --version
 # should fire during tests.
 banner build-propolis
 
+# Compile propolis-server so that it allows development features to be used even
+# though the `omicron-build` feature is enabled.
 export PHD_BUILD="true"
 ptime -m cargo build --verbose -p propolis-server \
 	--features omicron-build,failure-injection
--- a/.github/buildomat/jobs/phd-build.sh
+++ b/.github/buildomat/jobs/phd-build.sh
@@ -30,7 +30,10 @@ rustc --version
 # Build the Propolis server binary with 'dev' profile to enable assertions that
 # should fire during tests.
 banner build-propolis
-ptime -m cargo build --verbose -p propolis-server
+
+export PHD_BUILD="true"
+ptime -m cargo build --verbose -p propolis-server \
+	--features omicron-build,failure-injection
 
 # The PHD runner requires unwind-on-panic to catch certain test failures, so
 # build it with the 'dev' profile which is so configured.
--- a/lib/propolis/Cargo.toml
+++ b/lib/propolis/Cargo.toml
@@ -61,3 +61,5 @@ falcon = ["libloading", "p9ds", "dlpi", "ispf", "rand", "softnpu", "viona_api/fa
 # TODO until crucible#1280 is addressed, enabling Nexus notifications is done
 # through a feature flag.
 omicron-build = ["crucible/notify-nexus"]
+
+failure-injection = []
--- a/lib/propolis/Cargo.toml
+++ b/lib/propolis/Cargo.toml
@@ -61,5 +61,3 @@ falcon = ["libloading", "p9ds", "dlpi", "ispf", "rand", "softnpu", "viona_api/fa
 # TODO until crucible#1280 is addressed, enabling Nexus notifications is done
 # through a feature flag.
 omicron-build = ["crucible/notify-nexus"]
-
-failure-injection = []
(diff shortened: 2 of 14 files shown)


lib/propolis/Cargo.toml
```diff
@@ -61,3 +61,5 @@ falcon = ["libloading", "p9ds", "dlpi", "ispf", "rand", "softnpu", "viona_api/fa
 # TODO until crucible#1280 is addressed, enabling Nexus notifications is done
 # through a feature flag.
 omicron-build = ["crucible/notify-nexus"]
+
+failure-injection = []
```
pfmooney, line 65: Add a little comment here about the purpose of the feature?

iximeow: I'll do one better - I thought there was `omicron-build` code in `propolis` that would move to `failure-injection`, but it turns out that there isn't (yet?)

so I was premature in adding this flag in both crates, and I'll back this one out.

iximeow: https://github.com/oxidecomputer/propolis/pull/878/commits/cf02de70e02d59ecb8fbf48fc8537c42fd6a6755 - comment on `bin/propolis-server/Cargo.toml` seems good though right?

pfmooney: Yes, it's appreciated.

iximeow:
all the failures include propolis-server stderr like `"error_message_internal":"VM initialization failed: failed to join VM object creation task: failed to add low memory region"`

there are info logs for VMM allocations and from a passing build we have been saying `reservoir too small (0MiB) to use for guest memory`. so, now with `omicron-build` it's unconditionally trying and failing to use the zero-size reservoir, and every test fails. onward to wonder what's up with `buskin`.

xtask/src/task_phd.rs
```diff
@@ -292,7 +292,20 @@ impl Cmd {
...
+                server_build_env
+                    .insert("PHD_BUILD".to_string(), "true".to_string());
+                let bin = build_bin(
+                    "propolis-server",
+                    propolis_args.release,
+                    // Some PHD tests specifically cover cases where a component
+                    // in the system has encountered an error, so enable
+                    // failure-injection. Do not enable `omicron-build` like we
+                    // do in Buildomat because someone running `cargo xtask phd`
+                    // is not building a propolis-server destined for Omicron.
```
gjcolombo, line 305: Well, unless they are :) I think this was my use case earlier this week, and it may also come in handy if someone wants to try to reproduce a CI failure on a local machine.

Perhaps add an `omicron_build` switch to `PropolisArgs` to allow the user to opt into adding that feature?

gjcolombo (approved):
A couple little nits, but overall this looks good to me. Thanks for working on this!

.github/buildomat/jobs/phd-build.sh
```diff
@@ -30,7 +30,10 @@ rustc --version
 # Build the Propolis server binary with 'dev' profile to enable assertions that
 # should fire during tests.
 banner build-propolis
-ptime -m cargo build --verbose -p propolis-server
+
+export PHD_BUILD="true"
```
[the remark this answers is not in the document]
iximeow: oh yes, i'd meant to comment this once i'd figured out the exact semantics over in `propolis-server`. thanks!

xtask/src/task_phd.rs
iximeow: ..... i suppose `cargo xtask phd` does build a replacement `propolis-server` doesn't it.

but yeah, a build to get `omicron-build` enabled makes a lot of sense. lemme do that.

iximeow: in writing https://github.com/oxidecomputer/propolis/pull/878/commits/8affc6504ee4fc6663dd72e2339724d41b44c969 i discovered that `cargo xtask phd run` and `cargo xtask phd run --omicron-build` end up with artifacts in different directories. so if you run one, then the other, then the first one again expecting another `propolis-server` rebuild, it'll actually just run immediately. surprise!

but the test suite passes on my workstation with or without the flag (wherein i have a reservoir configured), everything fails (as expected!) when i set the reservoir too small.

PR #880, opened 2025-03-06
server: extend API request queue's memory of prior requests

### LINKED ISSUE #879: server: state driver request queue dispositions are difficult to manage
This arose from the discussion of #873.

## The problem

The server state driver's external request queue ([source](https://github.com/oxidecomputer/propolis/blob/6b5f2af796a3ea57405721407ab70520a93ec73f/bin/propolis-server/src/lib/vm/request_queue.rs)) receives requests to operate on a VM and decides whether to Enqueue, Ignore, or Deny them. The queue stores the current disposition for each kind of request it can receive. When the queue receives a new request, or when the state driver notifies it of a change in VM state, the queue runs its [`get_new_dispositions`](https://github.com/oxidecomputer/propolis/blob/6b5f2af796a3ea57405721407ab70520a93ec73f/bin/propolis-server/src/lib/vm/request_queue.rs) function to compute the new set of dispositions.

In general, the queue tries to deny requests that either can't be serviced right now or that might be preempted by some other, as-yet-unprocessed event. For example, the queue denies requests to reboot an instance that's not running, but it also denies reboot requests that are preceded by an unprocessed request to migrate or stop the instance: in the former case, the request may need to be redirected to the migration target, and in the latter case, the request won't be actionable by the time it gets to the front of the queue.

The problem is that `get_new_dispositions` computes the new dispositions as a function of the incoming event and the queue's previous dispositions, and the old dispositions don't always provide enough context to compute new ones accurately. Consider the case above: if a VM is rejecting reboot requests, but then it starts, should reboot requests now be accepted? The answer is "no" if a request to stop/migrate is pending and "yes" otherwise, but there's no way to distinguish these cases just from the prior disposition, because reboot requests on an un-started instance are rejected regardless.

As the state driver gets more complex, and begins to interleave handling of various kinds of requests (see #873), this is going to get more and more difficult to manage.

## Proposal

I think the queue can be improved in a couple of ways.

First, request dispositions should be a function of the queue's current state and its memory of the requests that have been queued but not acknowledged by the state driver. This prevents problems like the one described above: if a currently-starting VM queues a request to stop, the queue should remember that request and deny subsequent requests to reboot, even if the state driver successfully starts the VM in the meantime.

Second, it would be very handy for the queue to distinguish requests to change a VM's *state* from requests to change its *configuration*. This is useful primarily for #873, where we want the state driver to be able to process VCR change requests while a VM is starting without reordering or losing track of state change requests that should be processed only after the VM has successfully started.

I have a change ready to go that implements both of these behaviors, as well as a merge commit that extends #873 to take advantage of them. I hope to have a PR up soon.

gjcolombo:
Today the server's API request queue decides whether to accept or reject incoming requests by storing a fixed disposition for each kind of request and adjusting those dispositions as new requests are queued or as a VM's state changes. This is not always enough information to maintain the proper dispositions. Consider the following example:

1. A VM begins to start. Reboot requests are denied in this state (since a VM that hasn't booted yet can't be rebooted).
2. A caller queues a request to stop the VM after it starts. All subsequent reboot requests should now be denied because the VM will halt before they can be serviced.
3. The VM successfully starts. The correct reboot disposition is still "deny," because there's a pending stop request, but there's no way to discern this from the prior reboot disposition (it was also "deny" before the stop request was queued!).

To address this sort of problem, and to pave the way for better handling of Crucible VCR replacements during instance start, refactor the queue as follows:

- Remember what kinds of outstanding requests have not been processed and dispose of requests based on this state and the queue's notion of the instance's overall state.
- Break external API requests into "state change" and "component change" requests. This will allow them to be dequeued independently when the state driver wants to handle component changes without changing the order of state change requests.
- Tweak the language the state driver uses to communicate with the queue: instead of saying "the VM is now in state X," the driver says "I handled (or failed to handle) a request of type Y."

Tests: cargo test, PHD.

Fixes #879.

### DIFF
--- a/bin/propolis-server/src/lib/vm/request_queue.rs
+++ b/bin/propolis-server/src/lib/vm/request_queue.rs
@@ -1049,4 +1049,256 @@ mod test {
             }
         }
     }
+
+    /// An operation that can be performed during a [`QueueDequeueTest`].
+    #[derive(Clone, Copy, Debug)]
+    enum QueueOp {
+        Enqueue(RequestKind),
+        Dequeue,
+    }
+
+    fn queue_op_strategy() -> impl Strategy<Value = QueueOp> {
+        prop_oneof![
+            request_strategy().prop_map(QueueOp::Enqueue),
+            Just(QueueOp::Dequeue)
+        ]
+    }
+
+    /// A helper that queues and dequeues requests in a proptest-generated
+    /// order and that sends fake completion notifications back to the request
+    /// queue.
+    struct QueueDequeueTest {
+        /// The external request queue under test.
+        queue: ExternalRequestQueue,
+
+        /// The set of state change requests that the helper expects to see from
+        /// the external queue.
+        expected_state: VecDeque<RequestKind>,
+
+        /// The set of component change requests that the helper expects to see
+        /// from the external queue.
+        expected_component: VecDeque<RequestKind>,
+
+        /// True if the helper has queued a request to start its fake VM.
+        start_requested: bool,
+
+        /// True if the helper has successfully started its fake VM.
+        started: bool,
+
+        /// True if the helper has queued a request to stop its fake VM.
+        stop_requested: bool,
+
+        /// True if the helper has an outstanding request to reboot its fake VM.
+        reboot_requested: bool,
+
+        /// True if the helper has an outstanding request to migrate its fake
+        /// VM.
+        migrate_out_requested: bool,
+
+        /// True if the fake VM is halted (for any reason).
+        halted: bool,
+    }
+
+    impl QueueDequeueTest {
+        fn new() -> Self {
+            Self {
+                queue: ExternalRequestQueue::new(
+                    test_logger(),
+                    InstanceAutoStart::No,
+                ),
+                expected_state: Default::default(),
+                expected_component: Default::default(),
+                start_requested: false,
+                started: false,
+                stop_requested: false,
+                reboot_requested: false,
+                migrate_out_requested: false,
+                halted: false,
+            }
+        }
+
+        fn run(&mut self, ops: Vec<QueueOp>) {
+            for op in ops {
+                match op {
+                    QueueOp::Enqueue(request) => self.queue_request(request),
+                    QueueOp::Dequeue => {
+                        self.dequeue_request();
+                        if self.halted {
+                            return;
+                        }
+                    }
+                }
+            }
+        }
+
+        /// Submits the supplied `request` to the external request queue,
+        /// determines the expected result of that submission based on the
+        /// helper's current flags, and asserts that the result matches the
+        /// helper's expectation. If the helper expects the request to be
+        /// queued, it pushes an entry to its internal expected-change queues.
+        fn queue_request(&mut self, request: RequestKind) {
+            let result = self.queue.try_queue(request.into());
+            match request {
+                RequestKind::Start { .. } => {
+                    if self.halted || self.stop_requested {
+                        assert!(result.is_err());
+                        return;
+                    }
+
+                    assert!(result.is_ok());
+                    if !self.start_requested {
+                        self.start_requested = true;
+                        self.expected_state.push_back(request);
+                    }
+                }
+                RequestKind::Stop => {
+                    if self.halted || self.stop_requested {
+                        assert!(result.is_ok());
+                        return;
+                    }
+
+                    if self.migrate_out_requested {
+                        assert!(result.is_err());
+                        return;
+                    }
+
+                    self.stop_requested = true;
+                    self.expected_state.push_back(request);
+                }
+                RequestKind::Reboot => {
+                    if !self.started
+                        || self.halted
+                        || self.stop_requested
+                        || self.migrate_out_requested
+                    {
+                        assert!(result.is_err());
+                        return;
+                    }
+
+                    assert!(result.is_ok());
+                    if !self.reboot_requested {
+                        self.reboot_requested = true;
+                        self.expected_state.push_back(request);
+                    }
+                }
+                RequestKind::Migrate { .. } => {
+                    if (!self.started && !self.start_requested)
+                        || self.halted
+                        || self.stop_requested
+                        || self.migrate_out_requested
+                    {
+                        assert!(result.is_err());
+                        return;
+                    }
+
+                    assert!(result.is_ok());
+                    self.expected_state.push_back(request);
+                    self.migrate_out_requested = true;
+                }
+                RequestKind::ReconfigureCrucible => {
+                    if self.halted {
+                        assert!(result.is_err());
+                        return;
+                    }
+
+                    assert!(result.is_ok());
+                    self.expected_component.push_back(request);
+                }
+            }
+        }
+
+        /// Pops a request from the helper's external queue and verifies that it
+        /// matches the first request on the helper's expected-change queue. If
+        /// the requests do match, sends a completion notification to the
+        /// external queue.
+        fn dequeue_request(&mut self) {
+            let (dequeued, expected) = match (
+                self.queue.pop_front(),
+                self.expected_state
+                    .pop_front()
+                    .or_else(|| self.expected_component.pop_front()),
+            ) {
+                (None, None) => return,
+                (Some(d), None) => {
+                    panic!("dequeued request {d:?} but expected nothing")
+                }
+                (None, Some(e)) => {
+                    panic!("expected request {e:?} but dequeued nothing")
+                }
+                (Some(d), Some(e)) => (d, e),
+            };
+
+            match (dequeued, expected) {
+                (
+                    ExternalRequest::State(StateChangeRequest::Start),
+                    RequestKind::Start { will_succeed },
+                ) => {
+                    self.queue.notify_request_completed(
+                        CompletedRequest::Start { succeeded: will_succeed },
+                    );
+                    if will_succeed {
+                        self.started = true;
+                    } else {
+                        self.halted = true;
+                    }
+                }
+                (
+                    ExternalRequest::State(StateChangeRequest::Stop),
+                    RequestKind::Stop,
+                ) => {
+                    self.queue.notify_request_completed(CompletedRequest::Stop);
+                    self.halted = true;
+                }
+                (
+                    ExternalRequest::State(StateChangeRequest::Reboot),
+                    RequestKind::Reboot,
+                ) => {
+                    self.queue
+                        .notify_request_completed(CompletedRequest::Reboot);
+                    self.reboot_requested = false;
+                }
+                (
+                    ExternalRequest::State(
+                        StateChangeRequest::MigrateAsSource { .. },
+                    ),
+                    RequestKind::Migrate { will_succeed },
+                ) => {
+                    self.queue.notify_request_completed(
+                        CompletedRequest::MigrationOut {
+                            succeeded: will_succeed,
+                        },
+                    );
+                    self.migrate_out_requested = false;
+                    if will_succeed {
+                        self.halted = true;
+                    }
+                }
+                (
+                    ExternalRequest::Component(
+                        ComponentChangeRequest::ReconfigureCrucibleVolume {
+                            ..
+                        },
+                    ),
+                    RequestKind::ReconfigureCrucible,
+                ) => {}
+                (d, e) => panic!(
+                    "dequeued request {d:?} but expected to dequeue {e:?}\n\
+                    remaining queue: {:#?}\n\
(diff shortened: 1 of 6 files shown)


bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -164,248 +214,330 @@ enum RequestDisposition {
...
+            awaiting_stop: false,
             log,
         }
     }
 
-    /// Pops the request at the front of the queue.
+    /// Pops the next request off of the queue. If the queue contains both state
+    /// change and component change requests, the next state change request is
+    /// popped first (even if it arrived later in time than the next component
+    /// change request).
```
gjcolombo, line 321: I also drafted a version of this code that pops the oldest request regardless of its kind, but maintains a separate index of component change requests so that the state driver can ask to dequeue the oldest such request while leaving the state change requests alone. (The point of the exercise is to give the driver a way to pop VCR replacement requests while starting the VM without having to buffer any state change requests; see #873 for more discussion.)

I ended up not going with this because (a) it's simpler not to have the separate change request index, and (b) it also seemed to me to make some sense to privilege external API requests (like requests to stop) over internal housekeeping operations like VCR replacement. But I'd like to hear if others feel strongly here (it's not that hard to bring that code back).

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -164,248 +214,330 @@ enum RequestDisposition {
...
 #[derive(Debug)]
-pub struct ExternalRequestQueue {
-    queue: VecDeque<ExternalRequest>,
-    allowed: AllowedRequests,
+pub(super) struct ExternalRequestQueue {
+    /// The queue of unprocessed state change requests.
+    state_queue: VecDeque<StateChangeRequest>,
+
+    /// The queue of unprocessed component change requests.
+    component_queue: VecDeque<ComponentChangeRequest>,
```
hawkw, line 269: i note that this means that there's no longer a total ordering of state change and component change requests, but i think that...may not matter.

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -164,248 +214,330 @@ enum RequestDisposition {
...
+    /// completed by the state driver.
+    awaiting_reboot: bool,
+
+    /// True if this queue has enqueued a stop request that has not been
+    /// completed by the state driver.
+    awaiting_migration_out: bool,
+
+    /// True if this queue has enqueued a request to migrate out that has not
+    /// been completed by the state driver.
+    awaiting_stop: bool,
```
hawkw, line 285: the doc comments on `awaiting_migration_out` and `awaiting_stop` appear to have been swapped.
```suggestion
    /// True if this queue has enqueued a request to migrate out that has not
    /// been completed by the state driver.
    awaiting_migration_out: bool,

    /// True if this queue has enqueued a stop request that has not been
    /// completed by the state driver.
    awaiting_stop: bool,
```

bin/propolis-server/src/lib/vm/request_queue.rs
hawkw: I think I agree with your rationale here. I don't know if the property of having one total ordering across state-change and component-change requests is actually important, given that the whole point of this change is to allow component change requests to be serviced whilst a state change is in progress, and the disposition to subsequent component change requests based on in-flight state changes is now tracked outside of the actual queue, rather than based on the order of requests in the queue. So I think the simpler solution of just having two queues is fine.

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -164,248 +214,330 @@ enum RequestDisposition {
...
-            ExternalRequest::Reboot => self.allowed.reboot,
-            ExternalRequest::ReconfigureCrucibleVolume { .. } => {
-                self.allowed.mutate
+        // If the queue is in a terminal state, deny the request straightaway
+        // (unless it's a stop request, which can be ignored for idempotency).
+        let disposition = if let Some(reason) = self.state.deny_reason() {
+            if matches!(
+                request,
+                ExternalRequest::State(StateChangeRequest::Stop)
+            ) {
```
hawkw, line 350: nit, feel free to ignore: seems like `ExternalRequest` could just derive `PartialEq`?

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -164,248 +214,330 @@ enum RequestDisposition {
...
+                ExternalRequest::State(
+                    StateChangeRequest::MigrateAsSource { .. },
+                ) => {
+                    if self.awaiting_migration_out {
+                        Disposition::Deny(
+                            RequestDeniedReason::AlreadyMigrationSource,
+                        )
+                    } else if self.awaiting_stop {
+                        Disposition::Deny(RequestDeniedReason::HaltPending)
+                    } else if matches!(self.state, QueueState::NotStarted) {
```
hawkw, line 389: nit: seems `QueueState` could also derive `PartialEq`...or, we could use `if let` here as we do above? but i'd just derive `PartialEq`...

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -164,248 +214,330 @@ enum RequestDisposition {
...
+            // In general, try to make state change requests idempotent by
+            // ignoring new requests when a request of the appropriate kind is
+            // already on the queue, and deny requests to reach a state that is
+            // precluded by an earlier state change request.
+            match request {
+                // Interpret start requests as requests to reach the Running
+                // state.
+                ExternalRequest::State(StateChangeRequest::Start) => {
+                    if self.awaiting_stop {
+                        Disposition::Deny(RequestDeniedReason::HaltPending)
```
hawkw, line 369: part of me wonders if the `awaiting_stop` logic common to every state change request except `stop` could be factored out, but...idk if it's worth it; i think expressing the complete logic for each possible state request in the match arm for that request is probably clearer than trying to be less repetitive...

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -164,248 +214,330 @@ enum RequestDisposition {
...
-            // Requests to stop always succeed. Note that a request to stop a VM
-            // that hasn't started should still be queued to the state worker so
-            // that the worker can exit and drop its references to the instance.
-            ExternalRequest::Stop => self.allowed.stop,
+                // Always enqueue component change requests, even if the VM has
+                // a pending request to stop or migrate out. This allows the
+                // state driver to process these requests during a state change,
+                // which may be necessary to complete that state change. If the
+                // change request is canceled by a later state transition, the
+                // queue can use the request data to notify the requestor.
```
hawkw, line 443: :clap: YOU :clap: LOVE :clap: TO :clap: SEE :clap: IT :clap:

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -164,248 +214,330 @@ enum RequestDisposition {
...
-            ChangeReason::ApiRequest(ExternalRequest::Start) => {
-                let reason = DenyReason::StartInProgress;
-                AllowedRequests {
-                    start: Disposition::Ignore,
-                    migrate_as_source: Disposition::Deny(reason),
-                    reboot: Disposition::Deny(reason),
-                    mutate: Disposition::Deny(reason),
-                    stop: self.allowed.stop,
+            Disposition::Ignore => Ok(()),
+            Disposition::Deny(reason) => Err(reason),
```
hawkw, line 458: soooo...i'm not convinced that the disposition enum is really pulling its weight anymore. Why not just have the arms of the above match that presently return `Disposition::Ignore` and `Disposition::Deny` just `return Ok(())`/`return Err(reason)`? i guess the `Disposition`s do get logged in one place, which is maybe nicer than having all the places where we would return log separately...

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -164,248 +214,330 @@ enum RequestDisposition {
...
+                        assert!(!self.awaiting_reboot);
+                        self.awaiting_reboot = true;
+                    }
+                    ExternalRequest::State(StateChangeRequest::Stop) => {
+                        assert!(!self.awaiting_stop);
+                        self.awaiting_stop = true;
+                    }
+                    ExternalRequest::Component(_) => {}
                 }
-            }
```
hawkw, line 301: we could avoid one tabstop of rightward drift here by instead writing:
```rust
        Disposition::Enqueue => {},
    };

    match request {
        ExternalRequest::State(StateChangeRequest::Start) => {
            assert!(matches!(self.state, QueueState::NotStarted));
            self.state = QueueState::StartPending;
        }
        ExternalRequest::State(
            StateChangeRequest::MigrateAsSource { .. },
        ) => {
            assert!(!self.awaiting_migration_out);
            self.awaiting_migration_out = true;
        }
        ExternalRequest::State(StateChangeRequest::Reboot) => {
            assert!(!self.awaiting_reboot);
            self.awaiting_reboot = true;
        }
        ExternalRequest::State(StateChangeRequest::Stop) => {
            assert!(!self.awaiting_stop);
            self.awaiting_stop = true;
        }
        ExternalRequest::Component(_) => {}
    }
```

although i dunno if that matters.

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -459,48 +590,89 @@ mod test {
...
+
+        fn assert_migrate_as_source(&self) {
+            assert!(
+                matches!(
+                    self,
+                    Self::State(StateChangeRequest::MigrateAsSource { .. })
+                ),
+                "expected migrate as source request, got {self:?}"
+            );
         }
```
hawkw, line 640: this is cute! i'll be borrowing this for future tests! however, one note is that wrapping the assertions in these methods will change the panic location to the method definition, rather than to the line in the test where we made the particular assertion that failed. you can fix that by adding `#[track_caller]` to these:
```suggestion
    impl ExternalRequest {
        #[track_caller]
        fn assert_start(&self) {
            assert!(
                matches!(self, Self::State(StateChangeRequest::Start)),
                "expected start request, got {self:?}"
            );
        }

        #[track_caller]
        fn assert_stop(&self) {
            assert!(
                matches!(self, Self::State(StateChangeRequest::Stop)),
                "expected stop request, got {self:?}"
            );
        }

        #[track_caller]
        fn assert_reboot(&self) {
            assert!(
                matches!(self, Self::State(StateChangeRequest::Reboot)),
                "expected reboot request, got {self:?}"
            );
        }

        #[track_caller]
        fn assert_migrate_as_source(&self) {
            assert!(
                matches!(
                    self,
                    Self::State(StateChangeRequest::MigrateAsSource { .. })
                ),
                "expected migrate as source request, got {self:?}"
            );
        }
```

bin/propolis-server/src/lib/vm/request_queue.rs
```diff
@@ -459,48 +590,89 @@ mod test {
...
     }
 
-    #[tokio::test]
-    async fn start_requests_become_idempotent_after_first_request() {
+    #[test]
+    fn start_requests_become_idempotent_after_first_request() {
```
hawkw, line 644: it occurs to me that if we wanted to be particularly fastidious, a lot of these tests _could_ probably be expressed as property tests, where we would generate a random sequence of requests and assert that the property holds across all possible requests once we have entered into a particular state...

hawkw: but, you know, what you've got seems good enough.

hawkw (approved):
This change looks great to me! I left some fairly minor notes, but I'm down with the overall design here.

bin/propolis-server/src/lib/vm/request_queue.rs
gjcolombo: Unfortunately the websocket connection in the `MigrateAsSource` variant gets in the way :( This match is wordy enough, though, that I think it makes sense to have a (private) `is_stop` method on `ExternalRequest`. I'll add that.

bin/propolis-server/src/lib/vm/request_queue.rs
gjcolombo: (Un?)fortunately, I was just thinking the other day about how it'd be nice to exercise these muscles a bit. Added some property tests in 58cb957 and 81dde61.

bin/propolis-server/src/lib/vm/request_queue.rs
gjcolombo: Done via 29ea5fc and 0178609.

PR #883, opened 2025-03-10
Add initial block TRIM/UNMAP/discard support

pfmooney:


gjcolombo:
I *think* rebasing onto the latest `master` will resolve the `migrate-from-base` failure. (The base `phd_build` image now has the `omicron-build` feature turned on, so it needs a reservoir; but the PHD execution scripts come from the commit under test, which doesn't have the corresponding change to run `rsrvrctl` before starting the tests.)

lib/propolis/src/block/file.rs
```diff
@@ -238,30 +268,200 @@ impl block::Backend for FileBackend {
...
+        }
+    }
+
+    pub(crate) fn disk_info(fp: &File) -> DiskInfo {
+        match fp.metadata() {
+            Ok(ft) if ft.file_type().is_char_device() => {
+                // Continue on to attempt DKIOC lookups on raw disk devices
+            }
+            Ok(ft) if ft.file_type().is_file() => {
+                // Assume fcntl(F_FREESP) support for files
```
gjcolombo, line 329: Optional, up to you: I would be somewhat tempted to put the `target_os = "illumos"` check here (returning `None` if it doesn't hold) and then marking the F_FREESP case as `unreachable!` on other OSes below instead of using the invalid `fcntl` command code.

gjcolombo (approved):

PR #889, opened 2025-03-26
propolis-server: collect and report VSS minus VM mappings

iximeow:
collect VSS from `/proc/self/psinfo` and report it as an Oximeter metric for the VM. VSS is maintained as mappings are adjusted so it's about as close to free as a syscall can get. Propolis' VSS includes guest-mapped regions; this change subtracts out the size of a guest's `ASpace` so we only get the amount Propolis has a say in. no RSS measurement here, but that's really only useful to tell us if the virtual mappings are overly-large (like, do we have a 128 MiB heap when we're only using 4 MiB?) - that'd be nice to know in the limit, but somewhat less interesting now. most of our VSS comes from `propolis-server` itself anyway.

the original approach here was walking through `/proc/{getpid()}/xmap` to get both RSS and VSS, and letting us exclude mappings that are part of the guest's accessible memory - in some sense this is aesthetically pleasant, but collecting and iterating the thousands of mappings meant this took up to seconds(!) of time that potentially could block the guest VM (!) if it needed the process lock at any point while we were measuring.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -182,14 +182,17 @@ zerocopy = "0.7.34"
 #
 [patch."https://github.com/oxidecomputer/omicron"]
 #internal-dns = { path = "../omicron/internal-dns" }
-nexus-client = { path = "../omicron/clients/nexus-client" }
-omicron-uuid-kinds = { path = "../omicron/uuid-kinds" }
-nexus-types = { path = "../omicron/nexus/types" }
-illumos-utils = { path = "../omicron/illumos-utils" }
-omicron-common = { path = "../omicron/common" }
-oximeter-instruments = { path = "../omicron/oximeter/instruments" }
-oximeter-producer = { path = "../omicron/oximeter/producer" }
-oximeter = { path = "../omicron/oximeter/oximeter" }
+nexus-client = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+omicron-uuid-kinds = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+nexus-types = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+illumos-utils = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+omicron-common = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+oximeter-instruments = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+oximeter-producer = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+oximeter = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+
+# i've just picked what Omicron happens to currently point to.
+tufaceous-artifact = { git = "https://github.com/oxidecomputer//tufaceous.git", rev = "04681f26ba09e144e5467dd6bd22c4887692a670" }
 # [patch."https://github.com/oxidecomputer/crucible"]
 # crucible = { path = "../crucible/upstairs" }
 # crucible-client-types = { path = "../crucible/crucible-client-types" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -180,13 +180,16 @@ zerocopy = "0.7.34"
 # It's common during development to use a local copy of various complex
 # dependencies.  If you want to use those, uncomment one of these blocks.
 #
-# [patch."https://github.com/oxidecomputer/omicron"]
-# internal-dns = { path = "../omicron/internal-dns" }
-# nexus-client = { path = "../omicron/clients/nexus-client" }
-# omicron-common = { path = "../omicron/common" }
-# oximeter-instruments = { path = "../omicron/oximeter/instruments" }
-# oximeter-producer = { path = "../omicron/oximeter/producer" }
-# oximeter = { path = "../omicron/oximeter/oximeter" }
+[patch."https://github.com/oxidecomputer/omicron"]
+#internal-dns = { path = "../omicron/internal-dns" }
+nexus-client = { path = "../omicron/clients/nexus-client" }
+omicron-uuid-kinds = { path = "../omicron/uuid-kinds" }
+nexus-types = { path = "../omicron/nexus/types" }
+illumos-utils = { path = "../omicron/illumos-utils" }
+omicron-common = { path = "../omicron/common" }
+oximeter-instruments = { path = "../omicron/oximeter/instruments" }
+oximeter-producer = { path = "../omicron/oximeter/producer" }
+oximeter = { path = "../omicron/oximeter/oximeter" }
 # [patch."https://github.com/oxidecomputer/crucible"]
 # crucible = { path = "../crucible/upstairs" }
 # crucible-client-types = { path = "../crucible/crucible-client-types" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -182,14 +182,14 @@ zerocopy = "0.7.34"
 #
 [patch."https://github.com/oxidecomputer/omicron"]
 #internal-dns = { path = "../omicron/internal-dns" }
-nexus-client = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
-omicron-uuid-kinds = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
-nexus-types = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
-illumos-utils = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
-omicron-common = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
-oximeter-instruments = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
-oximeter-producer = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
-oximeter = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+nexus-client = { git = "https://github.com/oxidecomputer//omicron.git", rev = "da5f7706119b3f39dc5dfb2fde823c2294476b5a" }
+omicron-uuid-kinds = { git = "https://github.com/oxidecomputer//omicron.git", rev = "da5f7706119b3f39dc5dfb2fde823c2294476b5a" }
+nexus-types = { git = "https://github.com/oxidecomputer//omicron.git", rev = "da5f7706119b3f39dc5dfb2fde823c2294476b5a" }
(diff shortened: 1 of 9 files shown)


bin/propolis-server/src/lib/stats/process.rs
```diff
@@ -0,0 +1,166 @@
+use std::io::Read;
```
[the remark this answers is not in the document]
iximeow: thank you license eye

bin/propolis-server/src/lib/stats/process.rs
```diff
@@ -0,0 +1,170 @@
...
+        self.pr_mflags & MA_SHARED == MA_SHARED
+    }
+}
+
+pub fn process_stats() -> anyhow::Result<ProcessStats> {
+    let stats_read_start = Instant::now();
+
+    // Safety: getpid() does not alter any state, or even read mutable state.
+    let mypid = unsafe { libc::getpid() };
+    let xmap_path = format!("/proc/{}/xmap", mypid);
```
jclulow, line 100: I think this could just be `"/proc/self/xmap"` ?

iximeow: i hoped and thought so, but i don't have a `/proc/self*` at all locally. i was kinda curious about it: if you have it maybe it's added by something downstream of illumos, and helios doesn't get the fun?

iximeow: pmooney informs me that `/proc/self/` in fact there and just not in `ls /proc` (or, consequently, tab completion for `/proc/s` :) )

pfmooney: > the s-e-l-f is silent

iximeow:
this now collects useful information with at most 30 microseconds holding locks we'd care about - i'm not certain it would get through to oximeter yet (at least i haven't seen it work locally) but this is not outright harmful anymore!

Cargo.toml
```diff
@@ -180,13 +180,19 @@ zerocopy = "0.7.34"
...
+omicron-uuid-kinds = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+nexus-types = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+illumos-utils = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+omicron-common = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+oximeter-instruments = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+oximeter-producer = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+oximeter = { git = "https://github.com/oxidecomputer//omicron.git", rev = "3950a44ba51c1696b19392d20fd2cbb9457a000d" }
+
+# i've just picked what Omicron happens to currently point to.
+tufaceous-artifact = { git = "https://github.com/oxidecomputer//tufaceous.git", rev = "04681f26ba09e144e5467dd6bd22c4887692a670" }
```
iximeow, line 195: this all specifically is not intended to merge. to get a new Oximeter metric here, that's a change the the schema (and bump of `oximeter`). having Oximeter at one rev while Omicron is at other ones got me to a state where Propolis wouldn't build. bumping all of Omicron here left Propolis and Crucible using different Omicrons which again did not build because `Cargo.toml` had an old `tufaceous-artifact`.

this has somehow ended up with the three test failures in the last build, which all look like:
```
  phd-runner: [INSTANCE_ENSURE_INTERNAL - EVENT] permanent error from instance_spec_ensure 
    e = Error Response: status: 400 Bad Request; headers: {
      "content-type": "application/json", "x-request-id": "58e1ca30-797a-48bb-9191-a36b956e243b", 
      "content-length": "533", "date": "Wed, 30 Apr 2025 08:59:40 GMT"
    }; value: Error {
      error_code: None, 
      message: "unable to parse JSON body: init.value.spec.components.boot-disk.type: unknown variant `NvmeDisk`,
      expected one of `virtio_disk`, `nvme_disk`, `virtio_nic`, `serial_port`, `pci_pci_bridge`, `qemu_pvpanic`,
        `boot_settings`, `soft_npu_pci_port`, `soft_npu_port`, `soft_npu_p9`, `p9fs`, `migration_failure_injector`,
        `crucible_storage_backend`, `file_storage_backend`, `blob_storage_backend`, `virtio_network_backend`,
        `dlpi_network_backend` at line 1 column 180",
      request_id: "58e1ca30-797a-48bb-9191-a36b956e243b"
    } 
```

and frankly i'm at a loss about what changed the casing of variants here. it is presumably related to my wanton version bumping.

anyway, this PR is at "i think it's good, would folks +1 the new metric?" - if others agree i'll go PR the schema change in Omicron and figure out getting that through to Propolis.

iximeow (commented):
the Omicron change this points to is https://github.com/oxidecomputer/omicron/pull/8071.

i've left a comment on the `Cargo.toml` diff but i do not intend for the `Cargo.toml` or most of the `Cargo.lock` changes to land here. the CI failure is due to three tests that seem downstream of the version bumping, but i... really do not understand why.

bin/propolis-server/src/lib/stats/mod.rs
```diff
@@ -78,15 +82,80 @@ struct ServerStatsInner {
...
+    fn refresh_vmm_vsz(&mut self) {
+        let last_process_stats = self.process_stats_rx.borrow();
+
+        if let Some(vm_objects) = self.vm_objects.upgrade() {
+            // The mutex guarding VmObjects is async. Since we're not in an
+            // async context, we have to be a bit more explicit with the future.
+            // It's possible both for us to block acquiring this lock (if the
+            // state driver had an exclusive lock to make some change) and for
+            // us to block the state driver by acquiring the lock (but we
+            // *should* drop the lock very quickly, here).
```
hawkw, line 130: Yuck...I kinda wonder if it's worth doing this virtual address space accounting by having an atomic integer that we add to when mapping ranges, and which we can read from here without having to lock the whole `VmObjects` and iterate over every mapped region every time? That way, we go from "a critical section that's _O_(_n_mapped_physical_mem_regions_)" to "not having a critical section at all...

bin/propolis-server/src/lib/stats/process.rs
```diff
@@ -0,0 +1,129 @@
...
+        // eprintln for demonstrative purposes but otherwise should be removed.]
+        eprintln!(
+            "Sending new stats at least.. {:?}, took {}ms",
+            new_process_stats,
+            new_process_stats.measurement_time.as_millis()
+        );
+        tx.send_replace(new_process_stats);
+
+        tokio::time::sleep(std::time::Duration::from_millis(15000)).await;
+    }
```
hawkw, line 128: turbo nit, not actually important: consider using `tokio::time::interval` rather than `sleep`, it should be a bit more efficient as it reuses the same entry in the tokio timer wheel every time:
```suggestion
    let mut interval = tokio::time::interval(std::time::Duration::from_millis(15000);
    while !tx.is_closed() {
        interval.tick().await;
        let new_process_stats = tokio::task::spawn_blocking(process_stats)
            .await
            .expect("collecting address space stats does not panic")
            .expect("process_stats() does not error");

        // [there isn't a nice way to plumb a slog Logger here, so this is an
        // eprintln for demonstrative purposes but otherwise should be removed.]
        eprintln!(
            "Sending new stats at least.. {:?}, took {}ms",
            new_process_stats,
            new_process_stats.measurement_time.as_millis()
        );
        tx.send_replace(new_process_stats);
    }
```

bin/propolis-server/src/lib/stats/process.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    stats.vsz = info.pr_size;
+    stats.rss = info.pr_rssize;
+
+    Ok(stats)
+}
+
+async fn process_stats_task(tx: watch::Sender<ProcessStats>) {
+    while !tx.is_closed() {
+        let new_process_stats = tokio::task::spawn_blocking(process_stats)
```
hawkw, line 113: why not use `tokio::fs` rather than `spawn_blocking`ing? not a big deal as `tokio::fs` is essentially `spawn_blocking` anyway, but might be a bit nicer...

bin/propolis-server/src/lib/stats/process.rs
```diff
@@ -0,0 +1,129 @@
...
+            .expect("collecting address space stats does not panic")
+            .expect("process_stats() does not error");
+
+        // [there isn't a nice way to plumb a slog Logger here, so this is an
+        // eprintln for demonstrative purposes but otherwise should be removed.]
+        eprintln!(
+            "Sending new stats at least.. {:?}, took {}ms",
+            new_process_stats,
+            new_process_stats.measurement_time.as_millis()
+        );
```
hawkw, line 124: hmm, is it really that painful to pass a `Logger` into `ServerStats::new`?

bin/propolis-server/src/lib/stats/process.rs
```diff
@@ -0,0 +1,129 @@
...
+    let mut stats =
+        ProcessStats { measurement_time: Duration::ZERO, vsz: 0, rss: 0 };
+
+    let mut info: psinfo = FromZeroes::new_zeroed();
+
+    let stats_read_start = Instant::now();
+
+    psinfo_file
+        .read(info.as_bytes_mut())
+        .context("reading struct psinfo from file")?;
```
hawkw, line 101: cute use of `zerocopy` here :3

bin/propolis-server/src/lib/stats/process.rs
```diff
@@ -0,0 +1,129 @@
...
+type timestruc_t = timespec_t;
+/// From `sys/types.h`
+#[allow(non_camel_case_types)]
+type dev_t = u64;
+
+/// `psinfo`'s definition depends on the data model reported by illumos. This is
+/// in line with the 64-bit version of the struct, and ignores a few fields at
+/// the end of the struct.
+#[derive(Copy, Clone, Debug, FromZeroes, AsBytes, FromBytes)]
+#[cfg(target_arch = "x86_64")]
```
hawkw, line 49: shouldn't the whole module have this `cfg`? none of the code that references `psinfo`, which is ... all of it ... will compile if this isn't there...

bin/propolis-server/src/lib/stats/mod.rs
```diff
@@ -78,15 +82,80 @@ struct ServerStatsInner {
...
     run_count: virtual_machine::Reset,
+
+    /// Mapped virtual memory for the Propolis managing this instance, excluding
+    /// VM memory.
+    vmm_vsz: virtual_machine::VmmVsz,
+
+    /// Process RSS is sampled in a standalone task. The `tokio::sync::watch`
+    /// here is to observe that sampling and update the `vmm_vsz` tracked
+    /// here.
+    process_stats_rx: tokio::sync::watch::Receiver<process::ProcessStats>,
```
hawkw, line 102: I'm not convinced that this actually needs to be a `tokio::sync::watch` --- a `watch` channel is basically just an `Arc`ed read-write lock plus a thing for notifying the readers when the value inside the `RwLock` changes, and we never use `watch::Receiver::changed` in this code, just `borrow()`. So, it's an `Arc<RwLock<ProcessStats>>` with extra overhead for functionality we're not using here...and, because there's only a single reader and a single writer, it may as well be a `Mutex` instead of a `RwLock`.

I would also maybe consider just stuffing `vsz` in an `AtomicUsize`, but Patrick will probably disagree with me...

bin/propolis-server/src/lib/stats/mod.rs
hawkw: ON FURTHER CONSIDERATION: I have been reminded that the lock around `VmObjects` is a `RwLock` rather than a `Mutex`, so we're not contending with other readers, and writes only occur on VM initialization/rundown. So...this probably isn't worth the effort.

bin/propolis-server/src/lib/stats/process.rs
iximeow: on the other hand, nothing else here is dependent on the target architecture - i was kind of thinking that given an aarch64 target or something we'd be able to fill in the right struct for `psinfo` on that arch. but, yeah, as-is this just breaks even compiling `propolis-server` on other architectures.

i'll `cfg(target_arch = "x86_64")` psinfo read and have it return 0 with a `warn!()` on other architectures. seeing that would imply running `propolis-server` on non-x86 targets which would be super neat..

hawkw: If memory serves, there's other stuff that will just panic on any other architecture (certainly I believe all the CPUID stuff will), but that seems fine to me!

PR #895, opened 2025-04-14
add digests to falcon artifacts

rcgoodfellow:


gjcolombo (approved):
I probably should've noticed this the first time--sorry about the extra lap!

PR #896, opened 2025-04-14
client: prepare instance spec types for use in sled-agent API

gjcolombo:
The next step toward implementing virtual platforms in Omicron is to move instance spec construction from sled-agent up to Nexus. This requires instance spec types to appear in sled-agent's API. To allow this:

- Direct the generated client to add a `schemars::JsonSchema` derive to all generated types.
- Add `rename = snake_case` attributes to instance spec enum types that sled-agent expects to reuse. Omicron requires all of its API specifications to pass the OpenAPI linting rules defined in the openapi-lint crate (https://github.com/oxidecomputer/openapi-lint); among these is a requirement that enum variants all have snake case names.

Also add a note explaining special rules for the use of the `SpecKey` type (generated clients that include this type need to make sure that they furnish the correct Serialize implementation for it).

Tests: cargo test; PHD; verified that it's possible to write the desired Omicron changes using this change; verified that the Falcon repo continues to build with its propolis-client dependency pointed to this commit.


gjcolombo:
The `migrate_from_base` failure is expected; the new PHD client is using snake-case type definitions that the old server doesn't expect.

Also, for the record: I looked into writing a test that runs the OpenAPI linter over the entire propolis-server API document (instead of just cherry-picking fixes for errors in sled-agent). This test is trivial to write, but it immediately turns up several dozen errors that we'd have to go fix, so I've punted that to another PR. (I may come to regret this as I keep working on my sled-agent changes...)

crates/propolis-api-types/src/instance_spec/mod.rs
```diff
@@ -191,6 +191,13 @@ pub mod v0;
 // This type derives `SerializeDisplay` and `DeserializeFromStr` so that it can
 // be used as a map key when serializing to JSON, which requires strings (and
 // not objects) as keys.
+//
+// WARNING: Progenitor-generated versions of this type will use serde's default
+// derived Serialize and Deserialize implementations, not the serde_with
+// implementations. This can cause clients to serialize spec keys in a form that
+// the server can't deserialize. To get around this, Progenitor clients that
+// use spec keys should use a `replace` directive to replace the generated
+// type with this type.
```
hawkw, line 200: It occurs to me that there might be an alternative solution to this that doesn't require "WARNING: remember to do this extra stuff when generating clients": we are presently deriving `JsonSchema`, and I assume the generated schema is for a tagged enum representation. Could we, instead, add `serde(untagged)` and hand-implement `JsonSchema` for `SpecKey`, the way [we do for `NameOrId` in `omicron-common`](https://github.com/oxidecomputer/omicron/blob/aeb27f08c883c04360883fbb9ffd5f1222ceeede/common/src/api/external/mod.rs#L394-L414)? That way, I think we _should_ instead generate an OpenAPI spec that represents this as just a String but with hints suggesting it could be a UUID. I'm not 100% sure that will get us the desired behavior but it seems worth looking into.

hawkw (commented):
I had one thought about potentially changing the schema for `SpecKey`, but otherwise, this seems good to me!

crates/propolis-api-types/src/instance_spec/mod.rs
gjcolombo: I am embarrassed to admit that I did not see the parallel between this and `NameOrId`! I'll give this a try.

gjcolombo: I was able to get this to work in fc7a85f. It didn't exactly spark joy, but it does get rid of the `serde_with` dependency and makes things slightly nicer for sled-agent, so I think I'm leaning toward going with it anyway. Would be interested to hear what you think!

hawkw: I'll take a look!

crates/propolis-api-types/src/instance_spec/mod.rs
```diff
@@ -187,29 +186,20 @@ pub mod v0;
...
+// Direct serde to use an untagged enum representation for this type. Since both
+// Uuid and String serialize to strings, this allows other types that contain a
+// Map<K = SpecKey> to derive Serialize and successfully serialize to JSON.
+// (This doesn't work with a tagged representation because JSON doesn't allow
+// maps to be used as map keys.)
+//
+// Note that this makes the order of variants matter: serde will pick the first
+// variant into which it can successfully deserialize an untagged enum value,
+// and the point is to use the UUID representation for any value that can be
+// interpreted as a UUID.
```
hawkw, line 201: Ugh, that's ugly, but...not the end of the world, I guess... Perhaps we should add a test that would fail if someone changed the order of the variants, just in case?

crates/propolis-api-types/src/instance_spec/mod.rs
```diff
@@ -217,10 +215,7 @@ impl std::fmt::Display for SpecKey {
 }
 
 impl std::str::FromStr for SpecKey {
-    // This conversion is infallible, but the error type needs to implement
-    // `Display` for `SpecKey` to derive `DeserializeFromStr`.
-    type Err = &'static str;
-
+    type Err = ();
```
hawkw, line 218: should this be
```suggestion
    type Err = core::convert::Infallible;
```

crates/propolis-api-types/src/instance_spec/mod.rs
```diff
@@ -244,6 +239,62 @@ impl From<Uuid> for SpecKey {
...
+                            .to_string(),
+                    ),
+                    ..Default::default()
+                }
+                .into(),
+            ),
+            subschemas: Some(Box::new(SubschemaValidation {
+                one_of: Some(vec![
+                    label_schema("uuid", generator.subschema_for::<Uuid>()),
+                    label_schema("name", generator.subschema_for::<String>()),
```
hawkw, line 288: lovely

crates/propolis-config-toml/src/spec.rs
```diff
@@ -71,7 +71,7 @@ pub enum TomlToSpecError {
 #[derive(Clone, Debug, Default)]
 pub struct SpecConfig {
     pub enable_pcie: bool,
-    pub components: BTreeMap<SpecKey, ComponentV0>,
+    pub components: HashMap<SpecKey, ComponentV0>,
```
hawkw, line 74: why change this to `HashMap`? not a blocker, just courious

gjcolombo: I think this must have leaked from an earlier revision of this change. (This *can* matter because `propolis_client::types::InstanceSpecV0`'s `components` member is a `HashMap`, whereas the `propolis_api_types` version of the type is a BTreeMap.) But changing this back doesn't seem to break anything in the current version of the change, so I'll flip it back just to reduce the number of lines this churns.

crates/propolis-api-types/src/instance_spec/mod.rs
gjcolombo: The existing unit tests in this module cover this (past-me looking out for present-me for once!). Specifically they check that (a) serializing a `SpecKey::Uuid` and deserializing it produces another `SpecKey::Uuid`, and (b) serializing a `SpecKey::Name(Uuid::new_v4().to_string())` and deserializing that also gives you a `SpecKey::Uuid`. Both tests break if the `Uuid` variant isn't first in this enum.

hawkw (approved):
looks good to me

PR #897, opened 2025-04-16
Use latest Omicron zone packaging

pfmooney:
This updates our use of `omicron-zone-package` to the latest version, dropping a bunch of outdated deps from the tree in the process.

It should not be merged until oxidecomputer/omicron-package#75 is addressed, since the OVMF and Alpine blobs are otherwise missing from the resulting tarball.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -184,6 +184,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayref"
+version = "0.3.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "76a2e8124351fda1ef8aaaa3bbd7ebbcb486bbcd4225aca0aa0d84bb2db8fecb"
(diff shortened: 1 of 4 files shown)


pfmooney:
With omicron-zone-package 0.12.1 now published to crates.io, this should be ready to go.

pfmooney:
Testing notes:

Running the existing `propolis-package` and the "new" one (built from this change) yielded output tarballs which were practically identical.  The only difference (as checked with `rsync` between extracted directories) was the time attributes on the blob directory.
```
sending incremental file list
.d..t...... ./
.d..t...... root/opt/oxide/propolis-server/blob/

sent 538 bytes  received 37 bytes  1,150.00 bytes/sec
total size is 113,749,506  speedup is 197,825.23 (DRY RUN)
```

gjcolombo (approved):

PR #899, opened 2025-04-18
client: re-export instance spec types for use in sled-agent

gjcolombo:
Follow-up to #896. Using generated instance spec types in sled-agent's API has some substantial downsides:

1. Some component types (e.g. storage backends) have fields that we'd like to redact when logging their contents using their `Debug` impls. The native versions of these types manually impl `Debug` for these types, but their generated counterparts always derive Debug (and AFAIK there's no way to ask Progenitor not to add this derive).
2. The Dropshot-generated OpenAPI doc for sled-agent's API winds up having very gnarly type descriptions that include fully-escaped versions of the Progenitor-generated doc comments from propolis-client.[^1]

To get around these problems:

- Add a propolis-client submodule that re-exports all instance spec-related types from `propolis_api_types`.
- Add `replace` directives to the generated client that replace the generated versions of `InstanceSpecV0`, `VersionedInstanceSpec`, and `ReplacementComponent` with their re-exported equivalents.

This way, sled-agent can use the re-exported spec types internally and in its API. This reuses their manual `impl`s when they exist (solving problem 1) and avoids presenting generated JSON schema documentation when generating the sled-agent API document (solving problem 2).

Remove the `JsonSchema` implementations for generated spec types, since they're no longer needed. 

Finally, fix a one-line clippy nit that started showing up alongside these changes.

Tests: cargo test and PHD; picked up this change into oxidecomputer/omicron#8002 and verified that (a) that change still works (and that it produces a much cleaner sled-agent API document), and (b) log messages from sled-agent redact spec components that we want to be redacted.

[^1]: Concretely: suppose you have type `foo::Widget` with a doc comment, and you have a `foo_server` crate with a Dropshot API that makes use of this type. If you generate an OpenAPI document from this API, and then use Progenitor to generate a client from that document, `foo_client::types::Widget` will have a doc comment that includes both the original `foo::Widget` doc comment and a Progenitor-generated description of the type's JSON schema. This is all well and good and what we expect. Now suppose you then create crate `bar_server` and include `foo_client::types::Widget` in its Dropshot API. The description of `Widget` in `bar_server`'s API document will contain both the original `foo::Widget` description *and* the generated JSON schema text from `foo_client::types::Widget` in the type's "description" field. And if you then generate `bar_client::types::Widget`, the schema will be generated *again* and re-included in the generated type's doc comment. Needless to say: yuck.

### DIFF
--- a/crates/propolis-api-types/src/instance_spec/mod.rs
+++ b/crates/propolis-api-types/src/instance_spec/mod.rs
@@ -217,10 +217,16 @@ impl std::fmt::Display for SpecKey {
 impl std::str::FromStr for SpecKey {
     type Err = core::convert::Infallible;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Ok(match Uuid::parse_str(s) {
+        Ok(s.into())
+    }
+}
+
+impl From<&str> for SpecKey {
+    fn from(s: &str) -> Self {
+        match Uuid::parse_str(s) {
             Ok(uuid) => Self::Uuid(uuid),
             Err(_) => Self::Name(s.to_owned()),
-        })
+        }
     }
 }
 
--- a/lib/propolis-client/src/lib.rs
+++ b/lib/propolis-client/src/lib.rs
@@ -4,9 +4,28 @@
 
 //! A client for the Propolis hypervisor frontend's server API.
 
-// Re-export types from propolis_api_types where callers may want to use
-// constructors or From impls.
-pub use propolis_api_types::instance_spec::{PciPath, SpecKey};
+/// Re-exports of types related to instance specs.
+///
+/// These types are re-exported for the convenience of components like
+/// sled-agent that may wish to expose instance specs in their own APIs.
+/// Defining the sled-agent API in terms of these "native" types allows
+/// sled-agent to reuse their trait implementations (and in particular use
+/// "manual" impls of things that Progenitor would otherwise derive).
+///
+/// In the generated client, the native "top-level" instance spec and component
+/// types ([`VersionedInstanceSpec`], [`InstanceSpecV0`], and
+/// [`ReplacementComponent`]) replace their generated counterparts. This
+/// obviates the need to maintain `From` impls to convert between native and
+/// generated types.
+pub mod instance_spec {
+    pub use propolis_api_types::instance_spec::{
+        components::{backends::*, board::*, devices::*},
+        v0::*,
+        *,
+    };
+
+    pub use propolis_api_types::ReplacementComponent;
+}
 
 // Re-export Crucible client types that appear in their serialized forms in
 // instance specs. This allows clients to ensure they serialize/deserialize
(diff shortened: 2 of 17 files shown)


gjcolombo:
An explanation of ffeb813: `libc` defines the `F_WRLCK` flag as an i16 on `solarish` OSes but defines it as an i32 for Linux. For some reason, starting with this change, my local rust-analyzer-initiated Clippy runs started complaining about this seemingly-unnecessary cast. The mystery here (from my perspective) is not so much why the cast is there (you definitely need it on Linux!) as why my local dev environment started screeching about it today.

In any case, ffeb813 fixes the Linux build; I'll figure out what's going on with my dev environment separately. (I also see this on the latest master, fwiw.)

crates/propolis-api-types/src/instance_spec/mod.rs
```diff
@@ -217,10 +217,16 @@ impl std::fmt::Display for SpecKey {
 impl std::str::FromStr for SpecKey {
     type Err = core::convert::Infallible;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Ok(match Uuid::parse_str(s) {
+        Ok(s.into())
+    }
+}
+
+impl From<&str> for SpecKey {
```
hawkw, line 224: nice :)

lib/propolis-client/src/lib.rs
```diff
@@ -19,106 +38,21 @@ progenitor::generate_api!(
...
-        CpuidVendor = { derives = [ schemars::JsonSchema ]},
-        DlpiNetworkBackend = { derives = [ schemars::JsonSchema ]},
-        FileStorageBackend = { derives = [ schemars::JsonSchema ]},
-        GuestHypervisorInterface = { derives = [ schemars::JsonSchema ]},
-        I440Fx = { derives = [ schemars::JsonSchema ]},
-        HyperVFeatureFlag = { derives = [ schemars::JsonSchema ]},
-        MigrationFailureInjector = { derives = [ schemars::JsonSchema ]},
-        NvmeDisk = { derives = [ schemars::JsonSchema ]},
+        BootSettings = { derives = [ Default] },
+        CpuidEntry = { derives = [ PartialEq, Eq, Copy ]},
```
hawkw, line 52: obnoxious whitespace nits:
```suggestion
        BootSettings = { derives = [ Default ] },
        CpuidEntry = { derives = [ PartialEq, Eq, Copy ] },
```
(rustfmt presumably refuses to touch the macro...)

phd-tests/framework/src/test_vm/spec.rs
```diff
@@ -82,8 +85,12 @@ impl VmSpec {
...
 
             let backend_spec = disk.backend_spec();
-            let backend_name =
-                disk.device_name().clone().into_backend_name().into_string();
+            let backend_name = disk
+                .device_name()
+                .clone()
+                .into_backend_name()
+                .into_string()
+                .into();
```
hawkw, line 93: wow... :face_with_spiral_eyes:

hawkw (approved):
This approach makes sense to me, no major concerns!

PR #901, opened 2025-04-28
upgrade omicron-zone-package to 0.12.2

gjcolombo:
This fixes an issue with adding blob paths to zone archives that would otherwise prevent Propolis zones from booting.


smklein (approved):

gjcolombo:
Once the image job is done I'll check this in a local Omicron cluster and make sure it works before merging.

pfmooney (approved):
Thanks for fixing this.

PR #904, opened 2025-05-05
Bump MAX_PHYSMAP limit again

pfmooney:


iximeow (approved):
i looked around for where `MAX_PHYSMAP` gets used and this seems good, thanks.

PR #905, opened 2025-05-05
Fix up rustdoc warnings

pfmooney:
I noticed some warning output while looking at the output from some GHA jobs.


gjcolombo (approved):
I think just about all of these are mine--thanks for taking care of them!

PR #912, opened 2025-06-21
Support guest memory sizes >=1028 GiB

### LINKED ISSUE #907: VMs with >1TiB of memory confuse EDK2
@askfongjojo saw as disks no longer existing when starting larger instances in https://github.com/oxidecomputer/omicron/issues/7918#issuecomment-2882801985 .

the instance had two disks, but from the UEFI shell saw only the cloud-init disk:

```
UEFI Interactive Shell v2.2
EDK II
UEFI v2.70 (EDK II, 0x00010000)
Mapping table
      FS0: Alias(s):F0:;BLK0:
          PciRoot(0x0)/Pci(0x18,0x0)
Press ESC in 1 seconds to skip startup.nsh or any other key to continue.
Shell> 
```

`debug.out` was informative here:
```
PlatformBootManagerAfterConsole
Found Mass Storage device: PciRoot(0x0)/Pci(0x10,0x0)
NvmExpressDriverBindingStart: start
NvmeControllerInit: the controller doesn't support NVMe command set
NvmExpressDriverBindingStart: end with Unsupported
Found Mass Storage device: PciRoot(0x0)/Pci(0x11,0x0)
NvmExpressDriverBindingStart: start
NvmeControllerInit: the controller doesn't support NVMe command set
NvmExpressDriverBindingStart: end with Unsupported
Found Mass Storage device: PciRoot(0x0)/Pci(0x18,0x0)
```

which implies that `Cap.Css` was not `0x01` in [NvmeControllerInit](https://github.com/oxidecomputer/edk2/blob/907a5fd1763ce5ddd74001261e5b52cd200a25f9/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressHci.c#L792-L795).

skipping ahead a bit, it's clear that OVMF's view of `Mem64` is bogus:
```
PciHostBridge: NotifyPhase (AllocateResources)
 RootBridge: PciRoot(0x0)
  Mem64: Base/Length/Alignment = 800000000/100000/FFFFF - Success
  Mem: Base/Length/Alignment = C0000000/100000/FFFFF - Success
  I/O: Base/Length/Alignment = C000/1000/FFF - Success
```
there should not be MMIO against 0x8_0000_0000, that's 32 GiB and is just... memory.

there's another odd line in OVMF debug output: `GetFirstNonAddress: Pci64Base=0x800000000 Pci64Size=0x800000000`, `Pci64Size` is a PCD that defaults to 0x8_0000_0000, how `Pci64Base` ends up at that same value is a bit awkward. the log comes from [`GetFirstNonAddress`](https://github.com/oxidecomputer/edk2/blob/907a5fd1763ce5ddd74001261e5b52cd200a25f9/OvmfPkg/Bhyve/PlatformPei/MemDetect.c#L227-L228). `Pci64Base` first starts its life as `FirstNonAddress` from
```
FirstNonAddress = BASE_4GB + GetSystemMemorySizeAbove4gb ();
```
the result here is [rounded up twice](https://github.com/oxidecomputer/edk2/blob/907a5fd1763ce5ddd74001261e5b52cd200a25f9/OvmfPkg/Bhyve/PlatformPei/MemDetect.c#L205-L214). first, to 1G, then to the power of two that would contain `Pci64Size`. since that *is* a power of two, it's just rounded up to 0x8_0000_0000. to get here, `FirstNonAddress` could be any value less than 0x8_0000_0000`.

[`GetSystemMemorySizeAbove4gb`](https://github.com/oxidecomputer/edk2/blob/907a5fd1763ce5ddd74001261e5b52cd200a25f9/OvmfPkg/Bhyve/PlatformPei/MemDetect.c#L130-L150) is where it all goes off the rails. `CMOS[0x5b..0x5d]` provide three bytes of the amount of memory in the system above 4G. 1T in units of 64k wraps here:
```
>>> hex((1024 * (1 << 30)) >> 16)
'0x1000000'
```
conspicuously above 24 bits. in Propolis we'll dutifully populate the low 24 bits, and miss that [`high[3]` here](https://github.com/oxidecomputer/propolis/blob/master/lib/propolis/src/hw/bhyve/rtc.rs#L95-L102) is non-zero.

iximeow:
Fixes #907.

This change both produces an E820 table for OVMF to consult for highmem layout, and clamps the RTC CMOS bytes through which we communicate highmem size to OVMF. The E820 table supplants the RTC CMOS for OVMF's purposes, so in theory nothing will check those bytes anymore. Even so, clamping to them to `ffffff` seems more legible.

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -1112,6 +1112,28 @@ impl MachineInitializer<'_> {
         smb_tables.commit()
     }
 
+    fn generate_e820(&self) -> Result<Entry, MachineInitError> {
+        info!(self.log, "Generating E820 map for guest address space");
+
+        let mut e820_table = fwcfg::formats::E820Table::new();
+
+        for (addr, len, kind) in self.machine.map_physmem.mappings().into_iter()
+        {
+            let addr = addr.try_into().context("usize should fit into u64")?;
+            let len = len.try_into().context("usize should fit into u64")?;
+            match kind {
+                propolis::vmm::MapType::Dram => {
+                    e820_table.add_mem(addr, len);
+                }
+                _ => {
+                    e820_table.add_reserved(addr, len);
+                }
+            }
+        }
+
+        Ok(e820_table.finish())
+    }
+
     fn generate_bootorder(&self) -> Result<Option<Entry>, MachineInitError> {
         info!(
             self.log,
@@ -1212,6 +1234,10 @@ impl MachineInitializer<'_> {
                 MachineInitError::FwcfgInsertFailed("bootorder", e)
             })?;
         }
+        let e820_entry = self.generate_e820()?;
+        fwcfg
+            .insert_named("etc/e820", e820_entry)
+            .map_err(|e| MachineInitError::FwcfgInsertFailed("e820", e))?;
 
         let ramfb = ramfb::RamFb::create(
             self.log.new(slog::o!("component" => "ramfb")),
--- a/lib/propolis/src/hw/bhyve/rtc.rs
+++ b/lib/propolis/src/hw/bhyve/rtc.rs
@@ -94,7 +94,14 @@ impl BhyveRtc {
 
         // High-mem
         if high_mem > 0 {
-            let high = ((high_mem / CHUNK) as u32).to_le_bytes();
+            // If high_mem is 1TiB or larger, the division below produces a
+            // number that overflows the 24 bits available at
+            // `CMOS_OFF_MEM_HIGH`. Clamp the value so guests aren't subjected
+            // to arbitrary wrapping. OVMF is told about the highmem layout via
+            // E820 table anyway, so the only thing that might care about these
+            // bytes are guest OSes that check the RTC CMOS bytes directly.
+            let chunks = std::cmp::min((high_mem / CHUNK) as u32, 0xffffffu32);
+            let high = chunks.to_le_bytes();
 
             hdl.rtc_write(CMOS_OFF_MEM_HIGH, high[0])?;
             hdl.rtc_write(CMOS_OFF_MEM_HIGH + 1, high[1])?;
--- a/lib/propolis/src/hw/qemu/fwcfg.rs
+++ b/lib/propolis/src/hw/qemu/fwcfg.rs
@@ -1057,6 +1057,158 @@ mod test {
 pub mod formats {
     use super::Entry;
     use crate::hw::pci;
+    use zerocopy::{Immutable, IntoBytes};
+
+    /// A type for a range described in an E820 map entry.
+    ///
+    /// This is canonically defined as the ACPI "Address Range Types", though we
+    /// only define the types we use, which are a subset of the types that EDK2
+    /// is known to care about, which itself is a subset of types that ACPI and
+    /// OEMs define or guest OSes may care about.
+    #[derive(IntoBytes, Immutable)]
+    #[repr(u32)]
+    enum EfiAcpiMemoryType {
+        Memory = 1,
+        Reserved = 2,
+        // For reference, though these types are unused.
+        // Acpi = 3,
+        // Nvs = 4,
+    }
+
+    /// One address/length/type entry in the E820 map.
+    ///
+    /// This is... almost defined by ACPI's "Address Range Descriptor Structure"
+    /// table, under "INT 15H, E820". Critically, ACPI defines this structure
+    /// with an additional "Extended Attributes" field which EDK2 does not know
+    /// about and so we do not provide. Consequently the size of this struct is
+    /// 20 bytes as defined in `OvmfPkg/Include/IndustryStandard/E820.h` rather
+    /// than the ACPI definition's 24 bytes.
+    #[derive(IntoBytes, Immutable)]
+    #[repr(C, packed)]
+    struct E820Entry64 {
+        base_addr: u64,
+        length: u64,
+        ty: EfiAcpiMemoryType,
+    }
+
+    /// A list of E820 memory map entries.
+    ///
+    /// This is not defined by ACPI, but is an EDK2 implementation of a QMEU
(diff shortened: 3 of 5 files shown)


bin/propolis-server/src/lib/initializer.rs
```diff
@@ -1212,6 +1234,10 @@ impl MachineInitializer<'_> {
                 MachineInitError::FwcfgInsertFailed("bootorder", e)
             })?;
         }
+        let e820_entry = self.generate_e820()?;
+        fwcfg
+            .insert_named("etc/e820", e820_entry)
+            .map_err(|e| MachineInitError::FwcfgInsertFailed("e820", e))?;
```
iximeow, line 1240: I don't really see a reason to make this optional, at least. Regardless of offering up an e820 map the guest OS will use UEFI service and get the map that OVMF builds of the system.

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -1057,6 +1057,158 @@ mod test {
...
+        base_addr: u64,
+        length: u64,
+        ty: EfiAcpiMemoryType,
+    }
+
+    /// A list of E820 memory map entries.
+    ///
+    /// This is not defined by ACPI, but is an EDK2 implementation of a QMEU
+    /// construct to communicate an E820 map to the firmware. This table is not
+    /// directly presented to guest OSes; it is parsed by EDK2 and added to its
```
iximeow, line 1098: `This table is not directly presented to guest OSes` is poor phrasing as I read it back - I'm sure it's legible somehow if a guest OS really wanted to look for fwcfg bits.

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -1057,6 +1057,158 @@ mod test {
 pub mod formats {
     use super::Entry;
     use crate::hw::pci;
+    use zerocopy::{Immutable, IntoBytes};
```
iximeow, line 1060: this is like a hundred lines away from being large enough I'd kick it over to a `formats.rs`, but not quiiiite there yet..

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -1112,6 +1112,28 @@ impl MachineInitializer<'_> {
...
     }
 
+    fn generate_e820(&self) -> Result<Entry, MachineInitError> {
+        info!(self.log, "Generating E820 map for guest address space");
+
+        let mut e820_table = fwcfg::formats::E820Table::new();
+
+        for (addr, len, kind) in self.machine.map_physmem.mappings().into_iter()
+        {
+            let addr = addr.try_into().context("usize should fit into u64")?;
```
iximeow, line 1122: OVMF doesn't care about entries <4GiB, but I feel like it's slightly more useful to everyone to report lowmem ranges here anyway. I want to refresh myself on how PCIe MMIO ranges work out but given that I've run a guest with an NVMe disk it's presumably not _wrong_.

lib/propolis/src/hw/bhyve/rtc.rs
```diff
@@ -94,7 +94,14 @@ impl BhyveRtc {
 
         // High-mem
         if high_mem > 0 {
-            let high = ((high_mem / CHUNK) as u32).to_le_bytes();
+            // If high_mem is 1TiB or larger, the division below produces a
+            // number that overflows the 24 bits available at
+            // `CMOS_OFF_MEM_HIGH`. Clamp the value so guests aren't subjected
+            // to arbitrary wrapping. OVMF is told about the highmem layout via
+            // E820 table anyway, so the only thing that might care about these
+            // bytes are guest OSes that check the RTC CMOS bytes directly.
```
pfmooney, line 102: Maybe include mention that this clamping fails at 256TiB (where we roll over the 2^32 * 64KiB)?

lib/propolis/src/hw/qemu/fwcfg.rs
```diff
@@ -1057,6 +1057,158 @@ mod test {
...
+
+    /// A list of E820 memory map entries.
+    ///
+    /// This is not defined by ACPI, but is an EDK2 implementation of a QMEU
+    /// construct to communicate an E820 map to the firmware. It is parsed by
+    /// EDK2 and added to its EFI memory map; it is not, itself, the memory map
+    /// that OVMF presents via UEFI services. It is not required to be sorted,
+    /// and EDK2 ignores entries starting below 4 GiB. Adding additional
+    /// low-memory entries is not harmful, but not valuable to EDK2 either.
+    pub struct E820Table(Vec<E820Entry64>);
```
pfmooney, line 1102: I'd be tempted to just publicly expose the `E820Entry64` and `EfiAcpiMemoryType` types, and providing a `From <&[E820Entry64]> for Entry` impl for the conversion, since `E820Table` isn't really providing much in the way of value-add over just shimming the Vec methods.  Up to you. *shrug*

pfmooney (commented):
The broad strokes look good.

bin/propolis-server/src/lib/initializer.rs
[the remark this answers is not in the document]
iximeow: yeah i was 50/50 here since `MachineInitError::GenericError` pretty naturally holds this, but after the fact realizing that it's more for doing something useful with `std::io::Error` it's kinda rude to cram impossible conversions in there. switched this over.

lib/propolis/src/hw/bhyve/rtc.rs
iximeow: moved the peas around the plate instead (clamp to u32::MAX before converting to u32 so the only wrapping here is when highmem is more than 64b)

lib/propolis/src/hw/qemu/fwcfg.rs
iximeow: i should probably examine my dislike of `From`/`Into` one day, but i wonder if some of it is finding this easier to search for..

PR #913, opened 2025-06-26
server: build Tokio runtime using `oxide-tokio-rt`

hawkw:
As described in [RFD 579], we are now recommending that the [`oxide-tokio-rt`] crate be used to construct the Tokio runtime in all production software that uses Tokio. This crate provides common functions for constructing Tokio runtimes with the recommended configurations.

In particular, it currently does the following:

- On illumos, configures the runtime to emit DTrace probes, using [`tokio-dtrace`].
- Disables Tokio's LIFO slot optimization. This feature is intended to improve message-passing latency, but because tasks in the LIFO slot do not currently participate in work-stealing, it can result in extreme latency spikes in some cases (see omicron#8334 for a worked example).

Thus, this commit changes `propolis-server` to construct its Tokio runtimes using `oxide-tokio-rt` so that these common recommended configurations are included. This requires enabling Tokio's unstable features in the workspace.

[RFD 579]: https://rfd.shared.oxide.computer/rfd/0579
[`oxide-tokio-rt`]: https://github.com/oxidecomputer/oxide-tokio-rt
[`tokio-dtrace`]: https://github.com/oxidecomputer/tokio-dtrace

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -11,5 +11,5 @@ CARGO_WORKSPACE_DIR = { value = "", relative = true }
 # disabling the LIFO slot optimization.
 #
 # See here for details:
-# https://rfd.shared.oxide.computer/rfd/0579#_enabling_tokio_unstable_features
+# https://github.com/oxidecomputer/oxide-tokio-rt/blob/main/README.md#enabling-tokio_unstable-features
 rustflags = ["--cfg", "tokio_unstable"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,3 +5,6 @@ xtask = "run --package xtask --quiet --"
 # Currently required by Falcon due to
 # https://github.com/rust-lang/cargo/issues/3946#issuecomment-973132993
 CARGO_WORKSPACE_DIR = { value = "", relative = true }
+
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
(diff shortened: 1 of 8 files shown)


pfmooney (commented):
Apply it to standalone as well?

hawkw:
The Clippy thing seems like a new lint that's unrelated to this change: https://github.com/oxidecomputer/propolis/actions/runs/15913312178/job/44885677202?pr=913#step:5:1947

.cargo/config.toml
```diff
@@ -7,4 +7,9 @@ xtask = "run --package xtask --quiet --"
 CARGO_WORKSPACE_DIR = { value = "", relative = true }
 
 [build]
+# Tokio's unstable features are required by `tokio-dtrace` probes, and for
+# disabling the LIFO slot optimization.
+#
+# See here for details:
+# https://rfd.shared.oxide.computer/rfd/0579#_enabling_tokio_unstable_features
```
pfmooney, line 14: This would be a good reason to make the RFD public.

pfmooney (approved):

hawkw:
I'm gonna go clean up all the Clippy mess in a separate PR.

hawkw:
We _could_ also consider configuring Clippy to warn on all uses of `#[tokio::main]` in the repo, as I suggested [here](https://github.com/oxidecomputer/oxide-tokio-rt/blob/main/README.md#warning-on-use-of-tokiomain). I opted not to do this, since I felt like the remaining uses of `#[tokio::main]` in xtasks, examples, and `propolis-cli`, were all basically fine and it didn't feel necessary to add explicit `#[expect]` attributes for the lint on all of them. If others disagree, I'm happy to add this.

PR #914, opened 2025-06-27
Fix new lints

hawkw:
This branch fixes new Clippy and rustc lints that showed up in #913. In
particular, there are two warnings that needed to be fixed:

- `clippy::uninlined_format_args` (fixed in
  223d07221557f39170da09dad6d66079bf79915f): Clippy now suggests that
  all format strings with format arguments that could be inlined should
  be inlined. This is a very mechanical change, but the lint triggers on
  basically every format string in the project. Sigh.
- `dangerous_implicit_autorefs` (fixed in
  dc93665d20e2e2aee69e48622d957c17c9dddd0c). This is a new [compiler
  warning] that warns about the creation of implicit references to the
  dereference of a raw pointer, which may be invalid if the intermediate
  pointer is null or is not well-aligned. This one is actually somewhat
  interesting and requires some care.
  
  In this case, I've fixed this by using `ptr::addr_of!` and
  `slice::from_raw_parts` to construct the slice into `ma_addr`,
  avoiding the creation of temporary references through the raw pointer.
  In this case, the suggestions from the lint of just adding an
  additional explicit `&` reference as a way of saying "this is okay,
  because I know the pointer is non-null and well-aligned" would
  *probably* have been fine, but this felt less sketchy.

[compiler warning]:
    https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/autorefs/static.DANGEROUS_IMPLICIT_AUTOREFS.html

### DIFF
--- a/bin/propolis-server/src/lib/stats/virtual_machine.rs
+++ b/bin/propolis-server/src/lib/stats/virtual_machine.rs
@@ -495,9 +495,9 @@ mod test {
                 )
                 .unwrap_or_else(|| {
                     panic!(
-                        "kstat state '{kstat_state}' did not map to an oximeter state, \
-                        which it should have done. Did that state get \
-                        mapped to a new oximeter-level state?"
+                        "kstat state '{kstat_state}' did not map to an \
+                        oximeter state, which it should have done. Did that \
+                        state get mapped to a new oximeter-level state?"
                     )
                 });
                 *observed_states.entry(oximeter_state).or_default() += count;
--- a/bin/propolis-server/src/lib/stats/virtual_machine.rs
+++ b/bin/propolis-server/src/lib/stats/virtual_machine.rs
@@ -495,10 +495,9 @@ mod test {
                 )
                 .unwrap_or_else(|| {
                     panic!(
-                        "kstat state '{}' did not map to an oximeter state, \
+                        "kstat state '{kstat_state}' did not map to an oximeter state, \
                         which it should have done. Did that state get \
-                        mapped to a new oximeter-level state?",
-                        kstat_state
+                        mapped to a new oximeter-level state?"
                     )
                 });
                 *observed_states.entry(oximeter_state).or_default() += count;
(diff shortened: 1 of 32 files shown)


bin/propolis-server/src/lib/stats/virtual_machine.rs
```diff
@@ -495,10 +495,9 @@ mod test {
                 )
                 .unwrap_or_else(|| {
                     panic!(
-                        "kstat state '{}' did not map to an oximeter state, \
+                        "kstat state '{kstat_state}' did not map to an oximeter state, \
```
iximeow, line 498: this crests well past 80 columns but i don't think rustfmt knows how to reflow strings like this (sorry)

crates/dladm/src/lib.rs
```diff
@@ -140,7 +142,11 @@ impl Handle {
             state.n_seen += 1;
 
             if (*macaddr).ma_addrlen == (ETHERADDRL as u32) {
-                state.mac.copy_from_slice(&(*macaddr).ma_addr[..ETHERADDRL]);
+                let ma_addr = slice::from_raw_parts(
+                    ptr::addr_of!((*macaddr).ma_addr).cast::<u8>(),
```
iximeow, line 146: i think we can (should?) change this so that `Arg` has a `&'a mut [u8; ETHADDRL]` so we can side-step this and the question of "will copy_from_slice blow up somehow", but that's neither important nor better here :woman_shrugging: 

fwiw this apparently can also be `&raw const (*macaddr).ma_addr)` which i did not realize is a modern replacement for `ptr::addr_of!()`

iximeow (approved):
apparently there are a few places where rustfmt has given up on overly-long lines, hum.

crates/dladm/src/lib.rs
iximeow: apparently 1.82! neat

pfmooney (approved):

PR #915, opened 2025-07-02
Handle CPUID subleaves more precisely when using Bhyve defaults

iximeow:
We would report a few CPUID leaves as not having subleaves at all. Guests that try to interrogate CPUID leaves according to the APM then could end up very confused since later "subleaves" would just be repetitions of the single first leaf we queried from the host.

For leaves F, 10, and 8000_0020, I've chosen to hide the otherwise-unsupported-anyway PQM and PQE features from PQoS. If a guest tried to use the PQM/PQE features as-advertised, it should get a #GP anyway. Presumably #GP from a "supported" feature may not be gracefully handled. For 8000_0026, I'm less committed to returning zeroes, but it seems like the least-unreasonable thing to present.

I noticed this from `cpuid` in a Linux guest printing thousands of bogus 0x8000_0020 leaves on dogfood. That's PQoS feature information, and the APM happens to mention that the appropriate way to query this is to check individual bits in a bit field (implicitly that there are no more than 32 subleaves on that function). In that case it was technically a `cpuid` bug that resulted in it trying all the other 2^32-32 leaves, but we should be better here.

0x8000_0026 is more likely to matter, but it isn't present on Zen 3 parts.

(while in the product this will be avoided by using the CPUID definitions from RFD 314, this would still crop up when hacking on Propolis locally and it's annoying for "use the host defaults" to result in bogus leaves)

### DIFF
--- a/crates/cpuid-utils/src/host.rs
+++ b/crates/cpuid-utils/src/host.rs
@@ -114,14 +114,14 @@ fn collect_cpuid(
                     != 0;
                 set.insert(CpuidIdent::leaf(leaf), data)?;
             }
-            // Leaf 7 subleaf 0 eax indicates the total number of leaf-7
+            // Leaf 0x7 subleaf 0 eax indicates the total number of leaf-7
             // subleaves.
             0x7 => {
                 let mut data = query(leaf, 0)?;
 
-                // Leaf 7 EBX bits 12 and 15 indicate PQM and PQE support on AMD
-                // CPUs, and aspects of RDT support on Intel CPUs. In both
-                // cases, if the bits are set, leaves F and 10h are actually
+                // Leaf 0x7 EBX bits 12 and 15 indicate PQM and PQE support on
+                // AMD CPUs, and aspects of RDT support on Intel CPUs. In both
+                // cases, if the bits are set, leaves 0xF and 0x10 are actually
                 // subleaves with further capability information for the
                 // corresponding features. We don't support passing these
                 // features along, so mask out the bits.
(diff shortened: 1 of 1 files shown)


crates/cpuid-utils/src/host.rs
```diff
@@ -117,7 +117,16 @@ fn collect_cpuid(
             // Leaf 7 subleaf 0 eax indicates the total number of leaf-7
             // subleaves.
             0x7 => {
-                let data = query(leaf, 0)?;
+                let mut data = query(leaf, 0)?;
+
+                // Leaf 7 EBX bits 12 and 15 indicate PQM and PQE support on AMD
```
pfmooney, line 122: style nit: I think it would be nice to use the `h` suffix consistently, even for numbers (7, F) which do not require it for disambiguation.

crates/cpuid-utils/src/host.rs
```diff
@@ -165,6 +174,16 @@ fn collect_cpuid(
                     )?;
                 }
             }
+            // Leaf F describes Platform QoS Monitoring ("PQM").
```
hawkw, line 177: thank you for spelling out the acronym <3

hawkw (commented):
No concerns from me beyond Patrick's comment suggestions.

pfmooney (approved):
Thanks

crates/cpuid-utils/src/host.rs
iximeow: fair enough. it turns out we use `0x` prefixes for leaves elsewhere in Propolis anyway, so i've switched to those along the way.

PR #917, opened 2025-07-31
Added block_size for file backends in propolis_server (workers is optional)

leftwo:
Putting this out there to get an idea on what I'm doing wrong.
The "ZZZ" lines are my debugging.  

My goal here was to add `block_size` and `workers` as fields you could specify for a file based disk with `propolis-cli`
With these changes, the `--spec` option to `propolis-cli` is able to correctly parse and update the instance's `block_size` and `workers` filed if it finds those fields in the json.

What is not working the way I expect is the `--config-toml` option to `propolis-cli`.

I'm not sure where I missed it, but in  crates/propolis-config-toml/src/spec.rs:
https://github.com/oxidecomputer/propolis/blob/master/crates/propolis-config-toml/src/spec.rs#L154-L160

The `backend_config` that get back, it has only populated the `"path"` for `options`, not block_size or workers.
Here is the output of the log message after that call:
```
ZZZ try_from got config: BlockDevice { bdtype: "file", opts: BlockOpts { block_size: Some(4096), read_only: None, skip_flush: None, workers: Some(32) }, options: {"path": String("/dev/zvol/rdsk/oxp_247f8845-8f59-4669-a503-83e754c32327/crucible/local")} }
```

I modified `parse_storage_backend_from_config()` to look to the `opts` field for `block_size` and `workers` instead of looking in `options`, and that made all the rest work.

So, my question really is, what did I do wrong such that `block_size` and `workers` don't end up in the `options` field, or is what I did okay (and I'll remove the `TomlToSpecError`s I added).  Or, something else?

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -538,11 +538,13 @@ impl MachineInitializer<'_> {
                     );
                 }
 
-                let nworkers = NonZeroUsize::new(8).unwrap();
+                let nworkers =
+                    NonZeroUsize::new(spec.workers as usize).unwrap();
                 let be = propolis::block::FileBackend::create(
                     &spec.path,
                     propolis::block::BackendOpts {
                         read_only: Some(spec.readonly),
+                        block_size: Some(spec.block_size),
                         ..Default::default()
                     },
                     nworkers,
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -538,8 +538,19 @@ impl MachineInitializer<'_> {
                     );
                 }
 
-                let nworkers =
-                    NonZeroUsize::new(spec.workers as usize).unwrap();
+                let nworkers: NonZeroUsize = if spec.workers >= 1
+                    && spec.workers <= propolis::block::MAX_FILE_WORKERS
+                {
+                    NonZeroUsize::new(spec.workers).unwrap()
+                } else {
+                    slog::warn!(
+                        self.log,
+                        "workers must be between 1 and {} \
+                            Using default value of 8.",
+                        propolis::block::MAX_FILE_WORKERS
+                    );
+                    NonZeroUsize::new(8).unwrap()
+                };
                 let be = propolis::block::FileBackend::create(
                     &spec.path,
                     propolis::block::BackendOpts {
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -538,18 +538,23 @@ impl MachineInitializer<'_> {
                     );
                 }
 
-                let nworkers: NonZeroUsize = if spec.workers >= 1
-                    && spec.workers <= propolis::block::MAX_FILE_WORKERS
-                {
-                    NonZeroUsize::new(spec.workers).unwrap()
-                } else {
-                    slog::warn!(
-                        self.log,
-                        "workers must be between 1 and {} \
-                            Using default value of 8.",
-                        propolis::block::MAX_FILE_WORKERS
-                    );
-                    NonZeroUsize::new(8).unwrap()
+                let nworkers: NonZeroUsize = match spec.workers {
+                    Some(workers) => {
+                        if (1..=propolis::block::MAX_FILE_WORKERS)
+                            .contains(&workers)
+                        {
+                            NonZeroUsize::new(workers).unwrap()
+                        } else {
+                            slog::warn!(
+                                self.log,
+                                "workers must be between 1 and {} \
+                                    Using default value of 8.",
+                                propolis::block::MAX_FILE_WORKERS
+                            );
+                            NonZeroUsize::new(8).unwrap()
+                        }
+                    }
+                    None => NonZeroUsize::new(8).unwrap(),
                 };
                 let be = propolis::block::FileBackend::create(
                     &spec.path,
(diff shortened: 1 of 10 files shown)


bin/propolis-server/src/lib/stats/mod.rs
```diff
@@ -269,7 +269,7 @@ pub(crate) async fn track_network_interface_kstats(
         let network_interface_ids = interface_ids
             .iter()
             .map(|(uuid, device_id)| {
-                format!("{} (kstat-instance: {})", uuid, device_id)
+                format!("{uuid} (kstat-instance: {device_id})")
```
leftwo, line 272: This was just clippy cleanup, unrelated to the other changes

iximeow:
> I see I've broken the PHD test

the migrate-from-base test uses the current `propolis-client` against whatever the migration base is, and the PHD default for a test VM is to use a file-backed disk for the boot device. so PHD tries to use the new field against the older `propolis-server` to set up a VM and try migrating it, but we've changed the API and that old propolis-server doesn't know what to do.

i don't think there's a way for Omicron to produce file-backed disk requests right now, so in that sense "it's fine", but i'd wonder how this relates to migration..

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -538,11 +538,13 @@ impl MachineInitializer<'_> {
                     );
                 }
 
-                let nworkers = NonZeroUsize::new(8).unwrap();
+                let nworkers =
```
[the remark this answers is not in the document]
leftwo: Done.  I moved MAX_WORKERS to a place where we can validate against it here.

crates/propolis-config-toml/src/spec.rs
```diff
@@ -57,6 +58,12 @@ pub enum TomlToSpecError {
     #[error("failed to parse read-only option for file backend {0:?}")]
     FileBackendReadonlyParseFailed(String, #[source] ParseBoolError),
 
+    #[error("failed to parse block-size option for file backend {0:?}")]
```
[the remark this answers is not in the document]
leftwo: Removed

leftwo:
> > I see I've broken the PHD test
> 
> the migrate-from-base test uses the current `propolis-client` against whatever the migration base is, and the PHD default for a test VM is to use a file-backed disk for the boot device. so PHD tries to use the new field against the older `propolis-server` to set up a VM and try migrating it, but we've changed the API and that old propolis-server doesn't know what to do.

Will the "older" propolis-server change if this PR lands in main?  Or is it hard coded with a config that exists independent of what is in main?  And, go figure it out yourself is a fine answer here.
 
> i don't think there's a way for Omicron to produce file-backed disk requests right now, so in that sense "it's fine", but i'd wonder how this relates to migration..

I don't expect we will support migration for any file based storage.  Or, at most, the file based storage would be deleted and re-created anew on the other side, but that seems pretty ugly.  Most likely an instance with file based storage would just be halted and moved to failed if we were in a place where a sled was being expunged and a migration was attempted.

lib/propolis/src/block/mod.rs
```diff
@@ -38,6 +38,9 @@ pub type ByteLen = usize;
 /// is not choosing a block size, a default of 512B is used.
 pub const DEFAULT_BLOCK_SIZE: u32 = 512;
 
+// XXX: completely arb for now
+pub const MAX_FILE_WORKERS: usize = 32;
```
pfmooney, line 42: nit: I don't know why this needs to live in propolis-lib, as there's nothing in the underlying implementation which really "cares" about the upper bound on the worker threads.  We can probably just let proplis-server (and standalone) define its own upper limit?

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -538,11 +538,29 @@ impl MachineInitializer<'_> {
...
+                        {
+                            NonZeroUsize::new(workers).unwrap()
+                        } else {
+                            slog::warn!(
+                                self.log,
+                                "workers must be between 1 and {} \
+                                    Using default value of 8.",
+                                propolis::block::MAX_FILE_WORKERS
+                            );
+                            NonZeroUsize::new(8).unwrap()
```
pfmooney, line 554: nit: probably define a const somewhere in here with the default worker count, rather than repeating the "magic" `8` a few times here?

crates/propolis-api-types/src/instance_spec/components/backends.rs
```diff
@@ -47,6 +47,12 @@ pub struct FileStorageBackend {
 
     /// Indicates whether the storage is read-only.
     pub readonly: bool,
+
+    /// Block size of the backend
+    pub block_size: u32,
+
+    /// Optional worker threads for the file backend, exposed for testing only.
+    pub workers: Option<usize>,
```
pfmooney, line 55: Does the spec get grumpy if you use a NonzeroUsize here?

bin/propolis-server/src/lib/initializer.rs
leftwo: I've updated it to have a const

crates/propolis-api-types/src/instance_spec/components/backends.rs
leftwo: It seems to work, so I've updated it all to use `NonZeroUsize`

lib/propolis/src/block/mod.rs
leftwo: I've changed things so propolis-server and propolis-standalone use their own max (which happens to be the same currently).

pfmooney (approved):

PR #922, opened 2025-08-09
Bump crucible rev to latest

jmpesp:
Pick up the following PRs:

- Snapshots existing already are ok!
- Less verbose logging
- Remove unused `Vec
- Split "check reconciliation state" from "start reconciliation"
- Improve `ClientIoTask` start logic
- Use data-bearing enum variants pattern in negotiation
- Make Downstairs stoppable
- Don't log every region's metadata
- Compute reconciliation from `ClientMap` instead of three clients
- Make Offline -> Faulted transition happen without reconnecting
- Remove `WaitActive` state during negotiation
- Add explicit `UpstairsState::Disabled`
- Print version on startup for pantry and agent
- Update test mem to also show physical space used by regions.
- Add AllStopped command and endpoint for downstairs status
- Update tests to honor REGION_SETS env if provided.
- Fix panic in `set_active_request` when client is in Stoppingstate
- DTrace updates

Also updated dropshot.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -604,9 +604,9 @@ checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
 name = "chrono"
-version = "0.4.40"
+version = "0.4.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


faithanalog:
the clippy failure is new with rust 1.89.0. running `cargo xtask clippy --strict`, I don't get an error with 1.88.0, but I do with 1.89.0.

pfmooney (approved):

PR #929, opened 2025-08-13
Re-enable viona packet data loaning

pfmooney:
With the Chelsio driver now supporting bounded TX completion notifications, it should be safe to loan guest packet data in the mblk to avoid the extra alloc/copy.


pfmooney:
@askfongjojo was kind enough to re-run the tests from the aforementioned omicron ticket on bits I installed on `berlin`, featuring this change and up-to-date helios (including the TX completion work in cxgbe).  Performance results were the same as on dogfood, suggesting that the TX completions were working as expected, and that it's safe to re-enable packet loaning.

PR #931, opened 2025-08-14
update progenitor to 0.11.0

sunshowers:


pfmooney (approved):

sunshowers:
Thanks for the review @pfmooney!

iximeow:
@sunshowers my understanding is you no longer need the Progenitor version in Propolis to change since your later changes in https://github.com/oxidecomputer/newtype-uuid/pull/71, right? i figure we might as well bump the  Progenitor version here either way just it's no longer load-bearing?

PR #932, opened 2025-08-18
fix linux p9fs multi message reads

rcgoodfellow:
- Fix space left calculation to be based on maximum message size rather than client requested size.
- Fix tag propagation for some p9fs message types.
- Bump p9fs library that has message type tag fixes.
- Tie Rgetattr.valid to Tgetattr.mask

### DIFF
--- a/lib/propolis/src/hw/virtio/p9fs.rs
+++ b/lib/propolis/src/hw/virtio/p9fs.rs
@@ -32,7 +32,6 @@ use p9ds::proto::{
     self, Dirent, MessageType, P9Version, Qid, QidType, Rattach, Rclunk,
     Rgetattr, Rlerror, Rlopen, Rread, Rreaddir, Rstatfs, Rwalk, Tattach,
     Tgetattr, Tlopen, Tread, Treaddir, Tstatfs, Twalk, Version,
-    P9_GETATTR_BASIC,
 };
 use slog::{warn, Logger};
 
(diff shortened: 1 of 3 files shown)


taspelund:
The code changes seem sensible to me, but I don't have a lot of context around p9fs or filesystem implementations in general, so I don't know whether this is the right thing to do semantically

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -456,15 +467,21 @@ impl HostFSHandler {
         // mode: u32,
         let mode = metadata.mode();
         // uid: u32,
-        //let uid = metadata.uid();
         let uid = 0; //squash for now
                      // gid: u32,
                      //let gid = metadata.gid();
```
iximeow, line 472: strike this one too?

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -456,15 +467,21 @@ impl HostFSHandler {
...
         let gid = 0; //squash for now
                      // nlink: u64,
         let nlink = metadata.nlink();
         // rdev: u64,
-        let rdev = metadata.rdev();
+        let rdev =
+            if metadata.is_file() || metadata.is_dir() || metadata.is_symlink()
```
iximeow, line 478: you'd mentioned `p9fs parlance` above, is there a spec or do you mean parlance as in the implementation of p9fs somewhere else? I found https://ericvh.github.io/9p-rfc/rfc9p2000.html#anchor18 but it's not so helpful here :)

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -366,19 +365,26 @@ impl HostFSHandler {
...
-            - size_of::<MessageType>()  // Rread.typ
-            - size_of::<u16>()          // Rread.tag
-            - size_of::<u32>()          // Rread.data.len
-            - P9FS_VIRTIO_READ_HEADROOM;
-
-        let buflen =
-            std::cmp::min(space_left, (metadata.len() - msg.offset) as usize);
-
-        p9_write_file(&file, chain, mem, buflen, msg.offset as i64);
+        let space_left = u64::from(msize)
```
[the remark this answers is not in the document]
iximeow: taking this as an excuse to learn about p9fs, this is more like `MAX_READ_SIZE` now right? e.g. given the current `msize`, a read can safely be satisfied with up to this many bytes.

then the issue before was that if a client asked to read, say, 20 bytes, we'd calculate we could only return 9?

dancrossnyc, line 368: The formatting here feels funky? I understand that this is basically allocate space from `msize` for the message parameters for the `Rread` R-message (https://9p.io/magic/man2html/5/read); in this case, I'd recommend creating a constant for the size portion here something like:

    const RREAD_PARAM_SIZE: usize = size_of::<u32>() + size_of::MessageType>() + size_of::<u16>() + size_of::<u32>();`

Or, possibly just set it to a sum or something (4 + 1 + 2 + 4), as these values are well-defined in the protocol specification itself. An alternative might be to define a `repr(packed)` struct containing the R-message parameters and use its size.

Anyway, with that in place, this could become:

    let space_left = u64::from(msize) - u64::from(RREAD_PARAM_SIZE + P9FS_VIRTIO_READ_HEADROOM);

Which feels much cleaner.

iximeow (commented):
notionally :+1: but a smattering of nits to offer. Is this one of those things we _Ought To Test Better_ but we just haven't done w/ phd or something?

lib/propolis/src/hw/virtio/p9fs.rs
iximeow: @dancrossnyc mentioned on the side that what I'm looking for here is 9p2000.L, or https://github.com/chaos/diod/blob/master/protocol.md

lib/propolis/src/hw/virtio/p9fs.rs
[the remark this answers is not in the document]
rcgoodfellow: yes, that's correct

rcgoodfellow:
> we Ought To Test Better but we just haven't done w/ phd or something?

The p9fs stuff is behind the `falcon` feature and is use exclusively in falcon for testbeds. It's not compiled into the production binary. So it does not have the same level of testing as product features. Tbh most of it has been done as quickly as possible by me in service of creating a testing surface for an urgent product features.

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -954,6 +974,7 @@ pub(crate) fn write_error(ecode: u32, chain: &mut Chain, mem: &MemCtx) {
 fn p9_write_file(
```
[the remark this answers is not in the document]
rcgoodfellow: Updated to `p9_read_file`

lib/propolis/src/hw/virtio/p9fs.rs
rcgoodfellow: Pulled the struct out of `p9_write_file` and am using `size_of` on that struct.

iximeow (approved):

iximeow:
> use exclusively in falcon for testbeds

ah i'd forgotten it was fully sequestered like that, fair enough

dancrossnyc (approved):
LG! Thanks, Ry.

pfmooney (approved):

PR #937, opened 2025-08-27
Expand viona header pad to account for new Geneve options

FelixMcFelix:
With https://github.com/oxidecomputer/opte/pull/805 merged, we are pushing on an extra 8B for any rack-local TCP flow. Equally, we expect that any multicast packets will push Geneve options. This PR bumps the `header_pad` value to accomodate this extra requirement.

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -786,9 +786,9 @@ impl MachineInitializer<'_> {
                     // - ethernet: 14
                     // - IPv6: 40
                     // - UDP: 8
-                    // - Geneve: 8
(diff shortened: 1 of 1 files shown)


pfmooney (approved):
LGTM

For-later question: Is this something which would make sense for opte/xde to expose via an ioctl for consumers like propolis-server to query at runtime?

FelixMcFelix:
> For-later question: Is this something which would make sense for opte/xde to expose via an ioctl for consumers like propolis-server to query at runtime?

Agreed, I think the lower friction this can be the better. I've opened https://github.com/oxidecomputer/opte/issues/843 to track this.

Issue #921, opened 2025-08-08
Guest OS SMT Panic with Intel(r) Xeon(r) CPU E5-2660


qdzlug:
When attempting to stand up a4x2 on this system, the VM's start as expected but all helios vm's hit this panic:

<img width="3390" height="2072" alt="Image" src="https://github.com/user-attachments/assets/9ded1892-1bda-47c5-8413-d07210f9a469" />

The debian guest never shows anything on the console.

Diagnostic Details:

```
[jschmidt@sunny propolis]$ cpuid -d
CPU 0:
CPUID 00000000:00 = 0000000d 756e6547 6c65746e 49656e69 | ....GenuntelineI
CPUID 00000001:00 = 000206d7 00200800 1fbee3ff bfebfbff | ...... .........
CPUID 00000002:00 = 76035a01 00f0b2ff 00000000 00ca0000 | .Z.v............
CPUID 00000003:00 = 00000000 00000000 00000000 00000000 | ................
CPUID 00000004:00 = 3c004121 01c0003f 0000003f 00000000 | !A.<?...?.......
CPUID 00000004:01 = 3c004122 01c0003f 0000003f 00000000 | "A.<?...?.......
CPUID 00000004:02 = 3c004143 01c0003f 000001ff 00000000 | CA.<?...........
CPUID 00000004:03 = 3c07c163 04c0003f 00003fff 00000006 | c..<?....?......
CPUID 00000004:04 = 00000000 00000000 00000000 00000000 | ................
CPUID 00000005:00 = 00000040 00000040 00000003 00021120 | @...@....... ...
CPUID 00000006:00 = 00000075 00000002 00000009 00000000 | u...............
CPUID 00000007:00 = 00000000 00000000 00000000 9c000400 | ................
CPUID 00000008:00 = 00000000 00000000 00000000 00000000 | ................
CPUID 00000009:00 = 00000001 00000000 00000000 00000000 | ................
CPUID 0000000a:00 = 07300403 00000000 00000000 00000603 | ..0.............
CPUID 0000000b:00 = 00000001 00000002 00000100 00000000 | ................
CPUID 0000000b:01 = 00000005 00000010 00000201 00000000 | ................
CPUID 0000000c:00 = 00000000 00000000 00000000 00000000 | ................
CPUID 0000000d:00 = 00000007 00000340 00000340 00000000 | ....@...@.......
CPUID 0000000d:01 = 00000001 00000000 00000000 00000000 | ................
[...]

Issue #940, opened 2025-09-06
VMs should not have vCPU-many sockets


iximeow:
this is a bit silly:

```
debian@debian-11:~$ likwid-topology -g
--------------------------------------------------------------------------------
CPU name:	AMD EPYC 7713P 64-Core Processor
CPU type:	AMD K19 (Zen3) architecture
CPU stepping:	1
********************************************************************************
Hardware Thread Topology
********************************************************************************
Sockets:		8
Cores per socket:	1
Threads per core:	1
```

I (and I think everyone else) thought we're emulating a single-socket machine until we intend otherwise, with vCPU-many cores and one or two threads per core. [Looking into #921](https://github.com/oxidecomputer/propolis/issues/921#issuecomment-3260090080) I noticed we discard leaf `Ext1E` and then never repopulate it for the VM's CPU count, but we never clear the TopologyExtensions bit, so the all-zero leftovers is declared to be faithful. That gets us vCPU-many sockets, oops.

RFD 314 and https://github.com/oxidecomputer/omicron/pull/8728 in particular set leaf 8000_001E to have two threads per core, but I'm actually not sure that gets us to one socket with vCPU-many-(or vCPU/2) cores. Either way, out of the box Propolis should reconstruct the CPUID leaf StdB/Ext1E that it very eagerly discards.

This will change the CPUID profile of guests as-is, but I think that's good - I'm a little surprised that OSes are as nice about this as they are!

iximeow:
Trey saw that a Helios VM on colo has a mildly concerning-looking output related to this:
```
Oxide Helios Version helios-2.0.23866 64-bit
WARNING: There will be no MCA support on chip 32 core 0 strand 0

WARNING: There will be no MCA support on chip 33 core 0 strand 0

... incrementing through other chips ...

WARNING: There will be no MCA support on chip 63 core 0 strand 0
```

[...]

PR #941, opened 2025-09-06
Specialize CPUID leaves 4 and Bh for CPU topology

iximeow:
This is kind of two parts:

* teach `cpuid-utils` and propolis' CPUID specializer about the Intel-only leaf 4 (at least as it relates to core/thread counts)
* include `with_cpu_topo()` so we update to-be-specialized topo leaves instead of just discard them

this fixes https://github.com/oxidecomputer/propolis/issues/921 and is part of fixing https://github.com/oxidecomputer/propolis/issues/940 (the other part being https://www.illumos.org/issues/17529 or getting leaf B from an instance spec)

draft because I haven't given this a test with anything _other_ than OmniOS r151054 on Intel yet, but I'd be kind of shocked if this changes anything on AMD or if other OSes somehow find this intolerable where they were fine with the leaves before...

### DIFF
--- a/lib/propolis/src/cpuid.rs
+++ b/lib/propolis/src/cpuid.rs
@@ -43,6 +43,10 @@ pub enum SpecializeError {
     UnsupportedCacheLevel,
     #[error("missing vcpu count")]
     MissingVcpuCount,
+    #[error("missing vcpu id")]
+    MissingVcpuId,
+    #[error("unable to specialize leaf")]
+    IncompatibleTopology { leaf: u32, num_vcpu: u32, why: Option<&'static str> },
 }
 
 /// Specialize a set of cpuid leafs for provided attributes.
@@ -221,9 +225,7 @@ impl Specializer {
             // indicate that SMT is enabled, then vCPUs are presented as pairs
             // of sibling threads on vproc-many processors.
             let num_vproc = if self.has_smt {
-                if num_vcpu % 2 != 0 {
-                    panic!("SMT with an odd number of cores? how ... odd.");
-                }
+                // TODO: What if num_vcpu is odd?!
                 num_vcpu >> 1
             } else {
                 num_vcpu
@@ -251,14 +253,22 @@ impl Specializer {
                     // all-bits-set-and-use-another-topo-method condition, but
                     // bail out here and demand someone take a look.
                     if num_vproc >= 0b100_0000 {
-                        unimplemented!(
-                            "Don't know how to set CPUID leaf 4 processor \
-                             count if there are more than 64 processors!"
-                        );
+                        return Err(SpecializeError::IncompatibleTopology {
+                            leaf: 4,
+                            num_vcpu,
+                            why: Some(
+                                "Don't know how to set CPUID leaf 4 processor \
+                                 count if there are more than 64 processors!",
+                            ),
+                        });
                     } else if num_vproc == 0 {
-                        unimplemented!(
-                            "Cannot specialize CPUID leaf 4 for a 0 vCPU VM"
-                        );
+                        return Err(SpecializeError::IncompatibleTopology {
+                            leaf: 4,
+                            num_vcpu,
+                            why: Some(
+                                "Cannot specialize CPUID leaf 4 for a 0 vCPU VM"
+                            )
+                        });
                     }
 
                     // Cache types come in any order, but type 0 means there are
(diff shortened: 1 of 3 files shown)


lib/propolis/src/cpuid.rs
```diff
@@ -193,56 +197,202 @@ impl Specializer {
...
             let num_vcpu = self
                 .num_vcpu
                 .ok_or(SpecializeError::MissingVcpuCount)
                 .map(|n| u32::from(n.get()))?;
 
+            // The number of logical processors the guest should see. If we
+            // indicate that SMT is enabled, then vCPUs are presented as pairs
+            // of sibling threads on vproc-many processors.
+            let num_vproc = if self.has_smt {
+                // TODO: What if num_vcpu is odd?!
```
iximeow, line 228: well, on Linux if you're indicating SMT and then have an odd number of cores, at least with the RFD 314-like CPUID leaves, the OS decides that your topology is nonsense and asserts it's all one big core:
```
root@debian:~# cat /sys/devices/system/cpu/cpu0/topology/thread_siblings_list
0-2
root@debian:~# cat /sys/devices/system/cpu/cpu2/topology/thread_siblings_list
0-2
```
where the guest did in fact get reasonable leaf B bits:
```
root@debian:~# cpuid -r | grep 0000000b
   0x0000000b 0x00: eax=0x00000001 ebx=0x00000002 ecx=0x00000100 edx=0x00000000
   0x0000000b 0x01: eax=0x00000008 ebx=0x00000003 ecx=0x00000201 edx=0x00000000
   0x0000000b 0x00: eax=0x00000001 ebx=0x00000002 ecx=0x00000100 edx=0x00000001
   0x0000000b 0x01: eax=0x00000008 ebx=0x00000003 ecx=0x00000201 edx=0x00000001
   0x0000000b 0x00: eax=0x00000001 ebx=0x00000002 ecx=0x00000100 edx=0x00000002
   0x0000000b 0x01: eax=0x00000008 ebx=0x00000003 ecx=0x00000201 edx=0x00000002
```

if you have an even number of CPUs then Linux sees them all paired up (see the new test in `cpuid.rs`). it'd be kinda nice to just not allow odd vCPU-count guests until there's a reason to permit them. for the time being I think it's time to let this sit, do virtual platforms, and have reasonable topologies in the non-initial platform.

iximeow: actually the guest might be in a better state if the non-paired core does not have hyperthreading indicated in its CPUID. that makes me a little less worried about potential guest issues but is more work than this merits right now.

lib/propolis/src/cpuid.rs
```diff
@@ -193,56 +197,223 @@ impl Specializer {
...
+                    // VM is conceptually one socket) or set this to 7 or 8 like
+                    // a "normal" processor.
+                    //
+                    // Go with 8 for now and error if num_vcpu is above that:
+                    // you should trip over this quickly if you've bumped up
+                    // VM_MAXCPU and you may have hardware to compare against at
+                    // that point.
+                    if num_vcpu >= 256 {
+                        return Err(SpecializeError::IncompatibleTopology {
+                            leaf: 4,
```
pfmooney, line 402: leaf B, not 4

Should TopoKind have some accessor like `leaf()` which emits a u32, so it could be passed in here, rather than typed in by hand?

iximeow: doubly awkward: `leaf`, which is used at various points in the match, _is_ the u32. the enum is also `#[repr(u32)]` which is why that value comes from a simple `*topo as u32`. i've struck the explicit numbers in these `IncompatibleTopology` errors in favor of just `leaf,`.

lib/propolis/src/cpuid.rs
```diff
@@ -193,56 +197,223 @@ impl Specializer {
...
+                    for i in 0..32 {
+                        let leaf = set.get(CpuidIdent::subleaf(4, i)).cloned();
+                        let Some(mut leaf) = leaf else {
+                            // We've reached the end of provided subleaves, so
+                            // we're done here.
+                            break;
+                        };
+
+                        let ty = leaf.eax & 0b11111;
+                        let level = (leaf.eax >> 5) & 0b111;
```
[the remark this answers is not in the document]
iximeow: i was on the fence because names here are a bit silly ( .. yes, `let ty = leaf.eax & LEAF4_EAX_CACHE_TYPE;` ..), but it's kind of nice to line up the masks as much as rustfmt permits to see they don't overlap.

iximeow:
```
phd_tests::migrate::from_base::migration_from_base_and_back ... FAILED: building VM spec from VmConfig

Caused by:
    0: getting guest OS kind for boot disk
    1: Failed to download alpine.iso from any remote URI
```
=| interesting failure mode after 417 minutes.

PR #943, opened 2025-09-17
propolis-cli should be able to send TOML-defined CPU profiles

iximeow:
some plumbing I wrote up to help check the CPU profile from Omicron#8728 across more OS types. I think shortly after 8728 is sorted it makes sense to move a lot of the CPUID stuff to a crate shared with Propolis, one reason being we could bake in known-to-Nexus CPUID profiles in `propolis-cli` as well. Then we could make `--cpuid-profile milan_v1` _Just Work_ without having to copy around blobs of hex for well-defined profiles. but, this is sufficient for now.

### DIFF
--- a/bin/propolis-cli/src/main.rs
+++ b/bin/propolis-cli/src/main.rs
@@ -27,6 +27,7 @@ use propolis_client::types::{
     InstanceEnsureRequest, InstanceInitializationMethod, InstanceMetadata,
     InstanceSpecGetResponse,
 };
+use propolis_config_toml::spec::toml_cpuid_to_spec_cpuid;
 use propolis_config_toml::spec::SpecConfig;
 use serde::{Deserialize, Serialize};
 use slog::{o, Drain, Level, Logger};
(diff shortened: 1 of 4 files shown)


bin/propolis-cli/src/main.rs
```diff
@@ -288,22 +298,39 @@ impl VmConfig {
...
             from_toml.as_ref().map(|cfg| cfg.enable_pcie).unwrap_or(false);
 
+        let cpuid_profile = parsed_toml
+            .as_ref()
+            .and_then(|cfg| {
+                self.cpuid_profile.as_ref().map(|profile| {
+                    cfg.cpuid_profiles
+                        .get(profile)
+                        .ok_or_else(|| {
+                            anyhow!("CPUID profile not defined: {}", profile)
```
hawkw: turbo nit
```suggestion
                            anyhow!("CPUID profile not defined: {profile}",)
```
also, maybe it would be nice to have this say "not defined in path/to/toml.toml"?

iximeow: i'll internalize in-line format variables one day.. i was less sure about including the toml path at first because i'd like to check the name against a the kinds of profiles we've got in Nexus (after hoisting `cpu_platform.rs` out of Omicron, probably), but we can change the way this error is printed when that time comes

bin/propolis-cli/src/main.rs
```diff
@@ -288,22 +298,39 @@ impl VmConfig {
...
+                        })
+                        .and_then(|profile| {
+                            toml_cpuid_to_spec_cpuid(profile)
+                                .map_err(Into::into)
+                        })
+                })
```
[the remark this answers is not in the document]
iximeow: yes! closures turn off my `?` brain. the `Into` needed a bit of type help but this is nicer.

PR #947, opened 2025-09-19
Update dropshot and rust toolchain

pfmooney:
The dropshot update gets us past the clippy nits previously emitted by rust 1.89.  The rest for 1.90 are address in the follow-up.


pfmooney:
It seems the toolchain update has somehow broken how phd-runner locates its testcases via the `inventory` crate.

pfmooney:
Was opte bumped for a specific reason, or "just because"?

pfmooney:
Closed in favor of #952

PR #952, opened 2025-09-25
move phd-tests to linkme, update Rust to 1.90

sunshowers:
Followup to https://github.com/oxidecomputer/propolis/pull/947. While the inventory crate is busted on illumos with Rust 1.89 and above, it appears (based on my local testing) that the linkme crate works okay.

Also migrate `rust-toolchain.toml` to 1.90, bump Dropshot, and fix clippy warnings.


pfmooney (approved):

iximeow (approved):
ah neat, this presumably works because while the `used(linker)` attribute still gets a bogus bit set, all the sections for the linkme slice elements are merged together since their section flags are at the same with that bit set. so we end up with one contiguous slice of the elements after linking.

PR #954, opened 2025-10-08
update omicron-common and crucible

sunshowers:
Need to pull in updates for the `VERSION_HEADER` constant in omicron-common.

Probably going to fail CI because of the new dependency on libclang; will monitor.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,15 +77,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.18"
+version = "0.2.21"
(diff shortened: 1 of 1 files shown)


tools/install_builder_prerequisites.sh
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env bash
+
+# This file is adapted from Omicron's install_builder_prerequisites.sh.
```
sunshowers, line 3: This installs libclang as detailed at https://github.com/oxidecomputer/crucible/pull/1778#discussion_r2406836642.

pfmooney (approved):

PR #963, opened 2025-10-15
[ignore this, testing only] Bump crucible

leftwo:
We might want to get this into 17. so prepping a PR so I can test in Omicron

Brings up these changes
Add `crucible-downstairs validate` subcommand (#1792)
Remove incorrect `expect(unused)` annotation (#1789)
Handle errors in get-up-state.sh (#1787)
Handle final live-repair flush being skipped on all downstairs (#1783)
Offline downstairs go to faulted when LR starts (#1777) 
update drift to pick up fix for OpenAPI type graph cycles (#1785) 
make Crucible APIs versioned (#1782)


leftwo:
This test branch has reached the end of it's life.  Thank you test branch for your service

PR #966, opened 2025-10-28
Rudimentary NVMe emulation fuzzer

iximeow:
This tries doing a bunch of random operations against an NVMe device and checks the operations against a limited model of what the results of those operations should be.

The initial stab at this is what caught https://github.com/oxidecomputer/propolis/pull/965, and it caught a bug in an intermediate state of #953 (which other phd tests did notice anyway). This fuzzing would probably be best with actual I/O operations mixed in, and I think that *should* be relatively straightforward to add from here., but as-is it's useful!

This would probably be best phrased as a `cargo-fuzz` test to at least get coverage-guided fuzzing. Because of the statefulness of NVMe I think either way we'd want the model of expected device state and a pick-actions-then-run execution to further guide `cargo-fuzz` into useful parts of the device state.

The initial approach at this allowed for device reset and migration at arbitrary times via a separate thread. When that required synchronizing the model of device state it was effectively interleaved with "guest" operations on the device, and in practice admin commands are serialized by the `NvmeCtrl` state lock anyway. It may be more interesting to revisit with concurrent I/O operations on submission/completion queues.

### DIFF
--- a/lib/propolis/src/hw/nvme/test.rs
+++ b/lib/propolis/src/hw/nvme/test.rs
@@ -83,7 +83,7 @@ impl FuzzCtx {
     const SQE_SIZE: usize = std::mem::size_of::<SubmissionQueueEntry>();
 
     const ADMIN_SQ_ENTRIES: u16 = 1024;
-    const ADMIN_SQ_BASE: GuestAddr = GuestAddr(1 * MB as u64);
+    const ADMIN_SQ_BASE: GuestAddr = GuestAddr(MB as u64);
     const ADMIN_SQ_SIZE: usize =
         Self::ADMIN_SQ_ENTRIES as usize * Self::SQE_SIZE;
 


lib/propolis/src/hw/nvme/test.rs
```diff
@@ -0,0 +1,644 @@
...
+/// allowable successor states. One hope is for `FuzzCtx` and corresponding test
+/// state machine to adaptable to `cargo-fuzz`, even with the limitations of
+/// coverage-guided fuzzing.
+///
+/// `FuzzCtx` is too high-level for some tests. Because it manages shared state
+/// to drive the emulated device, `FuzzCtx` does not allow arbitrary resets at
+/// any point as `PciNvme` technically does. (In practice, reset immediately
+/// locks the inner `NvmeCtrl` to do the reset, for administrative options it is
+/// effectively serialized anyway.)
+struct FuzzCtx {
```
iximeow: I'd talked with Patrick just a bit about the idea of having a more composable fuzzing framework as part of Propolis (the library). the implementation in this file is simultaneously:

* an NVMe driver, which is driven ~randomly to exercise `propolis/src/hw/nvme`
* a model of the NVMe controller state, based on the driver's operations (what queues are created, is the device initialized, etc)
* a model of the synthesis of NVMe spec and controller model: given the current state, what NVMe operations should have what outcomes, and what state transitions are permissible from the current state?

you could swap out the word "NVMe" above for any other devices (the chipset would be interesting!). a reasonable thing to want would be fuzzing a pair of NVMe devices concurrently on the same PCI bridge. or poking a disk and NIC concurrently, or configuring a pair of NVMe devices to write on each others' queues, or ... 

in the limit this seems to me like assembling an increasingly chaotic VM and configuring how chaotic it should be. I don't plan on adjusting this in that direction right now, but it seems like an interesting future direction this could take.

hawkw (commented):
neat!

PR #977, opened 2025-11-25
Add support for programmable SMBIOS Type 1 table

andrewjstone:


bin/propolis-server/src/lib/spec/api_spec_v0.rs
```diff
@@ -69,6 +69,10 @@ impl From<Spec> for InstanceSpecV0 {
             migration_failure,
             #[cfg(feature = "falcon")]
             softnpu,
+
+            // Not part of `InstanceSpecV0`. Added in `InstanceSpec` in API
+            // Version 2.0.0.
+            smbios_type1_input: _,
```
iximeow: the interaction between this and `migration/{source,destination}.rs` as they are will introduce a bug in migration. we're going to be OK with this for the time being since we're not actually using it yet, but the issue is: in `migration/source.rs` we will send an `InstanceSpecV0` over to the destination `propolis-server`, and at this point we lose `smbios_type1_input` if it was present. on the destination side, we'll get a valid `InstanceSpecV0`, build a `Spec` from it without `smbios_type1_input`, may have a different operational `Spec` than the one from the source Propolis.

andrewjstone:
Thanks for all the discussion and review @iximeow!

PR #985, opened 2025-12-11
block/file: do pread/pwrite from Propolis heap instead of VM memory

iximeow:
As described in the comment on `MAPPING_IO_LIMIT_BYTES`, this is a gross hack: if VM memory is used as the iovec for `p{read,write}v` to a block device or zvol, we'll end up contending on `svmd->svmd_lock` pretty heavily in higher IOPS situations.

This might arise when doing fewer larger I/Os to zvols or block devices as well, when we'll be forwarding each of the 4k entries in guest PRP lists as a distinct iovec, but we haven't tested that.

Either way, operating from Propolis heap avoids the issue for now. One could imagine this "should" be a configurable behavior on file block backends, since many files (like disk images in propolis-standalone) would not need this. The hope is that we can fix this in the kernel so the hack is no longer needed for zvols or block devices, rather than leaning further into properly supporting this copy hack.

For good measure, I did some testing on a `propolis-standalone` with this locally. Of the 1% time in user code when doing large I/Os a `profile-1003` only turned up three stacks here, all in the malloc/free for the buffers.

### DIFF
--- a/lib/propolis/src/vmm/mem.rs
+++ b/lib/propolis/src/vmm/mem.rs
@@ -765,16 +765,32 @@ pub trait MappingExt {
 // which point it should be strictly worse than directly using the
 // requested mappings.
 //
-// Beyond this, either fall back to using iovecs directly (and trust the kernel
-// to handle bizarre vecs) or error. This is an especially gross hack because we
-// could/should just set MDTS, which more naturally communicates to the guest
-// the largest I/O we're willing to handle (and gives an out for too-large I/Os)
+// 1 MiB is an arbitrary-ish choice: `propolis-server` and `propolis-standalone`
+// set NVMe MDTS to "8", so the largest I/Os from NVMe will be
+// `2**8 * 4096 == 1048576 bytes == 1 MiB`. Beyond this, fall back to using
+// iovecs directly, potentially at increased OS overhead.
 //
 // The amount of memory used for temporary buffers is given by the number of
 // worker threads for all file-backed disks, times this threshold. It works out
-// to up to 128 MiB (8 worker threads) of buffers per disk by default as of
+// to up to 8 MiB (8 worker threads) of buffers per disk by default as of
 // writing.
-const MAPPING_IO_LIMIT_BYTES: usize = 16 * crate::common::MB;
+const MAPPING_IO_LIMIT_BYTES: usize = crate::common::MB;
+
+/// Compute the number of bytes that would be required to hold these mappings
+/// sequentially.
+///
+/// Ranges covered by multiple mappings are counted repeatedly.
+fn total_mapping_size(mappings: &[SubMapping<'_>]) -> Result<usize> {
+    mappings
+        .iter()
+        .try_fold(0usize, |total, mapping| total.checked_add(mapping.len))
+        .ok_or_else(|| {
+            Error::new(
+                ErrorKind::InvalidInput,
+                "Total mapping larger than a `usize`",
+            )
+        })
+}
 
 impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
     fn preadv(&self, fd: RawFd, offset: i64) -> Result<usize> {
@@ -789,16 +805,7 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
             ));
         }
 
-        let mut total_capacity: usize = 0;
-        for mapping in self.as_ref().iter() {
-            total_capacity =
-                total_capacity.checked_add(mapping.len).ok_or_else(|| {
-                    Error::new(
-                        ErrorKind::InvalidInput,
-                        "Total mapping too large",
-                    )
-                })?;
-        }
+        let total_capacity = total_mapping_size(self.as_ref())?;
 
         // Gross hack: see the comment on `MAPPING_IO_LIMIT_BYTES`.
         if total_capacity <= MAPPING_IO_LIMIT_BYTES {
@@ -813,13 +820,13 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
                 iov_len: buf.len(),
             }];
 
-            let read = unsafe {
+            let res = unsafe {
                 libc::preadv(fd, iov.as_ptr(), iov.len() as libc::c_int, offset)
             };
-            if read == -1 {
+            if res == -1 {
                 return Err(Error::last_os_error());
             }
-            let read: usize = read.try_into().expect("read is positive");
+            let read = res as usize;
 
             // copy `read` bytes back into the iovecs and return
             let mut remaining = &mut buf[..read];
@@ -888,16 +895,7 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
             ));
         }
 
-        let mut total_capacity: usize = 0;
-        for mapping in self.as_ref().iter() {
-            total_capacity =
-                total_capacity.checked_add(mapping.len).ok_or_else(|| {
-                    Error::new(
-                        ErrorKind::InvalidInput,
-                        "Total mapping too large",
-                    )
-                })?;
-        }
+        let total_capacity = total_mapping_size(self.as_ref())?;
 
         // Gross hack: see the comment on `MAPPING_IO_LIMIT_BYTES`.
         let written = if total_capacity <= MAPPING_IO_LIMIT_BYTES {


hawkw:
Interesting, this sure _seems_ like it should be worse than actually writing straight into the guest address space, but the contention making that worse makes sense. I agree that fixing the kernel seems like a better long term solution.

lib/propolis/src/vmm/mem.rs
```diff
@@ -754,6 +754,28 @@ pub trait MappingExt {
...
+// it, copy from guest memory to Propolis heap. The segment backing Propolis'
+// heap has an `as_page{,un}lock` impl that avoids the more
+// expensive/contentious `as_fault()` fallback.
+//
+// This is an optimization until stlouis#871 can get things sorted, at
+// which point it should be strictly worse than directly using the
+// requested mappings.
+//
+// Beyond this, either fall back to using iovecs directly (and trust the kernel
+// to handle bizarre vecs) or error. This is an especially gross hack because we
```
hawkw: "bizarre" here just meaning "in excess of 16 megabytes"?

lib/propolis/src/vmm/mem.rs
```diff
@@ -767,23 +789,92 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
...
+            total_capacity += mapping.len;
         }
 
-        Ok(read as usize)
+        // Gross hack: see the comment on `MAPPING_IO_LIMIT_BYTES`.
+        if total_capacity < MAPPING_IO_LIMIT_BYTES {
+            // If we're motivated to avoid the zero-fill via
+            // `Layout::with_size_align` + `GlobalAlloc::alloc`, we should
+            // probably avoid this gross hack entirely (see comment on
+            // MAPPING_IO_LIMIT_BYTES).
```
hawkw: fair!

lib/propolis/src/vmm/mem.rs
```diff
@@ -767,23 +789,92 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
...
-        if read == -1 {
-            return Err(Error::last_os_error());
+        let mut total_capacity = 0;
+        for mapping in self.as_ref().iter() {
+            total_capacity += mapping.len;
         }
 
-        Ok(read as usize)
+        // Gross hack: see the comment on `MAPPING_IO_LIMIT_BYTES`.
+        if total_capacity < MAPPING_IO_LIMIT_BYTES {
```
hawkw: i note that if we are doing a write of _exactly_ 16 MiB we shall spill over into the normal thing. might we want to use `<=` here instead given that the guest is likely to be doing megabyte-aligned-y IOs?

hawkw (approved):

lib/propolis/src/vmm/mem.rs
```diff
@@ -767,23 +789,92 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
...
+
+                // Safety: guest physical memory is READ|WRITE, and won't become
+                // unmapped as long as we hold a Mapping, which `self` implies.
+                //
+                // Sketchy: nothing guarantees the guest is not concurrently
+                // modifying this memory. Also, nothing prevents the guest from
+                // submitting the same ranges as part of another I/O where we
+                // might be doing this same operation on another thread on the
+                // same ptr/len pair. Because we are just moving u8's, it should
+                // be OK.
```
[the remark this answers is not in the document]
jclulow: > okay, so. one, regardless of the fact that we are "just moving u8s", i think that technically rustc is permitted to emit nasal demons in this case unless we use https://github.com/rust-lang/rust/issues/58599

What if you `unsafe { libc::memcpy(...`?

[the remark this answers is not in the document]
iximeow: oh, this should be `std::ptr::copy_nonoverlapping` actually. i forgot that exists at first.

lib/propolis/src/vmm/mem.rs
```diff
@@ -767,23 +789,92 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
...
+            let read: usize = read.try_into().expect("read is positive");
+
+            // copy `read` bytes back into the iovecs and return
+            let mut remaining = read;
+            let mut offset = 0;
+            for mapping in self.as_ref().iter() {
+                let to_copy = std::cmp::min(remaining, mapping.len);
+                if to_copy == 0 {
+                    break;
+                }
```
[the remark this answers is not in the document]
iximeow: it's incorrect in isolation, but an NVMe command couldn't produce a zero-size range like this without being a single zero-size iovec. i'll strike the if, going through the loop with `to_copy == 0` is fine (now. the pre-PR patch would have been less fine :grin: )

lib/propolis/src/vmm/mem.rs
```diff
@@ -767,23 +789,91 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
...
-        };
-        if read == -1 {
-            return Err(Error::last_os_error());
+        let mut total_capacity: usize = 0;
+        for mapping in self.as_ref().iter() {
+            total_capacity =
+                total_capacity.checked_add(mapping.len).ok_or_else(|| {
+                    Error::new(
+                        ErrorKind::InvalidInput,
+                        "Total mapping too large",
```
hawkw: nit, take it or leave it: perhaps
```suggestion
                        "Total length of all mappings would overflow a `usize`",
```
or something --- seeing this error would make me go "too large for _what_?"

lib/propolis/src/vmm/mem.rs
```diff
@@ -816,44 +822,37 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
...
-                // might be doing this same operation on another thread on the
-                // same ptr/len pair. Because we are just moving u8's, it should
-                // be OK.
-                let guest_slice = unsafe {
-                    std::slice::from_raw_parts_mut(
+                // The guest may be concurrently modifying this memory, we may
+                // be concurrently reading or writing this memory (if it is
+                // submitted for a read or write on another thread, for
+                // example), but `copy_nonoverlapping` has no compunctions about
+                // concurrent mutation.
```
hawkw: perhaps worth noting explicitly that if a guest _does_ decide it wants to screw up its own iovec for some reason, it would be pretty damn hard for LLVM to figure out a way to miscompile *us* in that case?

lib/propolis/src/vmm/mem.rs
```diff
@@ -889,28 +888,44 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
...
 
-        let mut total_capacity = 0;
+        let mut total_capacity: usize = 0;
         for mapping in self.as_ref().iter() {
-            total_capacity += mapping.len;
+            total_capacity =
+                total_capacity.checked_add(mapping.len).ok_or_else(|| {
+                    Error::new(
+                        ErrorKind::InvalidInput,
+                        "Total mapping too large",
```
hawkw: similarly, let's maybe say something like 
```suggestion
                        "Total length of all mappings would overflow a `usize`",

File: lib/propolis/src/vmm/mem.rs [hunk 9]
```diff
@@ -767,23 +789,91 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
...
+                iov_len: buf.len(),
+            }];
+
+            let read = unsafe {
+                libc::preadv(fd, iov.as_ptr(), iov.len() as libc::c_int, offset)
+            };
+            if read == -1 {
+                return Err(Error::last_os_error());
+            }
+            let read: usize = read.try_into().expect("read is positive");
```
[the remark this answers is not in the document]
iximeow: there are kind of a few things. first, yeah, I do care about panicking if `preadv()` returned a bogus error value, but that's easy enough to assert. I don't super like `usize::try_from` here because it seems weird to read this later as "if preadv() returned a number that couldn't be a `usize`, then it must have errored", versus "if pread() returned -1 that is an error, otherwise it must be a `usize`". I was hoping that maybe there's a less annoying construction of this but at the end of the day I consider this whole patch kinda temporary so not sweating it so much.

mechanically, from "before", I'd agree that both `rustix` and `nix` wouldn't hurt here. but particularly as this is "data plane" I'm very, very cautious about the system behaving exactly as we expect at all times. `nix` returns `Error::last_os_error()` if the returned value is a `-1`-like sentinel, otherwise [casts](https://github.com/nix-rust/nix/blob/ba63e90362e24dcebbf96783da668647ceace21c/src/sys/uio.rs#L107) whatever the bits are into the expected type. `rustix` in this case has a [`debug_assert!()`](https://github.com/bytecodealliance/rustix/blob/38327261924aa71059a4cdeeea2286ec88bd9146/src/backend/libc/conv.rs#L98). both, IMO, are insufficiently paranoid that the syscall completed as expected particularly when we're handling guest data.

lib/propolis/src/vmm/mem.rs
```diff
@@ -767,23 +789,91 @@ impl<'a, T: AsRef<[SubMapping<'a>]>> MappingExt for T {
...
-            })
-            .collect::<Vec<_>>();
-
-        let read = unsafe {
-            libc::preadv(fd, iov.as_ptr(), iov.len() as libc::c_int, offset)
-        };
-        if read == -1 {
-            return Err(Error::last_os_error());
+        let mut total_capacity: usize = 0;
+        for mapping in self.as_ref().iter() {
```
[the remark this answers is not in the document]
iximeow: ended up lifting this out into a function too, not that we're likely to have many other `&[SubMapping<'_>]`... https://github.com/oxidecomputer/propolis/pull/985/commits/4f42ff42a7d426ace0980421a139342081d67d6f#diff-da55090e4f645da5113d4b072b3b33b9284329afc56a081d2f22f1791709f535R779-R793

lib/propolis/src/vmm/mem.rs
iximeow: > if a guest _does_ decide it wants to screw up its own iovec

I'm definitely thinking more about "same page submitted in another I/O concurrently", where we're working on the same range on a different thread in Propolis, more than the guest happens to scribble on memory at the same time.

before, I'd broken the rules but hadn't seen a better option and really just needed to justify to myself why it _might_ be OK. now I just want to emphasize _things can be happening_ (more than a normal `copy_nonoverlapping`, since I expect that someone looking at this file wouldn't immediately think of the page-submitted-in-other-I/Os case), but the contract of `copy_nonoverlapping` doesn't care about those things.

but generally "operate on `SubMapping` exclusively through pointer operations, never create a reference to or slice of the referenced data" probably should be a comment on `SubMapping` (the `Safety` comments on `raw_readable` and `raw_writable` substantiate this: "The caller must never create a reference to the underlying memory region.")

> volatile copy intrinsic

would be a bit _too_ limiting, I think! that intrinsic would demand a byte-wise copy loop, not `memcpy()` and widen-to-u64-if-not-AVX paths there, but the memory here isn't _volatile_.

lib/propolis/src/vmm/mem.rs
iximeow: done in both-now-lifted-to-one-function cases, thanks

Issue #984, opened 2025-12-10
want vCPU binding


iximeow:
over in byhve, `vm_run` tries to make sure that vCPU-specific resources (cyclics, etc) are kept on the same CPU as whichever is current. `vm_localize_resources` makes this happen, and up at the top of that is `if (vcpu->lastloccpu == curcpu)`. if we're not on the same CPU, we'll have to move resources (such as `vlapic_localize_resources` moving a cyclic), and in the case of `vlapic_localize_resources` that's serialized through `cpu_lock`.

on a largeish system (256 threads) with a largeish VM (192 vCPUs), we're limited at somewhere in the range of 500-700k `vm_run()`/second because of this(!)

while we don't want to be doing 500-700k interrupts/second, ideally, we can save a bunch of system time by binding vCPUs to physical CPUs and ensuring `if (vcpu->lastloccpu == curcpu)`. at the very least, perhaps binding vCPU threads to a smaller CPU group so the odds are improved.

iximeow:
> and to do that allocation in the control plane as part of placing the VM

thinking about the future `propolis-server` API a bit more while writing #991, this is a bit subtle. the vCPU->CPU mapping the control plane will chose will not be consistent across sleds, so the "proper treatment" will have to include data like `MigrationTarget::replace_components` that we use to amend a source instance spec into something suitable for the destination `propolis-server`.

PR #991, opened 2025-12-17
bins: intial CPU binding support

iximeow:
this is the interim option I'd described in https://github.com/oxidecomputer/propolis/issues/984.

add bindings for `processor_bind()`, the mechanism to make this happen. then, in propolis-standalone, allow a choice of how binding should be done: allow vCPUs to roam with `cpu_binding = "any"` or pin vCPUs to CPUs 1:1 from the last processor downwards with
`cpu_binding = "from-last"`.

in `propolis-server`, pick between these two strategies automatically based on how the vCPU count compares to the number of available processors. a VM that would use more than half of available processors is bound in the same way as `propolis-standalone` with `cpu_binding = "from-last"`, but otherwise continues to have unbound vCPUs.

in the future, Nexus will dictate `propolis-server`'s vCPU bindings, but exactly how is a bit of an open question.

### DIFF
--- a/bin/propolis-standalone/src/main.rs
+++ b/bin/propolis-standalone/src/main.rs
@@ -298,11 +298,13 @@ impl Instance {
 
                 let first_bound_cpu = total_cpus - vcpu_count;
                 let bind_cpus =
-                    (first_bound_cpu..total_cpus).map(|v| Some(v)).collect();
+                    (first_bound_cpu..total_cpus).map(Some).collect();
                 slog::info!(
                     &log,
                     "Explicit CPU binding requested";
-                    "bind_cpus" => #?bind_cpus
+                    "last_cpu" => total_cpus,
+                    "first_cpu" => first_bound_cpu,
+                    "vcpu_count" => vcpu_count,
                 );
                 bind_cpus
             }
--- a/bin/propolis-server/src/lib/vm/ensure.rs
+++ b/bin/propolis-server/src/lib/vm/ensure.rs
@@ -601,12 +601,12 @@ async fn initialize_vm_objects(
     let cpu_threshold = total_cpus / 2;
     let bind_cpus = if vcpu_count > cpu_threshold {
         if vcpu_count > total_cpus {
-            anyhow::anyhow!("spec requested more CPUs than are online!");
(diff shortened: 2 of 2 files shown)


crates/pbind/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "pbind"
```
hawkw: hm, since this is a low-level binding to the C headers, should this perhaps be `pbind-sys`?

bin/propolis-standalone/src/main.rs
```diff
@@ -286,7 +286,29 @@ impl Instance {
...
         let machine = state.machine.as_ref().unwrap();
 
-        for vcpu in machine.vcpus.iter().map(Arc::clone) {
+        let bind_cpus = match this.0.config.main.cpu_binding {
+            Some(config::BindingStrategy::FromLast) => {
+                let mut bind_cpus = vec![None; machine.vcpus.len()];
+                let total_cpus =
+                    pbind::online_cpus().expect("can get processor count");
+                let vcpu_count: i32 =
+                    machine.vcpus.len().try_into().expect("<2^31 vCPUs");
```
hawkw: nit, considered not that big a deal: this is an error in the user's config file, rather than e.g. a syscall we expect should always work not working. can we make that a little clearer perhaps?

bin/propolis-standalone/src/config.rs
```diff
@@ -69,6 +72,16 @@ pub struct Main {
...
 }
 
+#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum BindingStrategy {
+    /// vCPUs are not bound to any particular physical processor.
+    Any,
+    /// vCPUs are bound to the highest-numbered processors in the system, one
+    /// vCPU per CPU, with the last vCPU bound to the last physical processor.
+    FromLast,
```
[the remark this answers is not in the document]
iximeow: i was really struggling with the name here.. "from last" (and descending) read to me like "first vCPU on last CPU, second vCPU on next-from-last, tihrd vCPU on the next, ...", so i've changed this to `UpperHalf` and included a little doodle that... might be better, i dunno.

PR #993, opened 2025-12-19
distinguish probes across NVMe devices

iximeow:
NVMe commands are unique in a device across queue ID (`qid` in various parts of Propolis) and command ID (`cid` in various parts of Propolis), but are ambiguous when looking at an entire VM: two different NVMe devices will have a "queue 1", and their distinct "queue 1" may also have "command 1234" submitted at the same time.

Pick a device ID for NVMe controllers, and plumb that through to distinguish queues similar to how we distinguish queues for block devices: a `devq_id` construction that makes a device and queue unique across the VM.

In the NVMe case, a `devq_id` is not uniquely identifying across the life of a VM, as the queues can be destroyed, recreated, the controller reset, etc. But absent such an administrative operation, this helps untangle what's happening across a VM with many disks.

Note that block attachment IDs and NVMe IDs are similar in name, and probably similar in practice, but are tracked distinctly. This is similar to the difference between block qeuue IDs and NVMe queue IDs: the architcture is similar, but the mapping between them is more arbitrary and may change as an implementation detail of Propolis.

---

a few of us (@ahl among others) were looking at the nvme probes as a "top-of-propolis" view of submission->completion time and found, to our surprise, that sometimes we'd get really bogus data. it took a bit of thinking to realize that we were probably seeing duplicate `(qid, cid)` tuples since we were loading a bunch of disks in the VM. I want to see that not happen any more for good measure, but from first principles the probes don't seem sufficient for analysis with multiple disks at the moment.

`nvme-trace.d` only looked at `cid`, not the queue ID, so it'd be even more ambiguous. it probably could use a sprinkling of `devq_id`/`arg0` too.

### DIFF
--- a/lib/propolis/src/hw/nvme/mod.rs
+++ b/lib/propolis/src/hw/nvme/mod.rs
@@ -32,14 +32,7 @@ mod requests;
 use bits::*;
 use queue::{CompQueue, QueueId, SubQueue};
 
-/// Static for generating unique NVMe device identifiers within a process.
-///
-/// Numbering across NVMe devices means that the combination of NVMe
-/// `device_id` and queue ID identify either an NVMe submission or completion
-/// queue across a VM.
-///
-/// Guests typically will configure submission queue N to complete into
-/// completion queue N, but this is not a mandated device configuration.
+/// Static for generating unique NVMe device identifiers across a VM.
 static NEXT_DEVICE_ID: AtomicU32 = AtomicU32::new(0);
 
 type DeviceId = u32;
@@ -184,6 +177,8 @@ const MAX_NUM_IO_QUEUES: usize = MAX_NUM_QUEUES - 1;
 struct NvmeCtrl {
     /// A distinguishing identifier for this NVMe controller across the VM.
     /// Useful mostly to distinguish queues and commands as seen in probes.
+    /// `device_id` is held constant across NVMe resets, but not persisted
+    /// across export and import.
     device_id: DeviceId,
 
     /// Internal NVMe Controller state
@@ -921,7 +916,7 @@ impl PciNvme {
                 }
             }));
 
-            // Peel out device ID before we move it into the Mutex below.
+            // Cache device ID before we move it into the Mutex below.
             let device_id = state.device_id;
 
             PciNvme {


lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -154,6 +182,10 @@ const MAX_NUM_IO_QUEUES: usize = MAX_NUM_QUEUES - 1;
 
 /// NVMe Controller
 struct NvmeCtrl {
+    /// A distinguishing identifier for this NVMe controller across the VM.
+    /// Useful mostly to distinguish queues and commands as seen in probes.
```
ahl:
```suggestion
    /// Useful mostly to distinguish queues and commands for parameters to DTrace USDT probes.
```

...more keyword-y for future lazy grepping?

lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -1188,9 +1237,12 @@ impl PciNvme {
                 })
             {
                 let block_qid = queue::sqid_to_block_qid(sqid);
+                let devsq_id = devq_id(self.device_id, sqid);
+                let block_devqid =
+                    block::devq_id(self.block_attach.device_id(), block_qid);
                 probes::nvme_block_notify!(|| (
-                    sqid,
-                    u16::from(block_qid),
+                    devsq_id,
```
ahl: what does the "s" signify? just asking for myself

lib/propolis/src/hw/nvme/queue.rs
```diff
@@ -20,13 +21,13 @@ use thiserror::Error;
...
-    fn nvme_cq_dbbuf_read(qid: u16, val: u32, tail: u16) {}
-    fn nvme_cq_dbbuf_write(qid: u16, head: u16) {}
-    fn nvme_cq_dbbuf_write_shadow(qid: u16, head: u16) {}
+    fn nvme_cqe(devcq_id: u64, idx: u16, phase: u8) {}
+    fn nvme_sq_dbbuf_read(devsq_id: u64, val: u32, tail: u16) {}
+    fn nvme_sq_dbbuf_write(devsq_id: u64, head: u16) {}
+    fn nvme_sq_dbbuf_write_shadow(devsq_id: u64, head: u16) {}
+    fn nvme_cq_dbbuf_read(devcq_id: u64, val: u32, tail: u16) {}
+    fn nvme_cq_dbbuf_write(devcq_id: u64, head: u16) {}
+    fn nvme_cq_dbbuf_write_shadow(devcq_id: u64, head: u16) {}
```
ahl: is the c or s distinction obvious to folks familiar with this code?

lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -31,13 +32,40 @@ mod requests;
...
+    fn nvme_doorbell(off: u64, devq_id: u64, is_cq: u8, val: u16) {}
     fn nvme_doorbell_admin_cq(val: u16) {}
     fn nvme_doorbell_admin_sq(val: u16) {}
     fn nvme_admin_cmd(opcode: u8, prp1: u64, prp2: u64) {}
-    fn nvme_block_notify(sqid: u16, block_qid: u16, occupied_hint: u16) {}
+    fn nvme_block_notify(devsq_id: u64, block_devqid: u64, occupied_hint: u16) {
+    }
+}
+
+/// Combine an NVMe device and queue ID into a single u64 for probes
```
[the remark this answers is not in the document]
iximeow: don't want to go quite that far: the NVMe controller can be reset, which will destroy queues, abandon any in-flight work, and probably have queues with the same ID re-created when the guest sets up the controller again. but I'll adjust this to say a bit more.

PR #998, opened 2025-12-23
VirtIO 1.0 and multi-queue support

dancrossnyc:


dancrossnyc:
Cc @papertigers  @rcgoodfellow

crates/viona-api/src/ffi.rs
```diff
@@ -0,0 +1,145 @@
...
+#![allow(non_camel_case_types)]
+
+use libc::size_t;
+use std::ffi::c_void;
+
+const fn vna_ioc(ioc: i32) -> i32 {
+    const V: i32 = b'V' as i32;
+    const C: i32 = b'C' as i32;
+    V << 16 | C << 8 | ioc
+}
```
hawkw: my  assumption upon reading this code (and perhaps I am wrong here) is that `ioc` will always be 8 bits wide, since everything above that will be clobbered, is that true? and if so, perhaps we ought to take it as a `u8` here?

crates/viona-api/src/ffi.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub _pad: [u16; 3],
+    pub rm_addr: u64,
+    pub rm_msg: u64,
+}
+
+#[repr(C)]
+#[derive(Default)]
+pub struct vioc_intr_poll {
+    pub vip_nrings: u16,
+    pub vip_status: [u32; howmany(VIONA_MAX_QPAIRS, 32)],
```
citrus-it: Since you have the padding explicitly shown in other structures here do you want to do the same for the hole here?

crates/viona-api/src/ffi.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+#[repr(C)]
+#[derive(Default)]
+pub struct vioc_nofity_mmio {
```
citrus-it:
```suggestion
pub struct vioc_notify_mmio {
```

lib/propolis/src/hw/virtio/mod.rs
```diff
@@ -16,24 +24,149 @@ mod queue;
...
 pub use viona::PciVirtioViona;
 
+bitflags! {
+    pub struct LegacyFeatures: u64 {
+        const NOTIFY_ON_EMPTY = 1 << 24;
+        const ANY_LAYOUT = 1 << 27;
+    }
+}
+
+/// Describes the VirtIO "mode" exposed by the device.  This
```
citrus-it: Truncated comment?

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -125,37 +161,125 @@ impl<D: PciVirtio + Send + Sync + 'static> pci::Device for D {
...
+                PCI_CFG_CAP_REGS.process(&mut rwo, |id, rwo| match rwo {
+                    RWOp::Read(ro) => vs.pci_cfg_cap_read(self, id, ro),
+                    RWOp::Write(wo) => vs.pci_cfg_cap_write(self, id, wo),
+                });
+            }
+            VirtioCfgCapTag::SharedMemory => {
+                unimplemented!("VirtIO Shared Memory is unsupported");
+            }
+            VirtioCfgCapTag::Vendor => {
+                unimplemented!("VirtIO Vendor capabilities are unsupprted");
```
citrus-it:
```suggestion
                unimplemented!("VirtIO Vendor capabilities are unsupported");
```

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -280,23 +452,444 @@ impl PciVirtioState {
...
+                // Read-only for driver
+            }
+            CommonConfigReg::QueueSelect => {
+                let mut state = self.state.lock().unwrap();
+                state.queue_select = wo.read_u16();
+            }
+            CommonConfigReg::QueueSize => {
+                let state = self.state.lock().unwrap();
+                let offered =
+                    VqSize::try_from(wo.read_u16()).expect("bad queue size");
```
citrus-it: If a guest sends a nonsensical queue size here (such as 0 or something which is not a power of two) we probably ought not panic.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -659,11 +1257,14 @@ impl MigrateMulti for dyn PciVirtio {
...
         vs.set_intr_mode(ps, ps.get_intr_mode().into(), true);
 
-        // Perform a (potentially spurious) update notification for the BAR
+        // Perform a (potentially spurious) update notification for the BARs
         // containing the virtio registers.  This ensures that anything
-        // interested in the placement of that BAR (such as the notify-port
-        // logic) is kept well aware
+        // interested in the placement of those BARs (such as the notify
+        // logic) is configured properly.
         self.bar_update(ps.bar(pci::BarN::BAR0).unwrap());
```
citrus-it: Although we aren't using it yet, the BAR creation logic only creates BAR0 if "Legacy" or "Transitional" mode was selected. We probably need the same existence check as for BAR2 here.

lib/propolis/src/hw/virtio/viona.rs
citrus-it: It seems we'll need to push at least the number of selected queue pairs into the migration state here.

citrus-it (commented):
I've reviewed this mostly from a VirtIO/viona perspective.

crates/viona-api/src/ffi.rs
dancrossnyc: Done.

crates/viona-api/src/ffi.rs
```diff
@@ -0,0 +1,145 @@
...
+    use super::*;
+
+    #[test]
+    fn test_vna_ioc() {
+        assert_eq!(vna_ioc(0x22), 0x00_56_43_22);
+    }
+}
+
+pub const VIONA_MIN_QPAIRS: usize = 1;
+pub const VIONA_MAX_QPAIRS: usize = 0x100;
```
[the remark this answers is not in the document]
dancrossnyc: Done.

crates/viona-api/src/lib.rs
```diff
@@ -129,16 +131,18 @@ impl VionaFd {
...
+                | VNA_IOC_RING_KICK
+                | VNA_IOC_RING_PAUSE
+                | VNA_IOC_RING_INTR_CLR
+                | VNA_IOC_VERSION
+                | VNA_IOC_SET_NOTIFY_IOP
+                | VNA_IOC_SET_PROMISC
+                | VNA_IOC_GET_MTU
+                | VNA_IOC_SET_MTU
+                | VNA_IOC_SET_PAIRS
+                | VNA_IOC_SET_USEPAIRS,
```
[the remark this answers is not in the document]
dancrossnyc: Done.

lib/propolis/src/hw/virtio/mod.rs
dancrossnyc: Rather, vestigial holdover from older wording.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -56,67 +60,99 @@ impl From<pci::IntrMode> for IntrMode {
...
-    /// register in the device BAR has changed.
-    fn notify_port_update(&self, state: Option<NonZeroU16>) {}
+    /// Handles notification that the IO port representing the queue
+    /// notification register in the device BAR has changed.
+    fn notify_port_update(&self, state: Option<NonZeroU16>) {
+        let _used = state;
+    }
+
+    /// Handles notification that the MMIO address representing the queue
+    /// notification register in the device BAR has changed.
```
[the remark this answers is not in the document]
dancrossnyc: Reworded; PTAL.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -56,67 +60,99 @@ impl From<pci::IntrMode> for IntrMode {
...
-                    bstate.value as u16 + LEGACY_REG_OFF_QUEUE_NOTIFY as u16,
-                )
-                .expect("addition does not wrap");
-                Some(notify_port_addr)
-            } else {
-                None
-            };
-            self.notify_port_update(port);
+        match bstate.id {
+            pci::BarN::BAR0 => {
```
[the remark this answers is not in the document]
dancrossnyc: I don't think that's true. If `bstate.decode_en` is false, then the `.then()` call will return `None`, which we then pass to `self.notify_port_update()`.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -125,37 +161,125 @@ impl<D: PciVirtio + Send + Sync + 'static> pci::Device for D {
...
+                COMMON_CFG_CAP_REGS.process(&mut rwo, |id, rwo| {
+                    if let RWOp::Read(ro) = rwo {
+                        vs.isr_cfg_cap_read(id, ro);
+                    }
+                });
+            }
+            VirtioCfgCapTag::Device => {
+                COMMON_CFG_CAP_REGS.process(&mut rwo, |id, rwo| {
+                    if let RWOp::Read(ro) = rwo {
+                        vs.device_cfg_cap_read(id, ro);
```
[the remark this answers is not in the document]
dancrossnyc: I added some words here, using a MAC address as an example: this is the device-independent code, so it feels a bit discordant, but as a motivating example is fine. What do you think?

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -280,23 +452,444 @@ impl PciVirtioState {
...
+                        current.desc_addr,
+                        current.avail_addr,
+                        offered_used_addr,
+                    );
+                }
+            }
+            CommonConfigReg::QueueNotifyData => {
+                // Read-only for driver
+            }
+            CommonConfigReg::QueueReset => self.set_needs_reset(dev),
```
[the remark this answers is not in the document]
dancrossnyc: ...and now I have blatantly plagiarized that into a comment. :-D

lib/propolis/src/hw/virtio/pci.rs
dancrossnyc: Done.

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -285,6 +431,10 @@ impl VirtQueue {
         avail.cur_avail_idx = Wrapping(info.avail_idx);
         used.used_idx = Wrapping(info.used_idx);
     }
+
+    /// Accummulates a sequence of available descrptors into a `Chain`.
```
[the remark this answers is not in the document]
dancrossnyc: Done.

lib/propolis/src/hw/virtio/viona.rs
dancrossnyc: We do; that information is bundled up into the `VirtioQueues` type and its migration machinery.

lib/propolis/src/hw/virtio/pci.rs
dancrossnyc: On reflection, it seems like maybe the proper thing to do here is signal that the device needs to be reset. Done now. PTAL.

lib/propolis/src/hw/virtio/pci.rs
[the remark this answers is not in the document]
citrus-it: Right, thanks.

lib/propolis/src/hw/virtio/mod.rs
```diff
@@ -16,24 +24,149 @@ mod queue;
...
+        self as u16
+    }
+
+    /// Maps a VirtIO Device ID to a PCI Device Class.
+    pub fn pci_class(self) -> u8 {
+        use crate::hw::pci as pci_hw;
+        match self {
+            Self::Network => pci_hw::bits::CLASS_NETWORK,
+            Self::Block | Self::NineP => pci_hw::bits::CLASS_STORAGE,
+            _ => pci_hw::bits::CLASS_UNCLASSIFIED,
```
iximeow: I realize this just moves the class definitions from `PciVirtioState::new()` to one more central lookup, but there's not really an authoritative spec definition for the device classes to use, is there? looking at QEMU I see that 9P is classed a `CLASS_NETWORK` (o_O?) and other devices seem to be `PCI_CLASS_OTHERS == 0xff` where `CLASS_UNCLASSIFIED` has us setting the class to 0.

not asking to change any of this, just wondering if there's a reference to note (or non-reference to lament :) )

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -207,27 +334,34 @@ pub struct PciVirtioState {
...
-            device_id: dev_id,
-            sub_vendor_id: VENDOR_VIRTIO,
-            sub_device_id: sub_dev_id,
-            class: dev_class,
+            vendor_id,
+            device_id,
+            sub_vendor_id,
+            sub_device_id,
+            device_class,
             ..Default::default()
```
iximeow: does it make sense to have a `device_type.pci_ident()` that does this shuffling? then it could return a Result and `pci_dev_id/pci_sub_dev_id/pci_class` can be Result-ful as appropriate for their mappings.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -236,38 +370,82 @@ impl PciVirtioState {
...
+        if mode == virtio::Mode::Transitional || mode == virtio::Mode::Modern {
+            builder =
+                builder.add_bar_mmio(pci::BarN::BAR2, 4 * PAGE_SIZE as u32);
+            builder = builder.add_cap_vendor(
+                VirtioCfgCapTag::Common.into(),
+                COMMON_CFG_CAP_SIZE,
+            );
+            // Note: we don't presently support a non-zero multiplier for the
+            // notification register, so we don't need to size this for the
+            // number of queues; hence the fixed size.
```
iximeow: me not really understanding the value of a non-zero multiplier: if I'm reading correctly, the guest notifies a device by writing a virtqueue number to this register, so it's not like a write to this register notifies _all_ queues. but with a non-zero multiplier you could do something like write to queue 1 to say that.. queue 0 has available buffers? is it just an error to notify queue N about a virtqueue other than N?

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -236,38 +370,82 @@ impl PciVirtioState {
...
-
+        // With respect to layout, for the time being, we are unconditionally
+        // transitional, meaning that we support both the legacy and common
+        // configuration layouts.
+        let common_config = RegMap::create_packed_passthru(
+            4 * PAGE_SIZE,
+            &[
+                (VirtioConfigRegBlock::Common, COMMON_REG_SIZE),
+                (VirtioConfigRegBlock::RazWi, PAGE_SIZE - COMMON_REG_SIZE),
+                (VirtioConfigRegBlock::DeviceConfig, cfg_size),
```
iximeow: if you don't want to add it here I'll add it later: we ought to assert `cfg_size < PAGE_SIZE`, and even stronger, that `cfg_size + LEGACY_REG_SIZE < 0x200` (for the I/O bar). 9P gets surprisingly (to me) close: `VIRTIO_9P_CFG_SIZE` is `0x102`!

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -830,43 +1451,241 @@ impl VirtioIntr for MsiIntr {
...
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+enum IsrStatusReg {
+    IsrStatus,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+enum LegacyConfigReg {
+    DeviceFeature,
+    DriverFeature,
+    QueueAddr4k,
```
iximeow: more pedantics: I'm +1 on `Feat*` -> `*Feature` as "Driver Features" is a phrase you'd find searching the VirtIO spec, but on that basis can we call this `QueueAddress` too? since that's the description in 4.1.4.10 `Legacy Interfaces: A note on PCI Device Layout`.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -830,43 +1451,241 @@ impl VirtioIntr for MsiIntr {
...
+
+lazy_static! {
+    static ref COMMON_CFG_CAP_REGS: RegMap<CommonCfgCapReg> = {
+        let layout = [
+            (CommonCfgCapReg::CapLen, 1),
+            (CommonCfgCapReg::CfgType, 1),
+            (CommonCfgCapReg::Bar, 1),
+            (CommonCfgCapReg::Id, 1),
+            (CommonCfgCapReg::Padding, 2),
+            (CommonCfgCapReg::Offset, 4),
```
iximeow: it is very funky to look at this and think we're padding to... 6??? :grin: remembering there's an implicit vendor capability and next pointer before this struct it makes much more sense.

could you note this is the layout described by `struct virtio_pci_cap` in 4.1.4?

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -830,43 +1451,241 @@ impl VirtioIntr for MsiIntr {
...
-    IsrStatus,
-    MsixVectorConfig,
-    MsixVectorQueue,
+    PciIsrStatus,
+    ConfigMsixVector,
+    QueueMsixVector,
+}
+
+lazy_static! {
+    static ref COMMON_REGS: RegMap<CommonConfigReg> = {
```
iximeow: just thinking aloud: feels like it would be pretty nice to have a macro like...

```rust
define_regmap! {
    device_feature_select: u32 {
        read: |state, ro| {
            let state = self.state.lock().unwrap();
            ro.write_u32(state.device_feature_select);
        },
        write: |state, ro| {
            let state = self.state.lock().unwrap();
            self.device_feature_select = ro.read_u32();
        },
.... etc
}
```
that does the combo of declaring this mapping, plus registers and their handlers.. I'm clearly still turning over the idea of reworking how we map these enum variants to their handlers since our conversation a bit ago.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -125,37 +161,128 @@ impl<D: PciVirtio + Send + Sync + 'static> pci::Device for D {
...
     fn bar_rw(&self, bar: pci::BarN, mut rwo: RWOp) {
         let vs = self.virtio_state();
-
-        assert_eq!(bar, pci::BarN::BAR0);
-        let map = match vs.map_which.load(Ordering::SeqCst) {
-            false => &vs.map_nomsix,
-            true => &vs.map,
+        let map = match bar {
+            pci::BarN::BAR0 => {
+                if vs.legacy_map_use_msix.load(Ordering::Acquire) {
```
iximeow:
```suggestion
                if vs.legacy_map_use_msix.load(Ordering::SeqCst) {
```

while I agree that Acquire here and Release below are not wrong here, none of this should be hot. so I also prefer keeping the ordering stronger until we need otherwise.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -280,23 +458,456 @@ impl PciVirtioState {
...
+        }
+    }
+
+    fn common_write(
+        &self,
+        pci_state: &pci::DeviceState,
+        dev: &dyn VirtioDevice,
+        id: &CommonConfigReg,
+        wo: &mut WriteOp,
+    ) {
```
iximeow: we're missing maintenance of `config_generation` here right? it seems like we'd want `common_read` to set a bit such that the next `common_write` (... to a not-ignored register) increments the generation and unset that bit (`4.1.4.3.1` talks about why not just increment on writes, which makes sense even if it's a bit silly and unfortunate)

I don't have a good sense of how much we _Need_ to maintain `config_generation` or how that interacts with guest drivers, but at the least if we're not doing it now it seems worth getting to quickly.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -280,23 +458,456 @@ impl PciVirtioState {
...
+                    Err(_) => {
+                        // Bad queue size.
+                        self.set_needs_reset(dev);
+                    }
+                    Ok(offered) => {
+                        if let Some(queue) = self.queues.get(state.queue_select)
+                        {
+                            let mut size = queue.size.lock().unwrap();
+                            *size = offered;
+                        }
```
iximeow:
```suggestion
                        let Some(queue) = self.queues.get(state.queue_select) else {
                            // Invalid queue: drop the write.
                            return;
                        }
                        let mut size = queue.size.lock().unwrap();
                        *size = offered;
```

here the wrapping might end up hideouser.. but in general this helps fight the rightward drift of getting the queue and gives us a clearer place to say why the non-behavior on `queue_select` is reasonable.

not demanding this change here, especially because this continues the prior pattern. but if you don't get to it here I'll probably do it in a followup next week.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -280,23 +458,456 @@ impl PciVirtioState {
...
+                    }
+                    // Everything else is read-only for the driver.
+                    _ => {}
+                }
+            }
+            PciCfgCapReg::PciData => {
+                // TODO: Handle the write.
+            }
+        }
+    }
```
iximeow: how TODO is this? it seems like leaving this unplumbed would be pretty unfortunate but clearly it's not necessary to see a working NIC?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -585,16 +856,22 @@ impl PciVirtio for PciVirtioViona {
     fn pci_state(&self) -> &pci::DeviceState {
         &self.pci_state
     }
+    // The notification addresses (both port and MMIO) for the device can change
+    // due to guest action, or other administrative tasks within propolis.  We
+    // want to update the in-kernel IO port hook only in the former case, when
+    // the device emulation is actually running.
```
iximeow: I wouldn't move this comment out onto the function itself: it's really "why is it correct to guard `hdl.set_notify_*_port` on `self.indicator.state() == IndicatedState::Run`".

maybe leave this one where it was and leave a `// Only update the kernel for the same reason as notify_port_update()` in `notify_mmio_addr_update`? I think what you're going for is that this really describes the `if` in both functions, at least.

lib/propolis/src/hw/virtio/pci.rs
citrus-it: The data written to the register can be more arbitrary if `VIRTIO_F_NOTIF_CONFIG_DATA` is negotiated, but that still doesn't need a multiplier.
If the multiplier was chosen to put each register on a separate cache line I suppose there could be a benefit there.
Qemu has a flag (`VIRTIO_PCI_FLAG_PAGE_PER_VQ`) that sets the multiplier to 0x1000, otherwise it uses 0x4.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -31,10 +37,133 @@ use tokio::task::JoinHandle;
 // Re-export API versioning interface for convenience of propolis consumers
 pub use viona_api::{api_version, ApiVersion};
 
+pub const RX_QUEUE_SIZE: u16 = 0x800;
+pub const TX_QUEUE_SIZE: u16 = 0x100;
+pub const CTL_QUEUE_SIZE: u16 = 32;
```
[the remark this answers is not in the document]
dancrossnyc: Done.

lib/propolis/src/hw/virtio/mod.rs
```diff
@@ -16,24 +24,149 @@ mod queue;
...
+    /// that space into PCI device IDs is dependent on the mode; for devices
+    /// following the VirtIO 1.0 and later specifications, this is straight
+    /// forward: just add 0x1040 to the VirtIO ID.
+    ///
+    /// However, for legacy and transitional mode devices, the mapping is
+    /// irregular, and defined by a table.  Moreover, not all virtio devices
+    /// have a well-defined transitional mapping.  This last thing is not
+    /// really an issue for us, since we only expose a handful of device
+    /// models; regardless, we provide mappings for everything defined in
+    /// the VirtIO spec.
```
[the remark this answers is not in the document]
dancrossnyc: I word-smithed it a bit, but think I've captured the spirit of what you're getting at. PTAL.

lib/propolis/src/hw/virtio/mod.rs
```diff
@@ -16,24 +24,149 @@ mod queue;
...
+                Self::Socket => Ok(0x1012), // Taken from QEMU, used by Linux
+                _ => Err(self),
+            },
+        }
+    }
+
+    /// Maps a VirtIO Device ID to a PCI Device Sub ID.
+    /// XXX: Check these mappings against some reference.
+    pub fn pci_sub_dev_id(self) -> u16 {
+        self as u16
```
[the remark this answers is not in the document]
dancrossnyc: Yeah, this one's a bit of a pickle.  As Andy said, c-bhyve does it the way we do, and it seems valuable to maintain parity with it; on the other hand, the spec _does_ contain that wording. In practice, I don't think any drivers really care. I find myself falling on the side of wanting to maintain parity with c-bhyve here.

lib/propolis/src/hw/virtio/pci.rs
dancrossnyc: I suppose it's also possible to support multiple queue notifications simultaneously if you use the multiplier, using multiple threads or something in the host.

lib/propolis/src/hw/virtio/mod.rs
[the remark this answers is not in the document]
citrus-it: I don't think this is doing it like c-bhyve at the moment.

What we want is this (for a network device - ID 1, dev 0x1000

### Legacy/Transitional:

- Vendor: `0x1af4`
- Device: `0x1000`
- Subvendor: `0x1af4`
- Subdevice: `0x1`
- PCI Revision: `0x0`

### Modern:

- Vendor: `0x1af4`
- Device: `0x1041`
- Subvendor: `0x1af4`
- Subdevice: `0x1041`
- PCI Revision: `0x1`

My inference from the spec wording is that we don't want, say, a modern network device having `0x1af4,1` in the subdevice fields to reduce the change of a guest attaching a legacy driver based on that. Anything >= 0x40 achieves it.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -236,38 +370,82 @@ impl PciVirtioState {
...
 
-        // XXX: properly size the legacy cfg BAR
-        builder = builder.add_bar_io(pci::BarN::BAR0, 0x200);
+        if mode == virtio::Mode::Transitional || mode == virtio::Mode::Legacy {
+            // XXX: properly size the legacy cfg BAR
+            builder = builder.add_bar_io(pci::BarN::BAR0, 0x200);
+        }
+        if mode == virtio::Mode::Transitional || mode == virtio::Mode::Modern {
+            builder =
+                builder.add_bar_mmio(pci::BarN::BAR2, 4 * PAGE_SIZE as u32);
```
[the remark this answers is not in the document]
dancrossnyc: As Andy said; it's arbitrary. Popolis is matching c-bhyve here.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -830,43 +1451,241 @@ impl VirtioIntr for MsiIntr {
...
+    ConfigMsixVector,
+    NumQueues,
+    DeviceStatus,
+    ConfigGeneration,
+
+    /// Configuration information for a specific queue.
+    QueueSelect,
+    QueueSize,
+    QueueEnable,
+    QueueMsixVector,
```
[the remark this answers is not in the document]
dancrossnyc: Done (I don't think I even realized it was out of order!).

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -830,43 +1451,241 @@ impl VirtioIntr for MsiIntr {
...
+#[repr(u32)]
+enum VirtioCfgCapTag {
+    Common = 1,
+    Notify = 2,
+    Isr = 3,
+    Device = 4,
+    Pci = 5,
+    SharedMemory = 8,
+    Vendor = 9,
+}
```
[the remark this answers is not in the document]
dancrossnyc: New! Improve! Order now! Operators are standing by!

(Done)

lib/propolis/src/hw/virtio/pci.rs
dancrossnyc: We haven't found a driver that requires it, so it's pretty "meh, should be done... _eventually_."  Given the time constraints, it didn't seem worth it at the moment.

lib/propolis/src/hw/virtio/pci.rs
dancrossnyc: The wrapping was worse, but I introduced a temporary to hold the queue select value and it was ok.

In general, I haven't stressed too hard on things like this because I think we ought to redo the entire mechanism.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -31,10 +37,133 @@ use tokio::task::JoinHandle;
...
+pub const fn max_num_queues() -> usize {
+    PROPOLIS_MAX_MQ_PAIRS as usize * 2
+}
+
 const ETHERADDRL: usize = 6;
 
-/// Viona's in-kernel emulation of the device VirtQueues is performed in what it
-/// calls "vrings".  Since the userspace portion of the Viona emulation is
+/// Types and so forth for supporting the control queue.
+pub mod control {
```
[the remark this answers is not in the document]
dancrossnyc: Oops; I missed this in my earlier swing. I added a comment here.

lib/propolis/src/hw/virtio/mod.rs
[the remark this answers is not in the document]
iximeow: yeah, i think that helps. thanks

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -249,14 +400,116 @@ impl PciVirtioViona {
...
+    fn ctl_queue_notify(&self, vq: &Arc<VirtQueue>) {
+        if let Some(mem) = self.pci_state.acc_mem.access() {
+            while !vq.avail_is_empty(&mem) {
+                let mut chain = Chain::with_capacity(4);
+                let intrs_en = vq.disable_intr(&mem);
+                while let Some((_idx, _len)) = vq.pop_avail(&mut chain, &mem) {
+                    let res = match self.ctl_msg(vq, &mut chain, &mem) {
+                        Ok(_) => control::Ack::Ok,
+                        Err(_) => control::Ack::Err,
+                    } as u8;
```
iximeow: minor Rust thing: I'd do the conversion as `let res = (<expr producing control::Ack>).as_u8()` or something. doing the `as u8` here has you jump back to `Ack` to double check that `Ack` variants are numbered, that it does actually have a `u8` repr (or discriminants at least are in range), ..

also fine to fix after.

iximeow (approved):
ok, I'm +1 on getting this in especially given the time constraints. I'd mentioned on the side I'd prefer to separate the renaming and the meatier changes for multiqueue into a pair of commits that remain distinct in git history - I still think this is worth doing and I'll come back here with a proposal branch splitting this into such a pair (or we get really itchy to get this landed and into dogfood)

I know when we've talked about it you'd mentioned that you've tested across Linux, Windows, OmniOS, and I _think_ a BSD, right? meaning that my questions re. virtio spec and TODOs above, we actually have pretty good confidence these don't leave guests in a risky spot.

Additionally, I recall you mentioned migrating guests in propolis-standalone and their NICs still working, so if I'm recalling correctly, that's great. I saw the conversation where Angela wasn't easily reproducing the relative improvement in net perf, do you know what's up with that?

on the whole, thanks for working on this (and Andy for the earlier review, and the kernel work!!). I hope to get back to you shortly with the proposed pair of commits.

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -737,62 +932,118 @@ pub struct Info {
...
-        if !(0..(u16::MAX as usize)).contains(&queues.len()) {
-            return Err(VirtQueuesError::BadQueueCount(queues.len()));
-        }
+        let len = AtomicUsize::new(len);
+        Self { len, queues }
+    }
 
-        Ok(Self { queues })
+    pub fn set_len(&self, len: usize) {
+        assert!(0 < len && len <= self.max_capacity());
```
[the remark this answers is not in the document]
dancrossnyc: Done.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -786,27 +1154,41 @@ impl VionaHdl {
...
+    /// Viona can install a hook in the associated VM at a specified address (in
+    /// either the guest port or physical address spaces) to recognize guest
+    /// writes that notify in-kernel emulated virtqueues of available buffers.
     ///
     /// With a non-zero argument, viona will attempt to attach such a hook,
-    /// replacing any currently in place. When the argument is None, any
+    /// replacing any currently in place.  When the argument is None, any
     /// existing hook is torn down.
-    fn set_notify_iop(&self, port: Option<NonZeroU16>) -> io::Result<()> {
+    fn set_notify_io_port(&self, port: Option<NonZeroU16>) -> io::Result<()> {
```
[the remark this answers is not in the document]
dancrossnyc: Ack.

PR #1002, opened 2026-01-05
Fix Intel CPUID leaf 4 cache topology for SMT

### LINKED ISSUE #1001: CPUID leaf 0x4 cache topology doesn't account for SMT
Linux guests with SMT enabled print this on boot
```
[    0.215000] smp: Brought up 1 node, 4 CPUs
[    0.215000] smpboot: Total of 4 processors activated (30011.39 BogoMIPS)
[    0.216000] BUG: arch topology borken
[    0.217000]      the SMT domain not a subset of the CLS domain
[    0.219000] BUG: arch topology borken
[    0.220000]      the SMT domain not a subset of the CLS domain
[    0.221000] BUG: arch topology borken
[    0.222000]      the SMT domain not a subset of the CLS domain
[    0.223000] BUG: arch topology borken
[    0.224000]      the SMT domain not a subset of the CLS domain
```
The problem is in https://github.com/oxidecomputer/propolis/blame/master/lib/propolis/src/cpuid.rs#L340. When specializing Intel's CPUID leaf 4 (Deterministic Cache Parameters),  L1/L2 caches are set as unshared
```
  if level < 3 {
      subleaf.eax &= !LEAF4_EAX_VCPU_MASK;
      // And leave that range 0: this means only one
      // vCPU shares the cache.
  }
```
As per the intel's documentation  https://cdrdv2-public.intel.com/775917/intel-64-architecture-processor-topology-enumeration.pdf, in Table 1-15. Reference for CPUID Leaf 04H 
> Bits 25-14: Maximum number of addressable IDs for logical processors sharing this
> cache**, ***
> ** Add one to the return value to get the result.
> ***The nearest power-of-2 integer that is not smaller than (1 + EAX[25:14]) is the number of
unique initial APIC IDs reserved for addressing different logical processors sharing this
cache.

The leaf 0xB tells the guest SMT is enabled with 2 threads per core. Linux uses leaf 4's `num_threads_sharing` field to figure out which CPUs share L2 cache, then builds scheduler domains expecting SMT siblings to be in the same cache sharing group.  So it complains when it sees "nobody shares L2" but also "you have 2 threads per core"

`calc_cache_topo_id()` https://github.com/torvalds/linux/blob/master/arch/x86/kernel/cpu/cacheinfo.c#L413, computes L2 cache ids from leaf 4,
`cpu_clustergroup_mask()` https://github.com/torvalds/linux/blob/master/arch/x86/kernel/smpboot.c#L670 returns CPUs sharing L2 and
 `build_sched_domain()` https://github.com/torvalds/linux/blob/master/kernel/sched/topology.c#L2472 then checks that SMT domain is subset of CLS domain

The same case for AMD in `fix_amd_cache_topo` is handled correctly
https://github.com/oxidecomputer/propolis/blob/f969fb7fd57c03bee03fe9a1d7458567a8e26b8d/lib/propolis/src/cpuid.rs#L174
```
  0b001 | 0b010 => {
      // L1/L2 shared by SMT siblings
      if self.has_smt { 2 } else { 1 }
  }
```

glitzflitz:
When SMT is enabled, L1/L2 caches should report being shared by 2 logical processors (the SMT siblings). Previously EAX[25:14] was always being set to 0, indicating no sharing which contradicts the SMT topology reported in leaf `0xB`. As per [1] EAX[25:14] indicates maximum number of addressable IDs for logical processors sharing this cache.

This mismatch causes linux guest to print "BUG: arch topology borken / the SMT domain not a subset of the CLS domain" during boot. Linux derives L2 cache sharing groups from leaf 4 and expects SMT siblings to share L2 but it was being informed that each vCPU has private L1/L2.

This brings the SMT handling logic in CPUID inline with whats being done for AMD in [`fix_amd_cache_topo()`](https://github.com/oxidecomputer/propolis/blob/f969fb7fd57c03bee03fe9a1d7458567a8e26b8d/lib/propolis/src/cpuid.rs#L174) which sets the sharing count to 2 when has_smt is true. This fixes #1001.

[1]: Table 1-15. Reference for CPUID Leaf 04H
https://cdrdv2-public.intel.com/775917/intel-64-architecture-processor-topology-enumeration.pdf

### DIFF
--- a/lib/propolis/src/cpuid.rs
+++ b/lib/propolis/src/cpuid.rs
@@ -337,12 +337,18 @@ impl Specializer {
                         // processor)
                         subleaf.eax |= (num_vproc - 1) << 26;
 
-                        // Present L1 and L2 caches as per-thread, L3 is across
-                        // the whole VM.
+                        // L1/L2 shared by SMT siblings, L3 shared by whole VM.
+                        //
+                        // Per Intel SDM, EAX[25:14] is "Maximum number of
+                        // addressable IDs for logical processors sharing this
+                        // cache". Add one to get the actual count; the nearest
+                        // power of 2 >= that value gives the APIC ID mask width.
                         if level < 3 {
                             subleaf.eax &= !LEAF4_EAX_VCPU_MASK;
-                            // And leave that range 0: this means only one
-                            // vCPU shares the cache.
+                            if self.has_smt {
+                                // 2 logical processors share L1/L2 so (1 + 1) = 2
+                                subleaf.eax |= 1 << 14;
+                            }
                         } else {
                             subleaf.eax &= !LEAF4_EAX_VCPU_MASK;
                             let shifted_vcpu = (num_vcpu - 1) << 14;
--- a/phd-tests/tests/src/cpuid.rs
+++ b/phd-tests/tests/src/cpuid.rs
@@ -3,7 +3,6 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use cpuid_utils::{CpuidIdent, CpuidSet, CpuidValues};
-use itertools::Itertools;
 use phd_framework::{test_vm::MigrationTimeout, TestVm};
 use phd_testcase::*;
 use propolis_client::instance_spec::{CpuidEntry, InstanceSpecStatus};
@@ -273,6 +272,26 @@ impl<'a> LinuxGuestTopo<'a> {
         }
         result.into_iter()
     }
+
+    /// Returns true if the guest has SMT enabled.
+    ///
+    /// This is determined by checking if cpu0's thread_siblings has more than
+    /// one bit set.
+    async fn has_smt(&self) -> bool {
+        let siblings = self
+            .vm
+            .run_shell_command(&format!(
+                "cat {}/thread_siblings",
+                Self::cpu_stem(0)
+            ))
+            .await
+            .expect("can get thread siblings");
+        let siblings = siblings.trim();
+
+        let value = u64::from_str_radix(siblings, 16)
+            .expect("thread_siblings should be valid hex");
+        value.count_ones() > 1
+    }
 }
 
 #[phd_testcase]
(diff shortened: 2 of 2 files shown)


iximeow:
oh nice, thanks! totally an oversight when I was in there. how visible is this under `/sys/devices/system/cpu/cpu*/cache/index*/shared_cpu_list` before the fix? I'm not sure how much Linux fixes this up, or if it faithfully communicates the wrong topology to userland. if this is legible to a booted guest, we probably should have been looking for this in [`guest_cpu_topo_test`](https://github.com/oxidecomputer/propolis/blob/f969fb7fd57c03bee03fe9a1d7458567a8e26b8d/phd-tests/tests/src/cpuid.rs#L279).

so, if this is legible in the guest, could you adjust that test to check `index*/shared_cpu_list` is reasonable too? that is, `index0` and `index1` are shared across two cores (if SMT is enabled! which, oh dear, the test doesn't assert..), that `index2` is shared across all cores, and that there aren't more cache levels?

(.. I also see that in retrospect that test assumes SMT siblings are adjacent in APIC ID, which is definitely wrong in general.)

glitzflitz:
This is what I get before the patch 
```
root@archiso ~ # cat /sys/devices/system/cpu/cpu*/cache/index*/shared_cpu_list
0
0
0
0-3
1
1
1
0-3
2
2
2
0-3
3
3
3
0-3
```
which indeed shows each vCPU as its own private L1 and L2 cache.

After the patch I get
```
root@archiso ~ # cat /sys/devices/system/cpu/cpu*/cache/index*/shared_cpu_list
0-1
0-1
0-1
0-3
0-1
0-1
0-1
0-3
2-3
2-3
2-3
0-3
2-3
2-3
2-3
0-3
```
Let me add a test for this

glitzflitz:
I fixed the `sibling_idx` calculation in [Fix thread_siblings assertion in guest_cpu_topo_test](https://github.com/oxidecomputer/propolis/pull/1002/commits/fcdac0e5638c00c03d1ed3cdfa1d8e4c4541c23a) and added the test in [Add cache topology verification to guest_cpu_topo_test](https://github.com/oxidecomputer/propolis/pull/1002/commits/291c8d4ecdb74c315a5415d2c9b598a9521e63f1). It passes for me locally now.

iximeow (commented):
hello! apologies again for taking forever to get back to this :)

I have a few notes here for references (for myself and others, things I wished I'd thought to cite earlier), so I'll push another commit that adjusts some of the docs a bit, squash it all, and merge shortly. thanks again for the fix and adjusting the test to suit :pray: 

separately I was taking a look at this locally with propolis-standalone and pretty confused by:
```
root@ubuntu:~# cat /sys/devices/system/cpu/cpu*/cache/index*/shared_cpu_list
0
0
0
0
1
1
1
1
2
2
2
2
3
3
3
3
```
which... comes down to `propolis-standalone` taking a different path to CPU profiles specifically if there is _no_ CPUID profile set. totally different issue! in that case we take the bhyve default literally and [bhyve indeed does not indicate that L1/L2 caches are shared](https://github.com/oxidecomputer/illumos-gate/blob/3baf9f2fefa23deda84c6c57e9f11ff5d25b83dc/usr/src/uts/intel/io/vmm/vmm_cpuid.c#L861-L875). on a quick skim that all pleads ignorance of SMT... except that CPUID_HTT is set in leaf 1 so at least I'm left wondering what all this looks like on a HT-capable processor with hyperthreading disabled. it's probably not terribly important over in bhyve the default CPUID behavior should probably represent `threads` as siblings in the topology here too; that'd be an issue over at illumos.org.

ps, to to myself from a few months ago:

> (.. I also see that in retrospect that test assumes SMT siblings are adjacent in APIC ID, which is definitely wrong in general.)

a more careful reading of `CACHE TOPOLOGY ENUMERATION USING CPUID LEAF 04H` from the topology enumeration doc you linked, and this is said similarly in the AMD APM as well, actually does require cache-sharing cores to be colocated in APIC ID. quote Intel:

> The CACHE_IDs for each cache level can be extracted from the x2APIC ID for processors that report 32-bit x2APIC ID, or from the initial APIC ID for processors that do not report x2APIC ID. ... The list of CACEH_Masks[n] of all types and levels when a bitwise AND is performed against its own x2APIC ID/APIC ID will provide the CACHE_ID[n] for that level and type that can be used to match other processors sharing the same cache ...

so cores sharing L1 or L2 cache must have APIC IDs that are the same outside bits corresponding to bits 25-14 in `cpuid 4.eax`. in a topology where three cores shared an L1 cache on a six-core processor, that yields APIC IDs 0, 1, 2, 4, 5, 6.

AMD describes APIC IDs having similar relatedness in Volume 3, appendix E "Obtaining Processor Information Via the CPUID Instruction", in the section about `Function 8000_001Dh`:

> `ShareId = LocalApicId >> log2(NumSharingCache+1)`

so same as Intel, you can slice up the APIC ID into bitmasks of who shares caches.

iximeow:
> is there anything else you need from me on this?

no - I'd wanted to see how the APIC IDs work out if the processor is hyperthreading-capable-but-disabled, but I'm clearly not going to get back to that promptly.. this is clearly more accurate in the typical case so I'll just hit the button. thanks again for the patch :pray:

PR #1009, opened 2026-01-08
Add DTrace script to monitor viona activity

citrus-it:


scripts/viona.d
```diff
@@ -0,0 +1,250 @@
...
+}
+
+viona_notify_mmio:entry/arg1/ {
+	printf("MMIO NOTIFY write %x+%x = %x\n", arg2, arg3, *args[4]);
+}
+
+viona_ioctl:return/self->cmd != 0 && arg1 != 0/ {
+	printf("ioctl(%s) returning 0x%x\n", self->cmd, arg1);
+	self->cmd = 0;
+}
```
iximeow: might be interesting to print errno if `arg1 == -1` yeah? presumably you have not had to care about that in working on multiqueue :)

PR #1012, opened 2026-01-09
avoid excessive Arc<Mapping> contention

iximeow:
Conceptually it wasn't wrong for submappings to `Arc::clone` the pointed-to base mapping, but in practice all the accessors off of MemCtx return SubMapping with lifetimes constrained by that MemCtx.

The sheer number of submappings being created and destroyed during I/O operations makes the refcount very hot under load. Lean into the de facto lifetime constraint on SubMapping and just take a borrow of the backing Mapping instead of Arc::clone (and implicit ref decrement in later drop).

---

this is actually one of the first bottlenecks that @jmpesp and I had noticed in November with noop-y backends but I'd incorrectly bundled this into an interrupt coalescing commit. then forgot about it between then and now, so I overlooked it Monday when getting in-flight stuff buttoned up. with the other changes we've gotten in, this a pretty substantial bottleneck: a sequential write `fio` without this saw *38%* of total CPU time in user code on the host. not in `memcpy`, all in `lock incq`/`lock decq`.

that extreme case come from us going through each PRP list entry, which cause a few `SubMapping::from_base` per. so the larger-but-fewer I/Os of large writes actually exercised this Arc worst of all benchmarks! with this change it's now 2% of total CPU time in user code and more than double the write bandwidth. seq-read, rand-read, rand-write are similar stories but less in magnitude.

### DIFF
--- a/lib/propolis/src/common.rs
+++ b/lib/propolis/src/common.rs
@@ -527,22 +527,18 @@ mod test {
         let mut buf = [0u8; 8];
         let mut ro8 = ReadOp::from_buf(0, &mut buf[0..1]);
         ro8.write_u8(1);
-        drop(ro8);
         assert_eq!(buf, [1, 0, 0, 0, 0, 0, 0, 0]);
 
         let mut ro16 = ReadOp::from_buf(0, &mut buf[0..2]);
         ro16.write_u16(0x2000);
-        drop(ro16);
         assert_eq!(buf, [0, 0x20, 0, 0, 0, 0, 0, 0]);
 
         let mut ro32 = ReadOp::from_buf(0, &mut buf[0..4]);
         ro32.write_u32(0x4000_0000);
-        drop(ro32);
         assert_eq!(buf, [0, 0, 0, 0x40, 0, 0, 0, 0]);
 
         let mut ro64 = ReadOp::from_buf(0, &mut buf);
         ro64.write_u64(0x8000_0000_0000_0000);
-        drop(ro64);
         assert_eq!(buf, [0, 0, 0, 0, 0, 0, 0, 0x80]);
     }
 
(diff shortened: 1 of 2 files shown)


lib/propolis/src/vmm/mem.rs
```diff
@@ -401,12 +401,9 @@ pub struct SubMapping<'a> {
 impl SubMapping<'_> {
     /// Create `SubMapping` using the entire region offered by an underlying
     /// `Mapping` object.
-    fn new_base<'a>(
-        _mem: &'a MemCtx,
-        base: &'_ Arc<Mapping>,
```
iximeow: `MemCtx` was here for the lifetime constraint, but `Mapping` *is* the backing structure that a `SubMapping`'s lifetime "should" be derived from. `MemCtx` is just a container for all the mappings of a guest. taking `&Mapping` directly means we're constraining callers of `SubMapping::new_base` a bit more than they theoretically could have been.

if we ever care about this, the caller can `Arc::clone` and call `new_base` freely, 'cause it's pretty close to free now.

lib/propolis/src/util/regmap.rs
```diff
@@ -130,7 +130,6 @@ impl<ID> RegMap<ID> {
             let mut sro = ReadOp::from_buf(0, &mut scratch);
 
             f(&reg.id, RWOp::Read(&mut sro));
-            drop(sro);
```
iximeow: `clippy` complains about this now, which was surprising. I think what happened is, `Backing` used to have an implicit `Drop` because of `Arc`, no longer does, and the implicit `Drop` had made this ineligible for the lint before. this used to be necessary because the lifetime on `ReadOp` would have `scratch` held for mutable access when we get to `write_bytes` below. the lint is, then, _not_ surprising in its relevance:

> call to `std::mem::drpo` witha value that doe not implement `Drop`. Dropping such a type only extends its contained lifetimes.

PR #1017, opened 2026-01-13
run header-check in CI

iximeow:
motivated by https://github.com/oxidecomputer/propolis/pull/1015, wherein I didn't pay close enough attention to the viona header changes because I thought header-check had run in CI and confirmed that they lined up. so, lets make that true.

this comes at the cost of one unfortunate-ish clone of illumos-gate per Propolis CI run, but that's not.. the worst thing, maybe? `git clone --depth 1 --branch stlouis https://code.oxide.computer/illumos-gate` yields about 500 MiB of stuff, though i'm not sure how much that was on the network.

@jclulow had suggested we could grab this from CTF, or headers from helios packages, and i think that's a good option particularly when we're comparing against `stlouis`. but i hope that if we find ourselves working on changes like #998 we'd nudge `GATE_REF` appropriately until OS changes are integrated alongside Propolis. in that case we'd need to be grabbing development headers out of git all the same, so here we are.

### DIFF
--- a/.github/buildomat/jobs/check-headers.sh
+++ b/.github/buildomat/jobs/check-headers.sh
@@ -18,7 +18,11 @@ set -e
 
 GATE_REF="$(./tools/check_headers gate_ref)"
 
-git clone --depth 1 --revision "$GATE_REF" \
-	https://github.com/oxidecomputer/illumos-gate ./gate_src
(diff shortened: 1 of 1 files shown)


jordanhendricks (approved):
this is awesome! i have definitely wished for it before!

Issue #1019, opened 2026-01-21
ASpace asserts when covering exactly one element


iximeow:
I've been looking at the cores from @askfongjojo in https://github.com/oxidecomputer/propolis/issues/827#issuecomment-3766106821 and the new ones are a different issue than originally reported in #827. figuring this out has mostly been an exercise in trusting what the computer is telling me, but straightforward enough..

so, the stack from Angela:
```
> $C ! demangle
fffff6ffe7bff2d0 libc.so.1`_lwp_kill+0xa()
fffff6ffe7bff300 libc.so.1`raise+0x22(6)
fffff6ffe7bff350 libc.so.1`abort+0x58()
fffff6ffe7bff360 ~std::sys::pal::unix::abort_internal::h515b31030a784ec5+8()
fffff6ffe7bff370 ~std::process::abort::h263ae6dd5f7f06e2+8()
fffff6ffe7bff380 ~__rustc::__rust_abort+8()
fffff6ffe7bff390 ~__rustc::__rust_start_panic+8()
fffff6ffe7bff3f0 __rustc::rust_panic+0xd()
fffff6ffe7bff4b0 std::panicking::rust_panic_with_hook::h24c1a837a734501c+0x224()
fffff6ffe7bff4f0 std::panicking::begin_panic_handler::{{closure}}::hebb980b0c684b90d+0x65()
fffff6ffe7bff500 ~std::sys::backtrace::__rust_end_short_backtrace::h4b651dba38b6cba9+8()
fffff6ffe7bff530 ~__rustc::rust_begin_unwind+0x1b()
fffff6ffe7bff560 ~core::panicking::panic_fmt::h717aa56ae600c892+0x1e()
fffff6ffe7bff5b0 ~core::panicking::panic::hf52b9b74c708168c+0x3a()
fffff6ffe7bff5c0 std::sync::poison::once::Once::call_once::_$u7b$$u7b$closure$u7d$$u7d$::h4aa1595ac930ccb3 (.llvm.7595959982290774484)+0x2c()
fffff6ffe7bff670 std::sys::sync::once::queue::Once::call::had06d6a18838cd97+0x1d6()
fffff6ffe7bff6a0 <propolis::hw::virtio::pci::ISR_STATUS_REGS as core::ops::deref::Deref>::deref::h5e01a69cbf2f7574+0x59()
[...]

PR #1020, opened 2026-01-21
ASpace should allow spaces with size 1

iximeow:
There was a test that explicitly rejected this, but because ASpace is inclusive it's actually fine for `start == end`. Adjust the test to use such a one-wide address space instead; it worked fine the whole time.

This (should?) fix https://github.com/oxidecomputer/propolis/issues/1019.

@dancrossnyc theoretically from my analysis in #1019 I believe any access to a VirtIO ISR should have immediately panic'd propolis, but I know we've both seen NICs work before and after landing #998. Do the two observations of "typically working Propolis" and "lazy-static initializer const-eval' to panic()" make sense to you? In particular, does it seem possible that a guest wouldn't touch this register except in some odd circumstance? The alternative to me seems like a guest OS/driver that we missed in testing which does touch this register when others don't?

### DIFF
--- a/lib/propolis/src/util/aspace.rs
+++ b/lib/propolis/src/util/aspace.rs
@@ -49,9 +49,9 @@ impl<T> ASpace<T> {
     ///
     /// # Panics
     ///
-    /// - Panics if start >= end.
+    /// - Panics if start > end.
     pub const fn new(start: usize, end: usize) -> ASpace<T> {
-        assert!(start < end);
+        assert!(start <= end);
         Self { start, end, map: BTreeMap::new() }
     }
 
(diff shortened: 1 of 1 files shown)


dancrossnyc:
Oophf. I thought I tested legacy fixed; clearly I missed this!

@iximeow yes, I think the analysis is sound. I wonder if this was because initially the upper bound was half-open, but it was changed to be inclusive later. I recall when I wrote the initial GPT code in preparation for the dirty bit tracker, I made page ranges inclusive, in order to handle pages at the very top of the address space, but Patrick had been doing things in terms of clopen ranges elsewhere, and we chatted about it a bit. I don't recall further details, though.

PR #1022, opened 2026-01-23
Panic with a clear message if the viona API is too old

citrus-it:
Currently if one runs propolis on a system which does not have a new
enough viona API, a stacktrace like this is generated.

    failed to enable promisc mode on vnic0: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
    called `Result::unwrap()` on an `Err` value: Os { code: 25, kind: Uncategorized, message: "Inappropriate ioctl for device" }
    stack backtrace:
       0: __rustc::rust_begin_unwind
       1: core::panicking::panic_fmt
       2: core::result::unwrap_failed
       3: propolis::hw::virtio::viona::PciVirtioViona::new_with_queue_sizes
       4: propolis::hw::virtio::viona::PciVirtioViona::new

Since we don't currently support anything older than the V6 API in
propolis we should at least produce a useful error message to the
operator. It would be possible to fall back to the older API if we
wish to implement that.

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -771,6 +771,12 @@ impl MachineInitializer<'_> {
             info!(self.log, "Creating vNIC {}", device_name);
             let bdf: pci::Bdf = nic.device_spec.pci_path.into();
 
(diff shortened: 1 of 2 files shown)


citrus-it:
Superseded by #1027

PR #1026, opened 2026-01-28
nvme: CQ db_buf EventIdx should slightly lag CQ tail

iximeow:
Doorbell buffers want the CQ event index to roughly track with the tail,
because most CQ doorbells are not interesting. If there are ten CQ
entries, space for 54 more, and the guest has acknowledged one, we don't
really care anything happened.

If there are 64 CQ entries, 0 available, and one SQ entry that we're
notified for, we'll try to reserve space in a full CQ, see it's at
capacity, and the SQ will idle with the CQ aware that that SQ is corked.
Without doorbell buffers, at some point the guest will process a CQE,
ring a CQ doorbell, uncork the CQ, and we'll nudge the parked SQ worker
thread into operation again.

With doorbell buffers, if the event index matches the CQ tail we'll
never get a doorbell when the CQ has been drained. If the guest never
submits another I/O on the corked SQ, the pending I/O may never get
processed.

A more careful treatment would only have the event index lag if we are
at risk of having a worker thread defer an I/O because this CQ would be
full. This would look something like not letting EventIdx advance so far
forward that there is less than one entry available for each SQ pointed
at this CQ.

Eating the unnecessary doorbell cost at least gets doorbell buffers
correct in this case and we can turn it back on. The more careful
treatment of EventIdx can come later.

### DIFF
--- a/lib/propolis/src/hw/nvme/admin.rs
+++ b/lib/propolis/src/hw/nvme/admin.rs
@@ -463,7 +463,6 @@ impl NvmeCtrl {
         }
     }
 
-    #[allow(dead_code)]
     pub(super) fn acmd_doorbell_buf_cfg(
         &mut self,
         cmd: &cmds::DoorbellBufCfgCmd,
--- a/lib/propolis/src/hw/nvme/mod.rs
+++ b/lib/propolis/src/hw/nvme/mod.rs
@@ -844,9 +844,8 @@ impl PciNvme {
             nn: 1,
             // bit 0 indicates volatile write cache is present
             vwc: 1,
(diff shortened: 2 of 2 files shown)


pfmooney:
> A more careful treatment would only have the event index lag if we are at risk of having a worker thread defer an I/O because this CQ would be full. This would look something like not letting EventIdx advance so far forward that there is less than one entry available for each SQ pointed at this CQ.

Might be worth recording that in the source commentary for future-you (or someone else).  It seems like implementing that wouldn't be _too_ onerous at some point down the line, and would get things back to nearly full elision.

PR #1027, opened 2026-01-29
Be more specific about version errors, raise min required viona API

iximeow:
The recent multiqueue work includes an unconditional ioctl that was introduced in V6 of the viona API. Propolis' expectation should have been raised then, but it was missed in review; instead, trying to create a vNIC on a too-old OS results in propolis panicking about `Inappropriate ioctl for device`.

So, raise the minimum viona API as required. propolis-server will exit at start if the OS is too old.

The message wasn't super clear though, complaining that:

> viona API version mismatch 4 != 6

so this comes with some adjustments to the version error reporting that now produce this error on a too-old host:

>    0: checking version of viona
>    1: API version 4 is not at or above 6. OS is too old?

propolis-standalone allows you to try running something on a too-old host on the expectation you're actively hacking on either the host or Propolis, so it just offers

> ERRO viona: API version 4 is not at or above 6. OS is too old?

before continuing on (and in this case, panicking about the an inappropriate ioctl).

--

this ended up not being really derived off of #1022 much at all so it's a distinct PR rather than a force push over @citrus-it's `vionaver` :grin:

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -771,32 +771,27 @@ impl MachineInitializer<'_> {
             info!(self.log, "Creating vNIC {}", device_name);
             let bdf: pci::Bdf = nic.device_spec.pci_path.into();
 
-            // Set viona device parameters if possible.
+            // Set viona device parameters. The parameters here (copy_data and
+            // header_pad) require `viona::ApiVersion::V3`, below Propolis'
+            // minimum of V6, so we can always set them.
             //
             // The values chosen here are tuned to maximize performance when
             // Propolis is used with OPTE in a full Oxide rack deployment,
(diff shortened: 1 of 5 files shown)


citrus-it:
We could also fairly trivially make propolis work with older versions, which was kind of in my mind with the other (temporary) PR. If that isn't something we want to do then this looks good to me and cleans stuff up too.

dancrossnyc (approved):

iximeow:
> could also fairly trivially make propolis work with older versions,

we talked about it a bit on the side and agree that Propolis supporting older viona (and imo bhyve) API versions isn't really important. unless someone really wants to run propolis on an older host OS, "bump the min version and move on" is probably the way to go here! if the need comes up in the future, we can figure out how to more gracefully handle older API versions.

PR #1031, opened 2026-02-05
Want virtio-vsock support

papertigers:
This PR implements the virtio-socket device from the [virtio spec](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html). The vsock device will be be used as the primary guest to host communication needed for VM attestation which is outlined in rfd 605.

This implementation functions by proxying guest vsock packets to host side TCP sockets. We currently do not
allow host to guest initiated connections.

With propolis-standalone one can enable the vsock device like so:
```toml
[dev.vsock0]
driver = "pci-virtio-vsock"
guest_cid = 16
pci-path = "0.6.0"

[[dev.vsock0.port_mappings]]
port = 80
addr = "127.0.0.1:3000"

[[dev.vsock0.port_mappings]]
port = 3001
addr = "127.0.0.1:3001"

[[dev.vsock0.port_mappings]]
port = 2222
addr = "10.0.1.22:22"
```

Fixes: #969


papertigers:
There will be a few follow up commits but I wanted to get this up for some review time while that stuff is in progress.

cc @dancrossnyc  if you have review cycles.

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -94,7 +94,7 @@ impl VqAvail {
         }
         if let Some(idx) = mem.read::<u16>(self.gpa_idx) {
             let ndesc = Wrapping(*idx) - self.cur_avail_idx;
-            if ndesc.0 != 0 && ndesc.0 < rsize {
+            if ndesc.0 != 0 && ndesc.0 <= rsize {
```
hawkw: is this a fix for a previous existing bug?

lib/propolis/src/vsock/buffer.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    fn free(&self) -> usize {
+        self.capacity() - self.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.head == self.tail
+    }
+
+    pub fn push(&mut self, data: Vec<u8>) -> Result<(), VsockBufError> {
```
hawkw: nit, take it or leave it: it seems like `Data` doesn't _need_ to be an owned `Vec` here, since we're just `copy_from_slice`ing its contents; i think this could be:
```suggestion
    pub fn push(&mut self, data: impl AsRef<[u8]>) -> Result<(), VsockBufError> {
        let data = data.as_ref();
```

that way, callers can call it with an owned `Vec` _or_ an `&Vec<u8>` or a slice into a vec or anything you want.

lib/propolis/src/vsock/buffer.rs
```diff
@@ -0,0 +1,210 @@
...
+            });
+        }
+
+        let head_offset = self.head.0 % self.buf.len();
+        let available_len = self.buf.len() - head_offset;
+
+        // If the data can fit in the remaining space of the ring buffer copy it
+        // in one go.
+        if data.len() <= available_len {
+            self.buf[head_offset..head_offset + data.len()]
```
hawkw: nit: a cute little trick @cbiffle showed me for slicing something to `offset..len` is that you can alternatively write:
```suggestion
            self.buf[head_offset..][...data.len()]
```

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+    }
+
+    /// This is the vsock event-loop. It's responsible for handling vsock
+    /// packets to and from the guest.
+    fn handle_events(&mut self) {
+        const MAX_EVENTS: u32 = 32;
+
+        let mut events: [MaybeUninit<libc::port_event>; MAX_EVENTS as usize] =
+            [const { MaybeUninit::uninit() }; MAX_EVENTS as usize];
+        let mut read_buf = vec![0u8; 1024 * 64];
```
hawkw: since this is zero-initialized and we never expect to grow it (i think) we might consider 
```suggestion
        let mut read_buf: Box<[u8]> = vec![0u8; 1024 * 64].into();
```
as it is often very slightly faster to index into. for proof, try running [this playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=1e6f7ab4298f240c52715ca616c2cc54) in release mode a few times:[^1]

```
     alloc::vec::Vec<u8>: 90ns
            [u8; 1000000]: 50ns
  alloc::boxed::Box<[u8]>: 60ns
```

[^1]: the timings are a bit variable but the vec is pretty consistently 30-50ns slower...

lib/propolis/src/vsock/packet.rs
```diff
@@ -0,0 +1,259 @@
...
+use crate::vsock::VSOCK_HOST_CID;
+
+pub const VIRTIO_VSOCK_OP_REQUEST: VsockPacketOp = 1;
+pub const VIRTIO_VSOCK_OP_RESPONSE: VsockPacketOp = 2;
+pub const VIRTIO_VSOCK_OP_RST: VsockPacketOp = 3;
+pub const VIRTIO_VSOCK_OP_SHUTDOWN: VsockPacketOp = 4;
+pub const VIRTIO_VSOCK_OP_RW: VsockPacketOp = 5;
+pub const VIRTIO_VSOCK_OP_CREDIT_UPDATE: VsockPacketOp = 6;
+pub const VIRTIO_VSOCK_OP_CREDIT_REQUEST: VsockPacketOp = 7;
+type VsockPacketOp = u16;
```
dancrossnyc: Why not make this a real enum?

```rust
#[derive(Clone, Copy, Debug)]
#[repr(u16)]
pub enum VsockPacketOp {
    Request = 1,
    Response = 2,
    Reset = 3,
    Shutdown = 4,
    ReadWrite = 5,
    UpdateCredit = 6,
    RequestCredit = 7,
}

impl TryFrom<u16> for VsockPacketOp {
    type Error = u16;
    fn try_from(raw: u16) -> Result<Self, Self::Error> {
        match raw {
            1 => Ok(Self::Request),
            2 => Ok(Self::Response),
            3 => Ok(Self::Reset),
            4 => Ok(Self::Shutdown),
            5 => Ok(Self::ReadWrite),
            6 => Ok(Self::UpdateCredit),
            7 => Ok(Self::RequestCredit),
            _ => Err(raw),
        }
    }
}

impl From<VsockPacketOp> for u16 {
    fn from(op: VsockPacketOp) -> u16 {
        op as u16
    }
}
```

lib/propolis/src/vsock/packet.rs
```diff
@@ -0,0 +1,259 @@
...
+    len: u32,
+    // Note this is "type" in the spec
+    socket_type: u16,
+    op: u16,
+    flags: u32,
+    buf_alloc: u32,
+    fwd_cnt: u32,
+}
+
+impl VsockPacketHeader {
```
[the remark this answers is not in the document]
dancrossnyc: I like this suggestion. The need for `packed` here at the moment is unfortunately, but Zerocopy handles the padding issue. My count is that this struct is 44 bytes long, but aligned to an 8-byte boundary, which means space at the end. Ugh.

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+            _ => Ok(()),
+        }
+    }
+
+    pub fn shutdown(&self) -> std::io::Result<()> {
+        self.port_send(VsockEvent::Shutdown)
+    }
+}
+
+bitflags! {
```
dancrossnyc: To whatever extent possible, I encourage avoiding direct use of `libc`. In this case, something form `nix` might be helpful: https://docs.rs/nix/latest/nix/poll/struct.PollFlags.html

@iximeow and I talked about this, and I of course defer to them; but I'll note that `nix` is already a dependency of propolis, though indirectly.

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+                    MAX_EVENTS,
+                    &mut nget,
+                    std::ptr::null_mut(),
+                )
+            };
+
+            if ret < 0 {
+                let err = std::io::Error::last_os_error();
+                // SAFETY: The docs state that `raw_os_error` will always return
+                // a `Some` variant when obtained via `las_os_error`.
```
jordanhendricks: "last_os_error"

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+        const IN = libc::POLLIN as i32;
+        const PRI = libc::POLLPRI as i32;
+        const OUT = libc::POLLOUT as i32;
+        const ERR = libc::POLLERR as i32;
+        const HUP = libc::POLLHUP as i32;
+    }
+}
+
+/// Set of `PollEvents` that signifies a readable event.
+fn fd_readable() -> PollEvents {
```
dancrossnyc: Instead of doing something like this, I'd consider implementing an `is_readable` method on `PollEvents`.

```rust
impl PollEvents {
    pub const fn is_readable(self, event: Self) -> bool {
        const READABLE: Self = PollEvents::IN | PollEvents::HUP | PollEvents::ERR | PollEvents::PRI;
        READABLE.intersects(event)
    }
}
```

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+impl VsockPoller {
+    /// Create a new `VsockPoller`.
+    ///
+    /// This poller is responsible for driving virtio-socket connections between
+    /// the guest VM and host sockets.
+    pub fn new(
+        cid: u32,
+        queues: VsockVq,
+        log: Logger,
+        port_mappings: IdHashMap<VsockPortMapping>,
```
dancrossnyc: It's probably not necessary, but I would perhaps consider using an `IdOrdMap` here to get in on some of that B-Tree cache locality goodness.

dancrossnyc:
I haven't gone through `poller.rs` in detail, but that seems like it's where the bulk of the logic lives?

lib/propolis/src/vsock/poller.rs
iximeow: (`READABLE.intersects(self)`?)

lib/propolis/src/hw/virtio/vsock.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        let packet = VsockPacket::parse(&mut chain, &mem);
+        vq.push_used(&mut chain, &mem);
+
+        Some(packet)
+    }
+}
+
+pub struct PciVirtioSock {
+    cid: u32,
```
iximeow: since the VirtIO spec describes this as a `u64` and as you've pointed out the upper half is reserved, i'm _super_ fine with making this `cid: u64` and asserting the upper half is zeroed in `PciVirtioSock::new` with a nod to the spec in that assert.

lib/propolis/src/vsock/packet.rs
```diff
@@ -0,0 +1,259 @@
...
+        //
+        // The upper 32 bits of src_cid and dst_cid are reserved and zeroed.
+        u64::from_le(self.src_cid) & u64::from(u32::MAX)
+    }
+
+    pub fn dst_cid(&self) -> u64 {
+        // The spec states:
+        //
+        // The upper 32 bits of src_cid and dst_cid are reserved and zeroed.
+        u64::from_le(self.dst_cid) & u64::from(u32::MAX)
```
iximeow: we should at the very least warn if these reserved bits are set. ideally if there's a way to tell the guest that this packet is invalid and respond with an error, we should do that instead of masking off the upper bits. as-is if a guest addresses a packet to `0x00000001_00000002` we'll treat it the same as a packet addressed to the host when it really is not.

I see Linux does not accept 64-bit CID in `sockaddr_vm`, but if someone did and their kernel checked for `CID == 2` as a way to disallow access to the host, we'd be undermining that check!

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+
+        loop {
+            let mut nget = 1;
+
+            let ret = unsafe {
+                libc::port_getn(
+                    self.port_fd.as_raw_fd(),
+                    events.as_mut_ptr() as *mut libc::port_event,
+                    MAX_EVENTS,
+                    &mut nget,
```
iximeow: my read of `port_get(3c)` is that we are saying here that we want to block for at least one event (`nget = 1` above), and that we are willing to receive up to 32 (MAX_EVENTS) events. and then `nget` is set to however many we actually got once there was at least one event to read. is that right? (I consider this "me learning `port_getn`" not like, needing a comment or anything)

then we `std::ptr::null_mut()` (! not const ! because this unfortunate [wrinkle in libc](https://github.com/rust-lang/libc/commit/d9208b3546c7da3ee9fc55e93b20e92db7279539)) to wait without a timeout. and that's fine because this is in a thread we spawn for the poller, where there's no runtime to be driving or other work to do? (having a remark that we don't wait with timeout and that's fine for such and such reasons seems helpful)

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+                        error!(&self.log, "vsock port_getn returned: {err}");
+                        continue;
+                    }
+                }
+            }
+
+            for i in 0..nget as usize {
+                let event = PortEvent::from_raw(unsafe {
+                    events[i].assume_init_read()
+                });
```
iximeow: i'd say this `unsafe` needs a `// Safety: port_getn has promised up to nget events are initialized`, but a different way to spell this might be...
```suggestion
            // Safety: port_getn has promised ... etc
            let events = unsafe {
                std::slice::from_parts(
                    events.as_ptr() as *const libc::port_event,
                    nget
                )
            };
```

... additionally, this is my own caution, but until we see a perf need I'd also consider zero-filling the whole thing instead of `MaybeUninit`. likewise, `assert!(nget <= events.len())` so that we have a check here that we have not suggested to the OS we might want more events than we should be reading, and that the OS has not suggested we should read more data than we can?

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+        log: Logger,
+        port_mappings: IdHashMap<VsockPortMapping>,
+    ) -> std::io::Result<Self> {
+        let port_fd = unsafe {
+            let fd = match libc::port_create() {
+                -1 => return Err(std::io::Error::last_os_error()),
+                fd => fd,
+            };
+
+            // Set CLOEXEC on the event port fd
```
iximeow: why? I don't think we'd ever legitimately `exec` from propolis..

iximeow (changes requested):
I was just gonna mention the `acc_mem` thing because I happened to spot it, and then I started reading more, sorry 😁

I very slightly skimmed the rest but mostly focused on the places we're sending data between guest and host, doing `unsafe {}`, etc.

"changes requested" because of the `acc_mem.access()` None handling, the packet validation-before-use, and guest-provided cid truncation bits. the rest are generally suggestions that I think would help but I'm not as pressed about one way or the other.

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+            if libc::fcntl(
+                fd,
+                libc::F_SETFD,
+                libc::fcntl(fd, libc::F_GETFD) | libc::FD_CLOEXEC,
```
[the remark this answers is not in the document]
papertigers: Interesting bit from the man page:
```
F_GETFD
              Value of flags defined in <fcntl.h>. The return value
              will not be negative.
```

Not sure if that means an invalid fd will just return an empty set?

bin/propolis-standalone/src/main.rs
```diff
@@ -1308,6 +1308,19 @@ fn setup_instance(
                         guard.inventory.register(&pvpanic);
                     }
                 }
+                "pci-virtio-vsock" => {
+                    // XXX MTZ: add the vsock device
```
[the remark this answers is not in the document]
papertigers: fixed in [98e35b5](https://github.com/oxidecomputer/propolis/pull/1031/commits/98e35b5be261fe1bc53d0111288ae117b6ffbfbb)

lib/propolis/src/hw/virtio/queue.rs
papertigers: Yup, turns out if you ask for n descriptors and the guest gives you them you should do the right thing!

lib/propolis/src/vsock/packet.rs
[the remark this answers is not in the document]
papertigers: done, as propolis already uses it!

lib/propolis/src/vsock/poller.rs
papertigers: Order doesn't really matter here and we currently will have 1 item in this mapping for propolis-server while propolis-standalone will allow you to define as many as you want. It's probably overkill for what it is already but makes the code nicer to deal with.

lib/propolis/src/hw/virtio/vsock.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+impl RxPermit<'_> {
+    /// Returns the maximum data payload that can fit in this descriptor chain.
+    pub fn available_data_space(&self) -> usize {
+        self.available_data_space
+    }
+
+    pub fn write(self, header: &VsockPacketHeader, data: &[u8]) {
+        let mem = self.vq.acc_mem.access().expect("mem access for write");
```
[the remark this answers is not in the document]
papertigers: yeah I took a look through and there's a mix of TODO, unwraps, and Errors in various places 😞.
I added the TODOs

lib/propolis/src/hw/virtio/vsock.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        let header_size = std::mem::size_of::<VsockPacketHeader>();
+        let available_data_space = self
+            .rx_chain
+            .as_ref()
+            .unwrap()
+            .remain_write_bytes()
+            .saturating_sub(header_size);
+
+        Some(RxPermit { available_data_space, vq: self })
```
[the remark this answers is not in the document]
papertigers: Ugh, I have made some of these changes but not all of them. There's an unfortunate design issue here where if we move the chain into `RxPermit` and then have a drop impl the compiler gets grump in the poller code because it can't determine that `permit.write()` has been used and it thinks self is mutably borrowed twice because of the Drop impl needing it.

Along with your other comment about handling `associate_fd` and bubbling up errors to the connection handing level, I think it would be best if I refactored a bunch of this. We can keep a lot of state in the `VsockProxyConn` and process all of the the state in one central function. I played around with this but I think it's more important to land this for R19 before we do additional refactoring here. It's my tech debt that I would like to fix once a few of the other pieces land.

lib/propolis/src/vsock/packet.rs
```diff
@@ -0,0 +1,259 @@
...
+        self
+    }
+
+    pub fn set_dst_cid(&mut self, cid: u32) -> &mut Self {
+        // The spec states:
+        //
+        // The upper 32 bits of src_cid and dst_cid are reserved and zeroed.
+        self.dst_cid = cid.to_le() as u64;
+        self
+    }
```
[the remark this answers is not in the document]
papertigers: Same as my other comment. I think dealing with u32 within propolis and using u64 on the wire seems fine and provides as a way to verify we are never using the reserved bits?

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1732 @@
...
+                        .set_dst_cid(*guest_cid)
+                        .set_src_port(key.host_port)
+                        .set_dst_port(key.guest_port)
+                        .set_len(nbytes as u32)
+                        .set_socket_type(VIRTIO_VSOCK_TYPE_STREAM)
+                        .set_op(VIRTIO_VSOCK_OP_RW)
+                        .set_buf_alloc(conn.buf_alloc())
+                        .set_fwd_cnt(conn.fwd_cnt());
+                    permit.write(&header, &read_buf[..nbytes]);
+                }
```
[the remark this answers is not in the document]
papertigers: Done, but this spilled over weirdly into the `VsockPacket` method for a new RW packet based on refactor Dan asked for. There's an assert there now and a comment that it's the callers responsibility to make sure the size of the data is not larger than `u32::MAX` as it should never be given the vsock buf_alloc field is a u32.

lib/propolis/src/vsock/packet.rs
```diff
@@ -0,0 +1,290 @@
...
+use crate::vsock::proxy::CONN_TX_BUF_SIZE;
+use crate::vsock::VSOCK_HOST_CID;
+
+bitflags! {
+    /// Shutdown flags for VIRTIO_VSOCK_OP_SHUTDOWN
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    #[repr(transparent)]
+    pub struct VsockPacketFlags: u32 {
+        const VIRTIO_VSOCK_SHUTDOWN_F_RECEIVE = 1;
+        const VIRTIO_VSOCK_SHUTDOWN_F_SEND = 2;
```
hawkw: personally, i would somewhat prefer this to be written as either
```suggestion
        const VIRTIO_VSOCK_SHUTDOWN_F_RECEIVE = 1 << 0;
        const VIRTIO_VSOCK_SHUTDOWN_F_SEND = 1 << 1;
```
or
```suggestion
        const VIRTIO_VSOCK_SHUTDOWN_F_RECEIVE = 0b01;
        const VIRTIO_VSOCK_SHUTDOWN_F_SEND = 0b10;
```
as "1" and "2" _happen_ to mean "bit 1" and "bit 2" respectively, but this wouldn't be the case for, say, 3, and if someone tried to add an additional bitflag for bit 3, they might misinterpret this and follow the existing pattern by adding `const VIRTIO_VSOCK_SOME_OTHER_FLAG = 3;`, thinking that this is defining the additional flag as being bit 3, but oops! it's actually the first two bits again!

lib/propolis/src/vsock/packet.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub fn new_response(guest_cid: u32, src_port: u32, dst_port: u32) -> Self {
+        let packet =
+            Self::new(guest_cid, src_port, dst_port, VsockPacketOp::Response);
+        packet
+    }
+
+    /// Create a new RW packet that sets the len field to the size of the data.
+    ///
+    /// Panics if the supplied data value is greater than u32::MAX as anything
+    /// larger would not fit within the peers buf_alloc which is defined as u32.
```
hawkw: nit, sorry:
```suggestion
    /// Panics if the supplied data value is greater than `u32::MAX`, as anything
    /// larger would not fit within the peer's `buf_alloc`, which is defined as `u32`.
```

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1736 @@
...
+            }
+        }
+    }
+
+    /// Handle the guest's VIRTIO_VSOCK_OP_RW packet.
+    fn handle_rw_packet(&mut self, key: ConnKey, packet: VsockPacket) {
+        if let Entry::Occupied(mut entry) = self.connections.entry(key) {
+            let conn = entry.get_mut();
+
+            // If we have a valid connection attempt to consume the guest's
```
hawkw:
```suggestion
            // If we have a valid connection, attempt to consume the guest's
```

"If we have a valid valid connection, attempt..." and "If we have a valid connection attempt," mean somewhat different things...

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1736 @@
...
+                        self.connections.remove(&key);
+                    }
+                    VsockPacketOp::Shutdown => {
+                        self.handle_shutdown(key, packet.header.flags());
+                    }
+                    VsockPacketOp::CreditUpdate => continue,
+                    VsockPacketOp::CreditRequest => {
+                        if self.connections.contains_key(&key) {
+                            self.rx.push_back(RxEvent::CreditUpdate(key));
+                        }
```
hawkw: shouldn't `self.connections.contains_key(&key)` always be `true`, since we are inside of the `if let Some(conn) = self.connections.get_mut(&key)` block? IIUC this second check is unnecessary.

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1736 @@
...
+                Ok(0) => break,
+                Ok(nbytes) => {
+                    conn.update_fwd_cnt(nbytes as u32);
+                    if conn.needs_credit_update() {
+                        self.rx.push_back(RxEvent::CreditUpdate(key));
+                    }
+                }
+                Err(e) if e.kind() == ErrorKind::WouldBlock => break,
+                Err(e) => {
+                    eprintln!("error writing to socket: {e}");
```
hawkw: should this `eprintln!` be a `slog` log instead?

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1736 @@
...
+                    // service the shutdown quiesce queue.
+                    std::ptr::null_mut(),
+                )
+            };
+
+            if ret < 0 {
+                let err = std::io::Error::last_os_error();
+                // SAFETY: The docs state that `raw_os_error` will always return
+                // a `Some` variant when obtained via `last_os_error`.
+                match err.raw_os_error().unwrap() {
```
hawkw: this is not unsafe so i would really prefer we not write `SAFETY` here.

lib/propolis/src/vsock/proxy.rs
```diff
@@ -0,0 +1,362 @@
...
+impl VsockBackend for VsockProxy {
+    fn queue_notify(&self, queue_id: u16) -> Result<(), VsockError> {
+        self.queue_notify(queue_id)
+            // Log the raw error in additon to returning the top level
+            // `VsockError`
+            .inspect_err(|_e| {
+                error!(&self.log,
+                    "failed to send virtqueue notification";
+                    "queue" => %queue_id,
+                )
```
hawkw: AFAICT, neither error actually gets logged hhere?

lib/propolis/src/vsock/packet.rs
```diff
@@ -2,258 +2,289 @@
...
 pub enum VsockPacketError {
-    #[error("Failed to read packet header from descriptor chain")]
+    #[error("failed to read packet header from descriptor chain")]
     ChainHeaderRead,
-    #[error("Packet only contained {remaining} bytes out of {expected} bytes")]
+
+    #[error("vsock packet header reported {hdr_len} bytes but the descriptor chain cont ains {chain_len}")]
+    InvalidPacketLen { hdr_len: usize, chain_len: usize },
+
+    #[error("descriptor chain only yielded {remaining} bytes out of {expected} bytes")]
```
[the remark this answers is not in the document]
papertigers: Done

lib/propolis/src/vsock/poller.rs
```diff
@@ -0,0 +1,1736 @@
...
+
+                    if let Entry::Occupied(mut entry) =
+                        self.connections.entry(key)
+                    {
+                        let conn = entry.get_mut();
+                        if let Err(e) = conn.set_established() {
+                            error!(
+                                &self.log,
+                                "cannot transition vsock connection state: {e}";
+                                "conn" => ?conn,
```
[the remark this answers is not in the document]
papertigers: Done in the error itself

lib/propolis/src/vsock/poller.rs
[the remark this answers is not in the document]
iximeow: spilled a little weird but on the whole this seems a lot nicer to read (imo!), thank you

lib/propolis/src/hw/virtio/vsock.rs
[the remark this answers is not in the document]
iximeow: > I think it would be best if I refactored a bunch of this. We can keep a lot of state in the `VsockProxyConn` and process all of the the state in one central function. I played around with this but I think it's more important to land this for R19 before we do additional refactoring here

that path makes sense to me too, yeah. mark this resolved if you want, or if you want to leave it as a reminder of the future refactor, have at it!

PR #1033, opened 2026-02-05
viona: reset device more completely

iximeow:
~~when I reboot a Debian 13 guest the NIC works again, but judging by viona.d output I think it's by.. accident? The reboot `SET_PAIRS 0xb` still returns 0x10 (EBUSY), and there's no `SET_USEPAIRS 0xb`. instead, there's a `SET_USEPAIRS 0x1`?~~ I think what I was seeing is that the guest negotiates multi-queue, configures the queues, then sets features again with the same bits (including MQ). We then try to SET_PAIRS to viona again but the queues are already out of reset because the guest just set them up. So our SET_PAIRS to the same value we'd set before fails, we don't SET_USEPAIRS to the same value again, and things continue "happily". This happens on first boot too, I just didn't see it above all the `RING_SET_MSI` and such.

@dancrossnyc suggested that it would be better to have a method on `VirtQueues` to go reset all the viona rings regardless of what `VirtQueues.len()` happens to be, which works out well.

I noticed along the way that we don't SET_PAIRS back down to 1 on reset. I think in the case that a guest would boot with a legacy driver, reconfigure the NIC for multiqueue, then reboot to a legacy driver (I'm thinking iPXE -> Linux -> reboot as a hypothesis), the driver would be pretty confused? So I'm SET_PAIRS and SET_USEPAIRS back to 1 on reset to get closer to the initial device state.

### DIFF
--- a/lib/propolis/src/hw/virtio/pci.rs
+++ b/lib/propolis/src/hw/virtio/pci.rs
@@ -1098,6 +1098,25 @@ impl PciVirtioState {
         }
     }
 
+    /// Reset all non-control queues as part of a device reset (or shutdown).
+    pub fn reset_queues(&self, dev: &dyn VirtioDevice) {
+        let mut state = self.state.lock().unwrap();
+        self.reset_queues_locked(dev, &mut state);
+    }
+
+    fn reset_queues_locked(
+        &self,
+        dev: &dyn VirtioDevice,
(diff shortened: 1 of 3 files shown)


dancrossnyc (approved):
I think this is right. Thanks!

Issue #1008, opened 2026-01-08
"rcu_preempt kthread starved for 10002 jiffies!"


iximeow:
@leftwo saw this on a slightly-but-not-very special build of Propolis on a Cosmo with slightly-but-not-very special host OS bits. on the Propolis side, it's commit 72a2f971ac812a79bb90fb31566dfe1b7be7b39c which is https://github.com/oxidecomputer/propolis/pull/1003. the host OS bits don't have any changes that would relate to VMs, but might relate to the VM's disks (`nvme1`..`nvme10`), excepting its boot disk. the VM was created like `./propolis-cli-1003 --server 0.0.0.0 new -c 192 -m 1048576 --config-toml ubuntu-instance-10-disks-wth.toml aln`. I've collected the `propolis-server`, configuration, kstats, and more recently some guest measurements (`/proc/stat` and `dmesg`), stashed them all at `/staff/core/propolis-1008`. nothing has crashed, so there's no core file :)

Here's the warning from the guest:
```
Ubuntu 24.04.3 LTS cosmomb2vm ttyS0

[  368.798927] rcu: INFO: rcu_preempt detected stalls on CPUs/tasks:
[  368.800199] rcu:     31-...0: (4 ticks this GP) idle=430c/1/0x4000000000000000 softirq=616/617 fqs=11246
[  368.801970] rcu:              hardirqs   softirqs   csw/system
[  368.803049] rcu:      number:        0          0            0
[  368.804127] rcu:     cputime:        0          0            0   ==> 30004(ms)
[  368.805472] rcu:     (detected by 28, t=60009 jiffies, g=6089, q=961 ncpus=192)
[  378.808588] nmi_backtrace_stall_check: CPU 31: NMIs are not reaching exc_nmi() handler, last activity: 4295042560 jiffies ago.
[  378.810855] rcu: rcu_preempt kthread starved for 10002 jiffies! g6089 f0x0 RCU_GP_DOING_FQS(6) ->state=0x0 ->cpu=17
[  378.812863] rcu:     Unless rcu_preempt kthread gets sufficient CPU time, OOM is now expected behavior.
[  378.814617] rcu: RCU grace-period kthread stack dump:
[  378.815647] rcu: Stack dump where RCU GP kthread last ran:
[  392.997356] Uhhuh. NMI received for unknown reason 20 on CPU 31.
[  392.997360] Dazed and confused, but trying to continue
[...]

PR #1038, opened 2026-02-06
awful (2): prevent alloc elimination and size better

iximeow:
I'd tested the commit introducing this hack with one additional detail I'd overlooked: an `eprintln!("balloon: {:p}", balloon.as_ptr())`. This prevented rustc/LLVM from optimizing out the allocation. Of course, it would be unsightly to have a random `eprintln!()` committed so I removed that in the PR, not realizing the allocation was optimized out entirely.

Along the way testing on a racklette had one VM where the heap grew to ~450 MiB, beyond the ~350 MiB the balloon would have gotten things. So grow the constant earmark a bit. RFD 413 estimates Propolis heap as 512 MiB at worst, versus the balloon's worst case of ~483 MiB (on top of the baseline 30 MiB or so of heap, so a bunch of VMs with 11 local disks are a bit more than budgeted. RFD 413 doesn't think too hard about "what if there are 256 propolises" so the whole area is a bit weird.)

I've surgically swapped this into London where `fill.fio` has been running for about 30 minutes without issue, and from here I'll let it keep going!

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -686,6 +686,11 @@ impl MachineInitializer<'_> {
                     &mut wanted_heap,
                 )
                 .await?;
+            info!(
+                self.log,
+                "raised balloon size";
+                "ballon_size" => wanted_heap
+            );
 
             self.block_backends.insert(backend_id.clone(), backend.clone());
             let block_dev: Arc<dyn block::Device> = match device_interface {
--- a/bin/propolis-server/src/lib/vm/ensure.rs
+++ b/bin/propolis-server/src/lib/vm/ensure.rs
@@ -664,9 +664,23 @@ async fn initialize_vm_objects(
     // This should be able to be removed when we are confident we can
     // actually brk() at runtime without starving ourselves out.
     //
-    // As one last step, include an extra 16 MiB of heap as space we might want
-    // when running `propolis-server`, handling HTTP requests, etc.
-    let balloon = vec![0u8; wanted_heap + 16 * propolis::common::MB];
+    // As one last step, include an extra 134 MiB in the balloon. In the happy
+    // case, even just 16 MiB was sufficient for Oximeter and regular check-ins
+    // with Propolis. In at least one case, though, we saw a Propolis with a
+    // heap about 126 MiB larger than the balloon had wanted. We don't know what
+    // happened to grow the heap like that, but size against the worst case for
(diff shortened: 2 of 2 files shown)


iximeow:
so, that fio ran fine up until...
```
[   24.635963] Loaded X.509 cert 'wens: 61c038651aabdcf94bd0ac7ff06c7248db18c600'
[   24.656652] NET: Registered PF_QIPCRTR protocol family
[  362.651626] hrtimer: interrupt took 3984548 ns
[17092.877781] INFO: task fio:4202 blocked for more than 122 seconds.
[17092.879381]       Tainted: P           O       6.8.0-90-generic #91-Ubuntu
[17092.881095] "echo 0 > /proc/sys/kernel/hung_task_timeout_secs" disables this message.
[17092.882996] task:fio             state:D stack:0     pid:4202  tgid:4202  ppid:3926   flags:0x00000002
[17092.883004] Call Trace:
[17092.883008]  <TASK>
[17092.883015]  __schedule+0x27c/0x6b0
[17092.883029]  schedule+0x33/0x110
[17092.883033]  schedule_preempt_disabled+0x15/0x30
[17092.883037]  __mutex_lock.constprop.0+0x42f/0x740
[17092.883044]  __mutex_lock_slowpath+0x13/0x20
[17092.883047]  mutex_lock+0x3c/0x50
[17092.883050]  bdev_release+0x40/0x240
[17092.883058]  blkdev_release+0x15/0x30
[17092.883062]  __fput+0xa3/0x2e0
[17092.883068]  __fput_sync+0x1c/0x30
[17092.883072]  __x64_sys_close+0x3e/0x90
[17092.883078]  x64_sys_call+0x1fec/0x25a0
```

that's about 4.7 hours of guest uptime, or about 3 hours ago. the propolis-server heap now looks like:
```
root@oxz_propolis:~# pmap 29126 | grep heap
0000000003D99000     429608K rw---    [ heap ]
000000001E123000      22632K rw---    [ heap ]
000000001F73D000     131888K rw---    [ heap ]
```

so after some Propolis uptime, the balloon is still too small. and in fact it's too small by about 100 MiB. we don't really know where that comes from, but we've done _better_ to hedge against #1008. this only got reached after a while of the pathological case of "uninterrupted I/O with novel guest addresses for hours at a time" - an improvement on the no-balloon outcome of "hit after two minutes of I/O with novel guest addresses". like with a few other commits recently, this will get unwound when we've fixed the underlying OS issue.

Issue #1052, opened 2026-02-09
virtio-nic: setting features should be idempotent?


iximeow:
I _think_ this might be the actual cause of https://github.com/oxidecomputer/propolis/issues/1048, but it might not be. either way, I'm pretty confident this is actually a bug in propolis:

https://github.com/oxidecomputer/propolis/blob/a28bee1d1b21a96344fe74b30c90361fa473471e/lib/propolis/src/hw/virtio/viona.rs#L716-L723

if a guest had already SET_FEATURES, and been using those queues, and then SET_FEATURES (maybe even the exact same bits) again, the second SET_FEATURES attempts to `set_pairs(PROPOLIS_MAX_MQ_PAIRS)` even though those queues had been enabled before and maybe even are in use.

in 1048 that is exactly what happens:
```
SET_FEATURES 0x1104388ab    # <--- somewhere early in boot
SET_PAIRS 0xb
SET_USEPAIRS 0xb
RING_INIT_MODERN 0x1 233496000/23349e000/23349f040 800
RING_SET_MSI 0x1 addr=0 msg=0
RING_INIT_MODERN 0x0 237ff1000/237ff2000/237ff2240 100
RING_SET_MSI 0x0 addr=0 msg=0
RING_INIT_MODERN 0x3 235002000/23500a000/23500b040 800
RING_SET_MSI 0x3 addr=0 msg=0
RING_INIT_MODERN 0x2 238158000/238159000/238159240 100
RING_SET_MSI 0x2 addr=0 msg=0
RING_INIT_MODERN 0x5 230505000/23050d000/23050e040 800
RING_SET_MSI 0x5 addr=0 msg=0
RING_INIT_MODERN 0x4 231468000/231469000/231469240 100
RING_SET_MSI 0x4 addr=0 msg=0
RING_INIT_MODERN 0x7 232057000/23205f000/232060040 800
RING_SET_MSI 0x7 addr=0 msg=0
RING_INIT_MODERN 0x6 232f99000/232f9a000/232f9a240 100
RING_SET_MSI 0x6 addr=0 msg=0
RING_SET_MSI 0x1 addr=fee0100c msg=49a1
RING_SET_MSI 0x0 addr=fee0100c msg=4991
RING_SET_MSI 0x3 addr=fee0200c msg=49b1
RING_SET_MSI 0x2 addr=fee0200c msg=49a1
RING_SET_MSI 0x5 addr=fee0400c msg=49b1
[...]

Issue #1048, opened 2026-02-07
Windows: spotty virtio nic functionality


iximeow:
Some boots of a Windows Server 2022 guest seem to be unable to use the virtio network device, while others can only use it for a few seconds before _something_ happens and the guest is unable to use the network.

here's the interesting viona bits as a Windows VM gets weird:
```
MMIO NOTIFY write c000a000+2 = 7       # <-- I'm pinging this VM from elsewhere on my network
MMIO NOTIFY write c000a000+2 = 7
MMIO NOTIFY write c000a000+2 = 1
MMIO NOTIFY write c000a000+2 = 1
MMIO NOTIFY write c000a000+2 = 1
MMIO NOTIFY write c000a000+2 = 1
MMIO NOTIFY write c000a000+2 = 1
MMIO NOTIFY write c000a000+2 = 1
MMIO NOTIFY write c000a000+2 = 3
MMIO NOTIFY write c000a000+2 = 3
MMIO NOTIFY write c000a000+2 = 3       # <-- and this is right as `ping` stops seeing the VM
RING_RESET 0x0                         # <-- it's reset the device, but why?
RING_RESET 0x1
RING_RESET 0x2
RING_RESET 0x3
RING_RESET 0x4
RING_RESET 0x5
RING_RESET 0x6
RING_RESET 0x7
RING_RESET 0x8
RING_RESET 0x9
RING_RESET 0xa
RING_RESET 0xb
RING_RESET 0xc
RING_RESET 0xd
RING_RESET 0xe
RING_RESET 0xf
RING_RESET 0x10
RING_RESET 0x11
RING_RESET 0x12
RING_RESET 0x13
RING_RESET 0x14
RING_RESET 0x15
RING_SET_MSI 0x1 addr=0 msg=0
RING_SET_MSI 0x0 addr=0 msg=0
RING_SET_MSI 0x3 addr=0 msg=0
RING_SET_MSI 0x2 addr=0 msg=0
RING_SET_MSI 0x5 addr=0 msg=0
RING_SET_MSI 0x4 addr=0 msg=0
RING_SET_MSI 0x7 addr=0 msg=0
RING_SET_MSI 0x6 addr=0 msg=0
SET_FEATURES 0x1104388ab
SET_PAIRS 0xb
SET_USEPAIRS 0xb
RING_INIT_MODERN 0x1 233496000/23349e000/23349f040 800
RING_SET_MSI 0x1 addr=0 msg=0
[...]

PR #1053, opened 2026-02-09
set-features should be idempotent?

iximeow:
I have *barely* tested this, but I built a propolis-standalone and threw it + a Windows Server 2022 onto a Cosmo and I'm no longer seeing the results of Windows repeatedly resetting the device. so this gets #1052 and apparently #1048 with it.

this probably should just be a mutex that we take somewhere in managing virtio device configuration but this was simple enough to throw together for a quick PoC/validation..

### DIFF
--- a/lib/propolis/src/hw/virtio/viona.rs
+++ b/lib/propolis/src/hw/virtio/viona.rs
@@ -8,6 +8,7 @@ use std::io::{self, Error, ErrorKind};
 use std::num::NonZeroU16;
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::sync::{Arc, Condvar, Mutex, Weak};
+use std::sync::atomic::{Ordering, AtomicBool};
 
 use crate::common::{RWOp, ReadOp};
 use crate::hw::pci;
@@ -308,6 +309,7 @@ pub struct PciVirtioViona {
     mac_addr: [u8; ETHERADDRL],
     mtu: Option<u16>,
     hdl: VionaHdl,
+    mq_active: AtomicBool,
     inner: Mutex<Inner>,
 }
 
@@ -396,6 +398,7 @@ impl PciVirtioViona {
             dev_features,
             mac_addr: [0; ETHERADDRL],
             mtu: info.mtu,
+            mq_active: AtomicBool::new(false),
             hdl,
             inner: Mutex::new(Inner::new(nqueues)),
         };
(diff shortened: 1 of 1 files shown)


iximeow:
well, I stand by my reading that we're not handling a reset of a virtio device via `device_status` write correctly, but the Windows circumstance is a lot more straightforward. we're handling the `3.1.1`-described device initialization poorly.

note that we `dev.set_features()` any time [a device status is set with FEATURES_OK](https://github.com/oxidecomputer/propolis/blob/master/lib/propolis/src/hw/virtio/pci.rs#L1053-L1067).

then, 3.1.1 says:

> The driver MUST follow this sequence to initialize a device:
> 1. Reset the device.
> 2. Set the ACKNOWLEDGE status bit: the guest OS has noticed the device.
> 3. Set the DRIVER status bit: the guest OS knows how to drive the device.
> 4. Read device feature bits, and write the subset of feature bits understood by the OS and driver to the device. During this step the driver MAY read (but MUST NOT write) the device-specific configuration fields to check that it can support the device before accepting it.
> 5. Set the FEATURES_OK status bit. The driver MUST NOT accept new feature bits after this step.
> 6. Re-read device status to ensure the FEATURES_OK bit is still set: otherwise, the device does not support our subset of features and the device is unusable.
> 7. Perform device-specific setup, including discovery of virtqueues for the device, optional per-bus setup,
reading and possibly writing the device’s virtio configuration space, and population of virtqueues.
> 8. Set the DRIVER_OK status bit. At this point the device is “live”.

so when a guest driver reaches step 5, we'll `set_features` and commit the negotiated features to the device. then when a guest reaches step 8, it sets a status including `FEATURES_OK` and set the same negotiated features again. then we see that MQ is an enabled feature and the whole `use_pairs() on rings outside of reset` thing happens. a "MUST NOT" tangent, if a guest wrote additional features between `FEATURES_OK` and `DRIVER_OK`, we'd accept them - we should probably warn and go to `NEEDS_RESET` for that case..

a lightly adjusted `viona.d` with probes for device status and reset events made this a lot more clear:
```
... nothing interesting ...
--- DEVICE STATUS SET: 00 ---
--- DEVICE STATUS SET: 01 ---
--- DEVICE STATUS SET: 03 ---
--- DEVICE STATUS SET: 0b ---
SET_FEATURES 0x1104388ab
SET_PAIRS 0xb
SET_USEPAIRS 0xb
RING_INIT_MODERN 0x0 234e15000/234e1d000/234e1e040 800
RING_SET_MSI 0x0 addr=0 msg=0
RING_INIT_MODERN 0x1 233258000/233259000/233259240 100
RING_SET_MSI 0x1 addr=0 msg=0
RING_INIT_MODERN 0x2 231289000/231291000/231292040 800
RING_INIT_MODERN 0x3 22d0f0000/22d0f1000/22d0f1240 100
RING_INIT_MODERN 0x4 22f8ce000/22f8d6000/22f8d7040 800
RING_INIT_MODERN 0x5 2294fd000/2294fe000/2294fe240 100
RING_INIT_MODERN 0x6 22bd5c000/22bd64000/22bd65040 800
RING_INIT_MODERN 0x7 21f5fd000/21f5fe000/21f5fe240 100
RING_SET_MSI 0x1 addr=fee0100c msg=49a1
RING_SET_MSI 0x0 addr=fee0100c msg=4991
RING_SET_MSI 0x3 addr=fee0200c msg=49b1
RING_SET_MSI 0x2 addr=fee0200c msg=49a1
RING_SET_MSI 0x5 addr=fee0400c msg=49b1
RING_SET_MSI 0x4 addr=fee0400c msg=49a1
RING_SET_MSI 0x7 addr=fee0800c msg=49b1
RING_SET_MSI 0x6 addr=fee0800c msg=49a1
--- DEVICE STATUS SET: 0f ---
SET_FEATURES 0x1104388ab
SET_PAIRS 0xb
```

but this also explains why I was seeing the "redundant" `SET_FEATURES` in a Linux guest and failing to find where in the Linux virtio driver that would actually happen..

dancrossnyc:
Oh yeah, I think you're right; it's write OK, read OK, write OK to acknowledge.

PR #1060, opened 2026-02-13
nvme: add oxide-specific feature to expose read-only attribute for device

### LINKED ISSUE #1059: instances with read-only boot disk fail to boot
Read-only disks were integrated as a control plane feature in oxidecomputer/omicron#9731. Today in dogfood I attempted to boot a variant of a debian [image](https://oxide.sys.rack2.eng.oxide.computer/images/vm-attest-proto-vm-rootfs) we've been using to test RFD 605 work. The instance (as well as future attempts to boot an instance with a RO disk) failed to boot, dropping into the UEFI shell, e.g.:

```
UEFI Interactive Shell v2.2
EDK II
UEFI v2.70 (EDK II, 0x00010000)
Mapping table
      FS0: Alias(s):HD0p:;BLK3:
          PciRoot(0x0)/Pci(0x10,0x0)/NVMe(0x1,00-00-00-00-00-00-00-00)/HD(15,GPT,6012ACAE-A51D-C14F-BB91-DBBFB6CF0F77,0x2000,0x3E000)
      FS1: Alias(s):F1:;BLK4:
          PciRoot(0x0)/Pci(0x18,0x0)
     BLK0: Alias(s):
          PciRoot(0x0)/Pci(0x10,0x0)/NVMe(0x1,00-00-00-00-00-00-00-00)
     BLK1: Alias(s):
          PciRoot(0x0)/Pci(0x10,0x0)/NVMe(0x1,00-00-00-00-00-00-00-00)/HD(1,GPT,E6842644-BB59-324A-BFF6-F7D01B8B2CB2,0x40000,0x3BFFDF)
     BLK2: Alias(s):
          PciRoot(0x0)/Pci(0x10,0x0)/NVMe(0x1,00-00-00-00-00-00-00-00)/HD(14,GPT,26E893FC-03C9-6049-B821-276C1F0B03A4,0x800,0x1800)
```

@hawkw and @jmpesp and I spent some time looking at this today. A collection of observations from today:

* booting from a read-only disk worked about a week ago (and eliza tested with this exact image; known good propolis sha is `2aa7f9d0ee84a1c45e821d6444b1d2f0e69b743e`)
* it is possible to boot from the same image as a normal (read-write) disk
* james was able to reproduce this failure trivially in the canada region
* the md5sum of the image, as uploaded to the control plane from james' desktop, matches the md5sum of the disk when it is attached as read-write disk to a working instance (implying crucible has not seen corruption of the disk)
* UEFI does not seem to believe there is a file system on the read-only disk (e.g., `ls FS0` shows nothing and no files), but there is valid data in the blocks on disk (e.g., seen via `dblk blk0 1` in the UEFI shell), and that data seems to match the image file locally
* the presence of a NIC on the instance did not change the behavior (i.e., I tried booting the image as a RO disk both with and without a NIC on the instance, and they failed the same way)
* this image is bootable as a read-only disk on qemu

luqmana:
Along with https://github.com/oxidecomputer/edk2/pull/4 this fixes #1059.

### DIFF
--- a/packaging/package-manifest.toml
+++ b/packaging/package-manifest.toml
@@ -14,6 +14,6 @@ output.type = "zone"
 [[package.propolis-server.source.buildomat_blobs]]
 repo = "edk2"
 series = "image_debug"
-commit = "8c60ff4c271b0442b9a506a5cf9eab04cd5554aa"
+commit = "bf64f45b1a58e69d126a3c6ca1e4512c88668132"
 artifact = "OVMF_CODE.fd"
-sha256 = "9ce2266aaefb671a1977ddabb5155f8235be134f797e6c831abeb6c97c705569"
+sha256 = "740187046a7267d0de72d3455070333547dbc0ea023531471fb2b2a61effa448"
--- a/bin/propolis-standalone/README.md
+++ b/bin/propolis-standalone/README.md
@@ -61,7 +61,7 @@ are some options to get up-and-running quickly:
 ### Guest bootrom
 
 The current recommended and tested guest bootrom is available
-[here](https://buildomat.eng.oxide.computer/public/file/oxidecomputer/edk2/image_debug/907a5fd1763ce5ddd74001261e5b52cd200a25f9/OVMF_CODE.fd).
+[here](https://buildomat.eng.oxide.computer/public/file/oxidecomputer/edk2/image_debug/bf64f45b1a58e69d126a3c6ca1e4512c88668132/OVMF_CODE.fd).
 
 Other UEFI firmware images built from the [Open Virtual Machine Firmware
 project](https://github.com/tianocore/tianocore.github.io/wiki/OVMF) may also
--- a/phd-tests/artifacts.toml
+++ b/phd-tests/artifacts.toml
@@ -13,5 +13,5 @@ kind = "bootrom"
 [artifacts.ovmf.source.buildomat]
 repo = "oxidecomputer/edk2"
 series = "image_debug"
-commit = "8c60ff4c271b0442b9a506a5cf9eab04cd5554aa"
-sha256 = "9ce2266aaefb671a1977ddabb5155f8235be134f797e6c831abeb6c97c705569"
+commit = "bf64f45b1a58e69d126a3c6ca1e4512c88668132"
+sha256 = "740187046a7267d0de72d3455070333547dbc0ea023531471fb2b2a61effa448"


lib/propolis/src/hw/nvme/admin.rs
```diff
@@ -403,6 +403,13 @@ impl NvmeCtrl {
                 )
             }
 
+            cmds::FeatureIdent::OxideDeviceFeatures => {
```
rmustacc: I don't see any logic really checking here on the feature selector. I was wondering if we had to check the data pointer, but it's specifically ignored when it's not used, so there's nothing to do there. It's not clear to me what other controllers do if you put a value in cdw14 when you don't support selecting by uuid, but it's probably fine to ignore.

rmustacc: One other thing now that I looked more closely here, it looks like we're going to commit to returning the same data regardless of namespace id. I think I'm fine with that, but just want to confirm that intent.

packaging/package-manifest.toml
```diff
@@ -14,6 +14,6 @@ output.type = "zone"
 [[package.propolis-server.source.buildomat_blobs]]
 repo = "edk2"
 series = "image_debug"
-commit = "907a5fd1763ce5ddd74001261e5b52cd200a25f9"
+commit = "8c60ff4c271b0442b9a506a5cf9eab04cd5554aa"
```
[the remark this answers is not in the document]
luqmana: I will squash/merge that PR first and update it here before submitting this PR. And yes, that SHA there was from buildomat. The image job on that repo publishes both the image and SHA. See [here](https://github.com/oxidecomputer/edk2/runs/63579446753) for example.

lib/propolis/src/hw/nvme/admin.rs
rmustacc: > Yes, although now I'm wondering if it's worth constraining one way or the other in the same vein as rejecting non-zero cdw11. Not a huge difference either way since we only support a single namespace today anyways.

Well, it's the distinction between using namespace id 0 being that it isn't valid or in the future using the broadcast namespace. I think I agree that we can change this in the future. I ultimately convinced myself it was probably fine last time I was looking at it. Ultimately we're the only consumer right now so it's okay to change htis if need be.

packaging/package-manifest.toml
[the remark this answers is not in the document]
rmustacc: Thanks for the confirmation.

lib/propolis/src/hw/nvme/admin.rs
```diff
@@ -404,10 +404,16 @@ impl NvmeCtrl {
...
-                cmds::Completion::success_val(
-                    cmds::OxideDeviceFeatures { read_only: self.read_only }
-                        .into(),
-                )
+                if cmd.cdw11 != 0 {
+                    // We don't currently accept any parameters for this feature
+                    cmds::Completion::generic_err(STS_INVAL_FIELD)
+                } else {
+                    cmds::Completion::success_val(
+                        cmds::OxideDeviceFeatures(0)
```
rmustacc: I'll defer to someone else on the change in construction from using `.into()` to the current construction. For my own edification, why did that change?

packaging/package-manifest.toml
```diff
@@ -14,6 +14,6 @@ output.type = "zone"
...
 artifact = "OVMF_CODE.fd"
-sha256 = "ff12d5cb021e34447b44301f70434e861b07d2779c16abe2f2efef49ff02fffb"
+sha256 = "9ce2266aaefb671a1977ddabb5155f8235be134f797e6c831abeb6c97c705569"
```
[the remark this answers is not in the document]
luqmana: I was waiting for the buildomat job to finish and forgot to push up the latest update referencing the merged PR. Should all match now.

PR #1064, opened 2026-02-27
virtio-nic: in-repo tests and a better device state FSM

iximeow:
VirtIO 1.2 states clearly that a driver MUST NOT clear bits from device_status (except, implicitly, by writing RESET aka 0 to device_status to.. do a device reset). Since this is the only way to clear a bit, bits that *are* set can be set at most once. Our old FSM accepted whatever status a guest happened to provide and stored that; a guest could intentionally or unintentionally clear NEEDS_RESET even, though it would be very much in error at that point.

Instead of this very lax treatment of device_status, be more clear that FEATURES_OK can be toggled to set only once (which fixes a bug on its own), and if a guest tries clearing a status bit by any means other than reset, set NEEDS_RESET and demand a reset outright.

Along with this, `viona.rs` gets a very simple driver to exercise some usage patterns we've seen in starting up virtio NIC devices.

`basic_operation_multiqueue` fails without the device state FSM improvements, a reflection of the bug mentioned above; since FEATURES_OK is set before DEVICE_OK, the final status write including DEVICE_OK would cause a second attempt to apply the negotiated features. We would then try to set up all the virtqueues again, which fails because the virtqueues were just taken out of the RESET state when the driver initialized them, and we come to rest with the NIC in NEEDS_RESET.

Linux (and BSDs, and illumos) clearly do not care about this, as such instances proceed to network seemingly without issue. Windows seems to have a periodic task that notices the NEEDS_RESET bit and obliges. It resets the NIC, which fails to ever come up correctly, and dutifully stops networking (#1048).

Commenting out parts of prior patches (957f5c4, 1df49a4, 45af0f7) also cause these new tests to fail; they are collectively as necessary as it seemed for correct operation in the face of resets.

Finally, one of the new tests verifies that a device in NEEDS_RESET cannot have the bit inappropriately cleared without reset. This tripped over a deadlock if a virtqueue is set with an inappropriate size. That is fixed in this patch as well.

### DIFF
--- a/lib/propolis/src/hw/virtio/viona.rs
+++ b/lib/propolis/src/hw/virtio/viona.rs
@@ -1430,10 +1430,7 @@ pub(crate) fn check_api_version() -> Result<(), crate::api_version::Error> {
 /// cirumstances carefully, and consider if these test expectations were correct
 /// in the first place; in some sense these tests function as a bespoke
 /// "virtio-nic driver" that lives only in Propolis' tests.
-///
-/// Finally, we'll actually create and destroy some vnics so not only do we need
-/// `dladm`, we need a recent enough viona and everything..
-#[cfg(all(test, target_os = "illumos"))]
+#[cfg(test)]
 mod test {
     use crate::common::{GuestAddr, RWOp, ReadOp, WriteOp, MB, PAGE_SIZE};
     use crate::hw::chipset::i440fx::{self, I440FxHostBridge};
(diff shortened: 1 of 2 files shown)


lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1420,3 +1420,656 @@ pub(crate) fn check_api_version() -> Result<(), crate::api_version::Error> {
...
+/// otherwise, the test expectations here are not grounded in any kind of
+/// observed behavior, just "the VirtIO spec says ... so ..."
+///
+/// If guests require changes that cause these tests to fail, please note the
+/// cirumstances carefully, and consider if these test expectations were correct
+/// in the first place; in some sense these tests function as a bespoke
+/// "virtio-nic driver" that lives only in Propolis' tests.
+///
+/// Finally, we'll actually create and destroy some vnics so not only do we need
+/// `dladm`, we need a recent enough viona and everything..
```
iximeow: well, not yet, but I want to. I was hesitant to automatically deleting vnics (such as in cleanup) because if your host OS is subject to https://www.illumos.org/issues/17853 then deleting the last vnic will panic your machine. to be clear, this is very fixed in illumos and helios, my OS is just on old bits and I haven't updated yet.

Issue #1069, opened 2026-03-09
Wire up vsock device to propolis-server


papertigers:
Currently the virtio-socket device is wired into `propolis-standalone` but it is not yet hooked up to `propolis-server`.
We need to extend the `InstanceSpec` to allow a vsock `Component`.

PR #1075, opened 2026-03-09
Wire up vsock device to propolis-server

papertigers:
This is the plumbing to allow virtio-socket devices to be usable in `propolis-server`.

Fixes: #1069

### DIFF
--- a/bin/propolis-server/src/lib/spec/builder.rs
+++ b/bin/propolis-server/src/lib/spec/builder.rs
@@ -285,8 +285,6 @@ impl SpecBuilder {
             return Err(SpecBuilderError::VsockInUse);
         }
 
-        // TODO validate guest_cid does not have reserved bits set
-
         self.register_pci_device(vsock.spec.pci_path)?;
         self.component_names.insert(vsock.id.clone());
         self.spec.vsock = Some(vsock);
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -501,9 +501,14 @@ impl MachineInitializer<'_> {
 
             let guest_cid = GuestCid::try_from(vsock.spec.guest_cid)
                 .context("guest cid")?;
+            // While the spec does not recommend how large the virtio descriptor
+            // table should be we sized this appropriately in testing so
+            // that the guest is able to move vsock packets at a reasonable
+            // throughput without the need to be much larger.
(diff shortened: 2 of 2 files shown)


bin/propolis-server/src/lib/spec/builder.rs
```diff
@@ -269,6 +273,27 @@ impl SpecBuilder {
...
+    ) -> Result<&Self, SpecBuilderError> {
+        if self.component_names.contains(&vsock.id) {
+            return Err(SpecBuilderError::ComponentNameInUse(vsock.id));
+        }
+
+        if self.spec.vsock.is_some() {
+            return Err(SpecBuilderError::VsockInUse);
+        }
+
+        // TODO validate guest_cid does not have reserved bits set
```
papertigers: Decide if we should return a `SpecBuilderError` or a `MachineInitError`

iximeow:
> by adjusting `PciVirtioSocket::new` to take a u64

the other option is to take a `u32` in the API ([this struct VirtioSocket](https://github.com/oxidecomputer/propolis/pull/1075/changes#diff-f0d0ab8a2ed23d7b29692ffcef7648e00d97e6a696fbced18bb2cb5c1b6a50efR15-R21)) and avoid the risk of taking a cid with bogus upper bits. we'll still want to error for `cid={0,1,2}` but the API taking the same type as `PciVirtioSock::new` avoids the conversions (what really stood out to me is that `guest_cid as u32` means we'd accept reserved bits in the API and build a device that uses a different truncated CID)

iximeow:
(fwiw looks like your test failures are VMs failing to boot which should be sorted as of https://github.com/oxidecomputer/propolis/pull/1080, if you want to rebase over that)

bin/propolis-standalone/src/main.rs
```diff
@@ -1319,12 +1320,14 @@ fn setup_instance(
                         guard.inventory.register(&pvpanic);
                     }
                 }
-                "pci-virtio-vsock" => {
+                "pci-virtio-socket" => {
```
iximeow: oh, i suppose VirtIO does call it a `Socket Device`, not ` Vsock Device` huh. we've some prior art in deviating from the specific spec name when we have good reason (`pci-virtio-viona` as a nod to specifically being a network device backed by viona rather than fully implemented in Propolis) but i fully agree with the name change here.

Issue #1035, opened 2026-02-05
a phd VM failed to boot?


iximeow:
in this test: 

`lspci_lifecycle_test_lifecycle_0.serial.log` is the serial output from the destination VM of the lifecycle test. the source VM is in `lspci_lifecycle_test.serial.log`. the source VM looks basically reasonable. the progress bar renders a bit funky because ANSI codes evaporated, and it seems like the test is wrong but in a way that _should_ pass:
```
localhost login: root
Welcome to Alpine!

The Alpine Wiki contains a large amount of how-to guides and general
information about administrating Alpine systems.
See <http://wiki.alpinelinux.org/>.

You can setup the system with the command: setup-alpine

You may change this message by editing /etc/motd.

localhost:~# stty -F `tty` cols 9999
localhost:~# sudo lspci -vvx
-ash: sudo: not found
localhost:~# sudo lshw -notime
-ash: sudo: not found
```

`lspci` and `lshw` are not doing what we wanted. that's not the problem here (https://github.com/oxidecomputer/propolis/issues/792 for the "normal" case of this test). the problem here is that in the destination VM, the serial output is:
```
                                                     The highlighted entry will be executed automatically in 1s.                    The highlighted entry will be executed automatically in 0s.                   Booting `Linux virt'

[   28.191936] watchdog: BUG: soft lockup - CPU#0 stuck for 26s! [kworker/0:1:15]
[   56.191890] watchdog: BUG: soft lockup - CPU#0 stuck for 52s! [kworker/0:1:15]
[   96.191826] watchdog: BUG: soft lockup - CPU#0 stuck for 89s! [kworker/0:1:15]
[  124.191780] watchdog: BUG: soft lockup - CPU#0 stuck for 115s! [kworker/0:1:15]
[  152.191735] watchdog: BUG: soft lockup - CPU#0 stuck for 141s! [kworker/0:1:15]
[  180.191689] watchdog: BUG: soft lockup - CPU#0 stuck for 167s! [kworker/0:1:15]
[  208.191644] watchdog: BUG: soft lockup - CPU#0 stuck for 193s! [kworker/0:1:15]
[  236.191598] watchdog: BUG: soft lockup - CPU#0 stuck for 219s! [kworker/0:1:15]
[  284.191520] watchdog: BUG: soft lockup - CPU#0 stuck for 264s! [kworker/0:1:15]
```

[...]

PR #1080, opened 2026-03-13
nvme: deadlock enabling doorbell buffer while doing I/O

iximeow:
9a3a93a introduced a bug in the previously-okay SubQueue::pop; before, acc_mem.access() would go through MemAccessor, acquire a reference on the underlying MemCtx (an Arc refcount bump), drop the lock, and continue. immediately after, SubQueue::pop would lock the SubQueue's state and actually perform the pop.

this ordering is backwards from set_db_buf, which locks the SubQueue's state *then* conditionally performs acc_mem.access() if the buffer is set up as part of an import.

after 9a3a93a, SubQueue::pop uses access_borrowed() to avoid contention on the refcount for MemCtx, the cost of retaining the lock on the SubQueue's MemAccessor node while taking the lock on the same queue's SubQueueState. at this point a concurrent SubQueue::pop and SubQueue::set_db_buf could deadlock; one holds the queue state lock, the other holds the accessor node lock, and both will try to take the other.

resolve this tension by picking a consistent ordering for the SubQueue state and acc_mem locks: SubQueueState first, then acc_mem.{access,_borrow}(). this was already the ordering in CompQueue::push and set_db_buf, so pop() gets the trivial change.

this fixes https://github.com/oxidecomputer/propolis/issues/1035 as experienced in https://github.com/oxidecomputer/propolis/runs/66881499333 but I'm left _very_ confused; why did this only come up after @hawkw switched to running these tests on the lab Gimlet? this issue has been present since the commit was landed on Jan 9, and we'd been testing Propolis on Cosmos with it for at least a few weeks before.

### DIFF
--- a/lib/propolis/src/hw/nvme/queue.rs
+++ b/lib/propolis/src/hw/nvme/queue.rs
@@ -113,7 +113,13 @@ struct QueueState<QS> {
     /// a `SubQueueState` for a Submission Queue.
     inner: Mutex<QueueInner<QS>>,
 
-    pub acc_mem: MemAccessor,
+    /// This queue's memory accessor node.
+    ///
+    /// Be careful about lock ordering when using this accessor; access_borrow()
+    /// holds this node's lock. If a user of this queue state requires both
+    /// `access_borrow()` and `QueueInner`, the protocol is to lock queue
+    /// state first and this accessor second.
+    acc_mem: MemAccessor,
 }
 impl<QS> QueueState<QS> {
     fn new(size: u32, acc_mem: MemAccessor, inner: QS) -> Self {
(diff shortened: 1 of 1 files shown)


iximeow:
> .. too fast to hit it ..

I don't think, mostly because this has also been on dogfood for like two (?) months and as far as I know no one's VMs there have seen an issue anything like this. we're definitely launching more VMs and testing their bootedness in PHD more often and specifically than people looking at their dogfood VMs but ..

I'm more wondering if there's something interesting to how the Alpine image is submitting I/Os that it's doing work while configuring doorbell buffers. I assume (but don't know!) that's OK by the NVMe spec but I'll take a closer look at all that tomorrow..

lib/propolis/src/hw/nvme/queue.rs
```diff
@@ -113,7 +113,13 @@ struct QueueState<QS> {
     /// a `SubQueueState` for a Submission Queue.
     inner: Mutex<QueueInner<QS>>,
 
-    pub acc_mem: MemAccessor,
+    /// This queue's memory accessor node.
+    ///
+    /// Be careful about lock ordering when using this accessor; access_borrow()
+    /// holds this node's lock. If a user of this queue state requires both
+    /// `access_borrow()` and `QueueInner`, the protocol is to lock queue
+    /// state first and this accessor second.
```
hawkw: thank you!

hawkw (approved):
great cool awesome okay let's go

PR #1093, opened 2026-03-30
[multicast]  viona CTRL_RX/CTRL_MAC receive filtering + softnpu egress lock scoping + dep updates

zeeshanlakhani:
## viona: CTRL_RX and CTRL_MAC table receive filtering

This work extends the minimal `VIRTIO_NET_F_CTRL_RX` support from #1125 to cover `CTRL_RX` promiscuity/all-multicast controls and `CTRL_MAC` tableset handling. Propolis records the guest-supplied unicast and multicast MAC tables and installs the multicast table on the in-kernel device via the new `VNA_IOC_SET_MAC_FILTERS` ioctl. The unicast table is used to decide when the device must fall back to full promiscuity, since the kernel MAC client already carries the primary unicast address and does not install additional unicast filters.

Filter state is cleared on device reset and carried across live migration via `VionaStateV1`. The payload is exported unconditionally, matching the import behavior on master since #1125, which takes the payload without consulting negotiated features. Import still tolerates a missing payload from sources predating `VionaStateV1`. The device narrows to classified delivery once the driver's first `MAC_TABLE_SET` since reset is accepted (empty or not). Until then, the all-multicast lower bound holds, because some drivers, including illumos `vioif`, negotiate `CTRL_RX` but never actually manage a filter table. That marker is carried across migration as an optional `multicast_table_managed` field, where older sources omit it (the target infers it from non-empty tables) and older targets ignore it. Table installation is tracked independently of promiscuity, so guest promiscuity toggles leave installed classified flows in place rather than unplumbing and replumbing them accordingly.

On falcon links attached to softnpu, promiscuity is pinned at construction rather than derived from guest filter state. A host without `ALL_VLAN` support pins to full promiscuity, because classified unicast delivery would drop transit frames for the emulated fabric. Other falcon VMs are no longer forced promiscuous.

## viona: guest MAC replacement via CTRL_MAC_ADDR

This adds `VIRTIO_NET_F_CTRL_MAC_ADDR` support, letting a driver replace the device's active unicast MAC address via `VIRTIO_NET_CTRL_MAC_ADDR_SET`. Propolis installs the replacement on the in-kernel device through the new `VNA_IOC_SET_MAC_ADDR` ioctl (viona interface version 7) and reflects it in the device configuration space the driver reads the MAC from. The kernel removes the current address and, with it, the installed multicast filters, before installing the replacement. Therefore, the device raises full promiscuity (or the falcon superset pin) to cover delivery across the swap, and then reconciles promiscuity and the filter table with the resulting state. 

The override is carried across live migration in `VionaStateV1` as an optional `mac_addr` field, and device reset swaps the nominal address back in.

Advertising the feature requires an explicit `allow_guest_mac_change` policy grant (default off), since viona performs the swap as directed and leaves validating the requested address against host policy to its consumer. propolis-server keeps the grant off until Nexus supplies an allowed-MAC policy, while propolis-standalone accepts it as a device option. A destination without the grant rejects migration imports that carry the negotiated bit or an installed override.

## softnpu: release pipeline lock before egress I/O

Every port worker and the guest Tx path serialize on the pipeline mutex. Previously, the guard was held across both the dlpi and virtio egress, meaning that one worker blocked in egress I/O stalled the others. This is especially costly for multicast, where a single input frame fans out to several ports for replication.

This change scopes the lock to pipeline evaluation (only). The egress packets borrow from the input frame rather than the pipeline, so they remain valid after the guard is dropped. Lock acquisition moves from the callers into `process_external_packet` and `process_guest_packet`, which now take the mutex directly.

## deps and callouts

- `softnpu` is pinned to its `zl/multicast` branch pending upstream.
- `oxide-tokio-rt` is bumped to match omicron.
- Kernel-side filter installation and guest MAC replacement use viona interface version 7 support as per [illumos-gate change 775](https://code.oxide.computer/c/illumos-gate/+/775) (stlouis#986, now merged). Without it, the device remains at its all-multicast fallback.


Cargo.toml
```diff
@@ -78,7 +78,7 @@ dlpi = { git = "https://github.com/oxidecomputer/dlpi-sys", branch = "main" }
 ispf = { git = "https://github.com/oxidecomputer/ispf" }
 libloading = "0.7"
 p9ds = { git = "https://github.com/oxidecomputer/p9fs" }
-softnpu = { git = "https://github.com/oxidecomputer/softnpu" }
+softnpu = { git = "https://github.com/oxidecomputer/softnpu", branch = "zl/multicast" }
```
FelixMcFelix: Calling this out to reset to `main` before merge.

lib/propolis/src/hw/virtio/mod.rs
```diff
@@ -227,6 +227,16 @@ pub trait VirtioDevice: Send + Sync + 'static + Lifecycle {
...
+
+    /// Notification that the virtio portion of the device has been reset.
+    ///
+    /// This fires for guest-initiated status resets as well as full device
+    /// resets, after the generic virtio state and virtqueues have been reset.
+    ///
+    /// Devices should use this to restore any state tied to feature
+    /// negotiation, since the next driver may negotiate a different feature
+    /// set.
+    fn device_reset(&self) {}
```
FelixMcFelix, line 240: I think this is already covered by `Lifecycle::reset()`?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -205,6 +205,17 @@ pub mod control {
             return Err(());
         }
 
+        // Bound the driver's entry count before it reaches an allocation.
+        // The limit sits well above viona's filter limits so that oversized
+        // tables still arrive intact and trigger the fallback to multicast
+        // promiscuity. Shipping guest drivers (Linux virtio_net, FreeBSD
+        // vtnet, Windows NetKVM) request all-multicast on their own long
+        // before this size.
```
FelixMcFelix: What are their known limits? Is this how you've chosen the 64-entry limit for viona?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -205,6 +205,17 @@ pub mod control {
...
+        const MAC_TABLE_MAX_ENTRIES: u32 = 1024;
+        if entry_count > MAC_TABLE_MAX_ENTRIES {
+            return Err(());
+        }
```
FelixMcFelix: I think this will just cause the guest to give up on plumbing the device altogether? It might be beneficial to just skip the list body and move into (all-multicast) promisc. The guest hasn't actually asked for anything illegal here as far as the spec goes.

lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -555,20 +548,44 @@ impl PacketHandler {
...
+    /// poisoned, in which case the caller should stop reading frames.
     fn process_guest_packet(
         mut pkt: packet_in<'_>,
         data_handles: &Vec<dlpi::DlpiHandle>,
-        pipeline: &mut Box<dyn Pipeline>,
+        pipeline: &Mutex<Option<LoadedP4Program>>,
         log: &Logger,
-    ) {
-        for (mut out_pkt, port) in pipeline.process_packet(0, &mut pkt) {
+    ) -> bool {
```
FelixMcFelix: Nit: I'd rather this return a `Result<(), E>` for some local `enum E`, just because it makes the semantics at the caller more obvious than needing to know what `true`/`false` represent here (i.e., `if let Err(e) = process_guest_pkt(..) { break; }`).

tools/check_headers
```diff
@@ -8,7 +8,15 @@ set -e
...
-HEADER_CHECK_REF="stlouis"
+#
+# This is temporarily pinned to the viona MAC filter table change (stlouis#986,
+# https://code.oxide.computer/c/illumos-gate/+/775), which adds the
+# VNA_IOC_SET_MAC_FILTERS ioctl and struct vioc_mac_filters checked by the
+# viona-api header-check.
+#
+# Restore to "stlouis" once that change integrates. Note: this ref names a
+# specific patchset and must be bumped if a new patchset is uploaded.
+HEADER_CHECK_REF="refs/changes/75/775/3"
```
FelixMcFelix: Commenting here so we don't forget to unset this!

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -498,6 +533,15 @@ pub struct PciVirtioViona {
...
     mtu: Option<u16>,
+    /// Whether the viona device supports `VNA_IOC_SET_MAC_FILTERS`
+    /// ([`viona_api::ApiVersion::V7`]).
+    kernel_mac_filters: bool,
+    /// Promiscuity pinned at construction (ALL_VLAN, or ALL where the host
+    /// lacks that support). Falcon links carry traffic for the entire
+    /// emulated fabric, which classified delivery would drop, so no guest
+    /// or migration state may downgrade below this level.
+    #[cfg(feature = "falcon")]
+    pinned_promisc: PromiscLevel,
```
FelixMcFelix: This construct is going to end up forcing all VMs created via falcon into `PromiscMode::All` all of the time. I don't think that's what we want? At least for the prior change the main way I investigated viona's config was to spin up debian/helios VMs _using falcon_ to establish all the links etc., and then prod at them in `mdb`.

Does `a4⨉2`/softnpu require this in some way to deliver multicast traffic? Where do VLANs come in there?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1123,11 +1281,6 @@ impl VirtioDevice for PciVirtioViona {
                 VIRTIO_NO_MQ_CTRL_Q_INDEX
             };
 
-            if (feat & VIRTIO_NET_F_CTRL_RX) != 0 {
-                let mut state = self.inner.lock().unwrap();
-                self.set_promisc(PromiscLevel::None, &mut state)?;
-            }
-
```
FelixMcFelix, line 1130: Why are we dropping this here? Defaulting to `PromiscLevel::AllMulti` is purely a concession for devices without `VIRTIO_NET_F_CTRL_RX`. This reads like we're relying on a later command from the guest to put us back to what should be the default state -- is this because illumos/vioif will *not* correctly manage the multicast table?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1053,13 +1145,79 @@ impl PciVirtioViona {
         let need_promisc =
             state.filter.contains(FilterState::PROMISCUOUS) || !filter_is_self;
 
-        if need_promisc {
+        // A guest whose multicast table is empty has not demonstrated that
+        // it manages its own filtering. illumos vioif negotiates CTRL_RX and
+        // forwards promiscuity toggles over the control queue, but its
+        // multicast entry point is a noop, so narrowing to classified
+        // delivery after a promiscuity cycle (snoop on/off) would silently
+        // drop its multicast traffic. Unrequested traffic is permitted by
```
FelixMcFelix: I think this is more likely an upstream illumos/vioif driver bug -- if it's advertising that feature flag, then it *should* be tracking the MAC addresses it needs and telling the device. I don't think any driver should *expect* to receive more packets than it asks for.

I also don't know that we can always rely on compliant guests requesting at least one multicast MAC -- I guess it's a reasonable proxy since `224.0.0.1`, `ff01::1` and `ff02::1` are often automatically joined on e.g. Linux, and we don't want to leave existing illumos guests worse off. But I don't know the full diversity of guest behaviours here and if any of them misbehave in the same way as illumos.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1053,13 +1145,79 @@ impl PciVirtioViona {
...
+
+        self.set_promisc(promisc, state)?;
+
+        // Promiscuity now meets or exceeds the required level, so a
+        // lingering table is redundant rather than harmful; clearing it is
+        // best-effort.
+        if !target.install_filters
+            && state.mac_filters_installed
+            && self.hdl.set_mac_filters(&[]).is_ok()
+        {
```
FelixMcFelix: I think between here and how `PciVirtioViona::rx_config` computes `install_filters`, I had imagined that we'd be leaving the filters in place when setting promisc on/off whenever the guest directly requests it. This would be mainly so that we're spending less time in MAC plumbing and unplumbing all the requisite flows and quiescing the link to do so? I think for a large enough filter table this could take a while.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1674,6 +1884,36 @@ impl VionaHdl {
...
+    ///
+    /// This requires [viona_api::ApiVersion::V7] or greater.
+    fn set_mac_filters(&self, multicast: &[MacAddr]) -> io::Result<()> {
+        assert!(multicast.len() <= viona_api::VIONA_MAX_MCAST_FILTERS);
+
+        let entries = multicast
+            .iter()
+            .filter(|mac| !mac.is_broadcast())
+            .map(|mac| mac.0)
+            .collect::<Vec<_>>();
```
FelixMcFelix: You could write these directly into `vmf.vmf_mcast` and set the length once you're done.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2814,6 +3190,276 @@ mod test {
...
+
+        // Clearing the tables returns to the all-multicast lower bound, and not
+        // classified delivery. An empty table gives no evidence the guest
+        // manages its own filtering.
+        let ack = driver.ctrl_mac_table_set(&[], &[]);
+        assert_eq!(ack, control::Ack::Ok as u8);
+        assert_eq!(
+            test_ctx.dev.inner.lock().unwrap().promisc,
+            PromiscLevel::AllMulti
+        );
```
FelixMcFelix: In this case we really do know that this guest is correctly behaving: since its last initialisation, it has sent a nonzero number of multicast mac addresses (and the device hasn't gone through a reset since doing so).

FelixMcFelix (commented):
Thanks Zeeshan; hopefully these comments makes sense, and thanks for the extra tests around the actual handling of control queue messages. I am partly looking at this from the angle of minimising the amount of things we need to do in viona/MAC, and from keeping as many (guest ⨉ test environment) pairs out of promiscuous mode as is doable.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: Good point.

lib/propolis/src/hw/virtio/mod.rs
zeeshanlakhani: Not quite as `Lifecycle::reset()` covers a full device reset, but a guest writing 0 to the virtio status register calls `set_status() -> virtio_reset()` without invoking `Lifecycle::reset()`. That path is used for driver unload/reload and must clear the state negotiated by the previous driver.

The existing notification is only `queue_change(VqChange::Reset)`, called once per queue while the virtio state lock is held. `device_reset()`, on the other hand, provides one device-level callback after the generic state and queues are reset and the lock is released.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: The 64 is QEMU's MAC_TABLE_ENTRIES (a single 64-entry table shared across unicast + multicast), It's a common interop convention, not a virtio requirement. The spec permits silently falling back to all-multicast or promiscuous mode. 

The driver comment overgeneralizes, sure. FreeBSD caps itself at 128 entries and falls back to all-multicast, while Linux sends the full multicast list and relies on the device to degrade. The 1024-entry limit is only an allocation-safety bound;  the viona filter limit remains 64. I'll make the comment more specific.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: Agreed. The incoming change ensures that an oversized but well-formed table is not rejected. The parser retains only the bounded prefix, consumes the remainder so the following list parses correctly, and acks the command. A saturated multicast table then forces all-multicast, while a saturated unicast table forces full promiscuity. A declared count without the corresponding body remains an error because the message is malformed.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: Agreed and changed. The pin now engages only when ALL_VLAN is successfully enabled, currently (just) on viona_vlans hosts. If that fails, the constructor uses the stock, unpinned all-multicast default with guest-managed filtering, as on the mainline. Falcon guests on stock hosts are therefore no longer forced into PROMISC_ALL.

Multicast itself doesn’t require the pin, as it is covered by the all-multicast floor or the guest’s filter table. ALL_VLAN is needed for the VLAN-tagged sidecar frames multiplexed by softnpu, which the other delivery modes may drop.

﻿﻿We cannot currently restrict the pin to NICs that face softnpu ports, as propolis only receives a VNIC name, and the simnet peering beneath it is topology knowledge that only Falcon has. Scoping the pin would force a Falcon-provided per-NIC setting, which I'd leave to a follow-up. Consequently, all Falcon NICs on an ALL_VLAN-capable host remain pinned.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: > is this because illumos/vioif will not correctly manage the multicast table?

Yes‚ this is an illumos vioif compatibility concession. vioif negotiates CTRL_RX and forwards promiscuity changes, but its multicast callback is a noop and never issues CTRL_MAC_TABLE_SET. Lowering to `PromiscLevel::None` here, based solely on feature negotiation, would then drop multicast for existing illumos guests.

Rather than relying on a later non-empty table, I'll retain `PromiscLevel::AllMulti`  until the first accepted MAC_TABLE_SET. An accepted table‚ including an empty one‚ shows that the driver manages filtering. This preserves vioif behavior while allowing conforming drivers to move to classified delivery.

I'll carry this marker in the migration payload as an optional, backward-compatible field. When importing an older payload without it, a non-empty table proves prior management, while an empty table remains conservatively all-multicast because the older payload cannot distinguish those cases.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: Agreed on both counts. Illumos vioif deliberately acknowledges multicast joins without programming the device and relies on hypervisor over-delivery, so this is a compatibility workaround for driver behavior rather than a virtio guarantee. The spec also permits an empty multicast table, so non-empty membership is not a reliable signal.

I'll make the distinction command-based instead, by retaining the all-multicast floor until the driver sends its first accepted MAC_TABLE_SET. After any accepted table update‚ we know the driver is managing the table, thus we can remove the compatibility floor and derive the receive configuration from its requested state. Reset will clear that marker, and migration will preserve it. This is essentially what I said above in https://github.com/oxidecomputer/propolis/pull/1093#discussion_r3618237476.

zeeshanlakhani: Changes aside, @FelixMcFelix, should we file the bug upstream? Is that something you want me to do?

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: Yep. Filter installation now follows the table independently of guest-requested promiscuity, so a filterable non-empty table remains installed across both PROMISC and ALL_MULTICAST toggles.

In the incoming update, I also added dirty tracking so the filter ioctl is issued only when the table changes or a previous installation may have been partial. Empty/oversized tables and reset still clear it, accordingly.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: I've opened https://www.illumos.org/issues/18280.

crates/viona-api/src/ffi.rs
```diff
@@ -135,6 +139,114 @@ pub const VIONA_PROMISC_ALL: i32 = 2;
...
+/// [`vmf_err`]: vioc_mac_filters::vmf_err
+/// [`vmf_erraddr`]: vioc_mac_filters::vmf_erraddr
+#[repr(C)]
+#[derive(Default)]
+pub struct vioc_mac_filters {
+    pub vmf_nmcast: u32,
+    pub vmf_err: u32,
+    pub vmf_erraddr: [u8; ETHERADDRL],
+    pub vmf_pad: [u8; 2],
+    pub vmf_addrs: u64,
```
FelixMcFelix, line 209: `uintptr_t` might get the semantic across a little better here, but I realise it's `uint64_t` in stlouis.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -192,10 +205,26 @@ pub mod control {
         Ack = 0,
     }
 
+    /// Parse bound for a driver-provided MAC filter table.
+    ///
+    /// This bounds what a driver-controlled entry count can demand of the
+    /// device before it reaches an allocation. It sits well above viona's
```
FelixMcFelix: "reaches an allocation limit"?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -205,10 +234,35 @@ pub mod control {
...
+            // The driver declared more entries than the message carries,
+            // which is malformed rather than merely oversized.
+            return Err(());
+        }
+
+        Ok(table.into_boxed_slice())
+    }
+
+    /// Advance the chain's read position by `len` bytes without copying.
+    fn skip_readable(chain: &mut super::Chain, len: usize) -> bool {
```
FelixMcFelix, line 268: Can we have a comment here on the meaning of `true`/`false`, or otherwise return a `Result<(),()>`?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -333,6 +419,24 @@ struct Inner {
...
+    /// because the guest changed it or a previous installation may have been
+    /// partial.
+    mac_filters_dirty: bool,
+    /// Whether (or not) the driver has issued an accepted `MAC_TABLE_SET` since
+    /// the last reset.
+    ///
+    /// An accepted table, including an empty one, demonstrates that the driver
+    /// manages its own filtering, which releases the all-multicast lower bound
+    /// in [`PciVirtioViona::rx_config`].
+    mac_table_set: bool,
```
FelixMcFelix, line 436: I'd maybe include a cursory note that we're doing this because of https://www.illumos.org/issues/18280.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,12 +1177,119 @@ impl PciVirtioViona {
...
+    /// Replace the device's active unicast MAC address on behalf of a
+    /// driver's `VIRTIO_NET_CTRL_MAC_ADDR_SET` command.
+    ///
+    /// The comments below distinguish the nominal address, the
+    /// [`PciVirtioViona::mac_addr`] the device was created with, from a
+    /// driver-installed override.
+    fn set_mac_override(&self, mac: MacAddr) -> Result<(), ()> {
+        // The policy is checked directly rather than through feature
+        // negotiation alone: the bit cannot be negotiated without the
+        // grant, but the swap must not depend on that indirection.
```
FelixMcFelix: I can't really see what this phrasing is getting at (e.g., grant?). Are we trying to say here that `VIRTIO_NET_F_CTRL_MAC_ADDR` is advertised only when `allow_guest_mac_change`, but we check both sides here for completeness?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1000,67 +1301,189 @@ impl PciVirtioViona {
...
             Ok(_) => {
                 state.promisc = level;
                 Ok(())
             }
             Err(_) => {
                 // Ensure that we return to the old level of promisc.
-                if self.hdl.set_promisc(state.promisc).is_err() {
-                    self.virtio_state.set_needs_reset(self);
-                }
+                let _ = self.hdl.set_promisc(state.promisc);
```
FelixMcFelix: Aren't we in a bad enough state to need a device reset if we can't revert to our earlier promisc level?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1000,67 +1301,189 @@ impl PciVirtioViona {
...
+            // level, meaning a guest promiscuity cycle does not unplumb and
+            // replumb the table's classified flows (which quiesces the
+            // link). Viona keeps the classified callback installed: under
+            // ALL it drops classified copies, while under MULTI it drops
+            // classified multicast copies and delivers multicast through
+            // the promiscuous callback.
+            //
+            // An empty table is never installed. Classified delivery with
+            // no table covers the device MAC and broadcast, which is
+            // exactly what an empty table requests.
```
FelixMcFelix: Guests do often start with an empty table from what I recall, then add the pile of all-routers, etc., related MACs. These tables are assumed to refer to MACs in _addition_ to the config space MAC and broadcast. I think it's just a wording thing for this comment, more than anything?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1224,15 +1717,16 @@ impl Lifecycle for PciVirtioViona {
...
-        self.hdl.set_pairs(1).expect("can set viona back to one queue pair");
+        // SET_PAIRS can fail while restoring the kernel's receive callbacks,
+        // and SET_USEPAIRS can fail independently. Attempt both restorations,
+        // and on failure surface NEEDS_RESET to the guest rather than
+        // panicking.
+        let use_pairs = self.set_use_pairs(1);
+        let pairs = self.hdl.set_pairs(1);
+        if use_pairs.is_err() || pairs.is_err() {
+            self.virtio_state.set_needs_reset(self);
+        }
```
FelixMcFelix: If we're at this point we *are* resetting the NIC, so I'm unsure that signalling `needs_reset` will be of much use. What exactly is the failure condition for these ioctls, now?

Leaving aside the addition/split of `VirtioDevice::device_reset` for now, why are we resetting the queuepair counts in a separate spot from the rest of the device logic? https://github.com/oxidecomputer/propolis/pull/1093#discussion_r3615543568 makes it seem as though this logic should live there too, if the guest forces a reset.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -222,6 +276,24 @@ pub mod migrate {
...
+        /// nominal one the device was created with, if any
+        /// (`VIRTIO_NET_CTRL_MAC_ADDR_SET`).
+        ///
+        /// This is optional for compatibility, as older sources omit it.
+        ///
+        /// An older destination ignores it and reverts config space to the
+        /// nominal address, while the guest continues using the replaced
+        /// one, which that host will not classify.
+        #[serde(default)]
+        pub mac_addr: Option<MacAddr>,
```
FelixMcFelix, line 295: My feeling is that this might have to be a V2 payload, because if we do migrate from a propolis with this capability (and which allows guest-driven MAC address config) to an older version, the guest will be left in a very strange state where it believes the host knows about its brand new MAC address but the kernel and PCI cfg-space are not thusly configured.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1674,6 +2278,142 @@ impl VionaHdl {
...
+            .collect();
+
+        let mut vmf = viona_api::vioc_mac_filters {
+            vmf_nmcast: addrs.len() as u32,
+            // A count of zero clears the table without the kernel reading
+            // the buffer. We pass no pointer rather than a dangling one.
+            vmf_addrs: if addrs.is_empty() {
+                0
+            } else {
+                addrs.as_bytes().as_ptr() as u64
```
FelixMcFelix:
```suggestion
                u64::try_from(addrs.as_ptr().addr()).expect("suitable justification for usize->uintptr_t/u64?")
```

I'm not sitting with `rust-analyzer`, so I'm not seeing how `as_bytes()` is resolved here, but I'd probably have written it like the above regardless.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1674,6 +2278,142 @@ impl VionaHdl {
...
+    /// Read the installed multicast filter table.
+    ///
+    /// This requires [viona_api::ApiVersion::V7] or greater.
+    #[cfg(test)]
+    fn get_mac_filters(&self) -> io::Result<Vec<MacAddr>> {
+        let mut addrs =
+            vec![MacAddr::default(); viona_api::VIONA_MAX_MCAST_FILTERS];
+        let mut vmf = viona_api::vioc_mac_filters {
+            vmf_nmcast: addrs.len() as u32,
+            vmf_addrs: addrs.as_mut_bytes().as_mut_ptr() as u64,
```
FelixMcFelix: I'd probably go with `addrs.as_mut_ptr().addr()`, again.

FelixMcFelix (commented):
Thanks Zeeshan. I think I'm generally happy with the actual plumbing of the filter tables, as well as the logic to set/unset various promisc levels in response to the error conditions aside from maybe one or two points of confusion. It looks good on the functionality front, though I haven't yet tested it myself and made sure that everything looks right in `mdb` for various guests yet.

I'd need to defer to @iximeow or @papertigers on the `VirtioDevice` changes, as well as the accuracy of the new tests which are actually driving the rings via `send_ctrl_cmd`.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: Agreed.

The missing distinction was between a failure that reverted cleanly, where `Inner::promisc` still describes the kernel and the rejected guest command is the only consequence, and one where the revert also failed, leaving the recorded level stale.

These outcomes are now represented by `PromiscError::Reverted` and `PromiscError::Diverged`, respectively.

`set_promisc` cannot safely raise the reset itself because `set_needs_reset` takes the virtio state lock, which is held across `queue_change` (while `Inner` is acquired). It therefore returns the outcome, and callers escalate after releasing the guard. The `apply_rx_config` callers retry reconciliation before escalating. The missing path was the `cover` raise in `set_mac_override`, which previously returned an error without either retrying or escalating.

I also documented why the level is reissued rather than short-circuited when it matches the recorded level. viona falls back to `VIONA_PROMISC_NONE` when `mac_promisc_add` fails during callback restoration, while retaining the requested mode in `l_promisc`. Reissuing the request, therefore, retries installation of the missing handler. The kernel elides a genuine no-op itself, returning early only when `l_rx_mode` matches as well: https://github.com/oxidecomputer/illumos-gate/blob/e73f2575430340bc2ebdfcb0b5d49ef8fa72d816/usr/src/uts/intel/io/viona/viona_main.c#L1631-L1646.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: Both points make sense. 

On the failure conditions: `VNA_IOC_SET_USEPAIRS(1)` reaches `viona_ioc_link_usepairs`. The request is within the valid range, so the validation failure paths do not apply here. `VNA_IOC_SET_PAIRS(1)` reaches `viona_ioc_link_setpairs`, which can fail if the requested count is below `l_usepairs`, if a ring is not in `VRS_RESET`, or if `viona_rx_set` fails while restoring receive callbacks.

The first two conditions are not reachable here: we lower `l_usepairs` first, and `virtio_reset` has already reset every ring before invoking `device_reset`. The `viona_rx_set` failure is recoverable, however, as it drops the active mode to `VIONA_PROMISC_NONE` while retaining the request in `l_promisc`, so the promiscuity restore below retries callback installation.

This makes `needs_reset` both unnecessary and misleading honestly, so it is now gone. The promiscuity restore in `device_reset` is the retry for that callback-restoration failure.

On placement, yeah..., it is a bug rather than an inconsistency. A guest writing zero to the status register goes through `set_status` to `virtio_reset` and then `device_reset`, but it does not go through `Lifecycle::reset`. A driver that negotiated multiqueue and then unloaded therefore left the kernel holding its old pair count. 

Resetting the pair count in `device_reset` fixes that path. The split itself was needed for resetting the `VIRTIO_NET_F_CTRL_RX` state. Restoring the pair count is an additional benefit of having the device-level reset callback.

Refs: 
- https://github.com/oxidecomputer/illumos-gate/blob/e73f2575430340bc2ebdfcb0b5d49ef8fa72d816/usr/src/uts/intel/io/viona/viona_main.c#L1403-L1423
- https://github.com/oxidecomputer/illumos-gate/blob/e73f2575430340bc2ebdfcb0b5d49ef8fa72d816/usr/src/uts/intel/io/viona/viona_main.c#L1426-L1432
- https://github.com/oxidecomputer/illumos-gate/blob/e73f2575430340bc2ebdfcb0b5d49ef8fa72d816/usr/src/uts/intel/io/viona/viona_main.c#L1034-L1066.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: ~~removed~~

crates/viona-api/src/ffi.rs
zeeshanlakhani: Yeah, I put the reasoning on the field here.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: I cleared this up.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
         }
     }
 
+    /// Replace the device's active unicast MAC address on behalf of a
+    /// driver's `VIRTIO_NET_CTRL_MAC_ADDR_SET` command.
+    ///
+    /// The comments below distinguish the nominal address, the
+    /// [`PciVirtioViona::mac_addr`] the device was created with, from a
+    /// driver-installed override.
+    fn set_mac_override(&self, mac: MacAddr) -> Result<(), ()> {
```
iximeow, line 1246: fwiw unless we think there's a reason we might want to allow guests to change their NIC's MAC in the future, I don't think there's a reason we should even theoretically plumb the command to viona, even if viona supports it? that is, why not implement this as `return Err(())` and be done with it?

if there's a way this is helpful then I don't mean to stand in the way, but if this is plumbed up only because the spec says the command has to be accepted with CTRL_RX being negotiated, this seems like all risk (because if you end up able to set your MAC to another guest's NIC...), plus implementation to support, for little gain?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -498,6 +687,21 @@ pub struct PciVirtioViona {
     dev_features: u64,
     mac_addr: MacAddr,
     mtu: Option<u16>,
+    /// Whether the viona device supports the MAC filtering ioctls of
+    /// [`viona_api::ApiVersion::V7`]: `VNA_IOC_SET_MAC_FILTERS` and
+    /// `VNA_IOC_SET_MAC_ADDR`.
+    kernel_mac_filters: bool,
```
iximeow, line 693: when `VIONA_CURRENT_INTERFACE_VERSION` is 7 this is true, and by making changing the API version expected in `check_api_version()` to V7 in this file you can cause Propolis to fail to start if the host bits are too old. lets do that here and not worry about feature-detecting with `kernel_mac_filters` and all that.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
+                let outcome = match self.hdl.set_promisc(state.promisc) {
+                    Ok(_) => PromiscError::Reverted,
+                    Err(_) => PromiscError::Diverged,
+                };
+                // The VIONA_PROMISC_* constants run 0 through 3, so neither
+                // level narrows.
+                probes::virtio_viona_promisc_err!(|| (
+                    usize::from(level) as u8,
+                    usize::from(state.promisc) as u8,
+                    e.raw_os_error().unwrap_or(0),
```
iximeow, line 1364: it occurs to me that we probably want an error variant out of ioctls that guarantees it can offer the OS error from the ioctl (if any). it would look something like this anyway but it at least is a way for us to know that 0 really means that the instance is gone rather than the ioctl returned 0. not a change for this PR, I'll do a separate issue about it.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1674,6 +2348,144 @@ impl VionaHdl {
...
+    /// rejected at the control queue, and every remaining failure is
+    /// answered the same way: by falling back to multicast promiscuity in
+    /// [`PciVirtioViona::apply_rx_config`].
+    ///
+    /// Diagnosis stays with the `virtio_viona_mac_filters_err` probe below,
+    /// which carries the code and the offending address.
+    ///
+    /// This requires [viona_api::ApiVersion::V7] or greater.
+    fn set_mac_filters(&self, multicast: &[MacAddr]) -> io::Result<()> {
+        assert!(multicast.len() <= viona_api::VIONA_MAX_MCAST_FILTERS);
```
iximeow, line 2375: are we obligated to limit the number of multicast MACs before issuing the `SET_MAC_FILTERS` ioctl? this assert seems weird: presumably the OS can tolerate too too many multicast MACs by returning a normal error, which would be an error either of the viona user (Propolis, in the current implementation), or an error for the guest driver to handle ("setting the MAC filters failed", if that's an idea that guests can handle).

in either case it seems inappropriate to assert here unless the device operation would be _incorrect_ if we proceeded?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1674,6 +2348,144 @@ impl VionaHdl {
...
+        let mut vmf = viona_api::vioc_mac_filters {
+            vmf_nmcast: addrs.len() as u32,
+            // A count of zero clears the table without the kernel reading
+            // the buffer. We pass no pointer rather than a dangling one.
+            vmf_addrs: if addrs.is_empty() {
+                0
+            } else {
+                u64::try_from(addrs.as_bytes().as_ptr().addr())
+                    .expect("usize fits in u64 on 64-bit targets")
+            },
```
iximeow, line 2392: nit: I'd mildly prefer if we could figure out `addrs.as_bytes().as_ptr().addr() else std::ptr::null()` above `vmf`, and just `vmf_addrs: u64::try_from(that_ptr).expect("..etc...")` here. just because that separates out "we're converting the data to a pointer" and "we're converting the pointer to the ABI-correct type", and the first part is not totally trivial like `addrs.len()` is above

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: > No Viona migration payload has shipped yet, and this PR introduces the first schema.

Sorry, what do you mean by this? I think `migrate::VionaStateV1` has been in the product since R21.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2317,7 +2387,8 @@ impl VionaHdl {
             vmf_addrs: if addrs.is_empty() {
                 0
             } else {
-                addrs.as_bytes().as_ptr() as u64
+                u64::try_from(addrs.as_bytes().as_ptr().addr())
```
FelixMcFelix, line 2390: I don't think you need the `.as_bytes()` here.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2405,7 +2476,8 @@ impl VionaHdl {
             vec![MacAddr::default(); viona_api::VIONA_MAX_MCAST_FILTERS];
         let mut vmf = viona_api::vioc_mac_filters {
             vmf_nmcast: addrs.len() as u32,
-            vmf_addrs: addrs.as_mut_bytes().as_mut_ptr() as u64,
+            vmf_addrs: u64::try_from(addrs.as_mut_bytes().as_mut_ptr().addr())
```
FelixMcFelix, line 2479: Ditto on not needing `.as_mut_bytes()`.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1644,10 +1722,26 @@ impl VirtioDevice for PciVirtioViona {
...
+        // SET_USEPAIRS precedes SET_PAIRS because `viona_link_qalloc` rejects
+        // a pair count below the in-use count. Failures are discarded: the
+        // guest must renegotiate before using the device again, and
+        // `set_features` restores both counts when it does.
+        probes::virtio_viona_mq_set_use_pairs!(|| (
+            MqSetPairsCause::Reset as u8,
+            1
+        ));
+        let _ = self.set_use_pairs(1);
+        let _ = self.hdl.set_pairs(1);
```
FelixMcFelix, line 1740: From looking at `stlouis` again, my read is that we can keep an unwrap or expect on the `set_use_pairs` call because we know that the input is well-founded. I think the call to `set_pairs` could do with at least a probe rather than discarding the value.

zeeshanlakhani: Gotcha, yeah. ok.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: @FelixMcFelix derp, yeah, I'm a dummy, it's just the initial one. I'll introduce a second-versioned payload with the filters, etc, and update `mac_addr` there to not be the option.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
+            // ALL it drops classified copies, while under MULTI it drops
+            // classified multicast copies and delivers multicast through
+            // the promiscuous callback.
+            //
+            // An empty table is never installed. A multicast table names
+            // addresses in addition to the device's own MAC and broadcast,
+            // both of which classified delivery already covers. An empty
+            // one needs nothing further.
+            install_filters: mcast_filterable
+                && !state.multicast_mac_filters.is_empty(),
```
iximeow, line 1479: I would most immediately ask that it seems like we could name the predicate `mcast_filterable && !state.multicast_mac_filters.is_empty()`, but more generally since `mcast_filterable` is effectively a question of `is multicast_mac_filters too big`, this is really a predicate saying "should I install these filters I have".

it seems like we'd want `state` to reflect the filters we've told viona about (or want to), rather than being a list that we may or may not install because of this secondary mechanism?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
+        // No table is required when it is empty, cannot be installed, or the
+        // link is pinned. Promiscuity has already been raised where needed,
+        // so clearing a lingering table is best-effort.
+        if !target.install_filters {
+            if state.mac_filters_installed
+                && self.hdl.set_mac_filters(&[]).is_err()
+            {
+                // The failed clear may leave entries behind, so a later
+                // apply retries it.
+                state.mac_filters_dirty = true;
```
iximeow, line 1537: what happens if we are set to be promiscuous but still have MAC filters installed because this failed? unless we have a reason to believe we should be tolerating transient failures to `set_mac_filters`, we should at _least_ have a probe here, and personally I think I'd `warn!()` with a `TODO` about this needing to be more visibly reported.

it seems like if we got here there is a bug either in Propolis or viona which we do need to find and fix (not to mention even find out exists..).

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
-                Err(())
+            Err(e) => {
+                // Return to the old level of promisc, reporting whether that
+                // succeeded so the caller can escalate a divergence.
+                let outcome = match self.hdl.set_promisc(state.promisc) {
+                    Ok(_) => PromiscError::Reverted,
+                    Err(_) => PromiscError::Diverged,
+                };
+                // The VIONA_PROMISC_* constants run 0 through 3, so neither
+                // level narrows.
```
iximeow, line 1360: @FelixMcFelix I'm curious about your thoughts on this, since I remember the "try changing things, try reverting if it fails" conversations we had in #1125: if we've tried to change viona configuration, and failed, why shouldn't we just operate as if the device is broken, `set_needs_reset()`, and move on?

in particular, if we say "set_promisc() failed and the device state is undefined, you may get fewer or greater packets than expected, you need to reset the device to return to a defined state, good luck", we don't even need to try reverting promiscuity, or plumbing PromiscError and branching on it or any of that. just, we failed to effect the configuration, you have to reset the device now. might guests typically try to set filter configurations which we should consider unacceptable?

I think this might be a safer approach overall anyway because when `set_promisc` fails we _do not_ know the state of the viona device. it is possible the device is, at that point, dropping packets the guest expected, or providing packets it did not. we may then "fix" things by resetting promiscuity here, and only return an error for that attempted reconfiguration with the device in a known good state. but even then, for the period between the two `set_promisc` calls the device may operate in a way the guest never expected and was never told it would.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
         let need_promisc =
             state.filter.contains(FilterState::PROMISCUOUS) || !filter_is_self;
 
-        if need_promisc {
+        // A guest that has never issued an accepted MAC_TABLE_SET has not
+        // demonstrated that it manages its own filtering. illumos vioif
+        // negotiates CTRL_RX and forwards promiscuity toggles over the
+        // control queue, but its multicast entry point is a noop, so
+        // narrowing to classified delivery after a promiscuity cycle
+        // (snoop on/off) would silently drop its multicast traffic.
```
iximeow, line 1451: I don't really understand what this is trying to communicate. what does it mean to say vioif's "multicast entry point is a noop"? are we really talking about a driver limitation in a guest OS here, and using that to inform our decision about what correct device implementation for spec-defined filtering controls?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
         } else {
             PromiscLevel::None
+        };
+
+        RxConfig {
+            promisc,
+            // Installation tracks the table alone, not the promiscuity
+            // level, meaning a guest promiscuity cycle does not unplumb and
+            // replumb the table's classified flows (which quiesces the
+            // link). Viona keeps the classified callback installed: under
```
iximeow, line 1469: is it typical to change the promiscuity of a link while under an interesting amount of traffic, and to want no impact to that traffic? I ask because if we're taking extra implementation complexity solely so that a (relatively rare?) device reconfiguration does not affect traffic that probably won't be concurrent with the reconfiguration, that seems funky.

if there's something I'm missing where guest OSes would be toggling promiscuity as part of normal operation while someone cares about the quality of traffic, then fair enough. that would be worth noting.

in any case, if this field needs to be here, this comment should be a doc comment on the field where `struct RxConfig` is defined, above, so it doesn't get lost if this code is shuffled around or someone else is cooking up `RxConfig`s later.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
+            install_filters: mcast_filterable
+                && !state.multicast_mac_filters.is_empty(),
         }
     }
+
+    /// Reconcile the in-kernel Rx configuration with the guest's filter
+    /// state, as computed by [`Self::rx_config`].
+    ///
+    /// Ordering follows the contract documented for
+    /// `VNA_IOC_SET_MAC_FILTERS`: a filter table is installed before
```
iximeow, line 1487: where is this contract documented? am I missing something in `viona_io.h`?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
+    /// promiscuity is lowered, and promiscuity is raised before a table is
+    /// cleared, so new arrivals always have a live delivery path across the
+    /// transition, with in-kernel dedup suppressing the overlap's duplicate
```
iximeow, line 1490: I really don't understand what "live delivery path" means; if the guest is reconfiguring a device that is actively handling traffic, then.. yes, the rx path must not be broken at any point because the device is being used. this is just saying "`apply_rx_config()` is careful to not be wrong", right..?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -983,84 +1216,352 @@ impl PciVirtioViona {
...
+                // replacement or clearing.
+                Err(_) => {
+                    state.mac_filters_installed = true;
+                    state.mac_filters_dirty = true;
+                    match target.promisc {
+                        PromiscLevel::None => PromiscLevel::AllMulti,
+                        covered => covered,
+                    }
+                }
+            }
```
iximeow, line 1519: if `set_mac_filters()` for example returned EFAULT, which is really something we should never ever ever ever get from an ioctl, we will lose that error and possibly limp along by having papered over it by setting promiscuity. if setting promiscuity also fails (maybe viona is in a bad spot?) we'll bail with that error, but we'll have lost this earlier operation and its fault.

that seems hard to debug, and it seems extremely difficult for a guest to determine while it might get flooded with packets from `PromiscLevel::AllMulti` when really something in viona has gone wrong. we should really just punt the error upwards and `set_needs_reset()` rather than trying to fix things here.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -333,6 +418,26 @@ struct Inner {
     filter: FilterState,
     unicast_mac_filters: Box<[MacAddr]>,
     multicast_mac_filters: Box<[MacAddr]>,
+    /// Whether (or not) a (possibly partial) multicast filter table is
+    /// installed on the in-kernel MAC client via `VNA_IOC_SET_MAC_FILTERS`.
+    mac_filters_installed: bool,
+    /// Whether (or not) the next installable multicast table must be written
+    /// because the guest changed it or a previous installation may have been
+    /// partial.
+    mac_filters_dirty: bool,
```
iximeow, line 427: needing this field seems to me like a sign we're doing something wrong. if it's ever true while we're not actively in a function that reconfigures viona, that means at some arbitrary point in the past something went wrong and viona state is only partially aligned with what we have in Propolis, but we think that viona probably is doing the right thing? if we _did_ have this, why _wouldn't_ we periodically try to reapply the partial state to bring viona into the expected configuration? we should definitely not do _that_, but having to ask if that would make sense again tells me that things are in a delicate (undesirable?) circumstance.

in the cases I see below where we'd have partial configuration like this, I think it's basically always cases where we should say the device is just not in a known state anymore, that for some period of time the device may have been _wrong_ with respect to the guest's following of the VirtIO spec, and that we should set NEEDS_RESET. the upside is that this all gets a bit simpler?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1329,6 +1905,20 @@ impl MigrateMulti for PciVirtioViona {
...
+
+        // A source that negotiated CTRL_MAC_ADDR granted its guest MAC
+        // replacement, a policy this host must extend as well because the
+        // guest keeps the negotiated bit and may issue further swaps here.
+        if (feat & VIRTIO_NET_F_CTRL_MAC_ADDR) != 0
+            && !self.allow_guest_mac_change
+        {
+            return Err(MigrateStateError::ImportFailed(
+                "source negotiated VIRTIO_NET_F_CTRL_MAC_ADDR but guest \
+                MAC replacement is not permitted on this host"
```
iximeow, line 1917: `allow_guest_mac_change` is not a host policy question. the documentation elsewhere in this change proposes it as a Nexus policy, though certainly one might imagine it is an AND of "Nexus permits .." + "Propolis on this host permits .."

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1337,7 +1927,12 @@ impl MigrateMulti for PciVirtioViona {
 
         let has_ctl_queue = (feat & VIRTIO_NET_F_CTRL_VQ) != 0;
         if (feat & VIRTIO_NET_F_MQ) != 0 {
-            self.hdl.set_pairs(PROPOLIS_MAX_MQ_PAIRS).unwrap();
+            self.hdl.set_pairs(PROPOLIS_MAX_MQ_PAIRS).map_err(|e| {
+                MigrateStateError::ImportFailed(format!(
+                    "error while restoring viona queue pairs \
+                    ({PROPOLIS_MAX_MQ_PAIRS}): {e:?}"
+                ))
+            })?;
```
iximeow, line 1935: this was an `unwrap()` because setting the viona device to the max queue pairs Propolis thinks viona supports should never fail. abstractly we probably shouldn't panic here so at least the source Propolis can learn the migration failed and maybe resume the VM? but a change for that reason is pretty distinct and would need a broader look at import handling.

(also the message here is missing a critical detail that we didn't just set the device to some queue pairs we liked, it is specifically actually `PROPOLIS_MAX_MQ_PAIRS`, in addition to being whatever specific number)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1358,23 +1953,102 @@ impl MigrateMulti for PciVirtioViona {
...
+            Err(e) => return Err(e),
+        };
+
+        let migrate::VionaStateV1 {
+            // The source's promisc level is a host mechanism, derived from
+            // the guest state below plus the source's own pin and kernel
+            // capabilities. It is not restored directly: doing so could
+            // demand a level this kernel does not support (ALL_VLAN on a
+            // stock host) or hold a level this host's pin does not require.
+            promisc: _,
```
iximeow, line 1975: "this host's pin"? I have generally thought when you say "this host" you mean "the kernel" or "the hardware", but "this host's pin" makes me think you mean "this Propolis process".

anyway, if you're migrating between Propolis versions that differ on being Falcon builds and not, a lot is going wrong and we actually should probably fail. instead, we're recomputing the promiscuity through `apply_rx_config` which is at least probably going to be what we had before. but then we're also potentially messing with promiscuity via a guest MAC override and so things are still kinda complicated.

this all seems like a complication we should not accept just to support migration between Falcon and non-Falcon Propolises.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1993,25 +2821,31 @@ mod test {
...
             let new_migrate = MigrateCtx { mem: &acc_mem };
-            <PciVirtioViona>::import(&new_ctx.dev, &mut offers, &new_migrate)
-                .expect("can import PciVirtioViona");
+            let res = <PciVirtioViona>::import(
+                &new_ctx.dev,
+                &mut offers,
+                &new_migrate,
+            );
+            // Start the target either way: a rejected import leaves the
+            // kernel state untouched, so it comes up as a cold boot would.
```
iximeow, line 2837: this is incredibly wrong. we may have gotten past `self.hdl.set_features()` for example, and failed at a later state (like setting queue pairs, like you changed here!). a failed import means the device state is undefined, and that device cannot be used in a productive manner without at least being reset.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2357,7 +3205,14 @@ mod test {
...
 
-            let avail_gpa = descriptor_table_gpa.next_multiple_of(page_u64);
+            // Give each ring a page of its own by stepping a full page per
+            // ring.
+            //
+            // Aligning with `next_multiple_of` would not work here:
+            // `next_queue_gpa` is already page-aligned, so it would be a
+            // noop and leave the descriptor table, available ring, and used
+            // ring all at the same address.
+            let avail_gpa = descriptor_table_gpa + page_u64;
```
iximeow, line 3215: I suppose you can see that the tests i'd written did not usefully use the queues beyond laying them out, lol. I had used `next_multiple_of` because the queue sizes are probably not exactly one page, and I wanted to keep all of these addresses page-aligned to help catch obviously-wrong values when debugging. what was `some_gpa.next_multiple_of(page_u64)` should really be `(some_gpa + len).next_multiple_of(page_u64)`.

if one of these items is _larger_ than a page, this change will be wrong and make for very confusing test issues.

lib/propolis/src/hw/virtio/viona.rs
iximeow: this should either expect or _maybe_ set NEEDS_RESET, yeah. if we're failing to set the device back to one queue pair we're operating viona wrong and that's a Propolis bug.

lib/propolis/src/hw/virtio/mod.rs
```diff
@@ -228,6 +228,16 @@ pub trait VirtioDevice: Send + Sync + 'static + Lifecycle {
...
     }
+
+    /// Notification that the virtio portion of the device has been reset.
+    ///
+    /// This fires for guest-initiated status resets as well as full device
+    /// resets, after the generic virtio state and virtqueues have been reset.
+    ///
+    /// Devices should use this to restore any state tied to feature
+    /// negotiation, since the next driver may negotiate a different feature
+    /// set.
```
iximeow: "should" is not appropriate here; devices *must* use this to reset any state VirtIO describes as not persisting across device reset. that is probably feature negotiation, and generally _probably_ is everything (that is not PCI state?), but implementors *must* follow VirtIO here, not what they think should or shouldn't be persistent.

lib/propolis/src/hw/virtio/mod.rs
iximeow: the reset logic that is in `viona.rs`' `device_reset` would also be fine in `viona.rs`' `set_features`. I'll comment more on `device_reset` in viona.rs.

lib/propolis/src/hw/virtio/pci.rs
```diff
@@ -1190,6 +1190,10 @@ impl PciVirtioState {
         self.reset_queues_locked(dev, &mut state);
         state.reset();
         let _ = self.isr_state.read_clear();
+        // Release the state lock before calling into the device so that its
+        // reset handling may query generic virtio state.
+        drop(state);
+        dev.device_reset();
```
iximeow, line 1196: we should not release the mutex here, or `device_reset` implementations must be cautious that while they are resetting device state, the corresponding virtio, queues, and PCI state _may not_ be the state at reset anymore. we'd probably want to pass the mutex along here, but this is the kind of complication that has me leaning towards pushing `set_features` maybe a bit too far (see below)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1213,26 +1720,91 @@ impl VirtioDevice for PciVirtioViona {
...
     }
-}
-impl Lifecycle for PciVirtioViona {
-    fn type_name(&self) -> &'static str {
-        "pci-virtio-viona"
-    }
-    fn reset(&self) {
-        self.virtio_state.reset(self);
+
+    fn device_reset(&self) {
```
iximeow, line 1724: generally, `set_features` is equivalently "called once per device initialization", and _might_ be a better place to handle filters for now? mainly, `FEATURES_OK` must be set _before_ enabling queues, so we know that if any has persisted through to `set_features`, it won't cause guest-visible misbehavior because there won't be queues configured to discover it.

it's a bit funky to overload `set_features` like that, but I think keeping ourselves to reasoning about the local "viona set_features() + virtio device reset" helps avoid the complication of who can reset what state with what guarantees we'd have with `device_reset()` (though in general it's not a _bad_ idea, and if we had other kinds of reset it's probably worth thinking through)

concretely, I think we already have a bug that if you initialize a device, negotiate F_CTRL_RX, set a promiscuity, reset, and did not negotiate the feature a second time, you end up with viona still in a promiscuous mode that you'd never actually asked for. that's unfortunate!

it seems to me that we should always `set_promisc(PromiscLevel::None)`, and clear filters, and ... etc to clear _all_ configurable state, rather than doing it piecemeal based on what we think the negotiated features were at the last time we reset. here, if we kept MAC overrides, that simplifies things a lot: just set PromiscLevel::None and be done with it. likewise, `self.hdl.set_mac_filters(&[])` always when we're going through here, maybe a guest negotiates the feature to change it again in the future, but that's the end of it and we never have to ask if there were filters before.

it probably _does_ still make sense to parcel out the state resetting into a standalone function, then you "only" have the weirdness of describing why it would happen in `set_features` rather that something that's a first-class part of a virtio device lifecycle. and later if/when lifecycle bits for it are more thought through we'll just move that call to reset out from the start of `set_features` and to wherever someone poked that bit into the status register.

that's more work on the reset path that might not be strictly necessary, but if someone is bound on our ability to reset the virtio-net device, uhhhh.... what is happening :) otoh it's all really simple to think through and extend in the future.

iximeow (changes requested):
I've really only looked at most of the viona/virtio-net changes so far, but I have a few other questions about the change at large, please bear with me..

how does this relate to multicast? the MAC filtering is "just" more comprehensive support for virtio features which guests could use regardless of multicastness, and iiuc aren't necessary for a guest _to_ work with multicast traffic?

the dependency updates are totally incidental, right? we don't need to bump `oxide-tokio-rt` here for any particular reason, nor `p4` or `softnpu`? if yes, since that's trivial to piece out, lets just do that?

likewise, the softnpu egress stuff is .. a performance-related change for softnpu right? so that is both a really clear distinct change we could make and review in its own right, and not necessary for the MAC filtering bits, nor related to dependency updates?

my understanding is that this was a draft while you've been working on multiqueue generally, and so this is an accumulation of Propolis changes along the way. that's fine of course, but since the boundaries between the changes are really clear, lets not tie all it together?

part of why I'm asking is, after the conversation yesterday about VMM/viona interface version numbers, it has become clear to me that it is just outright _wrong_ for a Propolis to have a different "current" interface version than the kernel it's running on. I'm actually thinking about putting a patch together that bumps the viona interface version to 7 to match stlouis, and masks CTRL_RX so that guests do not think the feature is available until we've landed the corresponding changes here. as-is, I actually do not understand why a TUF would be able to be installed and start a Propolis, because stlouis has already been updated and there is simply no corresponding Propolis that could safely run there; if racklettes do work on current TUFs, they should not. not to mention Propolis CI being broken, but that's less concerning to me than "the product works when it seems like it should not" :)

zeeshanlakhani:
@iximeow Admittedly, over the last few weeks, I've been changing this code a lot just to get multicast to work in voxel, so it has been super prototype-y, but I've never written legit propolis code before (minus stats). But, this came about in how were we using promisicuous mode and delivering multicast packets for link-local (previously), which was the inherent issue for the duplication.

To your points, maybe we don't need the guest-MAC override ability presented here to get the work through. And, as you mentioned I conflated the dependency and locking issues, the former to match Omicron, to make the end-to-end demo work. 

As I've been doing in dendrite and other repos related to mcast changes, I'm going to clsoe this PR and split out the work in smaller chunks, given the questions and discussion pieces you've mentioned. Though, it was still good for a first look to know what to split out and do (and test for).

sion42x:
@iximeow I know this PR is closed now, but this comment concerns me:

> as-is, I actually do not understand why a TUF would be able to be installed and start a Propolis, because stlouis has already been updated and there is simply no corresponding Propolis that could safely run there; if racklettes do work on current TUFs, they should not. not to mention Propolis CI being broken, but that's less concerning to me than "the product works when it seems like it should not"

Can you elaborate?

iximeow:
@sion42x I since figured out why things work, and it's not really.. great, but at least explains why things work.

`check_api_version()` in `viona.rs` checks that the current viona API version is at least whatever it is that we consider the minimum, and will silently be OK with the host's viona version being newer than what we support. that's a different interpretation of the version number than the kernel and c-byhve have, where they simply must just match. the kernel's interpretation is because the kernel does things like offer virtio-net features that we pass through to the guest, so a different viona version can tell a guest that it should try doing things that Propolis does not support.

the lifecycle of all this is very similar for VMM_CURRENT_INTERFACE_VERSION but checked in slightly different places for the different API.

we should at least warn and probably in an `omicron-build` build of Propolis fail to run if the kernel's versions are not exactly equal to the ones Propolis expects to be operating on.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: This isn't something we need. But I think the feature negotiation wrt. this command specifically is described a bit strangely. We do have this feature flag defined:

> `VIRTIO_NET_F_CTRL_MAC_ADDR(23)` Set MAC address through control channel.

My interpretation then is that:
* `VIRTIO_NET_CTRL_MAC_TABLE_SET` requires `VIRTIO_NET_F_CTRL_{VQ | RX}`
* `VIRTIO_NET_CTRL_MAC_ADDR_SET` requires `VIRTIO_NET_F_CTRL_{VQ | RX | MAC_ADDR}`

So then we can offer `VIRTIO_NET_F_CTRL_RX` without accepting `MAC_ADDR_SET`.

The Linux virtio_net.c imposes the same requirement, as far as I can see. [[1](https://github.com/torvalds/linux/blob/bc35965f6940a9bf834d54187b6088b8eb09206d/drivers/net/virtio_net.c#L7075-L7087), [2](https://github.com/torvalds/linux/blob/bc35965f6940a9bf834d54187b6088b8eb09206d/drivers/net/virtio_net.c#L3636-L3644)] As does FreeBSD, after converting to `VTNET_FLAG_CTRL_MAC` [[3](https://github.com/freebsd/freebsd-src/blob/e3767e149b6d8674d9637999bc4153e6fe1a82b7/sys/dev/virtio/network/if_vtnet.c#L4361-L4366)]. 

But the spec text (at least as of v1.4) only mentions the flag in a roundabout way, through the above description in §5.1.3 and then only in the 'Legacy Interface' subsections to say that older devices may assume they can just do a write to the config-space MAC rather than negotiating this feature. ...I wonder if they would take a PR/patch to clarify that?

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Yeah, I think I agree in principle. I was asking for us to stick with #1125's behaviour in [this comment](https://github.com/oxidecomputer/propolis/pull/1093#discussion_r3897387615), if a little imprecisely. But I agree just outright calling for a device reset is simpler still, given that the operation should be practically infallible for the inputs we use. That _might_ need slightly better signposting around why we call `VionaHdl::set_promisc` during setup, versus `PciVirtioViona::set_promisc`.

~~I've realised looking at this again that calling `set_needs_reset` as-is is a bug though -- we already hold `self.state.lock`, so we need to expose `set_needs_reset_locked` and go via that instead. :/~~ EDIT: they're different objects, don't mind me.

PR #1101, opened 2026-04-02
[vsock] cleanup half-closed connections

papertigers:
There are two scenarios under which connections are considering "closing" and we need to make sure that we handle them.

Scenario 1: The host sends a SHUTDOWN packet with `VsockPacketFlags::VIRTIO_VSOCK_SHUTDOWN_F_SEND | VsockPacketFlags::VIRTIO_VSOCK_SHUTDOWN_F_RECEIVE`, under normal circumstances the guest is supposed to ack this with an eventual RST packet. If we don't see the SHUTDOWN in a timely manor we send our own RST and cleanup the connection.

Scenario 2: The guest sends us a SHUTDOWN with `VsockPacketFlags::VIRTIO_VSOCK_SHUTDOWN_F_SEND | VsockPacketFlags::VIRTIO_VSOCK_SHUTDOWN_F_RECEIVE`, however the internal ring buffer that is flushing to the underlying socket has not yet finished draining all data. After a reasonable amount of time we need to send a RST to ack the SHUTDOWN and we need to cleanup the tracked connection state.

### DIFF
--- a/lib/propolis/src/vsock/poller.rs
+++ b/lib/propolis/src/vsock/poller.rs
@@ -606,6 +606,11 @@ impl VsockPoller {
 
             match conn.socket.read(&mut read_buf[..max_read]) {
                 Ok(0) => {
+                    // TODO (propolis#1102):
+                    // This is an overly aggressive shutdown. Typically EOF
+                    // signals that the conenction has been closed, however one
+                    // can intentionally shutdown read|write halves
+                    // independently.
                     let packet = VsockPacket::new_shutdown(
                         VsockGuestAddr::from_conn_key(*guest_cid, key),
                         VsockPacketFlags::VIRTIO_VSOCK_SHUTDOWN_F_SEND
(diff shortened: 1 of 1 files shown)


lib/propolis/src/vsock/poller.rs
```diff
@@ -669,6 +687,21 @@ impl VsockPoller {
         self.rx.push_back(RxEvent::Reset(key));
     }
 
+    fn quiesce_connections(&mut self) {
+        #[allow(unstable_name_collisions)]
```
hawkw: um, what the heck is "unstable name collisions"? is this bad?

lib/propolis/src/vsock/poller.rs
```diff
@@ -669,6 +687,21 @@ impl VsockPoller {
...
+        while let Some(conn) = self.quiescing.pop_front_if(|conn| {
+            conn.started.elapsed() > DEFAULT_QUIESCE_TIMEOUT
+        }) {
+            // It's possible that the guest sent us a RST for the connection
+            // since we put it on the quiesce queue.
+            if let Some(_) = self.connections.remove(&conn.key) {
```
hawkw: i think clippy may whine that this should be
```suggestion
            if self.connections.remove(&conn.key).is_some() {
```

lib/propolis/src/vsock/poller.rs
```diff
@@ -669,6 +687,21 @@ impl VsockPoller {
         self.rx.push_back(RxEvent::Reset(key));
     }
 
+    fn quiesce_connections(&mut self) {
+        #[allow(unstable_name_collisions)]
+        while let Some(conn) = self.quiescing.pop_front_if(|conn| {
+            conn.started.elapsed() > DEFAULT_QUIESCE_TIMEOUT
+        }) {
```
hawkw: so, IIUC, there _shouldn't_ be any situation in which an entry that is _not_ at the head of the queue could have an elapsed quiesce timeout even if the head of the queue doesn't? But, it might be worth noting that explicitly, since it is representable here.

hawkw (commented):
This looks reasonable to me! I had some smallish suggestions, but no major issues.

lib/propolis/src/vsock/poller.rs
```diff
@@ -588,17 +608,17 @@ impl VsockPoller {
...
                         VsockGuestAddr::from_conn_key(*guest_cid, key),
                         VsockPacketFlags::VIRTIO_VSOCK_SHUTDOWN_F_SEND
                             | VsockPacketFlags::VIRTIO_VSOCK_SHUTDOWN_F_RECEIVE,
                         conn.fwd_cnt(),
                     );
                     permit.write(&packet.header, &packet.data);
+                    self.quiescing.push_back(ClosingConn {
+                        key,
+                        started: Instant::now(),
+                    });
```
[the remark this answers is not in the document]
papertigers: Discussed  this further over chat and added a comment in [bf80e3b](https://github.com/oxidecomputer/propolis/pull/1101/commits/bf80e3b7e6c64a463d4ed93c71881f624833aaa4)

lib/propolis/src/vsock/poller.rs
papertigers: I left an additional comment at the call site.

iximeow (approved):
🫡 🙏

Issue #990, opened 2025-12-17
nvme: support deallocating blocks


iximeow:
there are a few ways to deallocate a block in an NVMe device, which the NVM Command Set Specification talks a bit about:

> A logical block may be marked deallocated as the result of:
> * a Dataset Management command (refer to section 3.2.3); or
> * a Write Zeroes command addressing the logical block (refer to section 3.2.3.2.1); or
> * a sanitize operation.

in the limit it would be nice to plumb this through to Crucible and the OS for local volumes, as we're unaware of the filesystem-specifics happening in the guest's volume. but a starting point would be for the emulated NVMe controller to at least support whatever the common mechanism is that OSes use to deallocate blocks - I assume Dataset Management with AD set is more common than Write Zeroes with DEAC, but I haven't looked around. probably comes with setting the bits and limits across `ONCS`/`WZSL`/`DM{RS,SL,RSL}` as appropriate. I doubt Sanitize is too interesting right now.

NVM Command Set Specification also talks about the semantics of deallocated blocks in 3.2.3.2.1 `Deallocated or Unwritten Logical Blocks`:

> The value read from a deallocated logical block shall be deterministic; specifically, the value returned by subsequent reads of that logical block shall be the same until a write operation occurs to that logical block.

over in namespace features, `DLFEAT` describes support for Write Zeroes' deallocate bit and the behavior of deallocated logical blocks. It seems like we could reasonably deallocate blocks in general by setting `DLFEAT=000b` and ignoring deallocate for now, or `DLFEAT=001b` and write zeroes on the deallocated block. plumbing this through to Crucible seems a little more interesting, and plumbing this through to file backends (specifically for local volumes) has the interesting consequence of exposing the guest to the hardware's value for `DLFEAT`. in the local volume case we're already not thinking about migration, so maybe that's OK.

in theory this could help keep with performance and wear-leveling especially when local volumes are getting used, but I haven't gotten anywhere near testing any of that. a no-op `Dataset Management` might make sense even without numbers just so adding support for the behavior later isn't as much of a guest surprise.

ahrens:
fixed by https://github.com/oxidecomputer/propolis/pull/1105

PR #1105, opened 2026-04-08
support NVMe Deallocate

ahrens:
This implements https://github.com/oxidecomputer/propolis/issues/990.  It makes propolis advertise support for the "Dataset Management" NVMe command.  It uses `ioctl(DKIOCFREE)` to pass these requests through to local disks (FileBackend).  The requests are ignored on distributed disks (Crucible).

Note that our devices are NVME 1.0e, which specifies that the disk "may" deallocate all provided ranges, and "shall return all zeros, all ones, or the last data written to the associated LBA".  The 1.0e spec has no mechanism for telling the guest which of these semantics is actually happening.  Future work may include migrating to NVME 1.1 or later, which can use the DRB (Deallocation Read Behavior) field to tell the guest whether the blocks are actually zeroed or not.

This requires the changes for https://github.com/oxidecomputer/stlouis/issues/940 which are under review [here](https://code.oxide.computer/c/illumos-gate/+/719).

A few details to be aware of or provide input on:

1. I am always advertising support for the Dataset Management command (with a bit in the `oncs` field).  I think this is OK since there is no live migration currently (even if the VM has no local disks).  So each time a VM boots, it will be on a specific version of Propolis which either advertises Dataset Management, or not.
2. `block::Operation::Discard` now means to discard multiple ranges from the client-provided list, not just one range.  `probes::block_begin_discard` used to take an offset and length, but I have changed it to take the number of ranges instead.  Is this OK? Are there consumers that need to change?  We could fire a probe for each range, but then there would be multiple “begin” probes for one devqid, which could be confusing because begin/complete probes would not match up.  Similar for `probes::nvme_discard_enqueue`.
3. In `VirtualDiskStats`, should we add stats for Discard?  How do we monitor these?  Would we need to add support in consumers?

### DIFF
--- a/bin/propolis-server/src/lib/stats/virtual_disk.rs
+++ b/bin/propolis-server/src/lib/stats/virtual_disk.rs
@@ -77,9 +77,9 @@ impl VirtualDiskStats {
                 self.on_write_completion(result, len, duration)
             }
             Operation::Flush => self.on_flush_completion(result, duration),
-            Operation::Discard(..) => {
-                // Discard is not wired up in backends we care about for now, so
-                // it can safely be ignored.
+            Operation::Discard => {
+                // Discard is now wired up for local disks. We need to add support for it to the
+                // schema in Omicron before we can report stats for it. For now, just ignore it.
             }
         }
     }
--- a/lib/propolis/src/block/crucible.rs
+++ b/lib/propolis/src/block/crucible.rs
@@ -146,9 +146,11 @@ impl WorkerState {
                     let _ = block.flush(None).await?;
                 }
             }
-            block::Operation::Discard(..) => {
-                // Crucible does not support discard operations for now
-                return Err(Error::Unsupported);
+            block::Operation::Discard => {
+                // Crucible does not support discard operations for now, so we implement this as
+                // a no-op (which technically is a valid implementation of discard, just one that
+                // doesn't actually free any space).
+                return Ok(());
             }
         }
         Ok(())
--- a/lib/propolis/src/block/file.rs
+++ b/lib/propolis/src/block/file.rs
@@ -12,6 +12,7 @@ use std::sync::{Arc, Mutex};
 use crate::block::{self, SyncWorkerCtx, WorkerId};
 use crate::tasks::ThreadGroup;
 use crate::vmm::{MappingExt, MemCtx};
+use slog::warn;
 
 use anyhow::Context;
 
@@ -31,6 +32,7 @@ struct SharedState {
 
     info: block::DeviceInfo,
     skip_flush: bool,
+    log: slog::Logger,
 }
 struct WceState {
     initial: bool,
@@ -43,6 +45,7 @@ impl SharedState {
         skip_flush: bool,
         wce_state: Option<WceState>,
         discard_mech: Option<dkioc::DiscardMech>,
+        log: slog::Logger,
     ) -> Arc<Self> {
         let state = SharedState {
             fp,
@@ -50,6 +53,7 @@ impl SharedState {
             discard_mech,
             skip_flush,
             info,
+            log,
         };
 
         // Attempt to enable write caching if underlying resource supports it
@@ -113,12 +117,29 @@ impl SharedState {
                     self.fp.sync_data().map_err(|_| "io error")?;
                 }
             }
-            block::Operation::Discard(off, len) => {
+            block::Operation::Discard => {
                 if let Some(mech) = self.discard_mech {
-                    dkioc::do_discard(&self.fp, mech, off as u64, len as u64)
-                        .map_err(|_| {
-                        "io error while attempting to free block(s)"
-                    })?;
+                    for &(off, len) in &req.ranges {
+                        // There might be some performance benefits to combining the ranges into
+                        // one DKIOCFREE call, but ZFS will only issue one range to the
+                        // underlying disk at a time, so we expect the benefit to be minimal in
+                        // practice.
+                        if let Err(e) = dkioc::do_discard(
+                            &self.fp, mech, off as u64, len as u64,
+                        ) {
+                            if e.kind() == ErrorKind::Unsupported {
+                                // If the discard mechanism is unsupported, we should not have
+                                // advertised support for discard in the first place.  However, if
+                                // this happens, it likely means we're running on older ZFS bits that
+                                // don't support DKIOCFREE on raw zvols.  Since this is not a supported
+                                // configuration, but developer machines might be in this state, we
+                                // swallow errors from the ioctl rather than failing the command.
+                                warn!(self.log, "discard at offset {off} length {len} is unsupported; check ZFS version");
+                            } else {
+                                return Err("io error while attempting to free block(s)");
+                            }
+                        }
+                    }
                 } else {
                     unreachable!("handled above in processing_loop()");
                 }
@@ -159,6 +180,7 @@ impl FileBackend {
         path: impl AsRef<Path>,
         opts: block::BackendOpts,
         worker_count: NonZeroUsize,
+        log: slog::Logger,
     ) -> Result<Arc<Self>> {
         let p: &Path = path.as_ref();
 
@@ -202,6 +224,7 @@ impl FileBackend {
                 skip_flush,
                 wce_state,
                 disk_info.discard_mech,
+                log,
             ),
             block_attach,
             worker_count,
--- a/lib/propolis/src/block/mod.rs
+++ b/lib/propolis/src/block/mod.rs
@@ -50,7 +50,7 @@ mod probes {
     fn block_begin_read(devq_id: u64, req_id: u64, offset: u64, len: u64) {}
     fn block_begin_write(devq_id: u64, req_id: u64, offset: u64, len: u64) {}
     fn block_begin_flush(devq_id: u64, req_id: u64) {}
-    fn block_begin_discard(devq_id: u64, req_id: u64, offset: u64, len: u64) {}
+    fn block_begin_discard(devq_id: u64, req_id: u64, nr: u64) {}
 
     fn block_complete_read(
         devq_id: u64,
@@ -106,8 +106,8 @@ pub enum Operation {
     Write(ByteOffset, ByteLen),
     /// Flush buffer(s)
     Flush,
-    /// Discard/UNMAP/deallocate region
-    Discard(ByteOffset, ByteLen),
+    /// Discard/UNMAP/deallocate some ranges, which are specified in Request::ranges
+    Discard,
 }
 impl Operation {
     pub const fn is_read(&self) -> bool {
@@ -120,7 +120,7 @@ impl Operation {
         matches!(self, Operation::Flush)
     }
     pub const fn is_discard(&self) -> bool {
-        matches!(self, Operation::Discard(..))
+        matches!(self, Operation::Discard)
     }
 }
 
@@ -203,6 +203,10 @@ pub struct Request {
     /// A list of regions of guest memory to read/write into as part of the I/O
     /// request
     pub regions: Vec<GuestRegion>,
+
+    /// A list of byte ranges to discard as part of the I/O request.  This is only
+    /// relevant for discard operations, and is expected to be empty otherwise.
+    pub ranges: Vec<(ByteOffset, ByteLen)>,
 }
 impl Request {
     pub fn new_read(
@@ -210,7 +214,7 @@ impl Request {
         len: ByteLen,
         regions: Vec<GuestRegion>,
     ) -> Self {
-        Self { op: Operation::Read(off, len), regions }
+        Self { op: Operation::Read(off, len), regions, ranges: Vec::new() }
     }
 
     pub fn new_write(
@@ -218,17 +222,17 @@ impl Request {
         len: ByteLen,
         regions: Vec<GuestRegion>,
     ) -> Self {
-        Self { op: Operation::Write(off, len), regions }
+        Self { op: Operation::Write(off, len), regions, ranges: Vec::new() }
     }
 
     pub fn new_flush() -> Self {
         let op = Operation::Flush;
-        Self { op, regions: Vec::new() }
+        Self { op, regions: Vec::new(), ranges: Vec::new() }
     }
 
-    pub fn new_discard(off: ByteOffset, len: ByteLen) -> Self {
-        let op = Operation::Discard(off, len);
-        Self { op, regions: Vec::new() }
+    pub fn new_discard(ranges: Vec<(ByteOffset, ByteLen)>) -> Self {
+        let op = Operation::Discard;
+        Self { op, regions: Vec::new(), ranges }
     }
 
     pub fn mappings<'a>(&self, mem: &'a MemCtx) -> Option<Vec<SubMapping<'a>>> {
@@ -239,7 +243,7 @@ impl Request {
             Operation::Write(..) => {
                 self.regions.iter().map(|r| mem.readable_region(r)).collect()
             }
-            Operation::Flush | Operation::Discard(..) => None,
+            Operation::Flush | Operation::Discard => None,
         }
     }
 }
--- a/lib/propolis/src/hw/nvme/bits.rs
+++ b/lib/propolis/src/hw/nvme/bits.rs
@@ -4,6 +4,7 @@
 
 #![allow(dead_code)]
 
+use crate::block::{ByteLen, ByteOffset};
 use bitstruct::bitstruct;
 use zerocopy::{FromBytes, IntoBytes};
 
@@ -173,6 +174,50 @@ impl CompletionQueueEntry {
             true => self.status_phase |= 0b1,
             false => self.status_phase &= !0b1,
         }
+    }
+}
+
+/// A Dataset Management Range Definition as represented in memory.
+///
+/// See NVMe 1.0e Section 6.6 Figure 114: Dataset Management – Range Definition
+#[derive(Debug, Default, Copy, Clone, PartialEq, Eq, FromBytes, IntoBytes)]
+#[repr(C, packed(1))]
+pub struct DatasetManagementRangeDefinition {
+    /// The context attributes specified for each range provides information about how the range
+    /// is intended to be used by host software. The use of this information is optional and the
+    /// controller is not required to perform any specific action.
+    pub context_attributes: u32,
+
+    pub number_logical_blocks: u32,
+
+    pub starting_lba: u64,
+}
+impl DatasetManagementRangeDefinition {
+    pub fn new(
+        context_attributes: u32,
+        number_logical_blocks: u32,
+        starting_lba: u64,
+    ) -> Self {
+        Self { context_attributes, number_logical_blocks, starting_lba }
+    }
+
+    pub fn offset_len(
+        &self,
+        lba_data_size: u64,
+    ) -> Result<(ByteOffset, ByteLen), &'static str> {
+        // Check for overflow in the byte offset calculation
+        let byte_offset = self.starting_lba.checked_mul(lba_data_size).ok_or(
+            "Starting LBA and LBA data size multiplication overflowed",
+        )?;
+        // Check for overflow in the byte length calculation
+        let byte_len = (u64::from(self.number_logical_blocks))
+                .checked_mul(lba_data_size)
+                .ok_or("Number of logical blocks and LBA data size multiplication overflowed")?;
+        // Check for overflow of offset + length
+        byte_offset
+            .checked_add(byte_len)
+            .ok_or("Byte offset and byte length addition overflowed")?;
+        Ok((byte_offset as ByteOffset, byte_len as ByteLen))
     }
 }
 
@@ -539,6 +584,8 @@ pub const NVM_OPC_FLUSH: u8 = 0x00;
 pub const NVM_OPC_WRITE: u8 = 0x01;
 /// Read Command Opcode
 pub const NVM_OPC_READ: u8 = 0x02;
+/// Dataset Mangement Command Opcode
+pub const NVM_OPC_DATASET_MANAGEMENT: u8 = 0x09;
 
 // Generic Command Status values
 // See NVMe 1.0e Section 4.5.1.2.1, Figure 17 Status Code - Generic Command Status Values
@@ -1195,5 +1242,6 @@ mod test {
         assert_eq!(size_of::<IdentifyController>(), 4096);
         assert_eq!(size_of::<LbaFormat>(), 4);
         assert_eq!(size_of::<IdentifyNamespace>(), 4096);
+        assert_eq!(size_of::<DatasetManagementRangeDefinition>(), 16);
     }
 }
--- a/lib/propolis/src/hw/nvme/cmds.rs
+++ b/lib/propolis/src/hw/nvme/cmds.rs
@@ -2,7 +2,10 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use super::bits::{self, StatusCodeType, SubmissionQueueEntry};
+use super::bits::{
+    self, DatasetManagementRangeDefinition, StatusCodeType,
+    SubmissionQueueEntry,
+};
 use super::queue::{QueueCreateErr, QueueId};
 use crate::block;
 use crate::common::*;
@@ -28,6 +31,10 @@ pub enum ParseErr {
     /// An invalid value was specified in the FUSE bits of `CDW0`.
     #[error("reserved FUSE value specified")]
     ReservedFuse,
+
+    /// A reserved field was set to a non-zero value.
+    #[error("reserved field value specified")]
+    Reserved,
 }
 
 /// A parsed Admin Command
@@ -678,6 +685,8 @@ pub enum NvmCmd {
     Write(WriteCmd),
     /// Read data and metadata
     Read(ReadCmd),
+    /// Dataset Management Command
+    DatasetManagement(DatasetManagementCmd),
     /// An unknown NVM command
     Unknown(GuestData<SubmissionQueueEntry>),
 }
@@ -709,6 +718,22 @@ impl NvmCmd {
                 prp1: raw.prp1,
                 prp2: raw.prp2,
             }),
+            bits::NVM_OPC_DATASET_MANAGEMENT => {
+                if (raw.cdw11 & !0b111) != 0 {
+                    // Only the lowest 3 bits of CDW11 are used for Dataset
+                    // Management, so reject if any other bits are set.
+                    return Err(ParseErr::Reserved);
+                }
+                NvmCmd::DatasetManagement(DatasetManagementCmd {
+                    prp1: raw.prp1,
+                    prp2: raw.prp2,
+                    // Convert from 0's based value
+                    nr: (raw.cdw10 & 0xFF) as u16 + 1,
+                    ad: raw.cdw11 & (1 << 2) != 0,
+                    _idw: raw.cdw11 & (1 << 1) != 0,
+                    _idr: raw.cdw11 & (1 << 0) != 0,
+                })
+            }
             _ => NvmCmd::Unknown(raw),
         };
         Ok(cmd)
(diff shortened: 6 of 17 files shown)


lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -779,6 +795,95 @@ impl ReadCmd {
...
+    pub nr: u16,
+
+    /// Attribute – Deallocate (AD)
+    ///
+    /// If set to ‘1’ then the NVM subsystem may deallocate all provided ranges. If a read occurs
+    /// to a deallocated range, the NVM Express subsystem shall return all zeros, all ones, or
+    /// the last data written to the associated LBA.
+    ///
+    /// Note: The operation of the Deallocate function is similar to the ATA DATA SET MANAGEMENT
+    /// with Trim feature described in ACS-2 and SCSI UNMAP command described in SBC-3.
```
iximeow: super appreciate the note here, thanks!

lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -779,6 +795,95 @@ impl ReadCmd {
...
+pub struct DatasetManagementCmd {
+    /// PRP Entry 1 (PRP1)
+    ///
+    /// Indicates a data buffer that contains the LBA range information.
+    prp1: u64,
+
+    /// PRP Entry 2 (PRP2)
+    ///
+    /// This field contains the second PRP entry that specifies the location where data should be
+    /// transferred from (if there is a physical discontinuity).
```
iximeow: fwiw "... where data should be transferred from" seems like this was originally taken from `WriteCmd` maybe? but is kind of stale since PRP2 is where we would read any range definitions that didn't fit in the region described by PRP1.

also a nit (sorry), `if there is a physical discontinuity` is overly-precise since PRP2 could legitimately be whatever page is after PRP1. imagine a mildly-evil command with `prp1 = 0xff0, prp2 = 0x1000, nr = 3`: we'd read one range description at 0xff0 from PRP1, then two more descriptions at 0x1000 because we'd reached the end of PRP1 and continued on at PRP2.

iximeow (commented):
nice, this will be nice to have, thanks for working on this. hopefully none of the comments are too surprising!

lib/propolis/src/hw/nvme/requests.rs
```diff
@@ -142,6 +145,37 @@ impl block::DeviceQueue for NvmeBlockQueue {
...
+                    }
+                    probes::nvme_discard_enqueue!(|| (
+                        sq.devq_id(),
+                        idx,
+                        cid,
+                        cmd.nr,
+                    ));
+                    let ranges: Vec<_> = cmd
+                        .ranges(&mem)
+                        .map(|r| r.offset_len(params.lba_data_size))
```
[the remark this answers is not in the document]
ahrens: No, `zvol_ioctl()` ignores any (part of the) range that's past the end of the zvol.  This applies to both normal and raw zvols.

lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -779,6 +795,95 @@ impl ReadCmd {
...
+                    mapping.len()
+                        / size_of::<DatasetManagementRangeDefinition>(),
+                    Default::default,
+                );
+                if mapping.read_many(&mut ranges).is_err() {
+                    ranges.clear();
+                }
+            };
+
+            ranges.into_iter()
```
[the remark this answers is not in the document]
ahrens: Yeah, filling in the default values and then possibly clearing them away does seem a bit janky.  Are you thinking the signature of the new helper would be:
```rust
    /// Read the entire mapping as an array of `T` objects.
    /// The size of the mapping must be aligned to `size_of::<T>()`.
    pub fn read_many_owned<T: Copy + FromBytes>(&self) -> Result<Vec<T>> {
 ```
With an implementation similar to `read_many()`?

File: lib/propolis/src/block/file.rs [hunk 5]
```diff
@@ -113,12 +113,16 @@ impl SharedState {
...
                 }
             }
-            block::Operation::Discard(off, len) => {
+            block::Operation::Discard => {
                 if let Some(mech) = self.discard_mech {
-                    dkioc::do_discard(&self.fp, mech, off as u64, len as u64)
+                    for &(off, len) in &req.ranges {
+                        dkioc::do_discard(
+                            &self.fp, mech, off as u64, len as u64,
+                        )
```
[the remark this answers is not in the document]
iximeow: > I don't think those fields are present in NVMe 1.0e

oh, sorry about that. I was going off of the mention in "NVM Command Set Specification, Revision 1.0e" but that clearly does not pair with the corresponding "NVM Express Base Specification" version..

lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -709,6 +714,17 @@ impl NvmCmd {
...
             }),
+            bits::NVM_OPC_DATASET_MANAGEMENT => {
+                NvmCmd::DatasetManagement(DatasetManagementCmd {
+                    prp1: raw.prp1,
+                    prp2: raw.prp2,
+                    // Convert from 0's based value
+                    nr: (raw.cdw10 & 0xFF) as u16 + 1,
+                    ad: raw.cdw11 & (1 << 2) != 0,
+                    idw: raw.cdw11 & (1 << 1) != 0,
+                    idr: raw.cdw11 & (1 << 0) != 0,
```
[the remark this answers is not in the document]
iximeow: I was thinking mostly just that the guest should see an error completion with `STS_INVAL_FIELD`, rather than specifics on `ParseErr`.. but it looks like we ignore the specific error variant of `NvmCmd::parse` and `AdminCmd::parse` today, so that's unfortunate! A `ParseErr;:Reserved` for "Invalid Reserved Field" seems like the right balance?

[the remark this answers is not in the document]
ahrens: I'll add code to return `STS_INVAL_FIELD` for `ReservedFuse` and `Reserved`, let me know if it looks OK.

lib/propolis/src/block/crucible.rs
```diff
@@ -145,7 +145,7 @@ impl WorkerState {
                     let _ = block.flush(None).await?;
                 }
             }
-            block::Operation::Discard(..) => {
+            block::Operation::Discard => {
                 // Crucible does not support discard operations for now
                 return Err(Error::Unsupported);
```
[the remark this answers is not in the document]
ahrens: sounds good, I will change this to return `OK(())`.

lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -779,6 +795,95 @@ impl ReadCmd {
...
+        self.data(mem).flat_map(|region| {
+            let mut ranges = Vec::new();
+            if let Some(mapping) = mem.readable_region(&region) {
+                ranges.resize_with(
+                    mapping.len()
+                        / size_of::<DatasetManagementRangeDefinition>(),
+                    Default::default,
+                );
+                if mapping.read_many(&mut ranges).is_err() {
+                    ranges.clear();
```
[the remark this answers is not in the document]
ahrens: That makes sense.  However, I'm having trouble implementing the `Iterator<Item = Result<DatasetManagementRangeDefinition, SomeNewErrorType>>`.  Here are a few other options, let me know what you think:

Same semantics as current, but using the new helper makes it much more compact.  (The new helper checks the alignment so that error is caught as well.) Same problems with having the caller check if it got them all:
```rust
    pub fn ranges<'a>(
        &self,
        mem: &'a MemCtx,
    ) -> impl Iterator<Item = DatasetManagementRangeDefinition> + 'a {
        self.data(mem).flat_map(|region| {
            mem.readable_region(&region)
                .map(|mapping| mapping.read_many_owned().unwrap_or_default())
                .unwrap_or_default()
                .into_iter()
        })
    }
```

Do the length check for the caller, returning a `Result<Vec<DatasetManagementRangeDefinition>>`.  It copies the ranges more, but the performance impact is probably inconsequential:

```rust
    pub fn ranges2<'a>(
        &self,
        mem: &'a MemCtx,
    ) -> std::io::Result<Vec<DatasetManagementRangeDefinition>> {
        let vec: Vec<DatasetManagementRangeDefinition> = self
            .data(mem)
            .flat_map(|region| {
                mem.readable_region(&region)
                    .map(|mapping| {
                        mapping.read_many_owned().unwrap_or_default()
                    })
                    .unwrap_or_default()
                    .into_iter()
            })
            .collect();
        if vec.len() != usize::from(self.nr) {
            Err(std::io::Error::new(
                std::io::ErrorKind::InvalidInput,
                "Could not read all LBA ranges from guest memory",
            ))
        } else {
            Ok(vec)
        }
    }
```

Implemented procedurally, without using the new helper function.  Less copying of the range list.  Note that the more granular error strings here are not really useful, as they all get squashed to `STS_DATA_XFER_ERR`:

```rust
    pub fn ranges3<'a>(
        &self,
        mem: &'a MemCtx,
    ) -> std::io::Result<Vec<DatasetManagementRangeDefinition>> {
        let mut vec = Vec::with_capacity(self.nr.into());

        for region in self.data(mem) {
            let mapping =
                mem.readable_region(&region).ok_or(std::io::Error::new(
                    std::io::ErrorKind::InvalidInput,
                    "Invalid pointer to guest memory",
                ))?;
            let def_size = size_of::<DatasetManagementRangeDefinition>();
            if mapping.len() % def_size != 0 {
                return Err(std::io::Error::new(std::io::ErrorKind::InvalidInput,
                    "LBA range data buffer length is not a multiple of range definition size"));
            }
            let count = mapping.len() / def_size;
            mapping.read_many(&mut vec.spare_capacity_mut()[..count])?;
            // SAFETY: We just initialized `count` elements of `vec` with valid data, so it's safe to
            // set the length to include those elements.
            unsafe { vec.set_len(vec.len() + count) };
        }
        if vec.len() != usize::from(self.nr) {
            return Err(std::io::Error::new(
                std::io::ErrorKind::InvalidInput,
                "Could not read all LBA ranges from guest memory",
            ));
        }
        Ok(vec)
    }
```

lib/propolis/src/block/crucible.rs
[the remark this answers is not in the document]
jmpesp: My reading of this:

> Note that our devices are NVME 1.0e, which specifies that the disk "may" deallocate all provided ranges, and "shall return all zeros, all ones, or the last data written to the associated LBA".

agrees that Crucible can ignore the operation: the disk "may" deallocate (Crucible can ignore), and shall return the last data written to the associated LBA (so nothing changes). Ignoring seems fine here?

lib/propolis/src/hw/nvme/cmds.rs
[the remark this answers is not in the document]
ahrens: Iterated with ixi and came up with:
```rust
    pub fn ranges<'a>(
        &self,
        mem: &'a MemCtx,
    ) -> impl Iterator<
        Item = Result<DatasetManagementRangeDefinition, &'static str>,
    > + 'a {
        self.data(mem).flat_map(|region| {
            if let Some(Ok(defs)) = mem
                .readable_region(&region)
                .map(|mapping| mapping.read_many_owned())
            {
                defs.into_iter().map(Ok).collect::<Vec<_>>().into_iter()
            } else {
                vec![Err("Failed to read LBA range")].into_iter()
            }
        })
    }
```

lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -779,6 +795,95 @@ impl ReadCmd {
...
+        &self,
+        mem: &'a MemCtx,
+    ) -> impl Iterator<Item = DatasetManagementRangeDefinition> + 'a {
+        self.data(mem).flat_map(|region| {
+            let mut ranges = Vec::new();
+            if let Some(mapping) = mem.readable_region(&region) {
+                ranges.resize_with(
+                    mapping.len()
+                        / size_of::<DatasetManagementRangeDefinition>(),
+                    Default::default,
```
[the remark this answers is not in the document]
ahrens: fixed with the new `read_many_owned()` helper function, which enforces alignment.

lib/propolis/src/hw/nvme/requests.rs
```diff
@@ -142,6 +145,37 @@ impl block::DeviceQueue for NvmeBlockQueue {
                         Request::new_read(off as usize, size as usize, bufs);
                     return Some((req, permit, None));
                 }
+                Ok(NvmCmd::DatasetManagement(cmd)) => {
+                    // We only support the "deallocate" (discard/trim) operation of Dataset
+                    // Management.  XXX should we return success?
```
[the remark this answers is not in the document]
ahrens: OK, I'll keep Invalid Field here.

iximeow (approved):
FWIW I'm a solid +1 on this now, thanks! if you're going to add a test about `DatasetManagementCmd` with bogus `prp1/prp2` then I'll be happy to take a look at that here too. if not (or you have surprise issues), this is a nice improvement as-is.

lib/propolis/src/hw/nvme/bits.rs
```diff
@@ -176,6 +177,38 @@ impl CompletionQueueEntry {
...
+        number_logical_blocks: u32,
+        starting_lba: u64,
+    ) -> Self {
+        Self { context_attributes, number_logical_blocks, starting_lba }
+    }
+
+    pub fn offset_len(&self, lba_data_size: u64) -> (ByteOffset, ByteLen) {
+        (
+            (self.starting_lba * lba_data_size) as ByteOffset,
+            (self.number_logical_blocks as u64 * lba_data_size) as ByteLen,
```
rmustacc: Can we add a comment about why it's okay to not check for overflow on user controlled data here and simply cast with `as`? I would have probably expected this to result in a failure at some point. Maybe this would fail with `Invalid Field in Command` or maybe NVMe controllers are defined to do wrap around at disk size or something similar?

If we think it should fail, we should probably check what other NVMe devices do.

lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -779,6 +804,93 @@ impl ReadCmd {
...
+    ///
+    /// Note: this field is advisory, and we ignore it.
+    _idr: bool,
+}
+
+impl DatasetManagementCmd {
+    /// Returns an Iterator that yields [`GuestRegion`]'s which contain the array of LBA ranges.
+    pub fn data<'a>(&self, mem: &'a MemCtx) -> PrpIter<'a> {
+        PrpIter::new(
+            u64::from(self.nr)
```
rmustacc, line 862: Maybe it's worth a comment that the truncating `as` here is safe because `self.nr` is required to be a u8 and a u8 * 16 will generally not overflow a u64. It may be clearer to users if we don't have the ::from, though that may be required.

lib/propolis/src/hw/nvme/cmds.rs
```diff
@@ -779,6 +804,93 @@ impl ReadCmd {
...
+        Item = Result<DatasetManagementRangeDefinition, &'static str>,
+    > + 'a {
+        self.data(mem).flat_map(|region| {
+            if let Some(Ok(defs)) = mem
+                .readable_region(&region)
+                .map(|mapping| mapping.read_many_owned())
+            {
+                defs.into_iter().map(Ok).collect::<Vec<_>>().into_iter()
+            } else {
+                vec![Err("Failed to read LBA range")].into_iter()
```
rmustacc, line 885: For my own understanding, how do these and backend errors get translated into NVMe errors?

lib/propolis/src/block/mod.rs
```diff
@@ -203,32 +203,36 @@ pub struct Request {
...
+    pub ranges: Vec<(ByteOffset, ByteLen)>,
 }
 impl Request {
     pub fn new_read(
         off: ByteOffset,
         len: ByteLen,
         regions: Vec<GuestRegion>,
     ) -> Self {
-        Self { op: Operation::Read(off, len), regions }
+        Self { op: Operation::Read(off, len), regions, ranges: Vec::new() }
```
rmustacc, line 217: Do we want to be adding a new allocation to every read/write/flush here? Seems like this is a better use for an Option maybe to indicate that there's nothing there. Maybe not really appropriate for this change. @iximeow thoughts?

lib/propolis/src/block/file.rs
```diff
@@ -113,12 +113,16 @@ impl SharedState {
                     self.fp.sync_data().map_err(|_| "io error")?;
                 }
             }
-            block::Operation::Discard(off, len) => {
+            block::Operation::Discard => {
                 if let Some(mech) = self.discard_mech {
-                    dkioc::do_discard(&self.fp, mech, off as u64, len as u64)
+                    for &(off, len) in &req.ranges {
+                        dkioc::do_discard(
```
rmustacc, line 123: Can we get a comment about why we're not using the ioctl's ability to handle multiple ranges in a single go? I'm guessing because ZFS is going to chunk it up, but if this is a bunch of basically 4k ranges from an FS, that means a lot of ioctls when the drive would prefer to get more commands.

lib/propolis/src/block/file.rs
```diff
@@ -113,12 +113,16 @@ impl SharedState {
                     self.fp.sync_data().map_err(|_| "io error")?;
                 }
             }
-            block::Operation::Discard(off, len) => {
+            block::Operation::Discard => {
                 if let Some(mech) = self.discard_mech {
-                    dkioc::do_discard(&self.fp, mech, off as u64, len as u64)
+                    for &(off, len) in &req.ranges {
```
rmustacc, line 118: Who is checking the offset/length for overflow and that it's actually within the virtual device? Is that being left to the kernel?

bin/propolis-server/src/lib/stats/virtual_disk.rs
```diff
@@ -77,9 +77,9 @@ impl VirtualDiskStats {
                 self.on_write_completion(result, len, duration)
             }
             Operation::Flush => self.on_flush_completion(result, duration),
-            Operation::Discard(..) => {
-                // Discard is not wired up in backends we care about for now, so
-                // it can safely be ignored.
+            Operation::Discard => {
+                // Discard is only partially wired up in backends we care about, and it's not
+                // clear what stats (if any) to report yet.
```
rmustacc: Since these are real now, I think we should probably consider adding the following to match https://github.com/oxidecomputer/omicron/blob/main/oximeter/oximeter/schema/virtual-disk.toml:

* discards
* failed-discards
* add this to io-latency and io-size and make sure this is a new io-kind of operation

rmustacc (commented):
Thanks for the work on this.

lib/propolis/src/block/mod.rs
iximeow: just a `Vec::new()` (and `Vec::with_capacity(0)`, iirc) won't allocate on their own, so yeah this is fine IMO. we're growing the size of `struct Request` by three usize to hold the Vec itself, but that's all in the typical read/write/flush case.

bin/propolis-server/src/lib/stats/virtual_disk.rs
iximeow: since the schema lives in Omicron, I think we'd want to land this, update the schema in Omicron, update the Omicron ref here in Propolis, then add the metrics. sound right to you too?

bnaecker: @iximeow That's right! LMK if you need any help getting that sorted.

lib/propolis/src/hw/nvme/cmds.rs
iximeow: they end up being slightly different, since the backend has the more generic `block` types in case the storage medium is behind say a virtio-block device.

for the NVMe command here, the caller collects these iterators into a `Result<Vec<DatasetManagementRangeDefinition>, _>` when translating to a `Request` in [hw/nvme/requests.rs](https://github.com/oxidecomputer/propolis/pull/1105/changes/BASE..0d67440e3ac9e3e2ba9f2dc82a30be7258ffbda9#diff-0cec6003ffb5e762ff0a8af9c12290fa0a2a2ac0e5ecfe02903862a4ad6fcaacR166-R177). that iteration stops at the first returned `Err`, returning that for the whole `try_collect()`. so an invalid range here turns into the NVMe `Completion::generic_err(bits::STS_DATA_XFER_ERR)` when the caller handles the error.

backend errors are more distant and just an opaque [Success or Failure](https://github.com/oxidecomputer/propolis/blob/0d67440e3ac9e3e2ba9f2dc82a30be7258ffbda9/lib/propolis/src/block/file.rs#L77-L80) which end up at [`DeviceRequest::complete`](https://github.com/oxidecomputer/propolis/blob/master/lib/propolis/src/block/minder.rs#L86-L93) -> `QueueMinder::complete` -> `self.complete_req_fn` there -> nvme's [`DeviceQueue::complete`](https://github.com/oxidecomputer/propolis/blob/bc489ddf0f38f75e0c194b86cf6f0de377f68845/lib/propolis/src/hw/nvme/requests.rs#L163-L187). at that last step we go through `impl From<block::Result> for Completion` to get an NVMe completion to actually write back to the guest.

lib/propolis/src/hw/nvme/bits.rs
iximeow: I imagine we'd see `LBA Out of Range` from an actual device? just from the NVMe spec that seems like the most relevant error for the offset/length being out of range. rather than the OS seeing the invalid range, returning an error, and that eventually being a Data Transfer Error? but I'm not sure what other devices do either.

lib/propolis/src/block/mod.rs
hawkw: Yeah, an empty `Vec` will only allocate if it's pushed to. 

I do somewhat wonder if we ought to be replacing the vecs of ranges with a bitmap or some such but that seems out of scope.

lib/propolis/src/hw/nvme/cmds.rs
ahrens: I think what's happening here is:
1. `self.nr` (type u16, value 256 or less) is converted to a u64 with `u64::from()`, which is lossless
2. `size_of::<DatasetManagementRangeDefinition>()` (type usize, value 16) is converted to a u64 with `as u64`.  In the most general case, a usize could be larger than u64; is that the case in any compilation environment we care about?
3. the two above u64's are multiplied together and passed to `PrpIter::new()`

The multiplication of two u64's can in general overflow, but in this case the maximum value is 256*16=4,096  which doesn't overflow.  I'll add a comment explaining that.

Side note: In a project I worked on previously, we made a helper function that would convert a usize to a u64, with a compile time check that this is lossless.  (We also used the [cast_possible_truncation](https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation) lint to disallow using `as` this way.)  Is there something similar in Propolis, or would it be worth adding that?

lib/propolis/src/block/file.rs
ahrens: Yeah it's true that this will result in more ioctls, but zfs will split it into one operation per range.  I think we could/should pass all the ranges in a single ioctl if that would substantially improve performance.  I can measure that if you'd like, is there any particular use case that I should look at?

lib/propolis/src/hw/nvme/cmds.rs
rmustacc: > size_of::<DatasetManagementRangeDefinition>() (type usize, value 16) is converted to a u64 with as u64. In the most general case, a usize could be larger than u64; is that the case in any compilation environment we care about?

No, I don't think so. I don't see us ever being on a non-LP64 platform. Given this basically relies on bhyve and other practical archs (ARMv8+, RV64) are going to basically have a `usize` and `u64` as the same in the ABI. I think @iximeow had something that basically forced that to be the case.

> Side note: In a project I worked on previously, we made a helper function that would convert a usize to a u64, with a compile time check that this is lossless. (We also used the [cast_possible_truncation](https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation) lint to disallow using as this way.) Is there something similar in Propolis, or would it be worth adding that?

Yeah, probably? I think @iximeow has something similar here that'd probably be worth considering.

lib/propolis/src/block/file.rs
rmustacc: If ZFS is going to split it into one underlying disk operation per range then I'm not sure there's much value in doing something different here. A comment to future us is probably worthwhile. I don't think we have enough to know about the different general guest FS patterns to say we'd be getting enough deallocates in the same request to an underlying device to allow it to better optimize internal FTL rewrites.

lib/propolis/src/block/file.rs
ahrens: `off` and `len` are type `usize`; can `as u64` truncate it on any systems we run on?

As for checking that it's actually within the virtual device, zfs will check and ignore any (part of the) range that's outside the zvol.

lib/propolis/src/hw/nvme/cmds.rs
iximeow: yeah, I'd mentioned to Robert elsewhere that I kind of want to pull [these helpers](https://codeberg.org/iximeow/asmlinator/src/branch/no-gods-no-/src/x86_64.rs#L14-L29) I'd written elsewhere into Propolis since we know we're only going to care about building this on 64-bit targets. then we can go shoo out all the `as u64`/`as usize` where it's really not as surprising to the reader as it could seem. I've started writing up a patch that moves a bunch of the storage-related stuff from `usize` to `u64` first though (it will conflict with these changes. that's fine. you should go first :) )

for here a comment would be great, I'll have a separate patch for those in a bit (and I can tag you for review if you want @ahrens :grin: )

lib/propolis/src/block/file.rs
rmustacc: `off` and `len` don't inherently. However, `off + len` certainly can overflow and represent things outside of the device. I don't think we should silently ignore invalid block ranges unless we have evidence that normal devices do. I'm surprised that ZFS would just ignore that and not error on that to be honest.

rmustacc: To be a bit more specific here, I think if someone gives us an invalid range that should cause an error and we shouldn't just acknowledge it. I'm not sure how hardware handles it and what the expected semantics should be, but this should maybe be all validated up front so we can distinguish it from a generic errno value. Overloaded non-semantic errnos for these aren't great, but that's long been the pattern of the dkio interfaces.

lib/propolis/src/block/file.rs
ahrens: Sounds good, I'll add a comment here.

lib/propolis/src/block/file.rs
ahrens: Good point about `off + len`.  I don't see anywhere that we calculate that in Propolis.  It is calculated in ZFS, and it doesn't look like overflow will be handled gracefully.  This seems like a longstanding bug, I'll investigate that and see about fixing that in Illumos.

The ZFS behavior of ignoring ranges outside of the zvol has always been there (since support for DKIOCFREE was [introduced](https://github.com/illumos/illumos-gate/commit/b77b9231da168bb31490f65bf2697f6031b7f601) 15 years ago), which makes me a little concerned about changing it upstream in illumos.  However, it looks like most if not all of the other implementers of DKIOCFREE use the `dfl_iter()` helper function which will return EINVAL if any part of any range is past the end of the device (as well as returning EOVERFLOW if any off+len wraps around the uint64_t).  So I think it could make sense to change the ZFS behavior, possibly by making it also use `dfl_iter()`.  Do you have thoughts on if we should do that upstream in illumos vs just for stlouis?

bin/propolis-server/src/lib/stats/virtual_disk.rs
ahrens: OK, sounds like there are no additional code changes needed for this commit.  I'll adjust the comment to indicate that we do intend to add the stat, and file an issue in Omicron.

lib/propolis/src/hw/nvme/bits.rs
ahrens: I agree the client could easily cause `starting_lba * lba_data_size` to overflow and then zero out unintended blocks.  I'll change this to return an error in that case, as well as if `number_logical_blocks * lba_data_size` overflows, or if `off+len` overflows.

I think "Invalid Field in Command" (`STS_INVAL_FIELD`) could make sense since this is an overflow regardless of the size of the device.  i.e. these field values don't specify any LBA; they are never valid, regardless of device state.  But I can also see that "LBA Out of Range" might be a better clue towards the LBA fields, for someone diagnosing why they get the error.

lib/propolis/src/block/file.rs
rmustacc: I don't think we want to have a long-term ZFS divergence in stlouis. If we are making changes to ZFS in stlouis, my general perspective is that's just because we want to soak it before we do it upstream. If we're concerned about it in ZFS, then I think we should start by implementing the check in Propolis and then we can come back to what makes sense in ZFS to account for the long history here.

ahrens: Sounds good.  
* The overflow (wrapping around u64) is now checked in Propolis in the `offset_len()` helper.  
* I'll work on a change to illumos to have ZFS return an error when trying to DKIOCFREE a range that extends past the end of the zvol (also covering the case where there's overflow).

bin/propolis-server/src/lib/stats/virtual_disk.rs
ahrens: Filed https://github.com/oxidecomputer/omicron/issues/10356

lib/propolis/src/block/file.rs
ahrens: Filed https://www.illumos.org/issues/18093 and https://github.com/oxidecomputer/stlouis/issues/970

PR #1114, opened 2026-04-13
Add project and silo ids to VM attestation

augustuswm:
This relies on [vm-attest#68](https://github.com/oxidecomputer/vm-attest/pull/68) to add support for including the project and silo ids in the VM instance configuration data provided by attestations. These PRs are meant as discussion points, and not intended to be merged yet. This change allows upstream callers to make richer policy decisions as project and silo ids are likely to be much more stable than VM ids. For instance if we are rolling deployments we may bring up multiple VMs and move IPs between them to roll a release forwards or backwards. It would be a lot of coordination to propagate that value to all of the policy deciders during a deploy when what we would really want is to specify a policy for all VMs in a project (or possibly with a tag in a future world).

There is not much interesting going on outside of piping the ids through. That said, it makes the API for `prepare_instance_conf` pretty nasty as we are passing three opaque Uuids that rely on their order to be correct. Happy to change this to something like a builder or some other pattern to avoid errors.

The primary focus of this PR though is to discuss if this the correct path to passing the data to the attestation. Notably we do **not** provide any access to sled metadata so as to split this conversion from concerns about exposing placement data.

### DIFF
--- a/lib/propolis/src/attestation/server.rs
+++ b/lib/propolis/src/attestation/server.rs
@@ -67,7 +67,14 @@ impl AttestationSockInit {
     /// Do any any remaining work of collecting VM RoT measurements in support
     /// of this VM's attestation server.
     pub async fn run(self) {
-        let AttestationSockInit { log, vm_conf_send, uuid, project, silo, boot_backend_ref } = self;
+        let AttestationSockInit {
+            log,
+            vm_conf_send,
+            uuid,
+            project,
+            silo,
+            boot_backend_ref,
+        } = self;
 
         let mut vm_conf = vm_attest::VmInstanceConf {
             uuid,
--- a/lib/propolis/src/attestation/server.rs
+++ b/lib/propolis/src/attestation/server.rs
@@ -57,9 +57,7 @@ enum AttestationInitState {
 pub struct AttestationSockInit {
     log: slog::Logger,
     vm_conf_send: oneshot::Sender<VmInstanceConf>,
-    uuid: uuid::Uuid,
-    project: uuid::Uuid,
-    silo: uuid::Uuid,
+    vm_instance_conf: vm_attest::VmInstanceConf,
     boot_backend_ref: Option<boot_digest::Backend>,
 }
 
@@ -70,18 +68,9 @@ impl AttestationSockInit {
         let AttestationSockInit {
             log,
             vm_conf_send,
-            uuid,
-            project,
-            silo,
+            mut vm_instance_conf,
             boot_backend_ref,
         } = self;
-
-        let mut vm_conf = vm_attest::VmInstanceConf {
-            uuid,
-            project,
-            silo,
-            boot_digest: None,
-        };
 
         if let Some(digest_backend) = boot_backend_ref {
             let boot_digest = match crate::attestation::boot_digest::compute(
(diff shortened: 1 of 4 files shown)


flihp:
:100: agree w/ policy element here. In your case you're looking for something akin to a group membership instead of a unique identifier for the instance. These are the only notion of "group" that we have currently but, as you note, a generic `tag` would be nice as a way to keep us from overloading the notion of `project` and `silo` too badly. Generally this seems like a reasonable evolution to me. I also agree that this makes some interfaces a bit precarious.

`prepare_instance_conf` takes all of the elements in the `VmInstanceConf` structure as parameters + the crucible volume that it uses to calculate the `boot_digest`. May be cleaner to instantiate & populate the `VmInstanceConf` structure around where we call `prepare_instance_conf` and just pass the `VmInstanceConf` through w/ the crucible volume. The `VmInstanceConf` passed to `prepare_instance_conf` would need to have the `boot_digest` set to `None` and `prepare_instance_conf` would eventually cause this to be set to `Some(digest)` if it can. This makes `prepare_instance_conf` a bit of a misnomer as it's not preparing the `VmInstanceConf` but instead just the `boot_digest` (which was all it did anyway i think?).

iximeow:
> May be cleaner to instantiate & populate the `VmInstanceConf` structure around where we call `prepare_instance_conf` and just pass the `VmInstanceConf` through w/ the crucible volume

totally agree that setting up an initial `VmInstanceConf` and passing that along to get "finalized" would make the UUIDs less confusion-prone and probably automagically Do The Right Thing when we add more later.

> This makes `prepare_instance_conf` a bit of a misnomer as it's not preparing the `VmInstanceConf` but instead just the `boot_digest` (which was all it did anyway i think?).

I'm pretty sure we just picked the name `prepare_...` because it's where it seemed we'd put any measurement of the guest VM that isn't statically derived from the instance spec. but yeah the only thing we're doing right now is measuring the boot digest. kinda getting ahead of ourselves on the naming!

augustuswm:
Thanks! I'll take another pass here and cleanup that interface.

augustuswm:
Turning this into a real PR now. I refactored the interface and renamed `prepare_instance_conf`, but please free to provide guidance there. I am still a bit distant from understanding the actual moving bits here.

augustuswm:
Updated dependency version pins and rebased against main.

lib/propolis/src/attestation/server.rs
```diff
@@ -271,9 +273,9 @@ impl AttestationSock {
         Ok(())
     }
 
-    pub fn prepare_instance_conf(
+    pub fn prepare_init_state(
```
iximeow: not to quibble on names too much but the idea was that this is where we can kick off whatever async work needs to happen to go from the initial derivable-from-instance-spec information to the actual conf we'd use for attestation.

maybe `measure_instance` is a better name? `init state` is kind of just an implementation detail tracking in getting the instance information through to the attestation side of things.

iximeow (approved):
I'd missed that the dice-util side of things is landed now, sorry about that - one naming thought which I'm not deeply wedded to..

merge if you want, if you leave the name as-is I'll probably take a swing at it later.

lib/propolis/src/attestation/server.rs
augustuswm: Happy to rename this, I definitely do not know much of the larger context here and have no attachment to it :)

augustuswm: Updated this fn name and fixed the related failure message so that it matches. If we are good otherwise I'll get this merged tomorrow.

PR #1118, opened 2026-04-17
wait for viona Poller to run before declaring device running

iximeow:
When `Lifecycle::run()` completes, the device should be fully running. `PciVirtioViona` did not uphold this, as `run()` would return after having asked the poller to start, but before confirming it *was* started. If the device was paused shortly after, the poller may have never actually started. `running` would not be set, and "pause" would immediately succeed only for the Poller to begin running shortly after!

Worse, the device may be halted and the underlying link connected to the viona fd may have been destroyed, all while the poller was starting. At this point the `AsyncFd::with_interest()` will fail with ENXIO and panic the poller task.

This is another "never happens in practice" bug. But in unit tests that rapidly set up a viona device, configure it, and tear it down, this is quite consistently hit.

Working backwards from the ENXIO, I eventually ended up at the following to diagnose what specifically was causing the ENXIO:
```
> cat debug.d
fbt::viona_chpoll:entry {
  self->interested = 1;
  printf("chpoll");
}

fbt::ddi_get_soft_state:return / self->interested / {
  printf("ss->link: %p", ((viona_soft_state_t*)arg1)->ss_link);
}
```

(note this currently merges into #1117 but is kinda independent except that I noticed it as part of writing up #1117)

### DIFF
--- a/lib/propolis/src/hw/virtio/viona.rs
+++ b/lib/propolis/src/hw/virtio/viona.rs
@@ -668,7 +668,10 @@ impl PciVirtioViona {
     fn poller_start(&self) {
         let mut inner = self.inner.lock().unwrap();
         let poller = inner.poller.as_mut().expect("poller should be spawned");
+        let wait_state = poller.state.clone();
         let _ = poller.sender.send(TargetState::Run);
+        drop(inner);
+        wait_state.wait_running();
     }
     fn poller_stop(&self, should_exit: bool) {
         let mut inner = self.inner.lock().unwrap();
@@ -689,7 +692,7 @@ impl PciVirtioViona {
     // Transition the emulation to a "running" state, either at initial start-up
     // or resumption from a "paused" state.
     fn run(&self) {
-        self.poller_start();
+        tokio::task::block_in_place(|| self.poller_start());
         if self.queues_restart().is_err() {
             self.virtio_state.set_needs_reset(self);
             self.notify_port_update(None);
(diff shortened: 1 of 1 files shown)


lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -689,7 +692,7 @@ impl PciVirtioViona {
     // Transition the emulation to a "running" state, either at initial start-up
     // or resumption from a "paused" state.
     fn run(&self) {
-        self.poller_start();
+        tokio::task::block_in_place(|| self.poller_start());
```
iximeow, line 695: sorry for more tokio crimes @hawkw 

`poller_start` now sleeps on a condition variable, so we'll block for however long on whatever runtime thread we were setting up this device on. seems right to `block_in_place()` waiting for this to happen instead of doing that.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2208,7 +2237,7 @@ mod test {
     #[test]
     #[cfg_attr(not(target_os = "illumos"), ignore)]
     fn run_viona_tests() {
-        let rt = tokio::runtime::Builder::new_current_thread()
+        let rt = tokio::runtime::Builder::new_multi_thread()
```
iximeow, line 2240: the other side of it: you can't `block_in_place` on a single-threaded runtime, so make this multi-threaded like server and standalone.

the real problem is that (as is kinda implicit in `create_test_ctx()/start()` we have to be able to create the device, import state, and _then_ set it running, so we can't just wait for the poller to be running out of the box. this is the least annoying way I see of stringing it up.

lib/propolis/src/hw/virtio/viona.rs
papertigers: At first I was like, wait this is not an async fn, but then I realized `start_vm` is. I think this is probably okay until we potentially fully utilize the `Indicator` stuff.  This made me think about the vsock stuff where we use a barrier for start and a oneshot channel for pause which technically also block the runtime.

In fact the viona `Lifecycle::pause` is going to end up in `wait_state.wait_stopped();` and we don't have a `block_in_place` there.

papertigers (approved):

PR #1120, opened 2026-04-18
Run viona unit tests in CI

iximeow:
something of a journey but we got there - leaving the commits for humor value but this all gets squashed for the merge of course. and tests pass! yay.

### DIFF
--- a/.github/buildomat/jobs/phd-build.sh
+++ b/.github/buildomat/jobs/phd-build.sh
@@ -17,6 +17,11 @@
 #: series = "phd_build"
 #: name = "propolis-server.sha256.txt"
 #: from_output = "/out/propolis-server-debug.sha256.txt"
+#:
+#: [[publish]]
+#: series = "propolis_tests"
+#: name = "propolis-tests.tar.gz"
+#: from_output = "/out/propolis-tests-debug.tar.gz"
 
 set -o errexit
 set -o pipefail
@@ -30,24 +35,13 @@ rustc --version
 banner prerequisites
 ptime -m ./tools/install_builder_prerequisites.sh -y
 
-# Before we commit to building propolis-server or PHD, there are a handful of
-# tests that are illumos-specific and not currently run in other test jobs. The
-# whole suite doesn't take much time so we'll just run all of them here..
-banner test-propolis
-
-# Get $TEST_FEATURES defined for below.
-. .github/buildomat/propolis-vars.sh
-
-# Set up an etherstub to use for VNICs in virtio-nic tests. We might want the
-# tests to run on a real link one day to do actual networking, but we don't need
-# that yet!
-pfexec dladm create-etherstub prop_viona_test0
-VIONA_TEST_NIC=prop_viona_test0 pfexec ptime -m \
-	cargo test --verbose --features "$TEST_FEATURES"
-
 # Build the Propolis server binary with 'dev' profile to enable assertions that
 # should fire during tests.
 banner build-propolis
+
+# We'll do a few cargo builds, keeping features the same means we reuse build
+# artifacts from crates these configure.
+TEST_FEATURES="omicron-build,failure-injection"
 
 # Compile propolis-server so that it allows development features to be used even
 # though the `omicron-build` feature is enabled. This should be a relatively
@@ -55,6 +49,16 @@ banner build-propolis
 # features, above.
 ptime -m cargo build --verbose -p propolis-server \
 	--features "$TEST_FEATURES"
+
+# Build Propolis test binaries, but they won't run here. The path to get here
+# is unfortunate:
+# * we don't have `git` on a Gimlet target.
+# * or `pkg`.
+# * `uname -m` is "oxide", which confuses rustup too.
+#
+# Setting this up by hand is "possible", but it's much easier to just build the
+# test binaries here and squirrel them off to *run* on a Gimlet. So do that.
+ptime -m cargo build --tests --verbose --features "$TEST_FEATURES"
 
 # The PHD runner requires unwind-on-panic to catch certain test failures, so
 # build it with the 'dev' profile which is so configured.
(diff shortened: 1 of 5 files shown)


.github/buildomat/jobs/phd-build.sh
```diff
@@ -34,11 +39,29 @@ ptime -m ./tools/install_builder_prerequisites.sh -y
...
+# Build Propolis test binaries, but they won't run here. The path to get here
+# is unfortunate:
+# * we don't have `git` on a Gimlet target.
+# * or `pkg`.
+# * `uname -m` is "oxide", which confuses rustup too.
+# * some tests need a VMM, but this job is probably run inside a VM.
+#
+# Doing the build on a bare gimlet host is "possible", but it's much easier to
+# just build the test binaries here and squirrel them off to *run* on a Gimlet.
+# So do that.
```
papertigers, line 63:
```suggestion
# just build the test binaries here and squirrel them off to *run* on a Gimlet,
# so do that.
```

papertigers: nit but fine either way

.github/buildomat/jobs/phd-build.sh
```diff
@@ -53,13 +76,17 @@ tar -czvf target/debug/phd-runner.tar.gz \
...
+	$(find target/debug/deps/ -perm -111 -type f -name 'propolis-*')
+
 banner copy
 pfexec mkdir -p $outdir
 pfexec chown "$UID" $outdir
 cp .github/buildomat/phd-run-with-args.sh $outdir/phd-run-with-args.sh
 mv target/debug/propolis-server-debug.tar.gz \
 	$outdir/propolis-server-debug.tar.gz
 mv target/debug/phd-runner.tar.gz $outdir/phd-runner.tar.gz
+mv target/debug/propolis-tests-debug.tar.gz $outdir/propolis-tests-debug.tar.gz
```
papertigers, line 89: nit: probably fine since we already don't quote `"$outdir"` anywhere else, but hey safety first

.github/buildomat/jobs/test-gimlet.sh
```diff
@@ -0,0 +1,51 @@
...
+#: variety = "basic"
+#: target = "lab-2.0-gimlet"
+#: rust_toolchain = false
+#: skip_clone = true
+#:
+#: [dependencies.phd-build]
+#: job = "phd-build"
+#
+# That buildomat frontmatter is kinda absurd. We're going to run Propolis
+# tests, but not build them? And `lab-2.0-gimlet` specifically?
```
papertigers, line 13: I am not sure if this comment was a note to yourself or if it is supposed to inform future readers.

.github/buildomat/jobs/test-gimlet.sh
```diff
@@ -0,0 +1,51 @@
...
+
+banner test-propolis
+
+# Set up an etherstub to use for VNICs in virtio-nic tests. We might want the
+# tests to run on a real link one day to do actual networking, but we don't need
+# that yet!
+pfexec dladm create-etherstub prop_viona_test0
+
+for testbin in ./target/debug/deps/propolis-*; do
+	VIONA_TEST_NIC=prop_viona_test0 pfexec ptime -m "$testbin"
```
papertigers: I suppose `prop_viona_test0` could become a variable, but okay with how it is considering it's used in 2 places.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2109,6 +2109,20 @@ mod test {
                     "Skipping viona tests as env does not have VIONA_TEST_NIC. \
                     Set this environment variable to an existing link that \
                     Propolis viona tests should create test vnics on.");
+                let uname = nix::sys::utsname::uname().unwrap();
+                if uname.machine() != std::ffi::OsStr::new("i86pc") {
+                    // If the tests are running on i86pc, this might be a dev
+                    // host that does not actually want us messing with devices
```
papertigers, line 2115: This comment mentions not wanting to mess with devices on a dev host, but I don't see any logic that determines that. Is that okay?

iximeow: the `uname.machine() != "i86pc"` kind of _is_ the logic, in that if you're on an illumos i86pc system I assume that's a dev machine (which does mean a future illumos `uname.machine() == aarch64"` dev system would be misunderstood, but I'm.. not sweating that for now)

it's possible that on a dev machine you _do_ want these tests to run, which is why this is all in handling the case of not having a test NIC set up.

I should probably have written the comment `Since the ...`, but I really meant to say "we're in the `true` arm, so it is the case that .. and so ...". I'll adjust the comment accordingly.

.github/buildomat/jobs/phd-build.sh
iximeow: yeah since `outdir` is fully under our control I'll leave this one as-is in this patch.

.github/buildomat/jobs/test-gimlet.sh
iximeow: I meant this as a lead-in to the comment below it, so definitely for future readers. is there a wording change that would help?

lib/propolis/src/hw/virtio/viona.rs
iximeow: adjusted wording in https://github.com/oxidecomputer/propolis/pull/1120/commits/cb99c4cf17f0035c4454c4fe6e2cb15a97ff4ba4

.github/buildomat/jobs/test-gimlet.sh
iximeow: went ahead and did that in https://github.com/oxidecomputer/propolis/pull/1120/commits/cb99c4cf17f0035c4454c4fe6e2cb15a97ff4ba4, `VIONA_TEST_NIC="$TEST_DEV"` is a bit funky but I'm keeping that as having a magic `VIONA_TEST_NIC` declared above that happens to configure the test binary feels way too magic :grin:

.github/buildomat/jobs/test-gimlet.sh
papertigers: I think just the question marks through me off - we are running the tests and the dependency is the build job.

It's not really a big deal.

papertigers (approved):
thanks for putting this together! I think the additional CI time is worth the additional test coverage.

Issue #1045, opened 2026-02-07
ubuntu 22.04 no longer likes the virtio nic


iximeow:
@leftwo spotted this on a minibar. I'd tested https://github.com/oxidecomputer/propolis/pull/1033 against Debian 13 because it's what I happened to reproduce the issue on, but that "fix" also made Ubuntu unable to drive the NIC at all (with or without a reboot).

dancrossnyc:
It's worth checking this against c-bhyve to see whether it fails in a similar way.

It's hard to imagine how the reset logic would prevent the driver from, er, driving the NIC initially, though.

iximeow:
> It's worth checking this against c-bhyve to see whether it fails in a similar way.

if you have a moment from other fires, I'd be curious to check, but with #1047 I think that's a pretty clear "Propolis was wrong, don't do that"

> It's hard to imagine how the reset logic would prevent the driver from, er, driving the NIC initially, though.

this is where I'm a bit frustrated, because _my_ 24.04 goes exactly like this. and all my test images are either:

* multiqueue aware, set up 11 queue pairs, reboot resets the whole device which gets all queues, shrinks PAIRS/USEPAIRS back down to 1 until multiqueue is negotiated again after reboot, everything's great
* not multiqueue aware, set up 1 queue pair, reboot resets the whole device which EINVAL's for the last 10 queue pairs but we're resetting the device so none of that actually matters

what happens in Alan's image, it looks like, is something before MQ negotiation pokes the reset bit in the virtio status register. it occurs to me it'd be nice to have DTrace probes around virtio device lifecycle like this, because I'm speculating from ioctls/returns and the fact that the guest does this early in boot. regardless we (incorrectly) try resetting all the rings, get everything after the first pair to `VRingState::Fatal`, and later multiqueue negotiation fails. it's not as bad as "all Linux X.Y is busted", probably, but.. jeez.

PR #1121, opened 2026-04-21
viona: multiqueue device should stay multiqueue across migration

iximeow:
We correctly export and import the propolis-side state given a multiqueue VirtIO device, but we did not communicate that through to viona. On import the virtio-nic device has only told viona it will use one queue pair; we've skipped the "normal" set_features() in favor of setting features on the handle directly. Setting an imported multi-queue device running at this point will reset the many queues, but viona rings are not in a resettable state, fail to reset, and set the device to NEEDS_RESET immediately.

Communicating the correct number of queue pairs to viona is a clear improvement, but we're not quite out of bugs yet..

### DIFF
--- a/lib/propolis/src/hw/virtio/viona.rs
+++ b/lib/propolis/src/hw/virtio/viona.rs
@@ -59,6 +59,7 @@ enum MqSetPairsCause {
     Reset = 0,
     MqEnabled = 1,
     Commanded = 2,
+    Import = 3,
 }
 
 #[usdt::provider(provider = "propolis")]
@@ -935,6 +936,25 @@ impl MigrateMulti for PciVirtioViona {
         self.hdl.set_features(feat).map_err(|e| {
             MigrateStateError::ImportFailed(format!(
                 "error while setting viona features ({feat:x}): {e:?}"
+            ))
+        })?;
+
+        if (feat & VIRTIO_NET_F_MQ) != 0 {
+            self.hdl.set_pairs(PROPOLIS_MAX_MQ_PAIRS).unwrap();
+        }
+        let queues = self.virtio_state.queues.count().get();
+        let Some(pairs) = (queues - 1).checked_div(2) else {
+            return Err(MigrateStateError::ImportFailed(format!(
+                "source queue count was not a number of pairs + 1: {queues}"
+            )));
+        };
+        probes::virtio_viona_mq_set_use_pairs!(|| (
+            MqSetPairsCause::Import as u8,
+            pairs
+        ));
+        self.hdl.set_usepairs(pairs).map_err(|e| {
+            MigrateStateError::ImportFailed(format!(
+                "error while restoring use pairs ({pairs}): {e:?}"
             ))
         })?;
 
@@ -2182,7 +2202,9 @@ mod test {
             &test_ctx.dev,
             drv_state,
         );
+        assert!(driver.status_ok());
         Lifecycle::reset(test_ctx.dev.as_ref());
+        assert!(driver.status_ok());
 
         let drv_state = driver.export();
         let test_ctx = test_ctx.migrate();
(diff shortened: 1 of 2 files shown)


iximeow:
between 70662f7 and 0b0566f I noticed an exciting issue very much like https://github.com/oxidecomputer/propolis/issues/1045: rebooting a guest would cause the PCI device to disappear. since `peak` wasn't retained across VirtQueues export/import, we'd assume any previously-initialized queues beyond `len` were never touched and not reset them. the guest then sees them as they were at import - enabled - after a "device reset" and rightfully refuses to operate the accursed device.

this would have been, if everything else was in a happy state, a bug introduced by https://github.com/oxidecomputer/propolis/pull/1047...

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1525,12 +1546,17 @@ mod test {
...
             let vnic_name = self.vnic_name.clone();
+            let underlying_nic = self.underlying_nic.clone();
             let test_name = self.test_name;
 
             std::mem::drop(acc_mem);
             std::mem::drop(self);
 
-            let new_ctx = create_test_ctx(test_name, &vnic_name);
+            delete_vnic(&vnic_name);
+            create_vnic(&underlying_nic, &vnic_name);
```
iximeow, line 1559: this is incidental but seems worth keeping: I was trying to chase out any sources of anything potentially sticking around across the "migrated" VMs, and the test vnic itself "could" stick around but really shouldn't.

the dladm commands won't (shouldn't?) block or be blocked by test operations, so I don't super love blocking the runtime like this but I'm also not worried about it..

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -937,6 +938,25 @@ impl MigrateMulti for PciVirtioViona {
...
 
+        if (feat & VIRTIO_NET_F_MQ) != 0 {
+            self.hdl.set_pairs(PROPOLIS_MAX_MQ_PAIRS).unwrap();
+        }
+        let queues = self.virtio_state.queues.count().get();
+        let Some(pairs) = (queues - 1).checked_div(2) else {
+            return Err(MigrateStateError::ImportFailed(format!(
+                "source queue count was not a number of pairs + 1: {queues}"
+            )));
+        };
```
lgfa29: I may be misreading this, but is the goal to check for values like `2 * p + 1`, where `p` is the number of queue pairs? If that's the case, we may need to verify `checked_rem` is `Some(0)` as well?

lgfa29 (approved):
Giving mostly a cerimonial ✅ since I'm not too familiar with this part of the code, but the general logic makes sense to me.

lib/propolis/src/hw/virtio/viona.rs
iximeow: that's a fair question, I'm not sure what would happen if a guest tried to set the device to use 0 tx/rx pairs.. obviously it wouldn't do useful networking, but that's maybe a legal configuration? and if so we should maintain it across migration. whatever happens, it would be good to have a test for!

I'll check out what the spec says, if anything. if viona tolerates this, then we should faithfully export/import it (and this should have a note), but if it's not allowed then.. yeah we should be more strict about it.

lgfa29: Yeah, 0 could be a problem, but reading the code I was thinking more of cases where there aren't X pairs + 1? For example, `queues = 4` (or any other even number) doesn't sound like a valid value based on the error message, but the code is only checking for overflow errors.

But I don't know how this part works, so I'm mostly going by the code and error messages. Feel free to ignore it if this doesn't make sense.

iximeow: oh. uh. yeah. I was expecting `checked_div` to return None numerator does not evenly divide by 2 but it just truncates. definitely needs a `%` or `checked_rem()` or something for that.

iximeow:
(the force push was to pick up the dropshot-api-manager bump)

PR #1125, opened 2026-04-24
viona: support (enough) `VIRTIO_NET_F_CTRL_RX` to leave all-multicast mode

### LINKED ISSUE #1122: Viona links never leave `VIONA_PROMISC_MULTI`
When `viona` creates a new link atop a mac provider, it is created with `link->l_promisc = VIONA_PROMISC_MULTI;` set so that guests which do not support the `VIRTIO_NET_F_CTRL_RX` capability can receive multicast traffic implicitly. `viona` expects that for compatible guests, the VMM will alter this via the `VNA_IOC_SET_PROMISC` ioctl as required. Propolis doesn't support this capability or unset promiscuous mode today; this results in the device doing some extra unnecessary work for every unicast packet on the receive side to verify that it is, in fact, a unicast packet.

There are some complications in advertising `VIRTIO_NET_F_CTRL_RX`. This feature covers two main pieces; the guest is allowed to explicitly request promiscuous mode (or all-multicast mode), but it can also push down arbitrary unicast/multicast filters in addition to its primary MAC. Viona itself doesn't have any plumbing to perform this filtering (or push it down to MAC, the VNIC, the NIC...). It's probably reasonable as a first pass that we treat the presence of a unicast/multicast filter table as just cause to move into one of the promiscuous modes, and the spec does suggest that devices are free to do this.

This capability does not require that we allow guests to alter their actual MAC address via the `VIRTIO_NET_F_CTRL_MAC_ADDR` capability. We might make that a toggleable feature of the interface, since OPTE relies quite heavily on guests sticking with the MAC address we give them. *Technically* a transitional/legacy driver is allowed to write straight into `NetReg::Mac` without this flag, but we ignore any writes to the config `struct` regardless.

### LINKED ISSUE #1127: Viona control queue index is incorrect
The control queue for a virtio network device can have one of two indices:

- `2`, if `VIRTIO_NET_F_MQ` is not negotiated.
- `2 * max_virtqueue_pairs`, if `VIRTIO_NET_F_MQ` is negotiated.

However, the logic in place today incorrectly assumes that the control virtqueue is the queue immediately after those the guests have asked for:

https://github.com/oxidecomputer/propolis/blob/96e772ab174bb38897c42aef6951fbe7b3e40642/lib/propolis/src/hw/virtio/queue.rs#L1047-L1060

https://github.com/oxidecomputer/propolis/blob/96e772ab174bb38897c42aef6951fbe7b3e40642/lib/propolis/src/hw/virtio/viona.rs#L431-L433

In context, `.len()` here (updated by `set_len`) refers to the number of Tx and Rx queues requested by the guest as part of `MqCmd::SetPairs`. `max_virtqueue_pairs` is in fact the data covered by `NetReg::MaxVqPairs` (e.g., which is used to derive `VirtQueues::max_capacity()`). It took me a while to find out that this is why advertising a new capability reliant on the control queue would hang the system -- nothing on queue 22 was ever being routed to the `ctl_msg` handler. 😅

FelixMcFelix:
This PR adds support for the `VIRTIO_NET_F_CTRL_RX` capability, which allows guests to request specific MAC address filters in addition to their primary MAC, and to explicitly inform the host whether the device should be shifted into promiscuous mode.

The hope is that this allows us to take a guest out of all-multicast promiscuous mode in viona itself, which generates extra per-packet work (checking "is this packet multicast?") and contention in the Tx and Rx paths as both reach for and refcount the same list of promisc callbacks on the client.

Testing in falcon makes it look as though Debian will just immediately insert multicast entries, although I don't know if that's purely an artefact of my environment. Using one of the helios-dev images appears to correctly move out of promiscuous mode and keep the filter tables empty.

Closes #1122. Closes #1127.


FelixMcFelix:
I've tested this so far on Alpine Linux 3.23.3, debian 13.2, and the helios-2.8 image we use for a4x2 testing. The first two are multiqueue-capable, whereas illumos is not -- control requests appear to be correctly served on queue 22 for the former and queue 2 for the latter.

The typical initialisation sequence looks to be that guests will send explicit promisc off signals for alongside every MAC filter update. `snoop -rd vioif0` will also have illumos explicitly move us in and out of promiscuous mode.

Running on top of OPTE and/or in omicron standalone doesn't really stop Linux from asking for multicast MACs (i.e., they're not a result of anything coming in link-local). So for most guests we will need to get explicit MAC filter support into viona proper, or they'll just put themselves back into all-multicast mode. illumos, at least, keeps the tables empty. On Debian, I'm seeing:

```
Got MAC list [
    MacAddr([33, 33, 0, 0, 0, 1]), MacAddr([1, 0, 5e, 0, 0, 1]), MacAddr([1, 80, c2, 0, 0, 0]),
    MacAddr([1, 80, c2, 0, 0, 3]), MacAddr([1, 80, c2, 0, 0, e]), MacAddr([33, 33, ff, e7, 9c, e4]),
    MacAddr([33, 33, 0, 1, 0, 3]), MacAddr([1, 0, 5e, 0, 0, fc]), MacAddr([33, 33, ff, 0, 6, da])
]
```
With visible group subscriptions on:
```
IPv6/IPv4 Group Memberships
Interface       RefCnt Group
--------------- ------ ---------------------
lo              1      224.0.0.1
enp0s6          1      224.0.0.252
enp0s6          1      224.0.0.1
lo              1      ff02::1
lo              1      ff01::1
enp0s6          1      ff02::1:ff00:6da
enp0s6          1      ff02::1:3
enp0s6          2      ff02::1:ffe7:9ce4
enp0s6          2      ff02::1
enp0s6          1      ff01::1
```
So the MAC table can be decoded as:
```
IPv6(ff01::1 [All Nodes, Node Local], ff02::1 [All Nodes, Link Local]),
IPv4(224.0.0.1 [All Systems on Subnet]),
STP/LLDP(additional),
LLDP(additional),
LLDP(primary)/PTP,
IPv6(ff02::1:ffe7:9ce4 [solicited node addr]),
IPv6(ff02::1:3 [LLMNR]),
IPv4(224.0.0.252 [LLMNR]),
IPv6(ff02::1:ff00:6da [solicited node addr])
```
`ip link set multicast off dev enp0s6` still leaves a bunch of active MACs including registered multicast addresses for all-/solicited-node, but those are sort of intrinsic to the stack. I think I'm happy at least that we're at parity with how cbhyve uses this machinery in `pci_viona_eval_promisc`.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -371,13 +572,14 @@ impl PciVirtioViona {
...
         // The vector is sized with the maximum number of rings/queues, but
-        // until the driver negotiates multiqueue, we only use the first two.
+        // until the driver negotiates multiqueue, we only use the first two
+        // for the datapath. The third will serve as the control queue if
+        // multiqueue is not negotiated, even if it is a little large for that
+        // purpose.
         let queues = VirtQueues::new_with_len(3, &queue_sizes);
-        if let Some(ctlq) = queues.get(2) {
-            ctlq.set_control();
-        }
```
FelixMcFelix, line 380: I don't know what effect putting `receiveq2` in this state would have, but we never unset this flag when the control queue index updated. So in theory this would have affected `ring_init/reset/kick/...` for this queue?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -937,7 +1328,18 @@ impl MigrateMulti for PciVirtioViona {
...
+        let mut state = self.inner.lock().unwrap();
+        state.filter = FilterState::from_bits_retain(input.filter);
+        state.unicast_mac_filters = input.unicast_mac_filters.into();
+        state.unicast_mac_filters = input.multicast_mac_filters.into();
+        self.set_promisc(input.promisc, &mut state).map_err(|_| {
+            MigrateStateError::ImportFailed(format!(
+                "Could not move device promisc level to {:?}.",
+                input.promisc
+            ))
+        })
```
FelixMcFelix, line 1378: I'm pretty unfamiliar with the propolis migration machinery, hopefully this does the job?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -177,6 +189,53 @@ pub mod control {
...
+        if !chain.read(&mut entry_count, mem) {
+            return Err(());
+        }
+
+        if entry_count == 0 {
+            return Ok(Box::new([]));
+        }
+
+        let mut space =
+            vec![MacAddr::default(); entry_count as usize].into_boxed_slice();
```
iximeow: if a guest has told us it will provide `u32::MAX` MACs we will do a truly unfortunate allocation here. it seems to me like we want a `read_many_owned()` here similar to the one that will imminently come in for `src/vmm/mem.rs` in #1105? ([for reference](https://github.com/oxidecomputer/propolis/pull/1105/changes#diff-da55090e4f645da5113d4b072b3b33b9284329afc56a081d2f22f1791709f535R560-R581))

if not a `read_many_owned` on VirtQueue here, lets please have an
```
if entry_count * mem::size_of::<MacAddr>() > chain.remain_read_bytes() {
    return Err(());
}
```
here?

(truly not pressed on which we go with in this change, but if you defer `VirtQueue:read_many_owned` please file an issue and I'll probably get to it)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -218,19 +277,79 @@ enum VRingState {
     Fatal,
 }
 
+bitflags! {
+    /// Packet receive filters requested by a virtio NIC driver.
```
iximeow, line 274: would it be right to add here something like: "filter configuration is only a binary state, and the device is a composition of those filter configurations. Note that VirtIO RX commands only operate on individual filters using consecutive integers and correspond to bit numbers here only for simplicity"?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -938,7 +1328,18 @@ impl MigrateMulti for PciVirtioViona {
             ))
         })?;
 
-        Ok(())
+        let input: migrate::VionaStateV1 = offer.take()?;
+
+        let mut state = self.inner.lock().unwrap();
+        state.filter = FilterState::from_bits_retain(input.filter);
+        state.unicast_mac_filters = input.unicast_mac_filters.into();
+        state.unicast_mac_filters = input.multicast_mac_filters.into();
```
iximeow:
```suggestion
        state.multicast_mac_filters = input.multicast_mac_filters.into();
```

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -938,7 +1328,18 @@ impl MigrateMulti for PciVirtioViona {
             ))
         })?;
 
-        Ok(())
+        let input: migrate::VionaStateV1 = offer.take()?;
+
+        let mut state = self.inner.lock().unwrap();
+        state.filter = FilterState::from_bits_retain(input.filter);
```
iximeow: definitely should be a `FilterState::from_bits()` and `MigrateStateError::ImportFailed` if we got a bit we don't know how to process in the input filter.

(a new bit would probably need to imply a `VionaStateV2` payload being _offered_ so we should really never see such an error)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1422,6 +1821,7 @@ pub(crate) mod bits {
     pub const VIRTIO_NET_CFG_SIZE: usize = 6 + 2 + 2 + 2 + 4 + 1 + 1 + 2 + 4;
 }
 use bits::*;
+use zerocopy::{FromBytes, Immutable, IntoBytes};
```
iximeow: super nit: editor get confused about the `use ...` most of the way through this file? oughtta have this up at around line 35.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -849,6 +1219,13 @@ impl Lifecycle for PciVirtioViona {
         self.set_use_pairs(1).expect("can set viona back to one queue pair");
         self.hdl.set_pairs(1).expect("can set viona back to one queue pair");
         self.virtio_state.queues.reset_peak();
+
+        let mut state = self.inner.lock().unwrap();
+        state.unicast_mac_filters = Box::new([]);
+        state.multicast_mac_filters = Box::new([]);
+        if let Err(_) = self.set_promisc(PromiscLevel::AllMulti, &mut state) {
+            eprintln!("failed to reset viona promiscuous state")
+        }
```
iximeow: is there a reason to not `.expect("...")` with this string? from looking at the viona side of things it really seems like there isn't any other link state that would prevent this from succeeding.

either way in general if we've failed to completely reset a device we should stop rather than potentially having viona in one state and the guest believing it's in another.

lib/propolis/src/hw/virtio/viona.rs
iximeow: yeah... among other things we'd entirely fumble the queue state across migration. pretty bad, thanks for noticing this.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -745,19 +1087,47 @@ impl VirtioDevice for PciVirtioViona {
...
+        } else {
+            VIRTIO_NO_MQ_CTRL_Q_INDEX
+        };
+
+        ctl_q_idx
+            .try_into()
+            .ok()
+            .and_then(|i| self.virtio_state.queues.get(i))
+            .map(|v| v.set_control())
+            .ok_or_else(|| ())?;
```
iximeow: on one hand, this is all well within its rights to be `expect`ful: the queue index should be in-range from above so `try_into` and `get` should both always succeed.

on the other.. `queues.get()` has a special-case for `max_capacity() - 1` in which case it returns `get_control()` getting us a "control" queue which we have not actually made the control queue yet? kinda odd!

leaning further into the `XXX:` in `VirtQueues::get` maybe we should have `set_len()` update some queues-internal state that says who is currently the control queue (and unsets `is_control` if there _was_ a previous one). then we avoid this whole "VirtQueues and virtio-nic have their own ideas of control queues" problem.

comparing with other VirtIO devices it seems like in the longer-run the question of "is this a control queue" needs to be answered wholly by the device impl, and that it's not really coherent as a generic property (some devices have 0 queues, some have two, most of the time they're the first/second except here, ... etc). so keeping this in `VirtQueues` is an OK step until we have other control-queue-having VirtIO devices?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -712,13 +931,135 @@ impl PciVirtioViona {
...
+        #[cfg(feature = "falcon")]
+        if state.promisc == PromiscLevel::AllVlan {
+            return PromiscLevel::AllVlan;
+        }
+
+        // We don't have an ioctl yet for viona to explicitly install a set of
+        // filters. For now, we need to apply some level of promiscuous mode
+        // to give the guest what it asks for.
+        let need_mcast = state.filter.contains(FilterState::ALL_MULTICAST)
+            || !state.multicast_mac_filters.is_empty();
```
iximeow: please bear with my rusty networking brain: if the filter has `ALL_MULTICAST | NO_MULTICAST` then do we still `need_mcast`? I'm not sure how broadcast fits in, is really the question. it seems more obvious to me that we shouldn't `need_mcast` if `ALL_MULTICAST | NO_MULTICAST | NO_BROADCAST`. in any case, having viona more promiscuous than required is not a correctness issue iiuc, so we don't need to worry about these cases?

(if that's right, imo worth a note on `fn needed_promisc` and I'm not worried about not filtering as much as silly combinations of filters could permit)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -712,13 +931,135 @@ impl PciVirtioViona {
...
+        std::mem::swap(&mut unicast, &mut state.unicast_mac_filters);
+        std::mem::swap(&mut multicast, &mut state.multicast_mac_filters);
+
+        match self.set_promisc(self.needed_promisc(&state), &mut state) {
+            Ok(()) => Ok(()),
+            Err(()) => {
+                state.unicast_mac_filters = unicast;
+                state.multicast_mac_filters = multicast;
+                Err(())
+            }
```
iximeow, line 993: I feel like I'm missing.. two things. if `set_promisc` errors, do we in general know anything about the state of viona? I think we would *want* that to result in the device having the status DEVICE_NEEDS_RESET, but right now this only results in an `Ack::Err`. I don't know if this is consistent with what cbhyve or other virtio-nic devices do.

the other thing being: shouldn't I expect to see somewhere we communicate these filters to viona? or no, and this is all fine because there's no _guarantee_ these are the only packets a guest would see, just the hope that the NIC/VNIC have prevented a guest from having to see them?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -712,13 +931,135 @@ impl PciVirtioViona {
...
+        }
+
+        let mut state = self.inner.lock().unwrap();
+
+        std::mem::swap(&mut unicast, &mut state.unicast_mac_filters);
+        std::mem::swap(&mut multicast, &mut state.multicast_mac_filters);
+
+        match self.set_promisc(self.needed_promisc(&state), &mut state) {
+            Ok(()) => Ok(()),
+            Err(()) => {
```
iximeow, line 989: if we're going to try rolling back state on error, should we try to `self.set_promisc(prev_promisc)` in the error case here? maybe failing to do _that_ is what should produce a severe enough error to NEEDS_RESET the deice?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -712,13 +931,135 @@ impl PciVirtioViona {
...
+        if level == state.promisc {
+            return Ok(());
+        }
+
+        match self.hdl.set_promisc(level) {
+            Ok(_) => {
+                state.promisc = level;
+                Ok(())
+            }
+            Err(_) => Err(()),
```
iximeow: same question on if we should set the old promisc level in the error case here. mostly looking at `viona_rx_set` in illumos and noticing we can `mac_rx_set()` and then error (in multi) or `mac_rx_clear()` and then error (in all), so `set_promisc` failing implies we've got viona in a really wonky state.

iximeow (commented):
sorry for the slow turnaround on this review! a couple of actual issues, a couple of me-trying-to-update-my-mental-model :)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -49,6 +50,13 @@ pub const fn max_num_queues() -> usize {
     PROPOLIS_MAX_MQ_PAIRS as usize * 2
 }
 
+/// The index of the control queue when multiqueue ([`VIRTIO_NET_F_MQ`]) has
+/// not been negotiated.
+///
+/// In this case, the driver will behave as though we have allocated only one
+/// Rx/Tx queue pair, followed by the control queue.
+pub const VIRTIO_NO_MQ_CTRL_Q_INDEX: usize = 2;
```
iximeow, line 59: fwiw part of the problem with picking `2` here on our own is that `VirtQueues::iter()` will still use `get_control()` to tack on a control queue at the end, so `reset_queues_locked()` will reset queues 0, 1, and propolis-VirtQueue-22 and I think leave queue 2 as it was before the reset. that's what has me thinking about keeping the notion of control-ness totally internal to `VirtQueues` for now, rather than splitting it between there and `virtio/viona.rs`.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -453,46 +649,70 @@ impl PciVirtioViona {
...
-        Err(())
+        use control::RxCmd;
+        let filter = match cmd {
+            RxCmd::Promisc => FilterState::PROMISCUOUS,
+            RxCmd::AllMulticast => FilterState::ALL_MULTICAST,
+            RxCmd::AllUnicast => FilterState::ALL_UNICAST,
+            RxCmd::NoMulticast => FilterState::NO_MULTICAST,
+            RxCmd::NoUnicast => FilterState::NO_UNICAST,
+            RxCmd::NoBroadcast => FilterState::NO_BROADCAST,
+        };
```
iximeow, line 686: awful Rust trivia: this does not optimize into the `1 << RxCmd` one might hope, it becomes a `movabs long_imm, reg; shr reg, RxCmd`. this is better w.r.t being legible source, but the compiler fully mangles the cute bitwise trick :(

FelixMcFelix: Sad, but understandable. Source legibility would be my main goal here since these operations are not frequently called.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Hmm, part of this is that a bunch of the features in `VIRTIO_NET_F_CTRL_RX_EXTRA` can't be expressed in the viona ioctl interface at all, so I didn't want them to complicate the logic there when we can only support the subset you've pointed out. I'll add a comment explaining why I'm not accounting for that block of features yet.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: I think I agree with you here, I'll change this over to mark the device as needing reset iff. we cannot rollback. cbyhve doesn't actually store the filters, and from what I can see it simply logs a warning then returns `VIRTIO_NET_CQ_ERR` (`Ack::Err`).

On the second point: yeah, we do want to do that, but viona can't handle that quite yet. I need to actually scope out what the right ioctl shape would be and have viona able to plumb it down on its MAC client. That will be a necessary part of actually making the promisc handler go away on the majority of guests. The spec does allow us to do this (virtio v1.4, §5.1.9.5.2), and it's the strategy used by cbhyve:

> The device can filter incoming packets by any number of destination MAC addresses [3]
> 
> The device SHOULD drop incoming packets which have a destination MAC which matches neither the mac
(or that set with VIRTIO_NET_CTRL_MAC_ADDR_SET) nor the MAC filtering table.
>
> [3] Since there are no guarantees, it can use a hash filter or silently switch to allmulti or promiscuous mode if it is given too many addresses.

Unfortunately 'too many addresses' is... one. But we're at least ensuring that they see that traffic! Nevermind that OPTE is unerringly strict about the MACs it will send and receive.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: I've tried to come up with something that works here. `VirtQueues` will always need additional information about what indices the control queues are going to be, otherwise we're baking in a heavily NIC-specific assumption across all devices. And it's a really weird one too, being that it needs to swap between `len() + 1` and `max_capacity() + 1`. I've set up an additional method for a device to tell `VirtQueue` what it considers as control devices -- let me know if what I have now works?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -505,28 +725,24 @@ impl PciVirtioViona {
             return Ok(());
         }
         self.hdl.set_usepairs(requested).unwrap();
-        self.virtio_state
-            .queues
-            .set_len(npairs * 2 + 1)
-            .expect("num queue pairs");
+        self.virtio_state.queues.set_len(npairs * 2).expect("num queue pairs");
```
FelixMcFelix: I'm a little torn on something here and in the concept of `len` as used between `VirtQueues` and `PciVirtioViona`, and would really like some more input. Conceptually it _is_ right that we have `2 * n_pairs + 1` active queue pairs if we've negotiated a control queue. But `VirtQueues` will treat the first `len` entries as being valid -- if we get a `set_use_pairs(2)` against a maximum `max_pairs = 11` we'll mistakenly treat queue `5` as enabled, and `iter`/`iter_all` will return queues `[0, 5] ∪ { 22 }`. So I've picked `n_pairs * 2` to at least return the right queues in the iterators.

But then both before and after this change, `queues.len()` and `queues.iter().len()` are different quantities. Is that something we care about? Do we need to document this difference, or store two counts in the VQ list?

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: The first part should be covered in [33aa6b2](https://github.com/oxidecomputer/propolis/pull/1125/commits/33aa6b27e66d68c5f1b256ba4764913734dd1e07). I'll open a stlouis ticket on the latter.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Done in [33aa6b2](https://github.com/oxidecomputer/propolis/pull/1125/commits/33aa6b27e66d68c5f1b256ba4764913734dd1e07).

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Done in [33aa6b2](https://github.com/oxidecomputer/propolis/pull/1125/commits/33aa6b27e66d68c5f1b256ba4764913734dd1e07).

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Done in [33aa6b2](https://github.com/oxidecomputer/propolis/pull/1125/commits/33aa6b27e66d68c5f1b256ba4764913734dd1e07).

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Done in [33aa6b2](https://github.com/oxidecomputer/propolis/pull/1125/commits/33aa6b27e66d68c5f1b256ba4764913734dd1e07).

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Whoops. Done in [33aa6b2](https://github.com/oxidecomputer/propolis/pull/1125/commits/33aa6b27e66d68c5f1b256ba4764913734dd1e07).

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: I've expanded this in [33aa6b2](https://github.com/oxidecomputer/propolis/pull/1125/commits/33aa6b27e66d68c5f1b256ba4764913734dd1e07), let me know if that suffices.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Done in [33aa6b2](https://github.com/oxidecomputer/propolis/pull/1125/commits/33aa6b27e66d68c5f1b256ba4764913734dd1e07).

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Done in [33aa6b2](https://github.com/oxidecomputer/propolis/pull/1125/commits/33aa6b27e66d68c5f1b256ba4764913734dd1e07).

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: I've tried to go for a mechanism where we provide a list of control queue indices separately from the main number of queues, and adapt `iter`/`iter_all` to account for the cases where our control queues might come after a gap. My thinking here is that the other device families look sorta... arbitrary in where they put their control queue(s), but viona in particular may not have a control queue, and it may be directly after the in-use queues or at the very end of the maximum. I think it's right that `PciVirtioViona` is the one to make that call, given it depends wholly on the negotiated features.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: Opened https://github.com/oxidecomputer/stlouis/issues/986.

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -326,6 +328,10 @@ impl VirtQueue {
         used.reset();
         self.live.store(false, Ordering::Release);
         self.enabled.store(false, Ordering::Release);
+
+        // We don't modify `self.is_control` here. This allows devices to avoid,
+        // e.g., attempting to plumb a later `VqChange::Reset` down to the
+        // kernel in the case of viona.
```
iximeow: I had to look at the code and think about this for a minute to realize, but this was/is really a bug in `reset_queues_locked`: we should do `dev.queue_change(.., VqChange::Reset)` before actually resetting the queue state so that if the device needs to do any per-queue teardown it still has the information available before we go and zero everything.

flipping that around means we're in the clear to reset `is_control` here as well, and I think that helps us keep a more consistent state of things in the between-reinit states

one of the cases on my mind: if we've just reset the device, but kept an `is_control` from the previous vnic configuration, a device export would pause all-but-one vring to get state out. that _should_ be fine, since if the queue was a control queue before reset then viona won't have state for it, so we're not missing anything. but it's an odd state. and we can avoid having to think about it differently from any other "export before features negotiated" case this way! yay

I'll push a commit to your branch in a second for this and a note or two below.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2233,6 +2644,7 @@ mod test {
         Lifecycle::reset(test_ctx.dev.as_ref());
         let mut driver = test_ctx.create_driver();
         driver.modern_device_init(expected_feats);
+        Lifecycle::reset(test_ctx.dev.as_ref());
```
iximeow: the absence of a `Lifecycle::reset()` here is actually important for the test: this models the case where we've had something like EDK2 or GRUB set up a virtio NIC driver, then booted into the guest (note: no PCI lifecycle change!) which set up its own driver. having a `Lifecycle::reset()` makes this more like "GRUB set something up, power-cycled the VM, then went directly into a guest that knows how to operate a multi-queue NIC".

that's not nothing, but "changing the device configuration with a power cycle" is where we've had more interesting failure (yay! the test works!)

I think you got here while debugging `ed23f4bbd` and maaaybe didn't mean to commit it? I'll back this out with my commit but shout if you do think we should keep this (or have another test with this flow)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2429,17 +2841,24 @@ mod test {
...
+                        .expect("can start viona device");
+                    let test_ctx = (test.test_fn)(test_ctx);
+                    drop(test_ctx);
+                });
                 delete_vnic(TEST_VNIC);
+                if let Err(_) = res {
+                    panic!("viona test '{}' was unsuccessful\n", test.name);
+                } else {
+                    eprintln!("viona test '{}' was successful\n", test.name);
+                }
```
iximeow, line 2904: this is a really great improvement to the humble test runner here, thank you.

lib/propolis/src/hw/virtio/viona.rs
iximeow: yes! very helpful, thank you.

lib/propolis/src/hw/virtio/viona.rs
iximeow: yeah, I think I'm pretty sold on this approach as least-fiddly from where we're at, thanks!

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -713,13 +929,149 @@ impl PciVirtioViona {
...
+
+        // Don't inflict promiscuous mode on drivers which request only their
+        // own MAC address.
+        let filter_is_self = !state.unicast_mac_filters.is_empty()
+            && state
+                .unicast_mac_filters
+                .iter()
+                .all(|mac| mac == &self.mac_addr);
+        let need_promisc =
+            state.filter.contains(FilterState::PROMISCUOUS) || !filter_is_self;
```
iximeow, line 1054: please double-check me on this one, but I noticed that `filter_is_self` was checking if the first MAC was `self.mac_addr` and that `unicast_mac_filters.is_empty() == false`, which would have meant that if a guest requested its MAC as well as another, we'd think it's `filter_is_self`.

in https://github.com/oxidecomputer/propolis/pull/1125/commits/7afb19deb9637779ffb3fc82a2605cc1dc7ad942 I've moved things around a little but more importantly it's `.iter().all()` rather than `.get(0).map(..)`.

I haven't tested this (I'm not really sure how I'd do the checks you mentioned [in this comment](https://github.com/oxidecomputer/propolis/pull/1125#issuecomment-4342601569)) so I'm working mostly on boolean reasoning: in the case of an empty list we'd have `(filter_is_self || state.unicast_mac_filters.is_empty())` be true because the former would get `bool::default() || is_empty()` for `false || true`, and that case now becomes `all(..) && !is_empty()` for `true && false`. non-empty lists are more obviously (to me) correct in this rewriting, but I was surprised to realize I couldn't think about the rewrite as "move expressions and apply De Morgan's" :)

lib/propolis/src/hw/virtio/viona.rs
iximeow: > Is that something we care about?

it's something _I_ care about, because I agree it's weird, but it's not a correctness issue anywhere that I know of.

I think this change is reasonable, but in the course of talking with you about this change I'm wondering if we can't actually have a `dyn VirtQueue` where the device is responsible for setting up queues with the appropriate operations for `queue_notify`, or `queue_change(.. VqChange::IntrCfg)`, etc. this would give a cleaner separation for the virtio/viona I/O queues vs other kinds of I/O, and lets us handle multiple different kinds of control queue. I'm not _totally_ sure how this works in the face of migration (it's not like we're exporting the vtable) but I think it's promising, and so I'm less worried about the numbers being a bit funky for now.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: No, that makes sense. I'd put that in while debugging, but also failed to understand that it was a deliberate omission there. Thanks for the explanation!

lib/propolis/src/hw/virtio/queue.rs
FelixMcFelix: Okay, that would explain why the len adjustment and `iter`/`iter_all` changes surfaced this. Thanks for catching the underlying issue.

lib/propolis/src/hw/virtio/viona.rs
FelixMcFelix: I think I'd originally skipped a `&& state.unicast_mac_filters.len() == 1`, so you're right that there's a bug here. The existence of this check is maybe a bit misleading though without an explanation -- I have never seen a guest pass down its own MAC here, so this is purely defensive programming (i.e., an empty list *does imply* `filter_is_self` because it implicitly includes the MAC in config space, `self.mac_addr`). We ideally always have an empty unicast table in that light.

(The checks in the linked  comment were mostly from shoving `eprintln!`s in various spots and inspecting stderr.)

I'll fix this up and amend the comment.

iximeow (approved):
In addition to the test improvements here, I know Kyle's given some additional looking since the reviews and I've given this a spin on Windows Server 2022 - with things looking good to both of us, seems like a good time to get this in and up to Omicron 🙏

PR #1128, opened 2026-04-27
Revert dropshot bumps, run all tests in ci

jmpesp:
Currently `cargo xtask openapi check` is failing for master. the `test_apis_up_to_date` catches this but was not running in CI. my theory is that picking up https://github.com/oxidecomputer/dropshot/pull/1554 caused openapi document generation to change, conflicting with the blessed version.

Reverting fixes things, so: downgrade dropshot to 0.16.7 and dropshot-api-manager related crates to 0.6.0, and add a CI job that runs `cargo test --locked`.


jmpesp:
Closing in favour of https://github.com/oxidecomputer/propolis/pull/1129

PR #1129, opened 2026-04-27
dropshot-api-manager 0.7.1

david-crespo:
Intended to supersede #1128, which undoes the Dropshot bump.

Dropshot 0.17 includes a change https://github.com/oxidecomputer/dropshot/pull/1554 that changes the OpenAPI schema for websocket endpoints. Normally to account for a schema change you just have to bump the API version, but dropshot-api-manager is thrown off because when it regenerates the schema for the blessed (old) version, it doesn't match what's on disk because the one on disk doesn't have the websocket change in it. This required https://github.com/oxidecomputer/dropshot-api-manager/pull/87, which tells dropshot-api-manager specifically to ignore that websocket schema change. You still have to add an API version.

Also had to bump progenitor for https://github.com/oxidecomputer/progenitor/pull/1320.


.github/workflows/rust.yml
```diff
@@ -50,6 +52,8 @@ jobs:
       run: cargo build -p propolis-mock-server --verbose
     - name: Test Libraries
       run: cargo test --lib --verbose
+    - name: Test everything
+      run: cargo test --locked
   # Build and test propolis-the-library on its own; `cargo test --lib` as used
```
david-crespo, line 62: This is copied from #1128

david-crespo:
It seems the one remaining check failure is expected.

https://github.com/oxidecomputer/propolis/blob/40dbfad52e9116e829f1b09919e8e4fb6eadc9f9/.github/buildomat/jobs/phd-run-migrate-from-base.sh#L16-L25

jmpesp:
> It seems the one remaining check failure is expected.
> 
> https://github.com/oxidecomputer/propolis/blob/40dbfad52e9116e829f1b09919e8e4fb6eadc9f9/.github/buildomat/jobs/phd-run-migrate-from-base.sh#L16-L25

In lieu of addressing this here, maybe we should open an issue and limit this PR to only adding `cargo text -- test_apis_up_to_date`?

jmpesp:
Nevermind, the Rust workflow isn't activating for PRs, disregard - though that may be its own problem to address

jmpesp:
However I think the original change from the other PR to add fetch depth isn't right. From omicron:

```
    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
      with:
        ref: ${{ github.event.pull_request.head.sha }} # see omicron#4461
        fetch-depth: 0 # git-stub-vcs needs full history
```

david-crespo:
ah, ew https://github.com/oxidecomputer/omicron/issues/4461. will update

.github/workflows/rust.yml
```diff
@@ -14,6 +14,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        # see https://github.com/oxidecomputer/omicron/issues/4461
```
jmpesp: I think you need the `with:` here

david-crespo: this is what I get for doing it by hand

jmpesp: also this needs to be under the `build-and-test` job :)

iximeow:
> Instead, it checks out a merge commit of the PR tip with the target branch as of the time the PR was created.

til and what in the world

iximeow (approved):
nice, thanks.

mote for later re. `steps.with.ref` being `github.event.pull_request.head.sha`: consensus is that we'd expect the tests to all pass regardless of the specific ref used for tests here. for the same reason as in Omicron#4461 it's better for GH Actions and buildomat to test changes the same way, so it doesn't hurt to include here too.

Issue #1126, opened 2026-04-27
`Chain::read`/`write` should require `T: FromBytes`/`IntoBytes`


FelixMcFelix:
The current trait bounds here are too lax, allowing the compiler to accept code such as:

```rust
fn ctl_test(
    chain: &mut Chain,
    mem: &MemCtx,
) -> Result<(), ()> {
    let mut a = false;
    if !chain.read(&mut a, &mem) {
        return Err(());
    }
    Ok(())
}
```

I think every use of this API is using well-formed types (so isn't liable [to hit UB](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#r-undefined.validity) on an invalid value), but we can rely on the typesystem to keep us right.

The underlying call to `MemCtx::read` already requires `FromBytes`, while `MemCtx::write` requires only `Copy`. It should probably also require `IntoBytes` for symmetry.

iximeow:
(done in #1130)

PR #1130, opened 2026-04-28
expand zerocopy IntoBytes/FromByes use in guest memory accesses

iximeow:
Forever ago (oct 2024) when I'd added FromBytes bounds for reads from guest memory in ff0b76d, I'd overlooked virtio queues and that the constraint is interesting on the write path too.

does the things @FelixMcFelix very rightfully suggested in #1126.


FelixMcFelix (approved):
Looks good, thanks for doing this!

PR #1132, opened 2026-04-28
Bump crucible rev to latest

jmpesp:
Pick up the following Crucible PRs:

- Add VolumeInfo (1928)
- Remove bonus Volume layer (1927)
- Add session and client id to panic messages (#1926)
- [crucible-agent-types] migrate to RFD 619 pattern (1899)
- Background read-only region creation (1919)
- [crucible-downstairs-repair] switch to RFD 619 pattern (1901)
- [crucible-pantry] switch to RFD 619 pattern (1900)
- Use separate in-memory types (1913)
- Remove old field from dtrace action script (1917)

Additionally, bump the server API and change the `disk_volume_status` endpoint to return the new VolumeInfo struct.

### DIFF
--- a/crates/propolis-server-api/src/lib.rs
+++ b/crates/propolis-server-api/src/lib.rs
@@ -22,6 +22,7 @@ api_versions!([
     // |  example for the next person.
     // v
     // (next_int, IDENT),
+    (5, CRUCIBLE_VOLUME_INFO),
     (4, DROPSHOT_BUMP_WEBSOCKET),
     (3, ADD_VSOCK),
     (2, PROGRAMMABLE_SMBIOS),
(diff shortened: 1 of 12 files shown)


crates/propolis-server-api/src/lib.rs
```diff
@@ -266,8 +267,21 @@ pub trait PropolisServerApi {
 
     /// Gets the status of a Crucible volume backing a disk
     #[endpoint {
+        operation_id = "disk_volume_status",
```
iximeow, line 270: oh I see, `disk_volume_status` is the automatically-chosen Id for `async fn disk_volume_status` below, and by setting it explicitly like this we're telling Dropshot that this is a different version of that API rather than a new `disk_volume_status_v1`? I'm not sure if I'm missing a detail to that, but if so then carry on 🫡

iximeow (approved):
just the one comment because I really Do Not Know What That Means, sorry for the slow turnaround here 🙏

crates/propolis-server-api/src/lib.rs
jmpesp: no you're right - this is done to maintain the operation ID for older versions

PR #1137, opened 2026-05-07
[meta] update to drift 0.1.4

sunshowers:
Drift 0.1.4 includes https://github.com/oxidecomputer/drift/pull/20 and https://github.com/oxidecomputer/drift/pull/22 -- the latter is particularly important because it can lead to false negatives where actually-incompatible APIs are marked as compatible.


iximeow (approved):
thanks for turning the crank on this. the rustfmt thing is ... a fascinating failure mode, sorry about that.

PR #1141, opened 2026-05-14
Bump crucible to latest, update Omicron, use explicit revs

jmpesp:
Pick up the following Crucible PRs:

- Use an explicit rev for oxidecomputer git deps (oxidecomputer/crucible#1936)
- Add Clone and Deserialize to VolumeInfo et al (oxidecomputer/crucible#1935)
- Update omicron/oximeter (oxidecomputer/crucible#1933)
- [meta] update to drift 0.1.4 (oxidecomputer/crucible#1932)
- Don't log if there is nothing to log (oxidecomputer/crucible#1930)

Also, similar to oxidecomputer/crucible#1936, switch the omicron related dependencies from 'branch = "main"' to an explicit rev. Previous to this commit, _two_ old versions of omicron were being pulled in: `becbbb61` and `b8efb9a0`. The first one is about 300 commits behind, and the second is about 700 commits behind. With explicit git revs, the rev being used moves to Cargo.toml, and is known without digging into the lockfile.

Related, the tokio dep had to be further specified in order to build.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,7 +151,7 @@ dependencies = [
 [[package]]
 name = "api_identity"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#b8efb9a08b366541c71eb6334b54768f3cbee724"
+source = "git+https://github.com/oxidecomputer/omicron?rev=becbbb616f5f18b59cc42e511c148734c2ba3831#becbbb616f5f18b59cc42e511c148734c2ba3831"
 dependencies = [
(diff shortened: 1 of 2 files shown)


iximeow (approved):
nice, thanks. I saw the `branch = main` Omicron deps when I tried nudging Omicron  last(?) week and was really surprised... good to have that fixed along the way.

iximeow:
(editorialized the title a bit for the future Omicron and if-anyone-skims-landed-PRs purposes)

PR #1143, opened 2026-05-18
add NVMe Discard Oximeter metrics

ahrens:
note, this depends on https://github.com/oxidecomputer/omicron/pull/10447 and https://github.com/oxidecomputer/omicron/pull/10419.  It was tested with these commits applied to a local omicron branch, and propolis's dependency on omicron patched to refer to the local branch.  I would not expect CI to pass until Omicron is updated.

Note that Crucible disks implement Discard as a no-op.  Local disks (for which Discard has an effect) currently do not report any metrics (including these new Discard metrics), which is addressed by https://github.com/oxidecomputer/propolis/pull/1144.

### DIFF
--- a/bin/propolis-server/src/lib/stats/virtual_disk.rs
+++ b/bin/propolis-server/src/lib/stats/virtual_disk.rs
@@ -26,8 +26,9 @@ use propolis::block::{self, Operation};
 
 pub use self::virtual_disk::VirtualDisk;
 use self::virtual_disk::{
-    BytesRead, BytesWritten, FailedFlushes, FailedReads, FailedWrites, Flushes,
-    IoLatency, IoSize, Reads, Writes,
+    BytesDiscarded, BytesRead, BytesWritten, Discards, FailedDiscards,
+    FailedFlushes, FailedReads, FailedWrites, Flushes, IoLatency, IoSize,
+    Reads, Writes,
 };
 
 /// Type for tracking virtual disk stats.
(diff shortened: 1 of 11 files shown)


bin/propolis-server/src/lib/stats/virtual_disk.rs
```diff
@@ -152,16 +181,18 @@ impl VirtualDiskStats {
...
 /// Number of I/O kinds we track.
-const N_IO_KINDS: usize = 3;
+const N_IO_KINDS: usize = 4;
 
 /// Indices into arrays tracking operations broken out by I/O kind.
 const READ_INDEX: usize = 0;
 const WRITE_INDEX: usize = 1;
-const FLUSH_INDEX: usize = 2;
+const DISCARD_INDEX: usize = 2;
+const FLUSH_INDEX: usize = 3; // Note that flush must be last since it does not have a size histogram (io_size)
```
bnaecker, line 190: Just curious, but why does this mean it needs to be last?

bnaecker (approved):
This looks great thanks!

bin/propolis-server/src/lib/stats/virtual_disk.rs
ahrens: These `*_INDEX` values are used to index into `VirtualDiskStats::io_size`, except for `FLUSH_INDEX`.  If I added `DISCARD_INDEX` at the end (value 3), then it would index off the end of `io_size`.

```rust
/// Number of I/O kinds we track.
const N_IO_KINDS: usize = 4;
...
struct VirtualDiskStats {
...
    /// Histogram tracking I/O sizes, by the I/O kind.
    ///
    /// Note that we have 1 fewer histogram here, since flush operations do not
    /// have a size.
    io_size: [IoSize; N_IO_KINDS - 1],
}
```

bnaecker: Ah right, we don't have an I/O size histogram for flushes. Thanks!

PR #1144, opened 2026-05-20
register Oximeter metrics for local disks

ahrens:
This will cause metrics to be reported for local disks (e.g. reads, writes, etc).

Tested by specifying a UUID in the instance spec toml file, and retrieving the metrics with `curl`, e.g.:

```toml
# Disk in slot 3
[block_dev.4e523bc0-cbaa-4b0e-8461-b7ebaa1053d4]
type = "file"
block_size = 4096
path = "/dev/zvol/rdsk/oxp_03/nocrypt/dump"
workers = 30

[dev.block3]
driver = "pci-nvme"
block_dev = "4e523bc0-cbaa-4b0e-8461-b7ebaa1053d4"
pci-path = "0.8.0"
```

```
$ curl http://0.0.0.0:37058/79c39656-34e2-4551-86a5-e94232bf6f06 | jq
...
  {
    "status": "ok",
    "info": [
      {
        "measurement": {
          "timestamp": "2026-05-20T18:00:42.931189546Z",
          "datum": {
            "type": "cumulative_u64",
            "datum": {
              "start_time": "2026-05-20T17:41:28.504981529Z",
              "value": 569 # <== number of read requests
            }
          }
        },
        "timeseries_name": "virtual_disk:reads", # <== metric name
        "timeseries_version": 1,
        "target": {
          "name": "virtual_disk",
          "fields": {
...
            "block_size": {
              "name": "block_size",
              "value": {
                "type": "u32",
                "value": 4096
              }
            },
            "disk_id": {
              "name": "disk_id",
              "value": {
                "type": "uuid",
                "value": "4e523bc0-cbaa-4b0e-8461-b7ebaa1053d4". # <=== matches block_dev UUID
              }
            },
...
          }
        },
        "metric": {
          "name": "reads",
          "fields": {}
        }
      },
```

### DIFF
--- a/bin/propolis-server/src/lib/initializer.rs
+++ b/bin/propolis-server/src/lib/initializer.rs
@@ -895,6 +895,8 @@ impl MachineInitializer<'_> {
                 }
             };
 
+            let block_size;
+            let volume_id;
             if let Some(crucible) = crucible {
                 let crucible =
                     match self.crucible_backends.entry(backend_id.clone()) {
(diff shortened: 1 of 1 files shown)


bnaecker (approved):

bin/propolis-server/src/lib/initializer.rs
```diff
@@ -957,7 +956,15 @@ impl MachineInitializer<'_> {
                     };
 
                     block_dev.attachment().set_metric_consumer(block_metrics);
-                };
+                } else {
+                    slog::error!(
```
leftwo, line 960: We still won't try to log cloud-init type backends, right?  As they have no UUID:
```
16:18:52.344Z INFO propolis-server (vm_state_driver): Creating storage device
    device_id = cloud-init-dev
    spec = Virtio(VirtioDisk { backend_id: Name("cloud-init-backend"), pci_path: PciPath { bus: 0, device: 24, function: 0 } })
```

leftwo (approved):

bin/propolis-server/src/lib/initializer.rs
ahrens: The type system allows virtio devices to have a Uuid (the `backend_id` could be a Uuid instead of a Name).  But if in practice they do not have uuid's (as shown in your log output), then we won't register metrics for them.

leftwo: > The type system allows virtio devices to have a Uuid (the `backend_id` could be a Uuid instead of a Name). But if in practice they do not have uuid's (as shown in your log output), then we won't register metrics for them.

Perfect.  It's what we discussed during the huddle, but I wanted to be sure we were not going to all of a sudden start logging all the cloud-init disks.

PR #1146, opened 2026-05-27
viona worker LWPs should not be CPU-pinned

iximeow:
VNA_IOC_RING_INIT* includes, as part of init'ing a ring, creating an LWP in the calling process for that ring's worker thread. This is all fine and good, but for that LWP inheriting the binding of the original ioctl-caller. The ioctl-caller, then, is functionally always a vCPU LWP, in service of an MMIO exit by a guest who is trying to start their virtio NIC.

Since one vCPU is probably handling a routine that loops over all desired virtio rings and enables them, the effective outcome is that all viona rings are handled on some sole arbitrary host CPU that is also responsible for a vCPU.

Avoid all of this by temporarily unbinding whatever LWP is getting the viona ring going, then rebinding before we've continued on.

### DIFF
--- a/crates/pbind/src/lib.rs
+++ b/crates/pbind/src/lib.rs
@@ -111,7 +111,9 @@ pub fn bind_lwp(
 /// On non-illumos targets, we're not actually running a VM. We do need the
 /// crate to compile to be nicer for blanket `cargo test` invocations on other
 /// platforms. So a no-op function will do.
-pub fn bind_lwp(_bind_cpu: processorid_t) -> Result<(), Error> {
+pub fn bind_lwp(
+    _bind_cpu: Option<processorid_t>,
+) -> Result<Option<processorid_t>, Error> {
     Ok(())
 }
 
--- a/crates/pbind/src/lib.rs
+++ b/crates/pbind/src/lib.rs
@@ -123,7 +123,7 @@ pub fn bind_lwp(
 ///
 /// If the function panics, the LWP's original processor binding will not be
 /// restored.
-pub fn with_unbound_lwp<T: Sized>(f: impl Fn() -> T) -> T {
+pub fn with_unbound_lwp<T: Sized>(f: impl FnOnce() -> T) -> T {
     let oldbind = bind_lwp(None).expect("can unbind this LWP");
 
     let res = f();
--- a/crates/pbind/src/lib.rs
+++ b/crates/pbind/src/lib.rs
@@ -114,7 +114,7 @@ pub fn bind_lwp(
 pub fn bind_lwp(
     _bind_cpu: Option<processorid_t>,
 ) -> Result<Option<processorid_t>, Error> {
-    Ok(())
+    Ok(None)
 }
 
 /// Run the provided function without any processor binding active on the
--- a/crates/pbind/src/lib.rs
+++ b/crates/pbind/src/lib.rs
@@ -63,8 +63,13 @@ pub fn online_cpus() -> Result<i32, Error> {
 }
 
 #[cfg(target_os = "illumos")]
-/// Bind the current LWP to the specified processor.
-pub fn bind_lwp(bind_cpu: processorid_t) -> Result<(), Error> {
+/// Bind this LWP to the specified processor.
+///
+/// Returns the previous processor binding for this LWP, if any was set,
+/// otherwise `None`
+pub fn bind_lwp(
+    bind_cpu: Option<processorid_t>,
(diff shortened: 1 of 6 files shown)


iximeow:
I haven't been able to get this on a racklette quite yet but I've tested the effect of this change by starting a large (68 vCPU) VM on dogfood, setting `iperf3 -s` up there, and then using a smaller (4-vcpu, not that it matters) VM elsewhere to send traffic to it on the VPC IP. status quo had the large VM receiving around 22gbit/sec. when I unbound the viona vring LWPs (incredibly scientific: `for lwp in [list]; do pbind -u $lwp; done`) the received throughput jumped to 26-31gbit/sec, stayed for a few seconds, and then fell to 22gbit/sec again.

restarting the iperf3 client saw a steady 31gbit/sec. I don't think any of this compares in an expected way with other benchmarks I know we've done, but I'm plenty convinced this is worth checking out in a more rigorous way..

crates/pbind/src/lib.rs
```diff
@@ -77,26 +82,53 @@ pub fn bind_lwp(bind_cpu: processorid_t) -> Result<(), Error> {
     // From `<sys/types.h>`.
     const P_MYID: id_t = -1;
 
+    // From `common/sys/processor.h`: sentinel indicating
+    // > "LWP/thread is not bound"
+    const PBIND_NONE: processorid_t = -1;
+
+    let newbind: processorid_t = bind_cpu.unwrap_or(PBIND_NONE);
+    let mut obind: processorid_t = PBIND_NONE;
```
hawkw, line 90: hm, what are the explicit `processorid_t` type annotations doing here? seems pretty clearly inferrable. is this just due to being sketched out by the cast in the unsafe block?

crates/pbind/src/lib.rs
```diff
@@ -77,26 +82,53 @@ pub fn bind_lwp(bind_cpu: processorid_t) -> Result<(), Error> {
...
+) -> Result<Option<processorid_t>, Error> {
+    Ok(None)
+}
+
+/// Run the provided function without any processor binding active on the
+/// current LWP. The LWP's original processor binding is restored after the
+/// function returns.
+///
+/// If the function panics, the LWP's original processor binding will not be
+/// restored.
```
hawkw, line 125: it seems this could be pretty easily achieved with a drop guard, but maybe it's not worth it?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1071,12 +1071,30 @@ impl VionaHdl {
...
+            // Because the guest probably goes and enables all the rings on one
+            // CPU as part of some driver operation, if we don't intervene here
+            // it's likely all the worker threads for the vNIC will be bound to
+            // the same core. We don't actually know, from the device side, if
+            // the guest will go and set up the rest of the rings right now, so
+            // we have to do the unbind/bind dance for each ring.
+            //
+            // Arguably one might not want to do such operations directly on a
+            // vCPU thread. Device setup isn't exactly on anyone's hot path so
+            // we'll live.
```
hawkw, line 1091: hm, what are you imagining here? just spawning off a short-lived thread to do one ioctl, or...?

crates/pbind/src/lib.rs
iximeow: yeah, in particular `&mut obind as *mut processorid_t` feels like an easy way to type confusion so I'm being a bit extra on the type annotations. and newbind got one because it felt weird to only type `obind` when they're the same.

crates/pbind/src/lib.rs
iximeow: since propolis-server/propolis-standalone are panic=abort i kinda don't care too much, but also even with a drop guard there's no guarantee that runs either right?

lib/propolis/src/hw/virtio/viona.rs
iximeow: maybe, or a separate thread to handle "control plane" operations on a device that these register reads/writes interact with by a channel. a lot of operations are allowed to be async, and we run then synchronously because it's simpler and decoupling the register access from taking a lock and bumping a field would be silly. but this is heavy enough you can see the thinking maybe?

iximeow:
having checked out that with this the viona tx/rx LWPs aren't unexpectedly bound to some CPU after this change, I'm going to call this "clearly better than the status quo" - from here the only question to me is "is this worth mentioning in release notes once it's in", so ... lets get it in.

PR #1148, opened 2026-05-28
propolis-cli: accept ip:port for server address

### LINKED ISSUE #1062: propolis-cli -s is very confused with a port number
in `propolis-server` we accept a [`SocketAddr`](https://github.com/oxidecomputer/propolis/blob/053a10c071da60b1e7869d1eda6a03587b3b1b1d/bin/propolis-server/src/main.rs#L81-L82) as a binary argument outright, meaning you can run `./propolis-server run path/to/ovmf.fd 0.0.0.0:12345` and you're off.

however, in `propolis-cli` we have `-s` for the server address, and `-p` for the port. this makes for some confusing UI: if you `./propolis-cli -s 127.0.0.1:12345` to talk to that server you just started, this actually just gets you

> failed to lookup address information: node name or service name not known

... because, i think, [resolve_host](https://github.com/oxidecomputer/propolis/blob/053a10c071da60b1e7869d1eda6a03587b3b1b1d/bin/propolis-cli/src/main.rs#L466-L473) is taking the string from the server parameter and forcing Rust to treat it as literally either IP address or hostname, with an explicit port 0 for the socket. `:3000` makes it clearly not an IP address so [Rust's ToSocketAddrs falls back to trying to look up the host](https://doc.rust-lang.org/src/std/net/socket_addr.rs.html#197-210) which it also won't do, and whoops name or service not known.

i think we should probably drop the `-p` flag and expect `ip:port` in all cases.

MorganaFuture:
`propolis-cli -s` parsed its argument as a bare host and discarded any port, so passing `-s 127.0.0.1:12345` (the form `propolis-server` prints and accepts) failed with "node name or service name not known". This parses the server argument as a full socket address and drops the now-redundant `-p` flag, so `-s` takes `IP:PORT` the same way `propolis-server run` already does. The `migrate` destination argument gets the same treatment, and the README/docs examples are updated to match.

Note this is a small breaking change to the CLI: a port is now required (there's no default `12400` anymore), per the issue's suggestion to "expect ip:port in all cases".

Closes #1062.

Tested with `cargo test -p propolis-cli` (added unit tests for the parser) and `cargo clippy`, built on Linux. I don't have an illumos host, so I couldn't exercise the CLI end-to-end against a running propolis-server.

### DIFF
--- a/bin/propolis-cli/README.md
+++ b/bin/propolis-cli/README.md
@@ -27,14 +27,14 @@ pci-path = "0.5.0"
 To create and run a Propolis VM using this configuration:
 
 ```
-# propolis-cli -s <server ip> -p <port> new --config-toml <path> <VM name>
-# propolis-cli -s <server ip> -p <port> state run
+# propolis-cli -s <server ip>:<port> new --config-toml <path> <VM name>
+# propolis-cli -s <server ip>:<port> state run
 ```
 
 To connect to the VM's serial console:
 
 ```
-# propolis-cli -s <server ip> -p <port> serial
+# propolis-cli -s <server ip>:<port> serial
 ```
 
 Run `propolis-cli --help` to see the full list of supported commands and their
--- a/bin/propolis-cli/src/main.rs
+++ b/bin/propolis-cli/src/main.rs
@@ -7,7 +7,7 @@ use std::fs::File;
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
 use std::{
-    net::{IpAddr, SocketAddr, ToSocketAddrs},
+    net::{SocketAddr, ToSocketAddrs},
     os::unix::prelude::AsRawFd,
     time::Duration,
 };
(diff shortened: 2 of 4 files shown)


iximeow:
hello! this'll be a nice, and thank you for including the docs in the change as well :pray:

fwiw I am realizing now that while I thought I'd spoken a bit off-the-cuff in saying "`ip:port` in all cases", that in fact `propolis-server` also does not have the idea of a default port (let alone 12400) - the fact that propolis-server ends up on 12400 is [from the SMF manifest](https://github.com/oxidecomputer/propolis/blob/master/packaging/smf/propolis-server/manifest.xml#L31) (other parts of the system assume [the same default](https://github.com/oxidecomputer/omicron/blob/c5c595dba5d3eb52eac9d030cbd6572936efd0db/nexus/src/app/sagas/instance_common.rs#L27-L28)). then when I've used 12400 locally it's just habit 'cause that's what the CLI used as a default.. we maybe _should_ formalize that the default port is 12400 both to `propolis-server` and `propolis-cli`.. anyway being able to use the same sockaddr on both CLIs will be nice.

you didn't hear this from me but if you do want to give this a try end to end: if you run an illumos guest on a Linux host, from that guest's perspective a nested VM will work well enough to at least see this change work with `propolis-cli`/`propolis-server`. nowhere near well enough to do any serious changes to Propolis or bhyve :D either way I'll give this a quick run on my illumos machine. _not_ asking you to do that unless you think it'd be fun!

PR #1150, opened 2026-06-03
fix nits on non-x86

ahrens:
Fix a couple of warnings when running `cargo check` on non-x86 machines, e.g. an ARM64 Mac.


iximeow (approved):
oh yep, super fair, thanks.

PR #1156, opened 2026-06-18
forget less NVMe state across migration

### LINKED ISSUE #981: NVMe devices don't tolerate migration
there are at least two bugs here, one much more obvious than the other.. I was trying to tickle a different but I _suspect_ exists, but have been stymied by these for the time being.

I got here via `./propolis-standalone -s ubuntu.22.04.5.toml`, ctrl+c, and `./propolis-standalone -r <snapshot_tar>` with a file-backed disk and a mem-async disk, but the disk types shouldn't really matter (both are wedged)

issue one is that NVMe devices now have an `is_enabled` to avoid locking `NvmeCtrl` to read `ctrl.cc` on hot paths. that's great, except when we import a device we don't peel out `is_enabled` again, so the imported controller has `cc` declaring it's enabled, but is functionally disabled. attempting to submit I/O (when run via `propolis-standalone`) then gets you `nvme  reg r/w failure, error: the submission queue specified (4) is invalid, register: IOQueueDoorBells, offset: 4128` in [`ring_doorbell`](https://github.com/oxidecomputer/propolis/blob/fd3636877061da7e951cb1fbce365f7cbf40933c/lib/propolis/src/hw/nvme/mod.rs#L1133-L1143).

adding a
```rust
if ctrl.ctrl.cc.enabled() {
    self.is_enabled.store(true, Ordering::Releae);
}
```
right after
https://github.com/oxidecomputer/propolis/blob/fd3636877061da7e951cb1fbce365f7cbf40933c/lib/propolis/src/hw/nvme/mod.rs#L1371

gets you past that and onto the second issue, which I've not debugged at all: I/Os seem to just not get acked (maybe not serviced?) on a migrated controller. it may be relevant that I've also got a local patch for `const MAX_NUM_QUEUES: usize = 5;` but I've not tested. we'll see how https://github.com/oxidecomputer/propolis/commit/cc43ca00c81e45d35a59155a5e17084bcff878d2 fares!

edit/ps: #966 didn't trip over this because we use `state.ctrl.cc.enabled()` on the admin SQ/CQ doorbells, but `self.is_enabled` on the I/O doorbells, and the rudimentary driver there doesn't actually do I/O. so.. probably worth making it do some I/Os..

iximeow:
add an additional test for propolis-to-itself migration and run the smoke tests there. extend the smoke test to try making sure disks are at least kind of basically still functional.

this should fix https://github.com/oxidecomputer/propolis/issues/981, but I have to admit I couldn't really reproduce the second failure mode that I alluded to there or @lgfa29 saw last week. I think that returning incorrect submission queue heads could have an outcome where the guest OS thinks it's submitting I/Os in a way we wouldn't get to, but beyond the status quo obviously being wrong, I'm not sure I've actually tickled that condition much at all.

I mostly really want to extend #966 to actually drive I/Os now, because either of the improvements here _should_ fail quickly through a fuzzer...

### DIFF
--- a/lib/propolis/src/block/attachment.rs
+++ b/lib/propolis/src/block/attachment.rs
@@ -1030,6 +1030,24 @@ impl WorkerCollection {
                 continue;
             };
 
+            {
+            // gross hack: we are going to reconfigure all of the queues, so any existing notify
+            // bits may become incorrect. Such as, imagine queue 0 had idled, indicated to minder
+            // 0 that it was idle, then reassigned to a fixed assignment to queue1. A wakeup for
+            // queue 0 could wake worker 0 based on notify_bits, which would not successfully
+            // discovery that work is ready on queue 0, as it would instead check queue 1.
+            //
+            // this is way more wordy than is worth PRing. anyway, clear all the notify bits
+            // because as part of updating assignments we will go wake every worker anyway. do this
+            // *before* waking *any* workers so we don't have TOCTOU issues on the notify bits and
+            // clear freshly-correct idle bits of just-reassigned queuees.
+            let ws = slot.state.lock().unwrap();
+            if let PollAssignment::Fixed(assigned_qid) = ws.assign_poll {
+                if qid_hint != Some(assigned_qid) {
+                    continue;
+                }
+            };
+            }
             if slot.wake(None, qid_hint) {
                 num_woken += 1;
             }
--- a/lib/propolis/src/block/attachment.rs
+++ b/lib/propolis/src/block/attachment.rs
@@ -232,6 +232,10 @@ impl QueueCollection {
         }
     }
 
+    // Callers must hold the worker's lock while polling the next request for
+    // the selected device queue. This ensures that if the worker is to have its
+    // queue assignment changed, this is mutually exclusive with the worker
+    // indicating idleness to a QueueMinder.
     fn next_req(
         &self,
         queue_select: QueueId,
@@ -742,11 +746,15 @@ impl WorkerSlot {
         WaitForReq::new(self)
     }
 
-    fn update_assignment(&self, assign: &Assignment) {
+    /// Reconfigure this worker for a polling strategy matching the current
+    /// assignment. Returns if the worker has actually been modified; `false` if
+    /// the current strategy is newer than provided (as in the case of racing
+    /// worker assignments).
+    fn update_assignment(&self, assign: &Assignment) -> bool {
         let mut state = self.state.lock().unwrap();
         if state.assign_strat.newer_than(&assign.strategy) {
-            // We already have a newer assignment
-            return;
+            // We already have a newer assignment; nothing has been updated.
+            return false;
         }
         state.assign_strat = assign.strategy;
         if assign.should_halt {
@@ -759,7 +767,8 @@ impl WorkerSlot {
                     PollAssignment::Idle
                 };
         }
-        self.wake(Some(state), None);
+
+        true
     }
 
     fn wake(
@@ -983,8 +992,67 @@ impl WorkerCollection {
             let generation = assign.strategy.generation() as u64;
             (devid.0, assign_name, generation)
         });
+
+        let mut any_updated = false;
         for slot in self.workers.iter() {
-            slot.update_assignment(&assign);
+            if slot.update_assignment(&assign) {
+                any_updated = true;
+            }
+        }
+
+        if !any_updated {
+            return;
+        }
+
+        // At this point we have updated at least one worker. The worker may
+        // already have been active, its assignment may have been Fixed or
+        // FreeForAll. It may have become idle, setting `notify_workers` on some
+        // queue, and if the strategy was changed from Fixed(A) to Fixed(B) the
+        // queue with a notify bit for this worker may actually not be the queue
+        // this worker will try pulling I/Os from! In short, we've made a mess
+        // of things. Cleaning this up is not all bad, but subtle.
+        //
+        // First, we'll clear all notification bits. At this point, if any
+        // workers idle again, they will indicate idleness on the correct
+        // queues, at least. We may blow away some worker idleness bits that are
+        // correctly placed, along with any bits that might be incorrectly
+        // placed.
+        //
+        // Then, kick all the workers. We probably don't actually have I/O for
+        // all of them. But, when workers idle this time, they'll indicate
+        // *correct* idle bits and notifications will engage the correct
+        // workers.
+        //
+        // The strangest case here is probably if a queue is moved from Fixed to
+        // FreeForAll; the worker should have set a bit for its ID in all
+        // minders. This procedure will result in bits for all idle queues as
+        // appropriate once the worker is kicked and deposits idleness across
+        // the minders.
+        for slot in self.workers.iter() {
+            // The mechanics of this are a little paranoid: only worker slots
+            // themselves have a queue collection, and it *should* be the case
+            // that all queue collections are the same. But even if it wasn't,
+            // we need to walk all the queues that all the reassigned workers
+            // could access to clear all the notify_workers bits.
+            //
+            // Currently this means we walk all the queue minders for the device
+            // NUM_WORKERS-many times. Queue reassignment is already terribly
+            // expensive what with having to kick all the workers. Please do not
+            // reconfigure device queues while sensitive to I/O latency or
+            // throughput..
+            let state = slot.state.lock().unwrap();
+            if let Some(qcol) = state.queues.as_ref() {
+                for queue in qcol.queues.iter() {
+                    let qs = queue.state.lock().unwrap();
+                    if let Some(minder) = qs.minder.as_ref() {
+                        let _ = minder.take_notifications();
+                    }
+                }
+            }
+        }
+
+        for slot in self.workers.iter() {
+            slot.wake(None, None);
         }
     }
     fn slot(&self, id: WorkerId) -> &WorkerSlot {
@@ -1030,24 +1098,6 @@ impl WorkerCollection {
                 continue;
             };
 
-            {
-            // gross hack: we are going to reconfigure all of the queues, so any existing notify
-            // bits may become incorrect. Such as, imagine queue 0 had idled, indicated to minder
-            // 0 that it was idle, then reassigned to a fixed assignment to queue1. A wakeup for
-            // queue 0 could wake worker 0 based on notify_bits, which would not successfully
-            // discovery that work is ready on queue 0, as it would instead check queue 1.
-            //
-            // this is way more wordy than is worth PRing. anyway, clear all the notify bits
-            // because as part of updating assignments we will go wake every worker anyway. do this
-            // *before* waking *any* workers so we don't have TOCTOU issues on the notify bits and
-            // clear freshly-correct idle bits of just-reassigned queuees.
-            let ws = slot.state.lock().unwrap();
-            if let PollAssignment::Fixed(assigned_qid) = ws.assign_poll {
-                if qid_hint != Some(assigned_qid) {
-                    continue;
-                }
-            };
-            }
             if slot.wake(None, qid_hint) {
                 num_woken += 1;
             }
--- a/lib/propolis/src/hw/nvme/mod.rs
+++ b/lib/propolis/src/hw/nvme/mod.rs
@@ -1418,6 +1418,14 @@ impl MigrateMulti for PciNvme {
         output.push(ctrl.export().into())?;
         drop(ctrl);
 
+        // We leave `is_enabled` and `device_id` behind here:
+        // * is_enabled is just a mirror of the controller's enabled bit. We'll
+        //   recover it when importing controller state on the other side.
+        // * `device_id` is, like other `define_id` items, tied to non-migrated
+        //   statics in the Propolis process. Migrating it is, absent other
+        //   work, a bug waiting to happen (imagine ID gaps where devices with
+        //   ID 0, 1, and 4 are migrated)
+
         MigrateMulti::export(&self.pci_state, output, ctx)?;
 
         Ok(())
@@ -1432,6 +1440,9 @@ impl MigrateMulti for PciNvme {
 
         let mut ctrl = self.state.lock().unwrap();
         ctrl.import(input, self)?;
+        // Now that the controller is imported, update our mirror of its
+        // enablement bit.
+        self.is_enabled.store(ctrl.ctrl.cc.enabled(), Ordering::Release);
         drop(ctrl);
 
         MigrateMulti::import(&self.pci_state, offer, ctx)?;
--- a/phd-tests/tests/src/migrate.rs
+++ b/phd-tests/tests/src/migrate.rs
@@ -318,6 +318,12 @@ async fn multiple_migrations(ctx: &TestCtx) {
     );
 }
 
+#[phd_testcase]
+async fn migration_smoke_test(ctx: &TestCtx) {
+    let vm = ctx.spawn_default_vm("migration_smoke_test_0").await?;
+    run_smoke_test(ctx, vm).await?;
+}
+
 async fn run_smoke_test(ctx: &TestCtx, mut source: TestVm) -> Result<()> {
     source.launch().await?;
     source.wait_to_boot().await?;
@@ -325,7 +331,9 @@ async fn run_smoke_test(ctx: &TestCtx, mut source: TestVm) -> Result<()> {
         source.run_shell_command("ls foo.bar 2> /dev/null").check_err().await?;
     assert_eq!(lsout, "");
 
-    // create an empty file on the source VM.
+    // create an empty file on the source VM. `/` might be a tmpfs, and the
+    // guest may have no writable filesystem even. validation post-migration is
+    // a bit tricky, below.
     source.run_shell_command("touch ./foo.bar").await?;
     source.run_shell_command("sync ./foo.bar").await?;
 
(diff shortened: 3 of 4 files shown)


phd-tests/tests/src/migrate.rs
```diff
@@ -334,13 +342,72 @@ async fn run_smoke_test(ctx: &TestCtx, mut source: TestVm) -> Result<()> {
...
                 let lsout = target
                     .run_shell_command("ls foo.bar")
                     .ignore_status()
                     .await
                     .expect("can try to run `ls foo.bar`");
                 assert_eq!(lsout, "foo.bar");
+
+                // we have a small conundrum: it's possible that the home
+                // directory we wrote into is a tmpfs anyway. so by succeeding
+                // this much, we may have really just tested memory is intact.
```
iximeow, line 363: @hawkw i would like for you to look at this change just because i can imagine the eye roll and swear when you read this comment thanks in advance. it's so much more annoying than i thought it was.

iximeow:
hm. the failures are fascinating!

`phd_tests::migrate::smoke_test` yields:
```
localhost:~# dmesg | grep 'reset controller'
[   68.965446] nvme nvme0: I/O 82 QID 2 timeout, reset controller
localhost:~# echo $?
0
```
so.. the test detected a problem! also what in the world.

`migration_smoke_test` on the other hand:
```
localhost:~# ls /media/nvme0n1/boot
System.map-virt  config-virt      dtbs-virt        grub             initramfs-virt   modloop-virt     syslinux         vmlinuz-virt
localhost:~# echo $?
0
localhost:~# dmesg | grep 'reset controller'
localhost:~# echo $?
1
```
... I presumably should be ignoring the exit code of the grep, oops.

lgfa29:
I ran a quick test using the same setup I did previously and things have certainly improved, the migration actually finished now 😄 

But I still ran into issues after the migration, but each time it was a different.

The first time I ran a migration the disk seemed to be "stuck" for about 2min, but then I saw these kernel messages in the guest and the disk seemed to be fine:

```
[   89.895139] nvme nvme0: I/O 900 (I/O Cmd) QID 2 timeout, aborting
[   89.896745] nvme nvme0: Abort status: 0x0
[  120.103133] nvme nvme0: I/O 900 QID 2 timeout, reset controller
[  120.135440] nvme0n1: I/O Cmd(0x2) @ LBA 1336320, 32 blocks, I/O Error (sct 0x3 / sc 0x71)
[  120.137336] I/O error, dev nvme0n1, sector 1336320 op 0x0:(READ) flags 0x80700 phys_seg 4 prio class 2
[  120.151848] nvme nvme0: 2/0/0 default/read/poll queues
```

After the second test the disk seemed to be fine, but then I left the guest running while I was in a meeting and when I came it was unresponsive and I had several messages like these:

```
[   99.371306] nvme nvme0: I/O 263 QID 2 timeout, reset controller
[   99.418681] nvme nvme0: 2/0/0 default/read/poll queues
[  130.091326] nvme nvme0: I/O 263 (I/O Cmd) QID 2 timeout, aborting
[  130.092870] nvme nvme0: Abort status: 0x0
[  160.811324] nvme nvme0: I/O 263 QID 2 timeout, reset controller
[  160.858176] nvme nvme0: 2/0/0 default/read/poll queues
[  191.531317] nvme nvme0: I/O 263 QID 2 timeout, disable controller
[  191.563602] I/O error, dev nvme0n1, sector 2884112 op 0x1:(WRITE) flags 0x29800 phys_seg 1 prio class 2
[  191.565764] I/O error, dev nvme0n1, sector 2884112 op 0x1:(WRITE) flags 0x29800 phys_seg 1 prio class 2
[  191.567984] Aborting journal on device nvme0n1p1-8.
[  191.568026] nvme nvme0: unable to set dbbuf
[  191.570141] nvme nvme0: failed to mark controller live state
[  191.570198] I/O error, dev nvme0n1, sector 358320 op 0x1:(WRITE) flags 0x0 phys_seg 1 prio class 2
[  191.570972] nvme nvme0: Removing after probe failure status: -19
[  191.572323] EXT4-fs warning (device nvme0n1p1): ext4_end_bio:343: I/O error 10 writing to inode 110 starting block 44790)
[  191.574436] Buffer I/O error on device nvme0n1p1, logical block 12022
[  191.575283] I/O error, dev nvme0n1, sector 358408 op 0x1:(WRITE) flags 0x0 phys_seg 1 prio class 2
[  191.575302] EXT4-fs error (device nvme0n1p1): ext4_journal_check_start:83: comm systemd-journal: Detected aborted journal
[  191.576423] EXT4-fs warning (device nvme0n1p1): ext4_end_bio:343: I/O error 10 writing to inode 110 starting block 44801)
[  191.579136] Buffer I/O error on device nvme0n1p1, logical block 12033
[  191.579981] I/O error, dev nvme0n1, sector 527624 op 0x1:(WRITE) flags 0x0 phys_seg 1 prio class 2
[  191.581113] EXT4-fs warning (device nvme0n1p1): ext4_end_bio:343: I/O error 10 writing to inode 110 starting block 65953)
[  191.582495] Buffer I/O error on device nvme0n1p1, logical block 33185
[  191.583445] I/O error, dev nvme0n1, sector 355296 op 0x1:(WRITE) flags 0x0 phys_seg 1 prio class 2
[  191.584784] EXT4-fs warning (device nvme0n1p1): ext4_end_bio:343: I/O error 10 writing to inode 110 starting block 44412)
[  191.586373] Buffer I/O error on device nvme0n1p1, logical block 11644
[  191.587339] I/O error, dev nvme0n1, sector 355584 op 0x1:(WRITE) flags 0x0 phys_seg 1 prio class 2
[  191.588686] EXT4-fs warning (device nvme0n1p1): ext4_end_bio:343: I/O error 10 writing to inode 110 starting block 44448)
```

In my last test, the disk started OK for read, but writes would hang and then I started to get:

```
Last login: Thu Jun 18 18:17:43 UTC 2026 on ttyS0
root@localhost:~#
root@localhost:~# ls
mnt  test
root@localhost:~# [   37.163334] nvme nvme0: I/O 201 (I/O Cmd) QID 2 timeout, aborting
[   37.164890] nvme nvme0: Abort status: 0x0
[   37.215307] nvme nvme0: I/O 202 (I/O Cmd) QID 2 timeout, aborting
[   37.216783] nvme nvme0: Abort status: 0x0

root@localhost:~# ls
mnt  test
root@localhost:~# ls test
test
root@localhost:~# cat test
1
root@localhost:~# [   67.371330] nvme nvme0: I/O 201 QID 2 timeout, reset controller
[   67.416449] nvme nvme0: 2/0/0 default/read/poll queues

root@localhost:~# ls
mnt  test
root@localhost:~# cat ./test
1
root@localhost:~# [   99.371308] nvme nvme0: I/O 201 (I/O Cmd) QID 2 timeout, aborting
[   99.372610] nvme nvme0: Abort status: 0x0

root@localhost:~#
root@localhost:~# cat ./test [  130.091303] nvme nvme0: I/O 201 QID 2 timeout, reset controller
[  130.136615] nvme nvme0: 2/0/0 default/read/poll queues

1
root@localhost:~# echo 2 > ./test

[  160.811324] nvme nvme0: I/O 201 QID 2 timeout, disable controller
[  160.843610] I/O error, dev nvme0n1, sector 2884112 op 0x1:(WRITE) flags 0x29800 phys_seg 1 prio class 2
[  160.845169] I/O error, dev nvme0n1, sector 2884112 op 0x1:(WRITE) flags 0x29800 phys_seg 1 prio class 2
[  160.846325] Aborting journal on device nvme0n1p1-8.
[  160.846396] nvme nvme0: unable to set dbbuf
[  160.848017] nvme nvme0: failed to mark controller live state
[  160.848047] I/O error, dev nvme0n1, sector 532504 op 0x1:(WRITE) flags 0x0 phys_seg 1 prio class 2
[  160.849102] nvme nvme0: Removing after probe failure status: -19
[  160.850177] EXT4-fs warning (device nvme0n1p1): ext4_end_bio:343: I/O error 10 writing to inode 94 starting block 66563)
[  160.850961] EXT4-fs error (device nvme0n1p1): ext4_journal_check_start:83: comm kworker/u4:1: Detected aborted journal
[  160.852306] EXT4-fs error (device nvme0n1p1): ext4_journal_check_start:83: comm systemd-journal: Detected aborted journal
[  160.853598] EXT4-fs error (device nvme0n1p1): ext4_journal_check_start:83: comm kworker/u4:4: Detected aborted journal
[  160.854895] I/O error, dev nvme0n1, sector 2883584 op 0x1:(WRITE) flags 0x29800 phys_seg 1 prio class 2
[  160.857402] Buffer I/O error on dev nvme0n1p1, logical block 327680, lost sync page write
[  160.858386] JBD2: I/O error when updating journal superblock for nvme0n1p1-8.
[  160.859265] I/O error, dev nvme0n1, sector 262144 op 0x1:(WRITE) flags 0x23800 phys_seg 1 prio class 2
root@localhost:~#
[  160.861812] Buffer I/O error on dev nvme0n1p1, logical block 0, lost sync page write
root@localhost:~# [  160.863545] EXT4-fs (nvme0n1p1): I/O error while writing superblock
[  160.863557] EXT4-fs (nvme0n1p1): previous I/O error to superblock detected
[  160.864327] EXT4-fs (nvme0n1p1): Remounting filesystem read-only
[  160.866214] EXT4-fs (nvme0n1p1): ext4_writepages: jbd2_start: 1024 pages, ino 95; err -30
[  160.871292] nvme0n1: detected capacity change from 6291456 to 0
[  160.871327] EXT4-fs warning (device nvme0n1p1): ext4_end_bio:343: I/O error 10 writing to inode 95 starting block 67414)
[  160.872211] Buffer I/O error on dev nvme0n1p1, logical block 0, lost sync page write
[  160.873556] Buffer I/O error on device nvme0n1p1, logical block 34646
[  160.874499] EXT4-fs (nvme0n1p1): I/O error while writing superblock
[  160.875297] EXT4-fs warning (device nvme0n1p1): ext4_end_bio:343: I/O error 10 writing to inode 95 starting block 67369)
[  160.876050] EXT4-fs (nvme0n1p1): Remounting filesystem read-only
```

For reference, these are my test steps:

1. Start first `propolis-server`
    ```
    pfexec propolis-server run ./OVMF_CODE.fd 0.0.0.0:3000 0.0.0.0:4000
    ```
2. Start second `propolis-server` (same host in my case)
    ```
    pfexec propolis-server run ./OVMF_CODE.fd 0.0.0.0:3001 0.0.0.0:4001
    ```
3. Create a `debian.toml` file
    ```toml
    [main]
    name = "testvm-debian"
    cpus = 2
    memory = 4096
    bootrom = "./OVMF_CODE.fd"
    
    [block_dev.os]
    type = "file"
    path = "./debian-12-nocloud-amd64.raw"
    
    [dev.block0]
    driver = "pci-nvme"
    block_dev = "os"
    pci-path = "0.3.0"
    ```
4. Start the VM
    ```
    propolis-cli --server 127.0.0.1 --port 3000 new -t ./debian.toml -c 2 -m 2048 test-vm
    propolis-cli --server 127.0.0.1 --port 3000 state run
    propolis-cli --server 127.0.0.1 --port 3000 serial
    ```
5. Trigger a migration
    ```
    propolis-cli --server 127.0.0.1 --port 3000 migrate -p 3001 127.0.0.1
    ```
6. Attempt to read and write files in the guest

phd-tests/tests/src/migrate.rs
hawkw: :roll_eyes:🤬

iximeow:
after staring at this some more: it looks like there's a condition where we can import a VM, have a submission enqueued, resume devices, but incorrectly decide the queue is empty. the guest doesn't submit another I/O because the one it's waiting for is outstanding, so it discovers the disk is stuck. then all the normal "your disk stopped" happens. this happens with and without doorbell buffers, I've been looking with them disabled for simplicity.

as we saw earlier, migrating while the I/O is submitted but unprocessed does "unstick" it, but I don't quite follow how. where it's stuck, `nvme_block_notify!()` says queue 2 got a notify for one item, then block queue 0 (??? this should be 1!) polls, returns None, and that's it. then you migrate the VM, resume nudges all the queues, the I/O gets done, and we're back.

so... why does `nvme_block_notify!()` for queue 2 lead into a `block_poll!()` for I/O queue 0? it's clearly a fixed assignment strategy because when that worker goes idle it only checks the first queue, so it's in `next_req()` not `next_req_any()` where `None` would have it check other queues until there's a request.  .. huh? ..

iximeow:
https://github.com/oxidecomputer/propolis/pull/1156/commits/67c252497482b8ba568bfbec1d47be7c9b2d1e28 looks like it covers the issue from here. clippy hates it, i think the change itself is not the right way to spell it, but it describes the issue i see and *a* fix in the interim until i figure out plumbing in a QueueCollection or .... something?

lgfa29:
I ran a few tests using a similar setup as my previous test and the migrated disks seem to work fine now 🎉

lib/propolis/src/block/attachment.rs
```diff
@@ -1030,6 +1030,24 @@ impl WorkerCollection {
...
+            // because as part of updating assignments we will go wake every worker anyway. do this
+            // *before* waking *any* workers so we don't have TOCTOU issues on the notify bits and
+            // clear freshly-correct idle bits of just-reassigned queuees.
+            let ws = slot.state.lock().unwrap();
+            if let PollAssignment::Fixed(assigned_qid) = ws.assign_poll {
+                if qid_hint != Some(assigned_qid) {
+                    continue;
+                }
+            };
+            }
```
iximeow: so, at this point we'll drop `slot.state.lock()`. we could reconfigure the worker at this point, so that `slot.wake()` just below will wake the worker which then goes and checks a _different_ fixed queue anyway. an example here is if a guest has gone and enabled (or disabled) an I/O queue and moved assignments around.

this is an improvement only as long as queues are not being reassigned concurrently with I/O, and if there is I/O then we have the same issue of queues poking workers that have no hope of serving them.

I might want to revert this and land 1156 as a marginal improvement for now, and have an issue about the shape of this problem and notify_workers in the face of device reconfiguration...

iximeow: so, at this point we'll drop `slot.state.lock()`. we could reconfigure the worker at this point, so that `slot.wake()` just below will wake the worker which then goes and checks a _different_ fixed queue anyway. an example here is if a guest has gone and enabled (or disabled) an I/O queue and moved assignments around.

this is an improvement only as long as queues are not being reassigned concurrently with I/O, and if there is I/O then we have the same issue of queues poking workers that have no hope of serving them.

I might want to revert this and land 1156 as a marginal improvement for now, and have an issue about the shape of this problem and notify_workers in the face of device reconfiguration...

lib/propolis/src/block/attachment.rs
```diff
@@ -742,11 +746,15 @@ impl WorkerSlot {
         WaitForReq::new(self)
     }
 
-    fn update_assignment(&self, assign: &Assignment) {
+    /// Reconfigure this worker for a polling strategy matching the current
+    /// assignment. Returns if the worker has actually been modified; `false` if
```
hawkw, line 750: this is a super ultra nitpick, but perhaps
```suggestion
    /// assignment. Returns whether the worker has actually been modified; `false` if
```
since i feel "returns if" is misinterpretable as "does not return until"? even though it's probably obvious.

lib/propolis/src/block/attachment.rs
```diff
@@ -983,8 +992,67 @@ impl WorkerCollection {
             let generation = assign.strategy.generation() as u64;
             (devid.0, assign_name, generation)
         });
+
+        let mut any_updated = false;
+        for slot in self.workers.iter() {
+            if slot.update_assignment(&assign) {
+                any_updated = true;
+            }
```
hawkw, line 1000: could also be:
```suggestion
            any_updated |= slot.update_assignment(&assign);
```
but it's up to you whether this is better or worse

lib/propolis/src/hw/nvme/mod.rs
```diff
@@ -1432,6 +1440,9 @@ impl MigrateMulti for PciNvme {
 
         let mut ctrl = self.state.lock().unwrap();
         ctrl.import(input, self)?;
+        // Now that the controller is imported, update our mirror of its
+        // enablement bit.
```
hawkw, line 1444: mmmmmmaybe worth explicitly calling out that this is inside the lock?

phd-tests/tests/src/migrate.rs
```diff
@@ -334,13 +342,77 @@ async fn run_smoke_test(ctx: &TestCtx, mut source: TestVm) -> Result<()> {
...
+                    .expect("can check mounted filesystems and grep");
+
+                let root_is_tmpfs = !mount_grep.is_empty();
+
+                if root_is_tmpfs {
+                    // the guest is *probably* an Alpine. `/` is a figment of
+                    // its imagination; a tmpfs with the actual disk somewhere
+                    // under /media, like /media/nvme0n1. try looking there, and
+                    // if that fails we'll need to improve the test, phd, or
+                    // both.
```
hawkw, line 383: lmao this rocks alpine is such a distro

phd-tests/tests/src/migrate.rs
```diff
@@ -334,13 +342,77 @@ async fn run_smoke_test(ctx: &TestCtx, mut source: TestVm) -> Result<()> {
...
+
+                // the `ls` succeeded ... eventually. it's possible the disk
+                // *was* broken, the guest realized this, reset the disk, and
+                // that brought it back. if so, there's a line about resetting
+                // the controller in dmesg. so if we see that, we've actually
+                // failed to migrate reasonably.
+                //
+                // ignore the status from `grep` because it returns 1 when the
+                // regex matches no line. we don't want to `check_err()`
+                // because it's useful for debugging to see what *did* match.
```
hawkw, line 406: isn't there a grep flag that makes it not do that, or did i make that up?

PR #1158, opened 2026-06-23
Update crucible

leftwo:
Changes include:
- Drop Pantry locks before Volume operations
- update tokio-rustls to 0.26, rustls-pemfile to 2.2, dropshot to 0.17.1
- Handle read-only downstairs reaching WaitQuorum during deactivation
- pantest, a pantry test program (#1946)

Also updated dropshot from 0.17.0 to 0.17.1


iximeow (approved):
seems good! feel free to take this and take over https://github.com/oxidecomputer/omicron/pull/10664, James and I were realizing that I had not realized the NVMe Discard stats included a Crucible bump, so bumping Crucible in Omicron should also let you unwind my (not ideal) rename of `crucible_client_types::CrucibleOpts`.

PR #1162, opened 2026-06-26
standalone: ability to set viona rx/tx queue sizes

rzezeski:
While running local network performance benchmarks I find it useful to be able to change queues sizes.

### DIFF
--- a/bin/propolis-standalone/src/main.rs
+++ b/bin/propolis-standalone/src/main.rs
@@ -1324,6 +1324,18 @@ fn setup_instance(
                 "pci-virtio-viona" => {
                     let vnic_name =
                         dev.options.get("vnic").unwrap().as_str().unwrap();
+                    let rxqsz = match dev.options.get("rx-queue-size") {
+                        Some(toml::Value::Integer(v)) => {
+                            hw::virtio::VqSize::new(u16::try_from(*v).unwrap())
+                        }
+                        _ => hw::virtio::viona::RX_QUEUE_SIZE,
+                    };
+                    let txqsz = match dev.options.get("tx-queue-size") {
+                        Some(toml::Value::Integer(v)) => {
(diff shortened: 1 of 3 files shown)


bin/propolis-standalone/src/main.rs
```diff
@@ -1324,6 +1324,18 @@ fn setup_instance(
                 "pci-virtio-viona" => {
                     let vnic_name =
                         dev.options.get("vnic").unwrap().as_str().unwrap();
+                    let rxqsz = match dev.options.get("rx-queue-size") {
+                        Some(toml::Value::Integer(v)) => {
+                            hw::virtio::VqSize::new(*v as u16)
```
FelixMcFelix: The disposition inside `VqSize::new` is already to panic if we're not power-of-two aligned. In that light can we have `u16::try_from(*v).unwrap()` here? That feels better than doing a silent truncation of the most-significant bits.

bin/propolis-standalone/src/main.rs
```diff
@@ -1324,6 +1324,18 @@ fn setup_instance(
...
+                        }
+                        _ => hw::virtio::viona::RX_QUEUE_SIZE,
+                    };
+                    let txqsz = match dev.options.get("tx-queue-size") {
+                        Some(toml::Value::Integer(v)) => {
+                            hw::virtio::VqSize::new(*v as u16)
```
FelixMcFelix: Ditto here.

FelixMcFelix (commented):
Thanks Ryan, some minor questions around value validity.

bin/propolis-standalone/src/main.rs
rzezeski: Done.

bin/propolis-standalone/src/main.rs
rzezeski: Done.

FelixMcFelix (approved):
Thanks!

PR #1164, opened 2026-07-01
acpi: adjust DSDT device order

lgfa29:
Using a `DsdtGenerator` to generate the AML code for the PS2 keyboard in the v0 ACPI variant can cause its position in the DSDT table to be different from the original tables because the `DsdtGenerators` are read from a the devices `BTreeMap`, which also include the serial ports.

Since the map key for serial ports is defined in the VM spec, it can cause the device order to change based on user-input, which is not desirible for the v0 as we hope to match the original tables as close as possible.


lgfa29:
Closing in favour of https://github.com/oxidecomputer/propolis/pull/1165

PR #1167, opened 2026-07-07
add NVMe I/O timing statemap script

iximeow:
this was mostly useful in preparing an analysis of Propolis I/O timing which I kind of hope to publish on account of being modestly interesting, hopefully useful for future-us too.

### DIFF
--- a/scripts/nvme-io-timing.d
+++ b/scripts/nvme-io-timing.d
@@ -1,0 +1,285 @@
+#!/usr/sbin/dtrace -s
+
+/*
+ * nvme-io-timing.d     Print propolis emulated NVMe read/write latency.
+ *
+ * USAGE: ./nvme-io-timing.d -p propolis-pid -o your_io_trace.jsonl
+ *
+ * This script traces the activity of *writes* from Propolis' emulated NVMe
+ * controller through file-backed disks, to the host OS and back. The measured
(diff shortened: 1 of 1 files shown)


scripts/nvme-io-timing.d
```diff
@@ -0,0 +1,285 @@
...
+fbt::biowait:return / self->id && self->in_default_physio / {
+	/*
+	 * The biowait is done, so we are back to whatever it was we were doing before waiting.
+	 */
+	printf("{\"state\": %d, \"time\": \"%d\", \"entity\": \"%x\"}\n", self->state, timestamp - START, self->id);
+}
+
+tick-1s {
+	exit(0);
+}
```
leftwo, line 312: Does having the empty end here result in dtrace printing something or behaving differently than if it did not?

iximeow: no, its only benefit is to confuse reviewers (at one point I'd had something there..)

iximeow: removed in https://github.com/oxidecomputer/propolis/pull/1167/commits/fa8a1934b7781a25a9032ab8bac0aec4eb66293c (and added read states because those are useful too)

leftwo (approved):

PR #1171, opened 2026-07-14
viona: want F_EVENT_IDX support

rzezeski:
This adds F_EVENT_IDX support for all virtqueues, but for this commit we enable it for viona only. This is needed to support F_EVENT_IDX for the viona control queue, which is required so that the illumos viona device may offer it for the data queues. It was found that Tx completion interrupts are greatly reduced with this feature support, presumably due to the Linux driver's better use of `avail.used_event` for interrupt moderation.

Adding this support has shown me a 10x reduction in guest Tx completion interrupts when running iperf benchmarks.

We might want to consider another change that lands prior to this one that actively prevents the negotiation of F_EVENT_IDX. Currently, if the host offers it, propolis blindly passes it through to the guest which will happily enable it: leading to the condition where the host and guest use `avail_event`/`used_event`, but the viona ctrl queue implemented by propolis is still using the `flags` method, ultimately leading to a guest that hangs during boot because it cannot make progress.

We can either do that, or we can make extra sure that Omicron picks up the latest propolis before we land this feature in Helios.

### DIFF
--- a/lib/propolis/src/hw/virtio/queue.rs
+++ b/lib/propolis/src/hw/virtio/queue.rs
@@ -138,9 +138,7 @@ impl VqAvail {
         self.gpa_flags = GuestAddr(avail_addr);
         self.gpa_idx = GuestAddr(avail_addr + 2);
         self.gpa_ring = GuestAddr(avail_addr + 4);
-        self.gpa_used_event = GuestAddr(
-            avail_addr + (2 + sz as u64) * mem::size_of::<u16>() as u64,
-        );
+        self.gpa_used_event = self.gpa_ring.offset::<u16>(usize::from(sz));
     }
 
     /// Returns guest flags.
@@ -244,8 +242,7 @@ impl VqUsed {
         self.gpa_flags = GuestAddr(gpa);
         self.gpa_idx = GuestAddr(gpa + 2);
         self.gpa_ring = GuestAddr(gpa + 4);
-        self.gpa_avail_event =
-            GuestAddr(gpa + 4 + mem::size_of::<VqdUsed>() as u64 * sz as u64);
+        self.gpa_avail_event = self.gpa_ring.offset::<VqUsed>(usize::from(sz));
     }
 }
 
@@ -467,6 +464,7 @@ impl VirtQueue {
             },
             avail_idx: avail.cur_avail_idx.0,
             used_idx: used.used_idx.0,
+            last_chk_uidx: used.last_chk_uidx.0,
         }
     }
 
@@ -484,6 +482,7 @@ impl VirtQueue {
         used.valid = info.mapping.valid;
         avail.cur_avail_idx = Wrapping(info.avail_idx);
         used.used_idx = Wrapping(info.used_idx);
+        used.last_chk_uidx = Wrapping(info.last_chk_uidx);
     }
 
     /// Accummulates a sequence of available descriptors into a `Chain`.
@@ -602,9 +601,7 @@ impl VirtQueue {
         used.interrupt.as_ref().map(|x| x.read())
     }
 
-    /// Disables interrupts (notifications) on the `Used` ring.
-    ///
-    /// Returns `true` if notifications were previously enabled.
+    /// Disable interrupts (notifications) from driver to device.
     pub(super) fn disable_intr(&self, mem: &MemCtx) {
         let idx = if self.f_event_idx() {
             let aidx = self.avail.lock().unwrap().cur_avail_idx - Wrapping(1);
@@ -628,7 +625,7 @@ impl VirtQueue {
         self.used.lock().unwrap().disable_notify(idx, mem);
     }
 
-    /// Enables interrupts (notifications) on the `Used` ring
+    /// Enable interrupts (notifications) from driver to device.
     pub(super) fn enable_intr(&self, mem: &MemCtx) {
         let aidx = if self.f_event_idx() {
             let aidx = self.avail.lock().unwrap().cur_avail_idx;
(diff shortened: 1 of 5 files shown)


rzezeski:
I have only tested this with an Ubuntu Server 24.04 guest thus far. I'll see about trying to test some other platforms, but I don't have the slightest idea on testing Windows.

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -130,12 +131,16 @@ impl VqAvail {
...
         self.gpa_desc = GuestAddr(desc_addr);
-        // 16-bit flags, followed by 16-bit idx, followed by avail desc ring
+        // 16-bit flags, followed by 16-bit idx, avail desc ring
+        // (u16[sz]), followed by 16-bit used_event.
         self.gpa_flags = GuestAddr(avail_addr);
         self.gpa_idx = GuestAddr(avail_addr + 2);
         self.gpa_ring = GuestAddr(avail_addr + 4);
+        self.gpa_used_event = GuestAddr(
+            avail_addr + (2 + sz as u64) * mem::size_of::<u16>() as u64,
+        );
```
FelixMcFelix:
```suggestion
        self.gpa_used_event = self.gpa_ring.offset::<u16>(usize::from(sz));
```

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -221,11 +237,15 @@ impl VqUsed {
...
-        // 16-bit flags, followed by 16-bit idx, followed by used desc ring
+
+    fn map_split(&mut self, gpa: u64, sz: u16) {
+        // 16-bit flags, followed by 16-bit idx, followed by used desc
+        // ring, followed by avail_event
         self.gpa_flags = GuestAddr(gpa);
         self.gpa_idx = GuestAddr(gpa + 2);
         self.gpa_ring = GuestAddr(gpa + 4);
+        self.gpa_avail_event =
+            GuestAddr(gpa + 4 + mem::size_of::<VqdUsed>() as u64 * sz as u64);
```
FelixMcFelix:
```suggestion
        self.gpa_avail_event = self.gpa_ring.offset::<VqdUsed>(usize::from(sz));
```

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -570,28 +605,123 @@ impl VirtQueue {
     /// Disables interrupts (notifications) on the `Used` ring.
     ///
     /// Returns `true` if notifications were previously enabled.
```
FelixMcFelix: We should probably drop this part of the comment.

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -570,28 +605,123 @@ impl VirtQueue {
...
+        used: &mut VqUsed,
+        mem: &MemCtx,
+    ) -> bool {
+        if self.f_event_idx() {
+            let used_event = Wrapping(avail.used_event(mem));
+            let uidx = used.used_idx;
+            let last_chk_uidx = used.last_chk_uidx;
+            let res =
+                (uidx - used_event - Wrapping(1)) < (uidx - last_chk_uidx);
+            used.last_chk_uidx = uidx;
```
FelixMcFelix, line 671: I've tried to double check this on paper, since wrapping arithmetic always throws me for a loop. My understanding is that we want to use `last_chk_uidx` as our effective zero, and we call always this *after* writing descriptors and incrementing `used_idx`. This gives us:
* The number of events since we last called this method, `x = uidx - last_chk_uidx` (modulo 2^16),
* The watermark relative to that zero to send an interrupt, `y = used_event + 1 - last_chk_uidx` modulo 2^16 (the spec says to signal if `idx` *was* equal to `used_event`, so we either reduce `uidx` or raise the watermark).

So we want to signal when `x >= y` -> `uidx - last_chk_uidx >= used_event + 1 - last_chk_uidx` -> `uidx - last_chk_uidx > used_event - last_chk_uidx` (modulo 2^16)? I'm unsure if I've gotten that wrong, but otherwise it's unclear to me what the quantity on the LHS of `res` currently represents without a comment.

FelixMcFelix (commented):
Thanks Ryan, Double checking some things to make sure I understand what's going on with the wrapping math 100%.

lib/propolis/src/hw/virtio/queue.rs
rzezeski: This code is simply following what's in the spec:

```
static inline int virtq_need_event(uint16_t event_idx, uint16_t new_idx, uint16_t old_idx) { 
    return (uint16_t)(new_idx - event_idx - 1) < (uint16_t)(new_idx - old_idx); 
}
```

* `uidx == new_idx`
* `last_chk_uidx == old_idx`
* `used_event = event_idx`

It's hard for me to explain the math itself, but I think it is some form of normalization (probably not the right term) that allows us to find out if the `event_idx/used_event` is contained inside the range of the published entries, which would indicate that the `event_idx` was crossed.

I can't really follow your example because I feel like you mixed up the equation. I think it would be `y = uidx - (used_event + 1)`.

lib/propolis/src/hw/virtio/queue.rs
rzezeski: Done. Also updated the comments for these two functions to be less vague.

lib/propolis/src/hw/virtio/queue.rs
rzezeski: Done

lib/propolis/src/hw/virtio/queue.rs
rzezeski: Done

lib/propolis/src/hw/virtio/queue.rs
FelixMcFelix: I did a quick test run on rust-playground and, at least for all choices of `new: u8, event: u8, old: u8` my expression above has the same truth table as that in the viona spec. I can't see the derivation yet but I am now fully confident here.

Issue #1170, opened 2026-07-10
nvme: volatile write cache presence should be configurable


iximeow:
I've been doing some looking at local storage I/O latency and one nice improvement will be to only have the emulated NVMe controller report a volatile write cache if that reflects the semantics of the backing medium. this is correct for Crucible storage, as Crucible fast-acks the writes, and is probably necessary for file-backed disks backed by .. normal files.

for file-backed disks using raw zvols (as in the case of local storage in the product), we use `/dev/zvol/rdsk/` paths which are not buffered. at that point, if the underlying storage device does not have a write cache, we the guest's NVMe disk doesn't need to report one either. the OS only completes the write once the underlying disk has completed the write, and Propolis only completes the write once the OS has completed the write, so when the guest sees the write complete then the device which promises the write is persistent has done the job.

I've done some fishing around in guest OSes and on the Linux side the OS short-circuits flushes directed towards block devices "without write caches" at `blk_insert_flush` in `blk-flush.c`, with a morally similar short-circuiting happening in illumos at `nvme_bd_sync` in `common/io/nvme/nvme.c`. some quick checking by simply clearing the `vwc` bit and rebuilding propolis-standalone show that workloads heavy on sync() are immediately ~3x faster (slightly more than I expected!), so without `vwc` guests do clearly do less work.

I don't think Propolis should try to decide if vwc is correct or not given a particular instance spec. for development (like my workstation, CI, etc) I probably shouldn't need a physical device without vwc to test how guests behave in such cases. more importantly in the product we may want to still report a write cache when hardware "doesn't have one". this is morally similar to how we may want to hide CPUID features for platform compatibility and do so by decision in Nexus, though not migrating local storage simplifies things somewhat.

(I'm already putting together a change for this one, it's fairly tiny all-told)

iximeow:
it is now! 1178 gets this configurable in the propolis-server API, and plumbs that to propolis-cli / toml specs (and propolis-standalone). see Omicron#10932 for the rest of the product plumbing.

Issue #1136, opened 2026-05-05
propolis-server: migration needs to learn about spec versions


iximeow:
I thought I had gotten these notes down, but I can't find them and it's not on the migration board! there are some relevant bits of `lib/migrate` to look at in the context of the HTTP API versioning work that has since landed:
https://github.com/oxidecomputer/propolis/blob/9343361f922d10c5825906a69c7bb05d139a3662/bin/propolis-server/src/lib/migrate/preamble.rs#L16-L53

and

https://github.com/oxidecomputer/propolis/blob/9343361f922d10c5825906a69c7bb05d139a3662/bin/propolis-server/src/lib/migrate/source.rs#L468-L473

these together mean that on the recipient side, `propolis-server` only knows how to accept a `v1::VersionedInstanceSpec::V0` and that spec must be whatever you can get from converting the active VM's `Spec` down to an `v1::InstanceSpec`. in turn, this means that instance specs must be convertible _to_ a `v1::InstanceSpec`. this all should be alarming, and maybe make your skin crawl.

ideally we can handle migration with an overall strategy like:

* try converting the current VM's `Spec` to each API type, with conversions as `TryFrom`, starting from the latest version.
* one of these should work; the spec came from the API at some point[1], so we should be able to convert back to one of those types.
* send a migration Prelude with that converted spec and a version indicator, something like `VersionedInstanceSpec` was intended to handle.
* ... and then handle device payloads and versioning thereof.

### converting from a current `Spec` down to `v1::InstanceSpec` is lossy! we lose information in migrating out!

it is, and [this is where the lossy conversion happens](https://github.com/oxidecomputer/propolis/blob/9343361f922d10c5825906a69c7bb05d139a3662/bin/propolis-server/src/lib/spec/api_spec_v0.rs#L59-L80).

- [x] we should eliminate `impl From<Spec> for v1::instance_spec::InstanceSpec`.

in general we can't `impl From<Spec> for [HTTP API Spec]`, because the API type is not guaranteed to be backwards compatible - it's possible we decide to remove a field from a device's configuration, for example.

### migration does need an alternative to `propolis_api_types::*::instance_spec::VersionedInstanceSpec`

[...]

Issue #1131, opened 2026-04-28
migrate-from-base tests have gotten long in the tooth


iximeow:
from the horse's mouth: https://github.com/oxidecomputer/propolis/blob/96e772ab174bb38897c42aef6951fbe7b3e40642/.github/buildomat/jobs/phd-run-migrate-from-base.sh#L16-L25

now, a test migrating from mainline propolis-server to the version under test (and back!) _is_ useful, so unlike the comment remarks, I haven't been particularly interested in _removing_ the test. but a few things have changed that make this test less useful than we'd want here..

first, we've stabilized HTTP API versioning across services. now, adding a new HTTP API version to Propolis implies this test _will_ fail, until the new HTTP API version is landed, which would get `master` and the build under test speaking the same version again. other tests that the version under test can migrate to and from _itself_ probably still pass in those cases! but by adding a new HTTP API version N, migrating to a propolis that only knows version HTTP API N-1 means the old `propolis-server` just returns an error and the test fails.

perhaps we should have `migrate-from-base` skip when we're bumping the HTTP API, and as a matter of policy(?) require that HTTP API plumbing not be changed alongside changes to propolis-standalone/propolis-server themselves? "policy" is a pretty subjective and unsatisfactory bar!

second, a VM can have a more storied life, especially once we start actually doing live migration. a VM first incarnated in a Propolis using the V1 API may live-migrate to a Propolis with a V2 InstanceSpec, which may live-migrate to a Propolis with a V3 instance spec. testing that _this_ can migrate to the version under test is important, and is currently untested.

migrating forward along this path should always work, regardless of API version that the propolis-server under test would use in talking to other Propolises.

PR #1178, opened 2026-07-18
untangle migration and instance spec versioning (+ configurable NVMe VWC bit)

iximeow:
this is something of an unfortunate two-fer of a change, in that this gets https://github.com/oxidecomputer/propolis/issues/1136 in service of making https://github.com/oxidecomputer/propolis/issues/1170 straightforward. I'd started by wanting to plumb the volatile write cache bit out to the `propolis-server` API, but I really couldn't stomach playing games with that bit in a `From` which really should be a _lossless_ conversion, so this evolved into not having incorrect `From` between `v1::instance_spec::InstanceSpec` and `propolis-server`'s `Spec`. at that point fixing instance versioning as handled by migration was trivial, so I did that here too.

in-all I think these together help explain why there are some mighty funky-looking names and impls in here: changing the structure of an existing device definition pushes the types and conversions in a way that wasn't super obvious to me at the outset, so reviewing this minus the new `v6` API would have been pretty confusing at least to me.

along the way I've taken this as an excuse to write down a bunch of thoughts about this approach vs some alternatives in `bin/propolis-server/src/lib/migrate/types.rs`.

the whole structure here could use proptests around spec transformation _actually_ being lossless. this doesn't get any of the testing bits out of the way in PHD (though now we can use new features in phd and migrate them successfully!). and this is still a draft because I've missed some obvious plumbing as it relates to getting `has_write_cache` through `propolis-cli` and into an `InstanceSpec`. but if anyone wants to take an early look, here it is!


bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,372 @@
...
+/// Architecturally, this bridges the very fixed HTTP API types and the
+/// possibility of having to migrate an arbitrarily old VM. See the doc comments
+/// on [`migrate`][crate::lib::migrate] for more about how this all fits
+/// together.
+//
+// If you're adding (or removing!?) API versions, you'll just want to adjust the
+// variants here, plus uses in `VersionedInstanceSpec::from_spec` and
+// `VersionedInstanceSpec::into_amended_spec`. shrimple as that.
+#[derive(Deserialize, Serialize, Debug)]
+pub(crate) enum VersionedInstanceSpec {
```
iximeow, line 111: this file and the functions on this enum are really the ones I'd like to hear from folks about, re. the intersection of API versioning and migration. if you are looking at this and are at all unsure how you'd add a new version, please shout.

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,372 @@
...
+    V1(v1::instance_spec::InstanceSpec),
+    V2(v2::instance_spec::InstanceSpec),
```
iximeow, line 113: (I need to figure out if v2 should or shouldn't be here. there's an InstanceSpec, so I think so, but my notes say that it maybe didn't end up in the API as such so we never actually have to express this rather than v1 or v3?)

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,372 @@
...
+        fn wrong_type_error(
+            id: &v1::instance_spec::SpecKey,
+            kind: &str,
+        ) -> MigrateError {
+            let msg =
+                format!("component {id} is not a {kind} in the source spec");
+            MigrateError::InstanceSpecsIncompatible(msg)
+        }
+
+        let amended_spec = match self {
```
iximeow, line 160: this match is kind of atrocious and I don't like what the future holds for it. mayyyyybe we can get a way with a per-API-version `amend()` that takes the replacements list and does the needful per-InstanceSpec? that at least means this grows only by a few lines per version with each `api_spec_*.rs` locked in amber.

bin/propolis-server/src/lib/spec/api_spec_v3.rs
```diff
@@ -0,0 +1,342 @@
...
+    fn from(value: StorageBackend) -> Self {
+        match value {
+            StorageBackend::Crucible(be) => Self::CrucibleStorageBackend(be),
+            StorageBackend::File(be) => Self::FileStorageBackend(be),
+            StorageBackend::Blob(be) => Self::BlobStorageBackend(be),
+        }
+    }
+}
+
+impl TryFrom<Spec> for v3::instance_spec::InstanceSpec {
```
iximeow, line 18: I don't _love_ having so much duplicate glue for each rev of `InstanceSpec`/`Component`,  but I think it's the least bad option.. open to ideas

bin/propolis-server/src/lib/migrate/types.rs
iximeow: this is somewhat less gross with the per-API-version bits pulled into the various `api_spec_v*` modules.. (in https://github.com/oxidecomputer/propolis/pull/1178/commits/a0dc3e5586d50e8b862894058220b0b8fd9b206b )

bin/propolis-server/src/lib/migrate/types.rs
iximeow: this was: I knew that there was no `v2::instance_spec::Component`, because we didn't change that in v2, but there is a new InstanceSpec. this makes the conversions to/from v2 trivial and the file is correspondingly not even 100 lines of glue.

bin/propolis-server/src/lib/spec/api_spec_v6.rs
```diff
@@ -0,0 +1,371 @@
...
+            key: SpecKey,
+            val: v6::instance_spec::Component,
+        ) {
+            assert!(
+                !spec.components.contains_key(&key),
+                "component name {} already exists in output spec",
+                &key
+            );
+            spec.components.insert(key, val);
+        }
```
iximeow, line 68: so.. `InstanceSpec->Spec` conversions work out kinda nicely with the approach of "remove whatever's new, convert it to the previous version, then in v1 produce a SpecBuilder we tack stuff onto in each revision".

this is me trying out the reverse for `Spec->InstanceSpec` and I'm really not sure how I feel about it. for v1, v2, and v3 this means the `TryFrom<Spec> for vWhatever::instance_spec::InstanceSpec` is legible, and I really like that! but for the latest version, where the conversion is realistically infallible, there's the really unsettling `unreachable!()` that will become reachable in a future v7 I'm sure.

there's _also_ `insert_component` which I don't see a way to _not_ have copies of for each version that touches components. since `SpecBuilder` is kinda the GCD of all instance specs we can just use it. but here we're editing some specific `InstanceSpec` with that spec's `Component`, so I think we end up having either this here, or
```rust
impl InstanceSpec {
    fn insert_component(&mut self, key, component) -> blah blah
}
```
on each version of InstanceSpec.

this is probably the lesser of the available evils? in any case this all needs proptests that the conversions are lossless in either direction, don't panic, etc. if I had that here I think I wouldn't care about this unreachable!(), because we'd just find it when adding new versions into proptests. maybe this is fine until then.

bin/propolis-server/src/lib/spec/api_spec_v3.rs
iximeow: I've since de-glued things and I think where this is at only makes sense with a proptest for the conversions, but with one is much better..

bin/propolis-server/src/lib/spec/builder.rs
```diff
@@ -71,6 +71,18 @@ pub(crate) enum SpecBuilderError {
...
+/// forcing function to canonicalize VM descriptions, where we can enforce
+/// invariants about components (such as "disk devices must reference backends
+/// that exist").
+///
+/// Note that the API type `Component` itself does not appear here: the
+/// expectation is that individual components' definitions change relatively
+/// rarely, so callers do the work of mapping components to the
+/// closer-to-internal definitions that `SpecBuilder` accepts. In theory,
+/// hopefully, this means `SpecBuilder` itself changes rarely and can be more
+/// reasily audited for semantic drift.
```
lgfa29:
```suggestion
/// easily audited for semantic drift.
```

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,198 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## Types to describe a VM which is being migrated.
+//!
+//! The API types that describe a VM are converted by `propolis-server` into a
+//! [`struct Spec`][crate::lib::spec::Spec] as a "how this version happens to
+//! describe VMs" internal structure. Early in migration we must convert this to
+//! some format that a `propolis-server` of a different version can instantiate
+//! an equivalent VM from, for device state everything else to be imported into.
```
lgfa29: > for device state everything else to be imported into.

Minor note, but I think there might be word missing here? I didn't quite understand this sentence.

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,198 @@
...
+                        MigrateError::PreambleParse(e.to_string())
+                    })?;
+
+                amended_spec
+            }
+            VersionedInstanceSpec::V2(mut source_spec) => {
+                api_spec_v2::amend(&mut source_spec, replacements)?;
+
+                let amended_spec: Spec =
+                    source_spec.try_into().map_err(|e: V1SpecError| {
```
lgfa29:
```suggestion
                    source_spec.try_into().map_err(|e: V2SpecError| {
```

Minor thing, so feel free to ignore. I think these are the same, but maybe consider explicitly using the same version to keep things consistent? Or would using `V2SpecError` cause other problems down the line? At first, this kind of looked like a copy-paste error.

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,198 @@
...
+                        MigrateError::PreambleParse(e.to_string())
+                    })?;
+
+                amended_spec
+            }
+            VersionedInstanceSpec::V3(mut source_spec) => {
+                api_spec_v3::amend(&mut source_spec, replacements)?;
+
+                let v6_spec: v6::instance_spec::InstanceSpec =
+                    source_spec.into();
```
lgfa29: I haven't read the details on amending specs for each version yet, so maybe the answer is somewhere there, but it could be useful to have a quick blurb on why we convert a v3 to a v6.

bin/propolis-server/src/lib/spec/api_spec_v1.rs
```diff
@@ -50,10 +49,46 @@ pub(crate) enum ApiSpecError {
 
     #[error("backend {0} not used by any device")]
     BackendNotUsed(SpecKey),
+
+    #[error("spec contains v1-incompatible component: {0}")]
```
lgfa29: I may be misunderstanding this, but since this enum is shared across versions, would every error message read `v1-incomplatible` even if attempting to transform it to other versions?

bin/propolis-server/src/lib/spec/api_spec_v2.rs
```diff
@@ -0,0 +1,79 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Conversions from the initial API version ([`propolis_api_types::v1`], aka
```
lgfa29:
```suggestion
//! Conversions from the initial API version ([`propolis_api_types::v2`], aka
```
I think?

bin/propolis-server/src/lib/spec/api_spec_v1.rs
```diff
@@ -254,175 +323,77 @@ impl TryFrom<v1::instance_spec::InstanceSpec> for Spec {
...
-                builder.set_p9fs(p9fs)?;
-            }
-            v1::instance_spec::Component::CrucibleStorageBackend(_)
-            | v1::instance_spec::Component::FileStorageBackend(_)
-            | v1::instance_spec::Component::BlobStorageBackend(_)
-            | v1::instance_spec::Component::VirtioNetworkBackend(_)
-            | v1::instance_spec::Component::DlpiNetworkBackend(_) => {
-                unreachable!("already filtered out backends")
-            }
+    crate::spec::api_spec_v3::v3_to_spec_builder(v3_spec)
```
lgfa29, line 306: This is mostly inferred from reading the code in isolation, so I may be missing hands-on experience like the compiler screaming at me. Feel free to ignore it if this doesn't make sense.

But it's not immediately obvious to me why we only go to v3 here. Looking at `v3_to_spec_builder` it says:

```rust
// Converting v3 to v6 is lossless so just do that and piggyback on the
// v6 `InstanceSpec->SpecBuilder`.
```

So I guess that's the answer?  But then I'm not sure if I would need to add a v7 conversion here if I added a new spec version.

I'm also not sure why we even need v3 here? Would this be an equivalent conversion?

```rust
pub(crate) fn v1_to_spec_builder(
    value: v1::instance_spec::InstanceSpec,
) -> Result<SpecBuilder, ApiSpecError> {
    let v2_spec: v2::instance_spec::InstanceSpec = value.into();

    crate::spec::api_spec_v2::v2_to_spec_builder(v2_spec)
}
```

And so the logic is effectively that each version just converts to the next, going up the chain until we get the latest version of a `SpecBuilder`.

bin/propolis-server/src/lib/spec/api_spec_v2.rs
```diff
@@ -0,0 +1,79 @@
...
+        value: v2::instance_spec::InstanceSpec,
+    ) -> Result<Self, Self::Error> {
+        Ok(v2_to_spec_builder(value)?.finish())
+    }
+}
+
+/// Parses a v1 instance spec into a [`SpecBuilder`], validating component
+/// names, PCI paths, and backend references along the way. Callers can add
+/// additional (non-v1) components to the builder before calling `finish()`.
+pub(crate) fn v2_to_spec_builder(
```
lgfa29:
```suggestion
/// Parses a v2 instance spec into a [`SpecBuilder`], validating component
/// names, PCI paths, and backend references along the way. Callers can add
/// additional (non-v2) components to the builder before calling `finish()`.
pub(crate) fn v2_to_spec_builder(
```

bin/propolis-server/src/lib/spec/api_spec_v6.rs
lgfa29, line 1: The way I understood this is that we have a chain of calls to convert specs backwards and forwards:

Image

This is pretty neat, and makes sense to me, but there are two aspects that I found a little confusing when going through a mental exercise of "what would I need to change to add a v7".

The first one is `ApiSpecError`. Each version seems to have its own, which is actually reused from other versions. This is fine, but I don't think it actually represents what they intend to communicate?

Looking at their definitions, the only difference is that `v1::ApiSpecError` has `IncompatibleComponent`, which can only happen when going backwards in the conversion, and `v6::ApiSpecError` can only happen when going forwards. 

If I were to add a v7, would I need to move `v6::ApiSpecError` to `v7::ApiSpecError`? Or would `v6::ApiSpecError` always stay there and be reused in v7?

Maybe the error types could be more generic and version independent, or maybe we don't actually need two different error types? In case we do need versioned errors, it may be useful to add some note on how they should evolve when new versions are added.

The second question I had was with regards to `v6_to_spec_builder`. Would a v7 need to copy this method over and update `v6_to_spec_builder` to call it?

If so, maybe we can take a note from the HTTP API versioning and have an `api_spec_latest` where this method is defined. And so adding a v7 means I just update the method in `api_spec_latest`, call `v7_to_spec_builder` from `v6_to_spec_builder` and `vlatest_to_spec_builder` from `v7_to_spec_builder`. Looking at the diagram, it would be like adding an anchor node at the end of the linked list with the latest spec conversion logic.

bin/propolis-server/src/lib/spec/api_spec_v6.rs
lgfa29: Hum...I don't have much to add here other that saying that this read fine to me. We may need to wait for a v7 to see how it holds-up, but it seems OK for now.

lgfa29 (approved):
Some minor notes and comments that are mostly me trying to understand things, but from what I do understand, this seems like a nice implementation to a hairy problem 😄

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,198 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! ## Types to describe a VM which is being migrated.
+//!
+//! The API types that describe a VM are converted by `propolis-server` into a
+//! [`struct Spec`][crate::lib::spec::Spec] as a "how this version happens to
+//! describe VMs" internal structure. Early in migration we must convert this to
+//! some format that a `propolis-server` of a different version can instantiate
+//! an equivalent VM from, for device state everything else to be imported into.
+//! We *kind of* use API types here, and the rest of this section gets into why
```
hawkw: i read this sentence and immediately thought to myself, "oh no..."

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,198 @@
...
+//! and what one should consider in adding future versions.
+//!
+//! Even for VMs that have been migrated many times, `propolis-server` must
+//! incarnate a VM that can be described by *some* HTTP API `InstanceSpec`
```
hawkw: maybe 
```suggestion
//! incarnate a VM that has been described by *some* HTTP API `InstanceSpec`
```
to make it a little more obvious that what you're saying is that, "if you were able to ever start the VM, this means that you have sent a start request with its spec AT SOME POINT BETWEEN NOW AND THE BEGINNING OF THE UNIVERSE"

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,198 @@
...
+//! version at some point in the past. We'll call this "oldest possible VM spec"
```
hawkw, line 17: shades of https://en.wikipedia.org/wiki/Last_universal_common_ancestor

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,198 @@
...
+//! the "import horizon" that `propolis-server` supports.  Further, the tooling
```
hawkw: nit: since we are defining a new piece of jargon which i assume you intend to use over and over again, i think we should italicize this:
```suggestion
//! the _import horizon_ that `propolis-server` supports.  Further, the tooling
```

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,198 @@
...
+//! This points us away from the naive "`propolis-server` simply produces an
+//! InstanceSpec from this version or last version" strategy one may imagine,
+//! and towards keeping HTTP API types around well after those HTTP API
+//! *endpoints* have been retired!
+//!
+//! So, we're finally at why the existing mechanics of getting a `Spec` out of
+//! `propolis-server` and transmitted are what they are: go through a list of
+//! `TryInto<Spec> for v*::instance_spec::InstanceSpec`, one of them will
+//! succeed, and send that over. This is the implementation you'll find in
+//! [`RonV0Runner::sync`][crate::lib::migrate::source::RonV0Runner::sync].
```
hawkw: nit: i think this link is broken and should instead be
```suggestion
//! [`RonV0Runner::sync`](crate::lib::migrate::source::RonV0Runner::sync).
```

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,198 @@
...
+//! conversion will fail to be downgradeable is if a VM has been created using
+//! only-in-newest API language. This means that some VMs created using a
+//! `latest::instance_spec::InstanceSpec` could end up with even `v1` types on
+//! the wire for migration, but as long as `From/TryFrom` use is correct and
+//! *not lossy*, that's fine!
+//!
+//! So, `VersionedInstanceSpec` is a container that is outside the HTTP API but
+//! only contains OpenAPI-described API types. A destination `propolis-server`
+//! is expected to gracefully reject new variants, and a source
+//! `propolis-server` is expected to emit oldest-supported forms of instances.
```
hawkw, line 66: i think this makes sense, yeah!

hawkw (commented):
just started reading through this with the doc comment, here are some small notes while i look at the way i tactually works

bin/propolis-server/src/lib/spec/builder.rs
iximeow: i've mangled readily and easily... oops :D

bin/propolis-server/src/lib/spec/api_spec_v1.rs
iximeow: you understand right! I was wondering about this as I wrote it. I think it's not terrible but not _wrong_ in a very literal sense. `api_spec_v1` is the only place we're _producing_ `ApiSpecError::IncompatibleComponent`, so whenever this happens it's true that the thing is v1-incompatible. the weirdest case is if you have a v6 NvmeDisk without a write cache. then that fails to convert to a v3 Component, which fails to convert to a v1 component.

if that is written out as a string it's something like: `spec contains v1-incompatible component: cannot convert component to v3: NvmeDisk with has_write_cache=false cannot be downgraded`. so.. we failed to convert the thing to v1, because we couldn't get it down to v3.

the guidance here, I think, is "if you need to produce an `ApiSpecError::<anything>`, define a new version of the enum". that happens to match the current use, where when the type name is written out it's just to say "I'm returning the error that other function returns" (except the v6->v3 conversion in api_spec_v3.rs)

bin/propolis-server/src/lib/spec/api_spec_v2.rs
iximeow: ow, yeah, and this isn't the initial API  version anymore. copy fail

bin/propolis-server/src/lib/spec/api_spec_v1.rs
iximeow: > I'm also not sure why we even need v3 here? Would this be an equivalent conversion?

that's right. this goes to v3 because I'd initially _not_ written in `api_spec_v2.rs`, because v2 was in a weird space of adding a new `InstanceSpec` type but _not_ a new `Component`. so I'd thought maybe we didn't need an `api_spec_v2.rs` at all. I argued myself into adding it, but didn't fix this forward conversion.

so yeah, the logic is each version just goes to the next `api_spec_v*.rs` until one of the conversions is fallible, which would probably mean we've deprecated a setting/component/etc. we haven't done that yet, so it's all infallible going to newer versions.

bin/propolis-server/src/lib/migrate/types.rs
iximeow: yeah, I was being a little cute with the wording because one possibility here is you start a VM with settings in API version 6 that are also describable in terms of API version 1. there's no implication it was ever _actually_ described in V1 terms until you discover it happens to be possible at migration time. that distinction isn't terribly important and I think `has been` reads much more naturally here. i might do a footnote or something.

bin/propolis-server/src/lib/migrate/types.rs
iximeow: so.. these _could_ be the same, but they're different in ways that I think end up differently confusing. to be able to write this, in `api_spec_v2.rs` you'd have to `pub use crate::api_spec_v1::ApiSpecError;` so you can write `use api_spec_v2::ApiSpecError as V2SpecError` here. that means we've got the same type available through two different crates and I know if I were chasing down the types I'd think they were different until I've followed the re-imports t o discover it's all just `V1SpecError`.

otoh, if this _was_ a copy-paste error, then to type check we'd have to have a `TryFrom<Spec> for v2::instance_spec::InstanceSpec { type = V1SpecError; .. }`. that's certainly something you could write, but at the least it would be a few places copy-paste errored together. and the actual impl would have to convert from V2 errors to V1. which since they're all the same that's fine here! but if it were a meaningful difference the type checker would have lots of reason to shoot down simple errors.

the bone I have to pick here, which might be part of why this read like a copy-paste error to you, is `.try_into().map_err()` doesn't tell you anything about what you're converting. it's all type inferenced glue. the other direction has explicit `TryInto::<v1::instance_spec::InstanceSpec>::try_into()` which at least helps me a lot..

I've taken a swing at [moving this to `vN_to_spec_builder()`](https://github.com/oxidecomputer/propolis/pull/1178/changes/cfdcc6baeb39c394bff0c38ada2a5fa77c88bc33#diff-44d71bfaa9b3b5c343a8931cfb991add42a8b8ff02e966d49eadf33e196a92baL154) which does a lot of things for us at once: the error type is concrete rather than an associated type, so rustc doesn't need it explicitly written at all. the input spec type is part of the function name, so it's a lot clearer what we're coming _from_. and since it's Just A Function it's not interesting to annotate the returned `amended_spec` type, so I just get the builder, convert the error, and `finish()` into a spec.

bin/propolis-server/src/lib/migrate/types.rs
iximeow: yeah i've kind of mangled English here. hopefully [this rephrasing](https://github.com/oxidecomputer/propolis/pull/1178/changes/cfdcc6baeb39c394bff0c38ada2a5fa77c88bc33#diff-44d71bfaa9b3b5c343a8931cfb991add42a8b8ff02e966d49eadf33e196a92baL11) is more legible?

bin/propolis-server/src/lib/spec/api_spec_v6.rs
iximeow: > If I were to add a v7, would I need to move v6::ApiSpecError to v7::ApiSpecError? Or would v6::ApiSpecError always stay there and be reused in v7?

ohhh.. yeah. you would probably move it _and then use v1::ApiErrorSpec in v6_. that super funky. I didn't want to commit to a generic "ApiSpecError" type because that would imply (from the types at least) that there could be a v6-incompatible component in a Spec. this probably works out much more legibly if there's a `latest` error type that doesn't know about incompatible components, and the v1 ApiErrorSpec which we use for later versions until we have some kind of novel error case in processing an InstanceSpec.

> with regards to v6_to_spec_builder. Would a v7 need to copy this method over and update v6_to_spec_builder to call it?

currently, yes. I bet that moving all this into a "api_spec_latest.rs" makes it make much more sense. I'll give that a try!

bin/propolis-server/src/lib/migrate/types.rs
```diff
@@ -0,0 +1,189 @@
...
+        } else if let Ok(v3_spec) =
+            TryInto::<v3::instance_spec::InstanceSpec>::try_into(spec.clone())
+        {
+            VersionedInstanceSpec::V3(v3_spec)
+        } else {
+            VersionedInstanceSpec::V6(
+                Into::<v6::instance_spec::InstanceSpec>::into(spec.clone()),
+            )
+        };
+        Ok(versioned)
```
iximeow, line 150: some thoughts from talking with James earlier: if _any_ previous `TryFrom` _incorrectly_ succeeded in accepting the current `Spec`, we can lose information that might have needed newer API terminology to migrate. starting from the oldest API type and going version to version means that for each new API version we're giving ourselves `O(versions)` ways to lose information when exporting out a Spec to wire types.

none of that is great, and in that lens this is about the riskiest way to write `VersionedInstanceSpec::from_spec`. particularly because in reality the only versions we're going from or to are the Propolises one version apart in released-rack-software-package terms. that might be `Propolis v20 <-> Propolis v25` if we have a few Propolis changes inside one release. but it's a very reduced set of conversions compared to "everything up to v25".

you could imagine we do something morally like Crucible, where when we connect to the other end we do a bit of version negotiation to ask the most recent version _they_ support, and try from converting to `v<them>` and older until one works. that would mean migration of VMs including latest-API items don't have to fail through O(versions) conversions before succeeding, at the very least.

proptests around the From/TryFrom impls would probably help on the "this is risky" front, at least.

bin/propolis-server/src/lib/migrate/types.rs
lgfa29: Yes, thank you!

bin/propolis-server/src/lib/migrate/types.rs
lgfa29: Oh nice. Yeah, I figured the compiler may yell at you if you use the wrong type, so not a real concern, but reading the new code it makes more sense to me.

bin/propolis-server/src/lib/spec/api_spec_v6.rs
iximeow: https://github.com/oxidecomputer/propolis/pull/1178/changes/64c032f46a67a7532602b38fa23990cff6de3c1e hopefully makes this a bit better all around: converting from the latest `InstanceSpec` to a `Spec` lives in `api_spec_latest.rs`, which is really just that conversion. that will just track "latest" now and into the future. if we have a new version where for some reason we can't go from `v<N-1>` to `v<N>`, we'll have to figure out either if we should copy-and-edit that conversion into `v<N-1>` or .. do something else.

converting a `Spec` into a `v6::instance_spec::InstanceSpec`, though, won't ever have something to copy into a newer API version. so it stays in `api_spec_v6`. that means we have both `api_spec_v<latest>` and `api_spec_latest`, but at least in my mind the line of "relevant to the specific `v<latest>`" and "relevant to the latest version" helps make that split not _super_ confusing..?

along the way I've pulled up `ApiSpecError` out of individual versions. it seems likely to me that `ApiSpecError` will probably be OK for the forseeable future in converting the latest spec - it will only grow new variants if we have new invariants on InstanceSpec that aren't expressed as types (like, "boot options disks in the spec"). "LegacyApiSpecError" is the new name for "ApiSpecError and also we can't convert a member of Spec to this old type", which I think maybe will work out for basically the same reason.

I also spent some time trying to figure out if there's a nice way to spell `LegacyApiSpecError::IncompatibleComponent` that gets a nice error message. the least confusing option I think is a `target_ver` field on the error to say "you were trying to convert to vX and this device can't. here's why: ..." . but that means if you've produced a `LegacyApiSpecError::IncompatibleComponent` from a conversion in v1, and the overall conversion is in support of converting to a v3 type, _something_ has to take the error and "rewrite" the target version.

`old_spec.try_into().map_err(|e| e.with_target_ver(3))?` seems _extremely_ to overlook in review if you copy-paste, and would mean the errors go from being a little confusing ("why did the conversion fail in v1? well, here's the error string, lets go look") to outright misleading ("this conversion failed in v2? but I was converting to v3, and the error message is from `api_spec_v1.rs`") so... I unwound that idea once I realized how fraught it is :( it could be less fiddly if you could write `.try_into().map_err(|e| e.set_target_ver(Self::API_VERSION))` but I'm not convinced that it gets us in a much less confusing place at the end of the day..

iximeow:
(my plan at this point: solicit any remaining opinions tomorrow, hopefully whack the button, then update Omicron #10932 to the squashed commit, make that a real PR, and onwards we go! `migrate-from-base` will fail because this adds a new API version and https://github.com/oxidecomputer/propolis/issues/1131 is yet unsolved)

bin/propolis-server/src/lib/spec/api_spec_latest.rs
```diff
@@ -0,0 +1,211 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Conversions from [`propolis_api_types::v6`] instance specs in the
```
lgfa29:
```suggestion
//! Conversions from [`propolis_api_types_versions::latest`] instance specs in the
```

I think the other doc links need to be updated to `propolis_api_types_versions` as well.

bin/propolis-server/src/lib/spec/mod.rs
```diff
@@ -13,10 +13,27 @@
 //! wire-format types in the [`propolis_api_types`] crate. This, in turn, allows
 //! [`Spec`] and its component types to take forms that might otherwise be hard
 //! to change in a backward-compatible way.
+//!
+//! Types and operations here are copied as-needed into new verson-specific
+//! modules as needed, as new versions of the `propolis-server` HTTP API are
+//! added.
```
lgfa29:
```suggestion
//! Types and operations here are copied as-needed into new version-specific
//! modules, as new versions of the `propolis-server` HTTP API are
//! added.
```

bin/propolis-server/src/lib/spec/mod.rs
```diff
@@ -13,10 +13,27 @@
...
+//! new defintion for `InstanceSpec`, you will probably want to:
+//!
+//! * add a new `api_spec_vN` module for the new version, with conversions from
+//!   `Spec` to the `vN::instance_spec::InstanceSpec` via the previous version
+//!   and from `vN::instance_spec::InstanceSpec` to `Spec` via
+//!   `latest_to_spec_builder()`
+//! * adjust the formerly-latest module's `vN_to_spec_builder` to call
+//!   `vN_to_spec_builder` and (try) converting `InstanceSpec` upwards to match.
+//! * adjust the formerly-latest module's `Spec` to `InstanceSpec` conversion to
+//!   be fallible, using `LegacyApiSpecError` instead of `ApiSpecError`
```
lgfa29, line 33: Neat! Yeah, these steps makes sense to me.

bin/propolis-server/src/lib/spec/mod.rs
```diff
@@ -49,55 +66,69 @@ use propolis_api_types::instance_spec::components::{
...
-    fn from(val: Spec) -> Self {
-        let smbios = val.smbios_type1_input.clone();
-        let vsock = val.vsock.clone();
+/// An error that can arise in converting
+/// `propolis_api_types::instance_spec::InstanceSpec` to a propolis-server
+/// `Spec`. To date, this is a subset of the errors that can arise in converting
+/// older versions of `InstanceSpec` as well.
+///
+/// When adding a new `api_spec_v*`, it may be appropriate to either move this
+/// error type into the formerly-latest module and add a nwe top-level type, or
```
lgfa29:
```suggestion
/// error type into the formerly-latest module and add a new top-level type, or
```

bin/propolis-server/src/lib/spec/api_spec_v1.rs
lgfa29: Ah I see, yeah, so I think it's fine to keep it like this. I didn't realize that we would get an intermediate `cannot convert component to v3` error, that helps explain the path that failed. We're also the most likely consumer of these error messages, so we know what mean here.

bin/propolis-server/src/lib/spec/api_spec_v6.rs
lgfa29: Nice! Thank you for taking a shot at this. I think the new code makes more sense, at least to me 😄 

With regards to error handling, I commented on https://github.com/oxidecomputer/propolis/pull/1178#discussion_r3648958983, but I think conversion error message is OK as it is.

lgfa29 (approved):

bin/propolis-server/src/lib/spec/mod.rs
iximeow: ow, yeah. this and the other typo in https://github.com/oxidecomputer/propolis/pull/1178/commits/78811b1ecc4913f2faaf0abd317b721756c8dcff

bin/propolis-server/src/lib/spec/api_spec_latest.rs
iximeow: huh. I think you're right, and I don't really know why I can't get `cargo doc` to warn about it. https://github.com/oxidecomputer/propolis/pull/1178/changes/36076baaa14e2bc4d62947cd7c32f440a370fc97 for the other crates, but something is very weird about the intra-doc links in the library crate here. https://github.com/iximeow/propolis/commit/3975b18b93dfac852ebd7fadcd73997f304c0264 is me messing with it a bit; it'll complain about `www.not.a.link.com` not being in `<>` to make it a proper hyperlink, but won't complain about `[other::broken::link]` no matter what I do.

I'm a little concerned this is cargo being confused about the difference between the propolis-server binary and propolis-server library crates... :D very odd.

PR #1179, opened 2026-07-20
propolis-cli should send boot_order along

iximeow:
would you believe that ever since Propolis has been able to set a boot order in fwcfg, we've never tried to use it in `propolis-cli`? `propolis-standalone` is where I use it often, and of course Nexus just uses the API directly with none of this "CLI program" nonsense.

well, @leftwo ran into a case where a VM was busted because the disk's PCI device was changed between where the image was created and the bench cosmo where it was being run, and the solution there is ... set a boot order and boot the thing at least once. we pretty quickly found that propolis-cli wasn't doing the needful, and now it does.

### DIFF
--- a/crates/propolis-config-toml/src/lib.rs
+++ b/crates/propolis-config-toml/src/lib.rs
@@ -21,6 +21,9 @@ pub mod spec;
 // configuration will likely become more dynamic.
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct Config {
+    #[serde(default, rename = "main")]
+    pub machine_settings: MachineSettings,
+
     #[serde(default, rename = "pci_bridge")]
     pub pci_bridges: Vec<PciBridge>,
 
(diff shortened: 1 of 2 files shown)


crates/propolis-config-toml/src/lib.rs
```diff
@@ -48,6 +52,15 @@ impl Default for Config {
     }
 }
 
+/// Settings covering the VM "at large".
+///
+/// This corresponds to (and is a subset of) the `main` block understood by
+/// `propolis-standalone`
+#[derive(Serialize, Deserialize, Debug, PartialEq, Default)]
+pub struct MachineSettings {
+    pub boot_order: Option<Vec<String>>,
```
iximeow, line 61: this gets a whole struct because I've thought for a minute that yoinking the `cpus/memory/name` fields from a standalone .toml would be really convenient if you're trying to reuse the file in both tools. as it turns out the toml is and has been a valid subset of the settings you'd need for the CLI, so my own `/rpool/devel/images/` is full of `"*.toml.standalone"` which often end up used against propolis-server...

every time I have to open one up to find `cpus =` and `memory = ` just to pass the CLI flags I think about this (which means about once every four months, to date)

leftwo (approved):
Thanks for fixing this

PR #1180, opened 2026-07-22
remove atty

iliana:
atty is unmaintained: https://github.com/advisories/GHSA-g98v-hv3f-hcfr

The only use is detecting if stdout is a terminal, which is supported in `std` since 1.70.

Not tested.


iximeow (approved):
get it outtya here (thanks)

PR #1186, opened 2026-07-28
Bump rust to 1.97.1

papertigers:
Bump rust-toolchain.toml to Rust version 1.97.1, and bump the MSRV to 1.93 so that we can drop the `VecDequeExt` trait which shamelessly copied the `pop_front_if` method from the rust standard library.


lib/propolis/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "propolis"
 version = "0.1.0"
 license = "MPL-2.0"
 edition = "2021"
-rust-version = "1.90"
+rust-version = "1.93"
```
papertigers, line 6: Without this we get the following at the `pop_front_if` call site
```
warning: current MSRV (Minimum Supported Rust Version) is `1.90.0` but this item is stable since `1.93.0`
```

bin/propolis-standalone/src/main.rs
```diff
@@ -406,14 +405,14 @@ impl Instance {
         match state {
             State::Run if first_boot => {
                 tokio::runtime::Handle::current().block_on(async {
-                    for (_name, be) in guard.inventory.block.iter() {
```
iximeow, line 409: now that's a good lint

iximeow (approved):
nice, thanks

.. though clippy has more words for phd, apparently

lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -520,7 +520,6 @@ impl VirtQueue {
                     false => ChainBuf::Readable(GuestAddr(desc.addr), desc.len),
                 };
 
-                count += 1;
```
jmpesp, line 523: this seems wrong, or at least an existing bug, where `count == self.size()` isn't considered past this point?

iximeow: it is an existing bug, yeah

papertigers: should we fix it here or in a PR stacked on top of this?

phd-tests/tests/src/boot_order/efi_utils.rs
```diff
@@ -198,7 +198,7 @@ impl DevicePath {
             }
             (ty, subtype) => {
                 bail!(
-                    "Device path type/subtype unsupported: ({ty:#02x}/{subtype:#02x})"
```
iximeow, line 201: i'm surprised by this one, does `:#02x` not format as `0x and "ty" as two hex digits padded with 0`?

papertigers: I was actually curious about this one as well, because the output was not what I expected.


This is a rust 1.97 lint I believe and didn't trigger on macOS with 1.96

```
sherma scratch
❯ cat src/main.rs
fn main() {
    let ty = 4u8;
    println!("{:#x}", ty);
    println!("{:#0x}", ty);
    println!("{:#02x}", ty);
    println!("{:#04x}", ty);
}

sherma scratch
❯ cargo clippy
    Checking scratch v0.1.0 (/Users/link/src/scratch)
warning: format width has no effect on the output for this format trait
 --> src/main.rs:5:15
  |
5 |     println!("{:#02x}", ty);
  |               ^^^^^^^
  |
  = help: consider removing the width or increasing it to at least 4
  = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#unused_format_specs
  = note: `#[warn(clippy::unused_format_specs)]` on by default

warning: `scratch` (bin "scratch") generated 1 warning
    Finished [`dev` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.04s

sherma scratch
❯ cargo r
    Finished [`dev` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.03s
     Running `target/debug/scratch`
0x4
0x4
0x4
0x04
```

[When used together with the # flag, a similar rule applies: padding zeros are inserted after the prefix but before the digits. The prefix is included in the total width.](https://doc.rust-lang.org/std/fmt/index.html#sign0)

papertigers: So, this preserves the old behavior but we could change it if you want.

lib/propolis/src/hw/virtio/queue.rs
papertigers: The "fix" is here #1188, however we really need to go do #1190 to address all of the issues in this code path.

phd-tests/tests/src/boot_order/efi_utils.rs
iximeow: oh i've written so many buggy format statements then. lmao. either way you want to take this here is fine imo.

PR #1188, opened 2026-07-29
validate number of indirect desc in pop_avail

papertigers:
The `pop_avail` function now validates that the number of indirect descriptors is not larger than the size of the virtio queue.

Stack:
- #1186


lib/propolis/src/hw/virtio/queue.rs
```diff
@@ -520,10 +520,16 @@ impl VirtQueue {
                     false => ChainBuf::Readable(GuestAddr(desc.addr), desc.len),
                 };
 
+                count += 1;
                 len += desc.len;
                 chain.push_buf(buf);
 
                 if flags.contains(DescFlag::NEXT) {
+                    if count == self.size() {
+                        // XXX: signal error condition?
```
papertigers: We need to address #1190 to properly solve this.
Currently return None and we also fail to return the descriptor to the used ring, this can leave the queue stalled or in an overall bad state.

iximeow: in particular we can gate this on `if count >= self.size() { .. }` because VirtIO says

> A driver MUST NOT create a descriptor chain longer than the Queue Size of the device

and it's probably useful to future readers to know this is a "driver MUST NOT" when actually signalling an error (rather than a "we'd really rather you not" like "the table shouldn't be gargantuan either please" - i don't see VirtIO describing a max size of the dscriptor _table_, just the length of the chainthrough it...)

iximeow (approved):
small thought on the comment but it probably applies to all the `XXX: signal error condition?` in here anyway. thanks for the improvement 🙏

lib/propolis/src/hw/virtio/queue.rs
papertigers: If the updated language here, if it looks good we can probably merge this

iximeow: yeah, seems fair!

papertigers:
I am an idiot and I merged this into the wrong branch so the real commit will come in via #1196

PR #1198, opened 2026-08-14
Fix vsock test flake

### LINKED ISSUE #1197: Test flake in credit_update_sent_after_flushing_half_buffer
As  seen in [this test run]( https://buildomat.eng.oxide.computer/wg/0/details/01KZSX1TCN1DMJE9VFN45RVGM7/s54V2ABDWunK03p9CRKcMwW8I19AjI5NWevQqSgbQB3yTufj/01KZSX2CDQ1JCTH7FD3EP084SV) we have a vsock test failure:

```
	test vsock::poller::test::credit_update_sent_after_flushing_half_buffer ... FAILED
	
	failures:
	
	---- vsock::poller::test::credit_update_sent_after_flushing_half_buffer stdout ----
	
	thread 'vsock::poller::test::credit_update_sent_after_flushing_half_buffer' (189) panicked at lib/propolis/src/vsock/poller.rs:1503:39:
	called `Result::unwrap()` on an `Err` value: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
	
	thread 'vsock-event-loop' (215) panicked at lib/propolis/src/hw/virtio/vsock.rs:127:45:
	mem access for write
	
	
	failures:
	    vsock::poller::test::credit_update_sent_after_flushing_half_buffer
	
	test result: FAILED. 193 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.73s
```

papertigers:
This fixes #1197.

The test is a simple off by one error which leads to a race condition. We end up waiting for
`harness.tx_used_idx()` to reach one less descriptor than we should be, which means calling
`reset_tx_cursors` at the top of the loop resets the `QueueWriter` too soon. This opens a window for
the poller to not read all the RW packets that we are sending. The log in the test confirms this,
and explains why our `read_exact()` fails.

```
Aug 14 02:51:12.719 WARN dropping invalid vsock packet: vsock packet header reported 8192 bytes but the descriptor chain contains 0, component: vsock-test
```

### DIFF
--- a/lib/propolis/src/vsock/poller.rs
+++ b/lib/propolis/src/vsock/poller.rs
@@ -1477,11 +1477,7 @@ mod test {
         let payload = vec![0xAB_u8; chunk_size];
         let total_sent = num_chunks * chunk_size;
 
-        // This is the vsock poller copying the REQUEST data and calling
-        // `push_used` which bumps the tx ring's used index.
-        let initial_tx_used = harness.tx_used_idx();
-
-        for tx_consumed in (1u16..).take(num_chunks) {
+        for _ in 0..num_chunks {
             // Reuse descriptor slots each iteration. This is only safe because
             // this test ensures that all tx_avail descriptors are pushed back
             // to us via tx_used.
@@ -1501,17 +1497,18 @@ mod test {
 
             let d_hdr = harness.add_tx_readable(hdr_as_bytes(&rw_hdr));
             let d_body = harness.add_tx_readable(&payload);
+
+            // The vsock poller calls `push_used` which bumps the tx ring's
+            // used index after processing the `harness.publish_tx`.
+            let expected_tx = harness.tx_used_idx() + 1;
             harness.chain_tx(d_hdr, d_body);
             harness.publish_tx(d_hdr);
-            notify.queue_notify(VSOCK_TX_QUEUE).unwrap();
 
             // Wait for the poller to release the desc chain back to us via
             // tx_used. This is the running total of the REQUEST and n RW
             // packets.
-            wait_for_condition(
-                || harness.tx_used_idx() >= initial_tx_used + tx_consumed,
-                5000,
-            );
+            notify.queue_notify(VSOCK_TX_QUEUE).unwrap();
+            wait_for_condition(|| harness.tx_used_idx() >= expected_tx, 5000);
         }
 
         // Drain the data from the accepted socket to confirm it arrived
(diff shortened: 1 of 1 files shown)


papertigers:
Before the fix:

```
Stress test [   5.027s] iteration 106/3000: 1 test run: 0 passed, 1 failed, 269 skipped                                                                                                    ────────────
      Summary [   7.561s] 106/3000 stress run iterations: 105 passed, 1 failed
         FAIL [   5.025s] [ 106/3000] propolis vsock::poller::test::credit_update_sent_after_flushing_half_buffer
error: test run failed

```

After the fix:

```
 Stress test iteration 3000/3000 (00:01:26 elapsed so far, 1 iteration remaining)
        PASS [   0.031s] [3000/3000] propolis vsock::poller::test::credit_update_sent_after_flushing_half_buffer                                                                             Stress test [   0.033s] iteration 3000/3000: 1 test run: 1 passed, 269 skipped
────────────
Summary [  86.578s] 3000/3000 stress run iterations: 3000 passed
```

papertigers:
One more follow up, I have been running a stress test for ~4hrs and ~8500 iterations and have not encountered any issues.

lib/propolis/src/vsock/poller.rs
```diff
@@ -1473,8 +1477,14 @@ mod test {
         let payload = vec![0xAB_u8; chunk_size];
         let total_sent = num_chunks * chunk_size;
 
+        // This is the vsock poller copying the REQUEST data and calling
+        // `push_used` which bumps the tx ring's used index.
+        let initial_tx_used = harness.tx_used_idx();
```
iximeow: the diff seems fine but.. if things are quiescent until we `publish_chain(); queue_notify()`, couldn't we do this at the start of the loop and have a mildly more straightforward check like..

```
let expected_tx = harness.tx_used_idx() + tx_consumed;

/* do tx */

wait_for_condition(|| harness.tx_used_idx() >= expected_tx, 5000);
```
and not worry about whatever the state was as we got to the loop?

iximeow (approved):
sgtm, one suggestion about how we figure out what to wait for but I'm not sure if I'm missing something and a fixed flake is a fixed flake.

lib/propolis/src/vsock/poller.rs
papertigers: I fixed the awkwardness here and I think it reads much more clearly now. If you are good with it, I will merge..


``` 
Summary [  95.536s] 3000/3000 stress run iterations: 3000 passed
 ```

iximeow: yeah, this is a lot nicer, thanks :pray:

iximeow:
for the sake of less CI flakiness, i'm ok with merging even though header-check fails (fixed by https://github.com/oxidecomputer/propolis/pull/1208, and only failing because CI now has a newer viona version than Propolis knows about. but that new version is only additive, so ... it's fine!)

PR #1200, opened 2026-08-22
[server] restore SMBIOS type 1 input on instance ensure

zeeshanlakhani:
An ensure request with an explicit SMBIOS Type 1 input silently got Propolis's defaults (`Oxide`/`OxVM`) instead. This issue came up in voxel and a4x2 testing, where the manufacturer string is used for how those infra stacks select hardware, and the product-name and serial are used to carry the sled's configured `BaseboardId` during the bootstrapping process. 

We ran into this after #1178 made it to the mainline, which moved API-to-internal conversion into per-version `SpecBuilder` paths and dropped the `smbios` -> `smbios_type1_input` assignment along the way.

This PR keeps `smbios` in `latest_to_spec_builder`, where all the versioned paths go through.


sion42x:
This looks great and I've verified it works right with voxel. Do we plan on updating the falcon pin from February? I've got some other propolis work I'll be making a PR for soon that would depend on this.

zeeshanlakhani:
> This looks great and I've verified it works right with voxel. Do we plan on updating the falcon pin from February? I've got some other propolis work I'll be making a PR for soon that would depend on this.

we should, though not sure what the proper ordering of deps updates should be.

sion42x (commented):
Confirmed this fixes the regression and will allow falcon to use a current propolis again. Recommend adding this to  #1204 for softnpu performance as well.

sion42x (approved):

zeeshanlakhani:
@iximeow thoughts on this btw?

sion42x:
@zeeshanlakhani would you be open to `SmbiosType1Input` growing the remaining SMBIOS Type 1 fields? `sku_number` and `family` (and arguably `uuid`). We may be able to use this for softnpu detection (similar to `tofino::get_tofino_from_devinfo()`). See https://github.com/oxidecomputer/omicron/issues/11202 for details on where this matters.

zeeshanlakhani:
> @zeeshanlakhani would you be open to `SmbiosType1Input` growing the remaining SMBIOS Type 1 fields? `sku_number` and `family` (and arguably `uuid`). We may be able to use this for softnpu detection (similar to `tofino::get_tofino_from_devinfo()`). See [oxidecomputer/omicron#11202](https://github.com/oxidecomputer/omicron/issues/11202) for details on where this matters.

~~That totally makes sense. Want me to add it? Or, do you want to extend this. I'd also like someone working on propolis to take a gander here.~~ We decided against the additional work.

zeeshanlakhani:
@iximeow @papertigers thoughts on this btw? We've kept the scope small vs what @sion42x was looking at in terms of including a larger portion of the spec.

bin/propolis-server/src/lib/spec/api_spec_latest.rs
```diff
@@ -216,3 +221,49 @@ pub(crate) fn latest_to_spec_builder(
...
+            }),
+        };
+
+        let spec = latest_to_spec_builder(api_spec).unwrap().finish();
+        let smbios =
+            spec.smbios_type1_input.expect("SMBIOS type 1 input preserved");
+        assert_eq!(smbios.manufacturer, "a4x2");
+        assert_eq!(smbios.product_name, "913-0000019");
+        assert_eq!(smbios.serial_number, "2FAKE000");
+        assert_eq!(smbios.version, 2);
```
iximeow: fwiw i'd throw a `PartialEq` on `SmbiosType1Input` and just make sure the original input is the same as the one we got back out. mostly so that if someone (hi Steve :) ) does add new type 1 fields later this correctly tests that those fields don't get set to some novel different default or something later on.

iximeow (commented):
sorry about the bug. this makes sense though I'd like to understand how I missed this in the refactor, particularly since there wasn't a `set_smbios_type1_input` before.

I'll convince myself of what happened there, probably push a commit that adjusts the test, and pull this in for the sake of getting this fixed for you; I realize this is an issue for voxel/a4x2/etc

bin/propolis-server/src/lib/spec/api_spec_latest.rs
zeeshanlakhani: Yep, I can definitely do that.

zeeshanlakhani:
Other than clippy's issues, I like the additional testing here @iximeow.

PR #1201, opened 2026-08-27
softnpu: Don't hold pipeline mutex across a blocking operation

sion42x:
Each softnpu port has an ingress that held a lock through multiple stages, and since sending is a syscall the lock holder was descheduled inside the critical section on every forwarded packet. This resulted in every waiting port thread to go to sleep and every packet handoff to go through a scheduler wakeup, which in turn results in a 6.6k pps rate on the whole switch.

On a voxel test, the emulated fabric topped out at 6.2k pps with a cross-sled TCP throughput of 100-340KB/s (loopback is 118MB/s on the same endpoint).

This change scopes the lock to `process_packet` and forwards it after it drops on both external and guest paths, and the management handler stops holding the lock while writing replies over UART. Single flow throughput go from 0.1-0.34 -> 5 MB/s, and retransmits droped from 346% -> 29% of transmitted bytes. Per flow ordering remains unchanged.

### DIFF
--- a/lib/propolis/src/hw/virtio/softnpu.rs
+++ b/lib/propolis/src/hw/virtio/softnpu.rs
@@ -346,27 +346,29 @@ impl PciVirtioSoftNpuPort {
             }
             push_used = true;
 
-            let pkt = packet_in::new(&frame[..n]);
+            let mut pkt = packet_in::new(&frame[..n]);
 
-            let mut pipeline = match self.pipeline.lock() {
-                Ok(pipe) => pipe,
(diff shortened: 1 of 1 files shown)


sion42x:
@zeeshanlakhani I'd appreciate your take on this. The change in performance was dramatic. While I agree with the TODO that we shouldn't be using a mutex at all, this is a good half measure whilst I work on a possible solution there.

sion42x:
Closed for now with the changes made by Ry for #1204

PR #1204, opened 2026-08-28
finagle a STREAMS rx buffer size for DLPI handles

rcgoodfellow:
STREAMS uses a high water mark as a back pressure mechanism. When we hit the high water mark, messages are dropped until we drain to the low water mark. This essentially makes the high water mark a receive buffer size.

For TCP on the STREAMS path (i.e. TPI consumers), it appears that the high water mark is set to SO_RCVBUF which is 128000 bytes. However, for DLPI the high water mark is not set and it defaults to 5120. This meaans we hit the mark at the first jumbo frame and thrash from there.

It seems the only way to influence this outside the kernel is pushing a passthrough bufmod STREAMS module. So that's what we do here for the time being.

It looks like the CI failure is addressed by #1093

### DIFF
--- a/lib/propolis/src/hw/virtio/p9fs.rs
+++ b/lib/propolis/src/hw/virtio/p9fs.rs
@@ -845,8 +845,7 @@ impl P9Handler for HostFSHandler {
 
         let mut entries: Vec<proto::Dirent> = Vec::new();
 
-        let mut offset = 1;
-        for de in &dir[msg.offset as usize..] {
+        for (offset, de) in (1..).zip(dir[msg.offset as usize..].iter()) {
             let metadata = match de.metadata() {
                 Ok(m) => m,
                 Err(e) => {
@@ -885,7 +884,6 @@ impl P9Handler for HostFSHandler {
 
             space_left -= dirent.wire_size();
             entries.push(dirent);
-            offset += 1;
         }
 
         let response = Rreaddir::new(entries);
--- a/lib/propolis/src/hw/virtio/softnpu.rs
+++ b/lib/propolis/src/hw/virtio/softnpu.rs
@@ -271,13 +271,18 @@ struct StrIoctl {
 }
 
 fn strioc<T>(fd: i32, cmd: i32, arg: &mut T) -> Result<()> {
+    #[cfg(target_os = "illumos")]
+    let rq = libc::I_STR;
+    #[cfg(not(target_os = "illumos"))]
+    let rq = panic!("softnpu only runs on illumos");
+
     let mut si = StrIoctl {
         ic_cmd: cmd,
         ic_timout: -1,
         ic_len: std::mem::size_of::<T>() as i32,
         ic_dp: arg as *mut T as *mut libc::c_char,
     };
-    if unsafe { libc::ioctl(fd, libc::I_STR, &mut si as *mut StrIoctl) } < 0 {
+    if unsafe { libc::ioctl(fd, rq, &mut si as *mut StrIoctl) } < 0 {
         return Err(Error::last_os_error());
     }
     Ok(())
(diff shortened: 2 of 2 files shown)


sion42x:
Ran tests and a/b checks on 64KB vs 4MB. Softnpu hits "fast enough" that way, with uploads and `recovery finish` beating race conditions and completing successfully. A combination of #1200 and #1204 would be required for voxel to use propolis main. Hoping we can get all three in soon!

lib/propolis/src/hw/virtio/p9fs.rs
```diff
@@ -845,8 +845,7 @@ impl P9Handler for HostFSHandler {
 
         let mut entries: Vec<proto::Dirent> = Vec::new();
 
-        let mut offset = 1;
-        for de in &dir[msg.offset as usize..] {
+        for (offset, de) in (1..).zip(dir[msg.offset as usize..].iter()) {
```
rcgoodfellow, line 848: Thsi is addressing a recent clippy complaint.

lib/propolis/src/hw/virtio/softnpu.rs
```diff
@@ -258,6 +262,82 @@ impl SoftNpu {
...
+    let fd = dlpi::fd(h)?;
+
+    // bufmod calculates the high water mark as 4*chunk + 512, do the
+    // inverse to get our target buffer size.
+    let mut chunk: u32 = size.saturating_sub(512) / 4;
+    // Put the bufmod in passthrough mode
+    let mut flags: u32 = SB_NO_HEADER | SB_NO_PROTO_CVT | SB_NO_DROPS;
+    // Disable chunking, still keeps the water mark that was applied for the
+    // supplied chunk. Yes this is relying on implicit behavior. Good motivation
+    // to get off DLPI entirely.
```
iximeow, line 321: I figured I dunno how this works and it'd be interesting to fish through, my best guess is that doing all this to push `q_hiwat` to be large enough to at least contain this chunk since we've set it `SB_NO_DROPS` preventing the kernel from just dropping it? in which case, presumably we can't somehow `strqset(.., QHIWAT, idk?, 0x120000)`?

iximeow (approved):
abstractly i Get How It Could Work and i trust you to have tested that softnpu works better with this change, but i'll admit i don't understand the mechanics of why These Flags do the needful..

anyhow, as you might have seen in the other PRs here, the header-check failure is an unfortunate consequence of when the viona API changes got into stlouis vs the Propolis side not being in yet, so feel free to merge even though that check's failing.

lib/propolis/src/hw/virtio/softnpu.rs
rcgoodfellow: Pushing the bufmod streams module is purely for the side-effect of impacting the high water mark. I set `SB_NO_DROPS` based on the comment in `uts/common/io/bufmod.c` that says

> SB_NO_DROPS - bufmod behaves transparently in flow control and propagates the blocked flow
conditions downstream.

With the idea being having the bufmod be a pure passthrough. Afaict `strqset` is only available in the kernel so this is the best I could come up with. The `snoop` program does the same trick to effectively get a decent size receive buffer.

rcgoodfellow:
> i trust you to have tested that softnpu works better with this change

Yep, with this change I'm able to push ~1.5 gbps through a softnpu port. Up from ~5 mbps.

lib/propolis/src/hw/virtio/softnpu.rs
iximeow: oh, as i was looking through kernel source and strqset looked vaguely like how i've seen ioctls get plumbed through, but i see now that it's under  `Kernel Functions for Drivers` so that explains the interface and why it wouldn't be usable here. welp. thanks for the pointers :saluting_face:

PR #1205, opened 2026-08-30
Add SKU number and family to the SMBIOS type 1 input

sion42x:
Follow up to #1200 per discussion in comments.

The type 1 input covers 4 fields, but the table also defines SKU and Family per DSP0134 7.2. Today we emit those empty and they can't be set. This PR adds optional `sku_number` and `family` to the input and plumbs them to the emitted table. Both default to unset.

Per RFD 619 this is a new API version. Worth reviewing: downgrades are lossless-or-fail, which may or may not be the desired method.

### DIFF
--- a/bin/propolis-server/src/lib/spec/api_spec_v2.rs
+++ b/bin/propolis-server/src/lib/spec/api_spec_v2.rs
@@ -9,7 +9,7 @@ use std::collections::BTreeMap;
 
 use propolis_api_types::instance_spec::SpecKey;
 use propolis_api_types_versions::{
-    v1, v1::instance::ReplacementComponent, v2, v3,
+    v1, v1::instance::ReplacementComponent, v2, v3, v7,
 };
 
 use super::{builder::SpecBuilder, LegacyApiSpecError, Spec};
@@ -26,8 +26,16 @@ impl TryFrom<Spec> for v2::instance_spec::InstanceSpec {
         // `smbios_type1_input`.  Emptying out the SMBIOS Type 1 input means
         // this either can be converted to a V1 spec which we can losslessly
         // make V2 by adding the SMBIOS table input back in, or we wouldn't be
-        // able to get to a V2 InstanceSpec either way.
-        let smbios = val.smbios_type1_input.take();
+        // able to get to a V2 InstanceSpec either way. The input itself must
+        // also downgrade: v7 added fields to it that V2 cannot express.
+        let smbios = val
+            .smbios_type1_input
+            .take()
+            .map(TryInto::try_into)
+            .transpose()
+            .map_err(|e: v7::instance_spec::SmbiosDowngradeError| {
+                LegacyApiSpecError::IncompatibleComponent(e.to_string())
+            })?;
 
         let v1::instance_spec::InstanceSpec { board, components } =
             val.try_into()?;
--- a/bin/propolis-server/src/lib/spec/api_spec_v2.rs
+++ b/bin/propolis-server/src/lib/spec/api_spec_v2.rs
@@ -9,7 +9,7 @@ use std::collections::BTreeMap;
 
 use propolis_api_types::instance_spec::SpecKey;
 use propolis_api_types_versions::{
-    v1, v1::instance::ReplacementComponent, v2, v3, v7,
+    v1, v1::instance::ReplacementComponent, v2, v3,
 };
 
 use super::{builder::SpecBuilder, LegacyApiSpecError, Spec};
@@ -28,14 +28,8 @@ impl TryFrom<Spec> for v2::instance_spec::InstanceSpec {
         // make V2 by adding the SMBIOS table input back in, or we wouldn't be
         // able to get to a V2 InstanceSpec either way. The input itself must
         // also downgrade: v7 added fields to it that V2 cannot express.
-        let smbios = val
-            .smbios_type1_input
-            .take()
-            .map(TryInto::try_into)
-            .transpose()
-            .map_err(|e: v7::instance_spec::SmbiosDowngradeError| {
-                LegacyApiSpecError::IncompatibleComponent(e.to_string())
-            })?;
+        let smbios =
+            val.smbios_type1_input.take().map(TryInto::try_into).transpose()?;
 
         let v1::instance_spec::InstanceSpec { board, components } =
             val.try_into()?;
--- a/bin/propolis-server/src/lib/spec/api_spec_v7.rs
+++ b/bin/propolis-server/src/lib/spec/api_spec_v7.rs
@@ -50,16 +50,6 @@ pub(crate) fn amend(
     spec: &mut v7::instance_spec::InstanceSpec,
     replacements: &BTreeMap<SpecKey, ReplacementComponent>,
 ) -> Result<(), MigrateError> {
-    for (id, replacement) in replacements {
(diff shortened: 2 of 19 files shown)


sion42x:
Worth mentioning I used `(6, NVME_WRITE_CACHE)` as a template for most of this, which was perfect since it also added one new optional field. https://github.com/oxidecomputer/dropshot-api-manager/blob/main/guides/new-version.md was also helpful, thanks @sunshowers

sion42x:
Interesting chicken and egg issue I'm not sure how (or whether) to resolve, the `phd-run-migrate-from-base` failure requires 7.0.0 to test 7.0.0. I guess the PHD side should pin version to the max supported by the base rather than the PR?

crates/propolis-api-types-versions/src/smbios_sku_family/instance_spec.rs
```diff
@@ -0,0 +1,232 @@
...
+    pub version: u64,
+
+    /// The SKU Number string in the emitted table. Unset means empty.
+    #[serde(default)]
+    pub sku_number: Option<String>,
+
+    /// The Family string in the emitted table. Unset means empty.
+    #[serde(default)]
+    pub family: Option<String>,
+}
```
zeeshanlakhani: Consider `#[serde(default, skip_serializing_if = "Option::is_none")]` here, per the compatibility notes in [instance_spec/mod.rs#L98-L101](https://github.com/oxidecomputer/propolis/blob/master/crates/propolis-api-types/src/instance_spec/mod.rs#L98-L101). Without it, any serialized SMBIOS input always carries `"sku_number":null,"family":null`, and pre-v7 `deny_unknown_fields` deserializers reject those field names even when null. Version negotiation keeps this off the wire while in normal operation, but if the from-base PHD check gets fixed by pinning the client's api-version header, a pinned client sending an SMBIOS input with the new fields unset would be rejected by a 6.0.0 server, despite the spec being v6-expressible. With `skip_serializing_if`, the serialized form matches v6 exactly whenever the fields are unset.

bin/propolis-server/src/lib/spec/api_spec_v2.rs
```diff
@@ -26,8 +26,16 @@ impl TryFrom<Spec> for v2::instance_spec::InstanceSpec {
...
+        // able to get to a V2 InstanceSpec either way. The input itself must
+        // also downgrade: v7 added fields to it that V2 cannot express.
+        let smbios = val
+            .smbios_type1_input
+            .take()
+            .map(TryInto::try_into)
+            .transpose()
+            .map_err(|e: v7::instance_spec::SmbiosDowngradeError| {
+                LegacyApiSpecError::IncompatibleComponent(e.to_string())
+            })?;
```
zeeshanlakhani: This error text is a bit misleading, right?  The SMBIOS input isn't a component and the incompatibility is v2-level. I'd suggest a dedicated `#[from] SmbiosDowngradeError` variant.

crates/propolis-server-api/src/lib.rs
```diff
@@ -131,6 +152,35 @@ pub trait PropolisServerApi {
...
+                        ClientErrorStatusCode::BAD_REQUEST,
+                        format!(
+                            "instance spec cannot be expressed to v6 \
+                             clients: {e}"
+                        ),
+                    )
+                },
+            )?;
+        Ok(HttpResponseOk(v6_response))
+    }
```
zeeshanlakhani, line 182: This 400s a read-only GET for a legally created VM whenever the new fields are set, including `Some("")`, which renders byte-identical to `None` in the emitted table (empty `SmbStrings`s get string index 0). I'd keep the 400 for non-empty values but normalize `Some("") -> None`, ideally at ingress so that it never reaches a `Spec`. Then, the 400 always means the VM carries data a v6 client genuinely can't see. Making the v6 view silently lossy instead would cut against our compatibility notes' own rationale and would break an old client's read/modify/write cycle.

bin/propolis-server/src/lib/spec/api_spec_v7.rs
```diff
@@ -0,0 +1,65 @@
...
+                "replacement component {id} not in source spec",
+            )));
+        };
+
+        // v7 reuses the v6 component types, so the v6 amendment logic applies.
+        api_spec_v6::amend_component(id, to_amend, replacement)?;
+    }
+
+    Ok(())
+}
```
zeeshanlakhani, line 55: Nit: This loop is identical to `api_spec_v6::amend` (without the spec type), and since v7 reuses v6's `Component` (the maps are the same type), the two could share one helper (something like):

```rust
pub(crate) fn amend_components(
    components: &mut BTreeMap<SpecKey, v6::instance_spec::Component>,
    replacements: &BTreeMap<SpecKey, ReplacementComponent>,
) -> Result<(), MigrateError> { ... }
```

... with both amend fns delegating via `&mut spec.components`.

zeeshanlakhani (changes requested):
Leaving a few comments / fixins here before we bring this upstream.

bin/propolis-server/src/lib/spec/api_spec_v7.rs
sion42x: Actually this works better because I can make `amend_component` private again.

bin/propolis-server/src/lib/spec/api_spec_v2.rs
sion42x: Ah good call, I ran into this elsewhere and neglected to fix it here.

crates/propolis-server-api/src/lib.rs
sion42x: Good to know on `Some("") -> None`. Thanks for that. This also might fix the phd errors on CI. At minimum it's good validation.

crates/propolis-api-types-versions/src/smbios_sku_family/instance_spec.rs
sion42x: Good to know, I'm fixing the attributes and will extend the serde test.

rmustacc:
What are the intended customer semantics of these new SMBIOS fields? How are they going to be populated and interpreted?

sion42x:
@rmustacc the goal was to keep this semantics free in propolis while writing the values into the table per DSP0134 (but emitted as empty when unset). Nexus (nexus/src/app/instance_platform/mod.rs) and sled-agent (sled-agent/src/instance.rs) explicitly set `smbios: None`. A customer guest could read them but we're not defining any Oxide contract.

The only intended producer/consumer pair is virtual test platforms, and a4x2 has used the table for a while now. A regression in #1178 dropped the SMBIOS input which broke sled identity and rack setup for a4x2. This PR's parent (#1200) fixes that. Since it was being corrected, I added this PR to make the table more complete against the spec. The new fields allow the emulator to declare the platform role (e.g. marking softnpu-based scrimlets for https://github.com/oxidecomputer/omicron/issues/11202). Today nothing in omicron reads the new fields, and the existing type 1 fields are only read in sled-agent's SMBIOS read path which is gated on the `a4x2` manufacturer string.

If I'm missing something or we'd rather not bump API version for a one-off a4x2-only change, let me know. At present, omicron can still infer the role from the existing manufacturer field (or funny enough, from the fact it's set at all).

PR #1206, opened 2026-08-31
[bug] Avoid holding the pipeline lock across management uart writes

zeeshanlakhani:
`handle_management_message` locked the P4 pipeline before dispatching the request. For dump and radix requests, for example, it kept that lock while waiting for the guest to drain the 1-byte UART FIFO.

In turn, if a queue notify message could send a vCPU into the `process_guest_packet` fn, where it waited on the same lock, that vCPU would not finish its PIO exit. This would lead to the guest not draining the UART, and, boom, deadlock.

## The Fix

Now, the dump path copies table state while under the lock and writes it after unlocking happens. Table updates still hold the lock while changing the pipeline. UART writes now have a deadline, where if a response times out partway through, the handler will try to write the missing newline while waiting for the next request.

*Note*: I ran into this issue on a long-running voxel session. The pre-fix binary reproduces it within minutes under a scadm polling loop, while the fixed one does not (A/B testing).

### DIFF
--- a/lib/propolis/src/hw/virtio/p9fs.rs
+++ b/lib/propolis/src/hw/virtio/p9fs.rs
@@ -845,8 +845,7 @@ impl P9Handler for HostFSHandler {
 
         let mut entries: Vec<proto::Dirent> = Vec::new();
 
-        let mut offset = 1;
-        for de in &dir[msg.offset as usize..] {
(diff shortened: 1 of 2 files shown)


lib/propolis/src/hw/virtio/p9fs.rs
zeeshanlakhani, line 848: A clippy find on the `falcon` feature.

PR #1208, opened 2026-08-31
don't header-check the bhyve or viona API version

iximeow:
since `5ffff4b` (nice hash!) stlouis has had a `VIONA_CURRENT_INTERFACE_VERSION` of 7, which differs from the existing minimum Propolis requires of 6. the last time this was changed was back when we added multi-queue support to viona, which predates me adding header-check as part of CI (... as a result of a bug that came in with that change!) ... when the CI illumos was updated in the last few weeks, the value we see in header-check finally got the update too. as-is, because the constant doesn't match between our definition and the kernel headers, `header-check` handles that as any other constant mismatch and fails CI. in any other case this would be right, such as if ioctl values were curiously changing around beneath our feet.. but here it's generally ok.

while viona's header-check now supports a notion of `API_VERSION`, I don't know of any items which have been removed and I don't want to add conflicts for new versions. so I'm adding it in here really in service of future use (perhaps in #1093?)

this gets CI green again, which really had no need to be failing. sorry for the troubles @zeeshanlakhani

### DIFF
--- a/crates/bhyve-api/header-check/build.rs
+++ b/crates/bhyve-api/header-check/build.rs
@@ -91,13 +91,15 @@ fn main() {
         // We expose our own copy for now for us as a constraint.
         "VM_MAXCPU" => true,
 
-        // Do not bother checking the version definition define if we are
-        // assuming the source is from a different version.
-        "VMM_CURRENT_INTERFACE_VERSION"
-            if !ver_eq(VMM_CURRENT_INTERFACE_VERSION) =>
-        {
-            true
-        }
+        // The bhyve interface is generally backwards-compatible; headers may be newer
+        // than Propolis knows about. In service of additive changes, ignore the
+        // interface version from headers.
+        //
+        // If items (structs, ioctl numbers) we know about changed, header-check
+        // will find and report those specifically. If items were removed, we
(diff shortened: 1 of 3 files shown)


crates/bhyve-api/header-check/build.rs
```diff
@@ -24,6 +24,7 @@ fn ver_lt(ver: u32) -> bool {
     CHECK_VERSION.load(Ordering::Relaxed) < ver
 }
 /// Source checked against has API version equal to `ver` argument
+#[allow(dead_code)]
```
hawkw: i guess it makes sense to not delete this since we might reach for it again later...but i _also_ feel like the presence of `ver_lt` and `ver_gt` make it pretty obvious how you would write `ver_eq` if you wanted it again. iunno!

iximeow:
I've misunderstood the semantic of the interface version numbers for a while now, it turns out. they're just Private interface from the OS's perspective, and it is incorrect to assume you might be able to get correct function using for example an old Propolis on newer host bits. I'll have a subsequent change that describes this a bit on the Propolis side, but we just need to update Propolis' VIONA_CURRENT_INTERFACE_VERSION to match.

PR #1210, opened 2026-09-07
[viona, mac-filter, mcast] install guest multicast MAC filter tables via V7 ioctls

zeeshanlakhani:
Guest multicast interest has been driven by `VIONA_PROMISC_MULTI`, delivering multicast frames for the guests to filter. We saw duplication issues at guest Rx when packets reached both the classified and promiscuous callbacks. This led to the viona interface V7 scoped changes in [illumos-gate](https://github.com/oxidecomputer/illumos-gate/blob/5ffff4b86e486e1f9d7860be1368386699a7829a) / [change 775](https://code.oxide.computer/c/illumos-gate/+/775).

Given the upstream changes, this work introduces `VNA_IOC_SET_MAC_FILTERS` to install guest multicast tables on the underlay MAC client backing viona. `VNA_IOC_GET_MAC_FILTERS` is used for reading back installed entries from the table. Classified delivery can replace all-multicast mode once the guest supplies a table that fits and no broader receive mode is necessary. All-multicast remains the fallback for guests that haven't sent a table request or have exceeded the kernel's capacity.

Most of the work here goes into handling ioctl failures, semantic result codes, and fallback. This includes explicit multicast table management, reset handling, and Falcon-build `AllVlan` logic.

Most importantly (most scary), this raises the minimum required host viona API to V7. We're keeping the V1 viona state payload for now, with import / export handling for the multicast table state.

**Note**: This work will fail on CI as don't have special handling. Chat with @iximeow for more. 

This is the first of a few different PRs upcoming, including
- [ ] CTRL_RX receive-mode handling
- [ ] send notification on `NEEDS_RESET` status


lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -332,7 +538,21 @@ struct Inner {
...
+    rx_config_failed: bool,
+    /// Whether the guest's MAC table management is active.
+    ///
+    /// An accepted table, including an empty one, allows
+    /// [`PciVirtioViona::rx_config`] to release the all-multicast
+    /// lower bound.
+    ///
+    /// This bound exists for illumos vioif, which negotiates
+    /// `VIRTIO_NET_F_CTRL_RX` but never programs a multicast
+    /// table: <https://www.illumos.org/issues/18280>.
```
iximeow, line 554: this is a bit confusing to me: is this field not necessary for correct operation anywhere else? I think as-is we would need it to correctly describe the device's behavior in the transitory period after the guest has negotiated `F_CTRL_RX` and before a driver has submitted a filter table. otherwise (for example) if you migrate in that period the destination VM would believe `multicast_mac_filters` represented the source's viona state and install an empty list, dropping all multicast traffic while the source was still passing it.

I figure that's a pretty small window, but the comment here looks like we're working around guest quirks when the bit is necessary to describe a legal (albeit strange) device state.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -332,7 +538,21 @@ struct Inner {
     promisc: PromiscLevel,
     filter: FilterState,
     unicast_mac_filters: Box<[MacAddr]>,
-    multicast_mac_filters: Box<[MacAddr]>,
+    multicast_mac_filters: Box<[MulticastMacAddr]>,
+    /// Whether a multicast filter table is installed on the in-kernel MAC
+    /// client via `VNA_IOC_SET_MAC_FILTERS`.
+    mac_filters_installed: bool,
+    rx_config_failed: bool,
```
iximeow, line 545: this seems kind of funky too: we're tracking if an rx configuration-related operation failed at _some point_ so we can reject future attempts to change the viona rx configuration? if we've failed to set the configuration and set NEEDS_RESET on the device, _should_ we be standing in the way of future viona rx changes?

my expectation is that viona makes no promises as to the device state if an ioctl fails. so, if an rx configuration errors for any reason, we'll already be in a situation where Propolis' understanding of the guest state may not match viona (e.g. we'll have changed Propolis state just after some `if state.rx_config_failed`, but possibly not have actually changed viona's configuration to match). i don't think we can avoid the possibility of some skew there either way, because if we wait until an ioctl succeeds to update Propolis' side of things, then an ioctl that fails may have applied a change while we don't update Propolis' side due to the failure. so i think we're just going to have some skew either way.

_that_ means that if a migration happens while `rx_config_failed` is set, all we know is the thing we're migrating might not be the state of viona that was migrated from. but we do _not_ know if there is actually config skew. so the best this can tell us is "the guest may see different device behavior after migration", which is probably good to know for debugging purposes but we'd probably want to build differently with that purpose as a goal?

namely: if we've set NEEDS_RESET then the guest `SHOULD NOT` rely on the completion of operations, but i don't _think_ there is a reason we should prevent the guest from trying. if the guest sees a NEEDS_RESET after trying to set a bogus set of filters and the guest "fixes it" by setting a different set of filters, i could imagine that manifesting as user observing "the driver does something weird, then a user changes a NIC setting and the hardware tolerates that", where `rx_config_failed` would probably just keep the device in a stuck and undesired state.

a guest _should not_ depend on any of that working, but i don't see a reason we should go out of our way to make it _not_ work.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -216,12 +350,69 @@ pub mod migrate {
...
+            //
+            // We export all-multicast mode when delivery currently depends
+            // on that table, so the migration does not drop multicast.
+            let promisc = if state.mac_filters_installed
+                && state.promisc == PromiscLevel::None
+            {
+                PromiscLevel::AllMulti
+            } else {
+                state.promisc
+            };
```
iximeow, line 401: so.. if a guest sets up a virtio-net device and provides one unicast MAC and no multicast MACs, we'll migrate it out at `PromiscLevel::AllMulti` regardless of what the promiscuity _was_, and on the other end .. ignore the promiscuity computed here? I don't follow why we're going through this.

also: if `state.promisc` was `AllVlan` before migrating, but the guest had set a MAC filter, we'll switch to `AllMulti` here and the destination's check that if the source promiscuity was `AllVlan` then the destination's promiscuity must be on import will be bypassed. but then in that case the destination would get to `AllVlan` after the next `rx_config()` so i think it would appear to work while all the plumbing points in different directions?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -216,12 +350,69 @@ pub mod migrate {
...
+        fn from(value: PromiscLevel) -> Self {
+            match value {
+                PromiscLevel::None => Self::None,
+                PromiscLevel::AllMulti => Self::AllMulti,
+                PromiscLevel::All => Self::All,
+                #[cfg(feature = "falcon")]
+                PromiscLevel::AllVlan => Self::AllVlan,
+            }
+        }
+    }
```
iximeow, line 371: why have this in addition to `PromiscLevel`? is this to permit migrating state between Falcon and non-Falcon builds of Propolis, where `AllVlan` would disappear from the wire? I don't think that's something we'd want to go out of our way to support, but I'll be the first to admit I don't really know much about Falcon.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -216,12 +350,69 @@ pub mod migrate {
...
+
+    // Note: `multicast_table_managed` comes after V1 has already shipped.
+    // A payload from older sources omits this field, defaulting to false and
+    // retaining all-multicast mode. Pre-MAC-filter targets will just ignore it.
```
iximeow, line 375: we still aren't live-migrating, there's not really any need to ensure compatibility for migration payloads. if we _were_ live-migrating, we wouldn't be changing old payloads (the general plan is to follow more along the lines of the HTTP API versioning where old versions are immediately frozen in time forever)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -71,10 +71,144 @@ enum MqSetPairsCause {
...
+    /// Re-establish a known kernel state.
+    Reinitialize,
+}
+
+/// The set of possible multicast table actions.
+#[derive(Debug, Eq, PartialEq)]
+enum McastTableAction {
+    Install,
+    Clear,
+    Keep,
```
iximeow, line 158: I'm trying to understand what we're optimizing for here, with `McastTableAction` and `Keep` in particular. I think what we're going for is: we don't _need_ to set the multicast table because the guest has either not set one, or has only changed the unicast filters. .. is that something guests are prone to do after starting using the device? my understanding is this is an operation that will happen once or twice when setting up a device and left as-is under load.

if so, we're adding a bunch of opportunities to cache state incorrectly in Propolis when we could probably just go off to viona? if there's a reason we shouldn't just repeat the filter config to viona redundantly, then I think we could handle the accounting a bit better here, but hopefully this initial question makes sense.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -71,10 +71,144 @@ enum MqSetPairsCause {
...
+
+impl McastTableAction {
+    /// Derive the appropriate kernel table operation needed to reconcile
+    /// installation with the `wanted` table state from `RxConfig`:
+    ///
+    /// - `TableReplaced` forces reinstall even when the local state
+    ///   has a table installed
+    /// - `Reinitialize` clears the table even when local state is absent.
+    /// - `FilterChange` keeps the existing table as is, unless `wanted`
+    ///   and `installed` differ.
```
iximeow, line 169: if we _do_ want to keep `McastTableAction`, these ought to be docs on that enum, no?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -71,10 +71,144 @@ enum MqSetPairsCause {
...
+        use RxReconcileCause::{FilterChange, Reinitialize, TableReplaced};
+
+        match (wanted, installed, cause) {
+            (_, _, Reinitialize) => Self::Clear,
+            (true, false, FilterChange | TableReplaced)
+            | (true, true, TableReplaced) => Self::Install,
+            (false, true, FilterChange | TableReplaced) => Self::Clear,
+            (true, true, FilterChange)
+            | (false, false, FilterChange | TableReplaced) => Self::Keep,
+        }
```
iximeow, line 180: I'm curious if you find this "exhaustive truth table" kind of pattern legible at a glance - it might just be me not being sufficiently Haskell-brained but it feels hard to tell what the desired derivation is if `cause == FilterChange` for example. versus something like...
```rust
match cause {
    Reinitialize => {
        assert!(!wanted, "reinitializing a device implies we want there to be no installed mcast filters");
        McastTableAction::Clear
    }

    TableReplaced => {
        assert!(!(wanted == false && installed == false),
            "mcast tables changing implies either tables were present, or should now be present");
        if wanted {
            McastTableAction::Install
        } else {
            McastTableAction::Clear
        }
    }

    FilterChange => {
        assert_eq!(wanted, installed, "the rx action only changed filter state, not multicast tables");
        McastTableAction::Keep
    }
}
```
is what I _think_ describes this function in as much as virtio-net needs to use it. clearly the exhaustive test you included would assert on all three of the arms, but I'm curious what you think of the overall idea?

to me, this was helpful because the `TableReplaced` assert seemed reasonable at first but would fire in an error path when we'd previously returned `Install`. describing _that_ situation in natural language feels pretty convoluted and what ultimately got me to thinking the primitives here are mismatched in ... some way. I'm still trying to put a finger on it.

my best way to describe that condition is something like: `we may want to replace a table while there is no table with an intention to end up with no table after having tried to install a table, failing specifically because of providing too many MACs, and using a lower-performance mechanism to effect the guest's request`.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -949,117 +1213,288 @@ impl PciVirtioViona {
...
+                        capacity,
+                    ));
+                    if effective_promisc == PromiscLevel::None {
+                        effective_promisc = PromiscLevel::AllMulti;
+                    }
+                    action = McastTableAction::derive(
+                        false,
+                        state.mac_filters_installed,
+                        &cause,
+                    );
```
iximeow, line 1442: the mushy overall point I think I'm trying to get between the comment on `derive()` above and here is that taking multiple passes through `McastTableAction::derive` in the process of one `apply_rx_config` means we're interleaving "determine an action then apply it". if that `McastTableAction::derive` ever in the future returns `McastTableAction::Install`, we would silently lose the obligation to take that action, because in that caller we've already gotten past trying to do the Install action.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -949,117 +1213,288 @@ impl PciVirtioViona {
...
+        );
+
+        if matches!(action, McastTableAction::Install) {
+            match self.hdl.set_mac_filters(&state.multicast_mac_filters) {
+                Ok(()) => state.mac_filters_installed = true,
+                // The kernel rejected the count with the installed filters
+                // untouched, reporting its capacity back through
+                // vmf_nmcast.
+                //
+                // The guest's request stays in its state.
```
iximeow: so, this is partially "I didn't understand the context and hadn't gotten to stlouis#986": since the only ways the userland can handle this are "delete multicast filters and set the device to `PromiscLevel::AllMulti`" or "error" ... should viona just.. do that? maybe this was talked about in the review there and I should go check the conversations.

we could still return the same error out to userland even, it just means the semantic of this error is "the device is configured in a way compliant with the request, just by having set the device promiscuous instead". otherwise we have to have support in Propolis and c-bhyve to do this fallback behavior. maybe that's fine.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -949,117 +1213,288 @@ impl PciVirtioViona {
...
+        // No table is required when it is empty, over capacity, or when host
+        // promiscuity takes precedence. `Reinitialize` also clears the table
+        // in order to re-establish known kernel state, even if local state has
+        // no multicast table installed.
+        if matches!(action, McastTableAction::Clear) {
+            let promisc_before_clear = (effective_promisc > state.promisc)
+                .then_some(effective_promisc);
+            if let Some(level) = promisc_before_clear {
+                self.set_promisc(level, state)?;
+            }
```
iximeow, line 1457: `McastTableAction` splits out a few specific states to end up in, but then we have this imperative approach actually applying that action, which gets us into these situations like "we know we need to clear the mcast table, but depending on why we might have to do that one way or another way". so we're not really in a split "determine/apply" approach, but it's kind of trying to be.. this also makes me think the verbs from `McastTableAction` aren't expressive enough?

I don't have a particular suggestion here, but I'm wondering if something might become more obvious alongside other changes here..

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -949,117 +1213,288 @@ impl PciVirtioViona {
...
+    /// the `Inner` guard avoids reversing lock order.
+    fn reconcile_rx_and_release(
+        &self,
+        mut state: MutexGuard<Inner>,
+        cause: RxReconcileCause,
+    ) -> Result<(), RxConfigError> {
+        let outcome = self.apply_rx_config(&mut state, cause);
+        drop(state);
+        if outcome.is_err() {
+            self.virtio_state.set_needs_reset(self);
```
iximeow, line 1482: trivial: you could `needs_reset_locked()` here and avoid having to manage `state` so explicitly:
```suggestion
        if outcome.is_err() {
            self.virtio_state.needs_reset_locked(self, &mut state);
```

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1358,24 +1793,77 @@ impl MigrateMulti for PciVirtioViona {
             ))
         })?;
 
-        let input: migrate::VionaStateV1 = offer.take()?;
+        let input: migrate::VionaStateV1 = match offer.take() {
+            Ok(input) => input,
+            // A source predating this payload has no guest Rx state to
+            // restore.
+            Err(MigrateStateError::DataMissing) => {
```
iximeow, line 1800: a source predating `VionaStateV1`...?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1672,8 +2164,105 @@ impl VionaHdl {
...
+        }
+
+        let installed = vmf.vmf_nmcast as usize;
+        if installed > addrs.len() {
+            return Err(io::Error::new(
+                ErrorKind::InvalidData,
+                "kernel installed filter count exceeded the readback buffer",
+            ));
+        }
+        addrs.truncate(installed);
```
iximeow, line 2263: is it typical for `VNA_IOC_GET_MAC_FILTERS` to return a higher `vnf_nmcast` than the size of the filter table? it seems like this is also an error (did someone else mutate the filters between our ioctls? or is this just a kernel bug, etc)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1672,8 +2164,105 @@ impl VionaHdl {
...
+            vmf_nmcast: 0,
+            vmf_addrs: 0,
+            ..Default::default()
+        };
+
+        unsafe {
+            self.0.ioctl(viona_api::VNA_IOC_GET_MAC_FILTERS, &mut vmf)?;
+        }
+
+        let mut addrs = vec![MacAddr::default(); vmf.vmf_nmcast as usize];
```
iximeow, line 2244: alternatively, we could size this against the maximum number of multicast MACs the kernel would accept and truncate accordingly (maybe good, maybe bad? depends on your goals really)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -1947,67 +2536,66 @@ mod test {
...
+        let mut offers =
+            PayloadOffers::new(payloads.iter().zip(desers.iter_mut()).map(
+                |((kind, version, _bytes), deser)| PayloadOffer {
+                    kind: kind.as_str(),
+                    version: *version,
+                    payload: Box::new(<dyn erased_serde::Deserializer>::erase(
+                        deser,
+                    )),
+                },
+            ));
```
iximeow, line 2583: fwiw i'd gotten to the point of building `PayloadOffers` as part of the "export" side of things to keep it morally similar to propolis-to-propolis migration wherien the thing that goes on the wire is a `PayloadOffers`. then importing gets a `PayloadOffers` and goes from there. that avoids having to juggle around a `(String, u32, String)` too, which i think a caller would pretty reasonably ask "which string means what, how do i call this correctly?"

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2191,10 +2779,193 @@ mod test {
...
+                    }
+                    _ => (0, requested),
+                };
+                assert_eq!(error.probe_args(requested), (code, addr, count, 0));
+            }
+        }
+    }
+
+    #[test]
+    fn test_mac_filters_ioctl_error_overrides_semantic_result() {
```
iximeow, line 2889: it feels.. kind of overbuilt to test a particular helper like this, shouldn't we test that `set_mac_filters`' errors are the `MacFilterError` that we want those errors to express?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2191,10 +2779,193 @@ mod test {
...
+            assert_eq!(
+                error.probe_args(requested),
+                (viona_api::VMF_OK, 0, requested, expected_errno.unwrap_or(0)),
+            );
+            let MacFilterError::Io(error) = error else {
+                panic!("expected ioctl failure: {error:?}");
+            };
+            assert_eq!(error.raw_os_error(), expected_errno);
+            assert_eq!(error.kind(), expected_kind);
+            assert_eq!(error.to_string(), expected_message);
```
iximeow, line 2916: and this is really just testing that `check_mac_filters` didn't invent a de novo `Error` and discard random bits from the ioctl right..? I guess that's.. fine to test, but does this give you more confidence in the `MacFilterError` machinery?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2191,10 +2779,193 @@ mod test {
...
+            "multicast_mac_filters": [[1, 0, 94, 0, 0, 1]]
+        }"#;
+
+        let state: super::migrate::VionaStateV1 =
+            serde_json::from_str(prev_payload)
+                .expect("payload without multicast_table_managed deserializes");
+        assert!(!state.multicast_table_managed);
+        assert_eq!(state.promisc, super::migrate::PromiscMode::AllMulti);
+        assert_eq!(state.multicast_mac_filters.len(), 1);
+    }
```
iximeow, line 2935: this is not something we should test, re. my comment on `VionaStateV1` above

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2191,10 +2779,193 @@ mod test {
...
+
+        state.multicast_mac_filters = Box::new([]);
+        state.mac_filters_installed = false;
+
+        let exported = VionaStateV1::from(&state);
+        let serialized = serde_json::to_value(&exported).unwrap();
+        assert_eq!(serialized["promisc"], "None");
+        assert!(exported.multicast_mac_filters.is_empty());
+        assert!(exported.multicast_table_managed);
+    }
```
iximeow, line 2963: likewise, this is not how we would handle device state migration across payload changes

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2218,6 +2989,56 @@ mod test {
...
+    fn discover_multicast_capacity(hdl: &VionaHdl) -> usize {
+        assert!(
+            hdl.get_mac_filters()
+                .expect("can read filters before capacity probe")
+                .is_empty(),
+            "capacity probe requires an empty kernel filter table",
+        );
+        let mut count = viona_api::VIONA_MAX_MCAST_FILTERS;
+        loop {
+            match hdl.set_mac_filters(&multicast_table(count)) {
```
iximeow, line 3026: this is really confusing to me. if this does not immediately `MacFilterError::Count`, then that means  `VIONA_MAX_MCAST_FILTERS` has diverged from the kernel and we should at least change it in propolis. so it should probably just fail, instead of being `Ok(())` and retrying below. and we can just create and program a ~1k table and get the capacity immediately, so.. why not do that? if that fails to overflow, then something Interesting has changed and this approach might not be so useful anyway?

(... should we just have an ioctl to read vionas' max mcast filters at runtime? then we can decide to `PromiscLevel::AllMulti` without trying and failing to `set_mac_filters`!)

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2814,6 +3793,531 @@ mod test {
...
+        assert!(test_ctx
+            .dev
+            .hdl
+            .get_mac_filters()
+            .expect("can read back filters")
+            .is_empty());
+
+        let all_hosts = MacAddr::from([0x01, 0x00, 0x5e, 0x00, 0x00, 0x01]);
+        let all_nodes = MacAddr::from([0x33, 0x33, 0x00, 0x00, 0x00, 0x01]);
+        let ack = driver.ctrl_mac_table_set(&[], &[all_hosts, all_nodes]);
```
iximeow, line 3827: ... `all_hosts -> all_ipv4`, `all_nodes -> all_ipv6`? mostly just me trying to understand what "node" or "host" means here

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2814,6 +3793,531 @@ mod test {
...
+        // All-multicast mode covers drivers that never manage a table.
+        assert_eq!(
+            test_ctx.dev.inner.lock().unwrap().promisc,
+            PromiscLevel::AllMulti
+        );
+        assert!(test_ctx
+            .dev
+            .hdl
+            .get_mac_filters()
+            .expect("can read back filters")
```
iximeow, line 3822: i'd consider making `get_mac_filters` a function on `TestCtx` which `expect()`s internally just because we're _always_ using it as `test_ctx.dev.hdl.get_mac_filters().expect(...)`. so as-is it just makes the tests wordier for.. if there's a reason i'd like to be sold!

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2814,6 +3793,531 @@ mod test {
...
+
+        assert_eq!(
+            test_ctx.dev.hdl.get_mac_filters().expect("can read back filters"),
+            vec![all_hosts, all_nodes],
+        );
+
+        // Duplicate and broadcast entries are dropped and the table is
+        // compacted by the kernel.
+        let bcast = MacAddr::from([0xff; ETHERADDRL]);
+        let all_routers = MacAddr::from([0x33, 0x33, 0x00, 0x00, 0x00, 0x02]);
```
iximeow, line 3854: ... what happens if the guest offers >MAX_MCAST_ADDRS of the same address?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2550,6 +3407,128 @@ mod test {
...
+            let vq = self
+                .dev
+                .virtio_state
+                .queues
+                .get(qidx)
+                .expect("control queue exists")
+                .clone();
+            self.dev.queue_notify(&vq);
+
+            // Completion happens before reading the ack byte.
```
iximeow, line 3491: it feels pretty important (e.g. at least worth a note here) that `queue_notify` on the control queue is synchronous, so we won't possibly have checked for a completion before the device had done it

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2550,6 +3407,128 @@ mod test {
...
+            {
+                assert_eq!(
+                    acc_mem.write_from(GuestAddr(cursor), data, data.len()),
+                    Some(data.len())
+                );
+                regions.push((cursor, data.len() as u32, 0));
+                cursor += data.len() as u64;
+            }
+
+            let ack_gpa = cursor;
```
iximeow, line 3456: morally I've been hoping the test driver kind of looks like "a normal driver, but used in a test context", and here I think I'd have expected we have some general virtqueue type talking about writing entries and reading completions, rather than happening to poke all the bits in memory (`GuestAddr(d + 12)` for example down below feels like we owe ourselves a little more infrastructure to make that legible than "because the value being written is `flags` this is probably flags on the descriptor...")

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2550,6 +3407,128 @@ mod test {
...
+                .read::<u32>(GuestAddr(loc.used_gpa + 4 + 8 * slot))
+                .expect("can read used element id");
+            assert_eq!(used_id, 0, "used element reports the chain head");
+
+            let used_len = *acc_mem
+                .read::<u32>(GuestAddr(loc.used_gpa + 4 + 8 * slot + 4))
+                .expect("can read used element length");
+            assert_eq!(used_len, 1, "device wrote only the ack byte");
+
+            *acc_mem.read::<u8>(GuestAddr(ack_gpa)).expect("can read ack byte")
```
iximeow, line 3507: similar re. expressions like `loc.used_gpa + 4 + 8 * slot + 4`. we're clearly indicating _something_ in a structure or in a structure after a structure, but it's pretty awkward to work this back to "what exactly is it we're doing in this test". typing this stuff and giving ourselves pointers will make this much more legible when someone (hi) has to come poke this to send data in these tests one day.

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2550,6 +3407,128 @@ mod test {
...
+                    buf.extend_from_slice(&mac.0);
+                }
+                buf
+            }
+
+            // Keep each table in its own descriptor.
+            let unicast = table(unicast);
+            let multicast = table(multicast);
+            self.send_ctrl_cmd(
+                1,
```
iximeow, line 3527: I really should have asked that the integers in `impl TryFrom<Header> for Command` are the virtio `VIRTIO_NET_CTRL_*` identifiers back when we added that, but here out of context we really should use the identifier instead of just `1`...

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2814,6 +3793,531 @@ mod test {
...
+            assert_eq!(state.promisc, promisc);
+            assert!(!state.rx_config_failed);
+        }
+
+        driver.modern_device_init(features);
+
+        assert!(
+            test_ctx.dev.hdl.1.load(Relaxed) > updates_before_reset,
+            "reinitialization did not restore the promiscuous mode",
+        );
```
iximeow, line 4032: so we're.. instrumenting ioctls on the test fd by adding a field under `#[test]`, and we're asserting that the device is not in promiscuous mode on the basis of "we didn't see _some ioctl_ on this fd"...? this seems pretty brittle and very easy to pass for the wrong reasons.

it seems to me what we really need is to add a `get_promisc` ioctl and just come back and use that here?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2814,6 +3793,531 @@ mod test {
...
+            .get_mac_filters()
+            .expect("can read back filters")
+            .is_empty());
+        {
+            let state = test_ctx.dev.inner.lock().unwrap();
+            assert!(!state.mac_filters_installed);
+            assert!(!state.mac_table_set);
+            assert_eq!(state.promisc, PromiscLevel::AllMulti);
+        }
+        assert!(driver.status_ok());
```
iximeow, line 4098: probably should assert that `get_mac_filters()` returns an empty list here too, right?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -2814,6 +3793,531 @@ mod test {
...
+            .is_empty());
+        assert!(driver.status_ok());
+
+        test_ctx
+    }
+
+    #[cfg(not(feature = "falcon"))]
+    fn mac_filters_host_promisc_pin(test_ctx: TestCtx) -> TestCtx {
+        test_ctx
+    }
```
iximeow, line 4319: this maybe could be `#[cfg(feature = "falcon")] testcase!(mac_filters_host_promisc_pin)` on the list below, instead?

lib/propolis/src/migrate.rs
```diff
@@ -149,9 +149,16 @@ impl<'a> PayloadOffers<'a> {
...
+        match self.take_schema(T::id()) {
+            Some(mut offer) => offer.parse(),
+            None => match self.0.iter().find(|offer| offer.kind == T::id().0) {
+                Some(offer) => Err(MigrateStateError::UnexpectedPayload(
+                    offer.kind.into(),
+                    offer.version,
+                )),
+                None => Err(MigrateStateError::DataMissing),
+            },
+        }
```
iximeow, line 161: i really don't understand what's going on here or in the tests below. this looks like plumbing that's still around from trying to support a `VionaStateV2`? i'm not super familiar with the device migration payload bits and i'm not sure either of us want to get into that for this change, at least...

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -949,117 +1213,288 @@ impl PciVirtioViona {
...
+        //
+        // Note: Unrequested traffic is permitted by the VirtIO spec.
+        let promisc = if need_promisc {
             PromiscLevel::All
-        } else if need_mcast {
+        } else if need_mcast_promisc || !state.mac_table_set {
             PromiscLevel::AllMulti
         } else {
             PromiscLevel::None
+        };
```
iximeow, line 1338: i'm even less familiar with the networking space than virtio-net, which i am very much learning as-needed; we're setting the device up to interpret an empty multicast table in two different ways based on if the guest has ever set one before. but if the guest wants to not get multicast traffic, shouldn't we support CTRL_RX_EXTRA and CTRL_RX_NOMULTI? is there guest driver support for that? or, is it typical to set an empty table and expect that to mean "no traffic"?

i don't really know what the spec suggests for this case. filtering is a "SHOULD", so passing extra traffic is always okay. but "no MAC in the empty list matched, so don't pass the traffic" also seems like a valid reading. do you have a sense of what guest drivers are expecting to work with here, or is this more "it makes illumos consistently work so we'll have some extra ifs in viona.rs"?

lib/propolis/src/hw/virtio/viona.rs
```diff
@@ -949,117 +1213,288 @@ impl PciVirtioViona {
...
-                state.multicast_mac_filters = multicast;
-                Err(())
-            }
+        if state.rx_config_failed {
+            return Err(());
         }
+        let table_changed = multicast != state.multicast_mac_filters;
+        state.unicast_mac_filters = unicast;
+        state.multicast_mac_filters = multicast;
+        state.mac_table_set = true;
```
iximeow, line 1252: since we're using `mac_table_set` to mean "has the guest ever given us MAC filters" ... maybe we should just have an `struct GuestMacFilters { unicast, multicast }` and `Option<GuestMacFilters>`?

iximeow (commented):
I've left a bunch of thoughts spanning a few levels of detail, sorry for the mix going up and down - I kind of read this top to bottom in diff order and had different levels of thought along the way.

since I'd really like to get Propolis to match the stlouis VIONA_CURRENT_INTERFACE_VERSION before we cut the next release, I think what I'll do is a separate PR that bumps the expected interface version, and we can decouple the specifics of using the new MAC filtering ioctls separately. my understanding of the stlouis change is that until we're using those ioctls, the viona behavior is not any different. so the extant behavior around compiling a list of filters into viona promiscuity is fine today, and would be fine to switch away from at some point in the future? I'll tag you on that impending change but please shout here or there (or chat of course) if it sounds like I've misunderstood the kernel side here.

zeeshanlakhani:
@iximeow, thanks for the updated review. 

> since I'd really like to get Propolis to match the stlouis VIONA_CURRENT_INTERFACE_VERSION before we cut the next release, I think what I'll do is a separate PR that bumps the expected interface version, and we can decouple the specifics of using the new MAC filtering ioctls separately. my understanding of the stlouis change is that until we're using those ioctls, the viona behavior is not any different. so the extant behavior around compiling a list of filters into viona promiscuity is fine today, and would be fine to switch away from at some point in the future? I'll tag you on that impending change but please shout here or there (or chat of course) if it sounds like I've misunderstood the kernel side here.

I do think it makes sense to cut the V7 version update separately. I mainly wanted to try and begin the work on what approaching this interface would be like, and I'll separately go through your review here for sure. Let me know when you have the bump up, so we can get that in as soon as we can.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: It will fall back to all-multicast in that case, but I'll add a test to cover this.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: Good point. I will actually reuse and extend the existing `QueueWriter` for this, or will try at least.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: I dropped the loop here altogether and added the capacity ioctl. 

We'll now check the limit before installing the table.

lib/propolis/src/hw/virtio/viona.rs
zeeshanlakhani: The meaning for `all_hosts` here refers to IPv4's 224.0.0.1 group address, which includes gateways (per spec), and `all_nodes` refers to IPv6's ff02::1 group address, which includes hosts and routers (per spec). Their Ethernet addresses are just sample multicast filter entries.

I've renamed these to `ipv4_all_hosts` and `ipv6_all_nodes` to make the protocol association more explicit.

PR #1216, opened 2026-09-16
bump viona API version to match stlouis

iximeow:
This brings ominous tidings for users of Propolis on illumos distributions that are not Helios: the illumos-gate viona API version is still 6, and will continue to be until https://code.oxide.computer/c/illumos-gate/+/775 is upstreamed. viona version 7 includes new ioctls which Propolis does not *yet* use, but will soon.

At that point, Propolis will not run on other illumos distributions until the MAC filter table ioctls are available upstream. or worse, some inadvertent change gives viona API version 7 a different meaning upstream...


iximeow:
> [header-check](https://github.com/oxidecomputer/propolis/pull/1216/checks?check_run_id=104901643695)Successful in 1m — Success!

its a miracle

PR #1218, opened 2026-09-18
standalone: should work without explicit CPUID profiles

iximeow:
93e04e189 added support for selecting specific hypervisor interfaces to propolis-standalone, rather than the prior state of "guest always sees bhyve because of the 0x4000_0000 leaf default in vmm_cpuid.c".

unfortunately, all my test VM configs include a `cpuid_profile`, so I missed that this change introduced a bug. given a configuration with the default CPUID leaves, the hypervisor interface leaves would get added and misconfigure guests to have all-zero CPUID leaves for every non-hypervisor leaf.

many things go wrong; EDK2 sometimes will do bogus rdmsr and panic propolis on the unimplemented MSR. *if* you get past EDK2, GRUB checks for the TSC's presence via leaf 1 EDX[4], which is also clear. failing to find a TSC, GRUB gives up trying to boot.

this change has propolis-standalone collect CPUID information more like propolis-server - do not rely on "default" resulting in an empty CPUID profile getting to bhyve, and instead at the point we decide to use a default profile, collect that from bhyve and explicitly provide it later.


iximeow:
Luiz tripped over this as the first person to `propolis-standalone` without a CPUID profile since I clicked merge on 1194, sorry about the troubles. this isn't an issue for server, because propolis-server does the thing I thought was happening here: get a full CPUID listing and edit in some additional leaves. separately, we should at least smoketest propolis-standalone in CI across a few configs, because breaking test tools is not something I'm very proud of :)

lgfa29 (approved):
Tested with the same setup that was failing before and it's working now 🥳 

Thanks!

